PR #43, opened 2016-11-17
Fix bad parameter when calling nmcli to get DNS

krepver:
With nmcli version 1.2.2 i got this error

nmcli_dev_list_error


seladb:
Apparently nmcli params were changed from version 1.0 forward, so `nmcli dev list` works only in versions < 1 and `nmcli dev show` works only in versions > 1. So we need to find a solution that works for all versions

seladb:
Closed this pull request and create issue #46 instead

PR #48, opened 2017-03-03
Improve the LinkLayerType and add LINKTYPE_RAW support

Gnnng:
This PR is developed mainly on macOS Sierra and Ubuntu 16.04 where two tests (`Packet++Test` & `Pcap++Test`) are all green in manual tests. Also, Two continuous integrations are passed, which means it will be able to run on Windows too.

It includes one enhancement and some fixes:

1. Added support of new LinkLayerType, `LINKTYPE_RAW`. Now it can parse capture file containing raw IP packets (e.g. [CAIDA-passive](https://www.caida.org/data/passive/)).
2. Fixed some values of enum LinkLayerType in `RawPacket.h`. This bug occurred in the development of adding support of `LINKTYPE_RAW`. The values corresponding to the names (`LINKTYPE_*`) should have values of `DLT_*` as shown in `RawPacket.h` and `DLTDef.h`. See the explanation below for more information.
3. Fixed some errors in the configuration script of macOS. 
4. Fixed some bugs in`Packet++Test` and `Pcap++Test` which prevent the compilation of these tests on macOS.

**Explanation of 2**

This is one usage of LinkLayerType is in the file `Pcap++/src/PcapFileDevice.cpp`.
```
m_PcapLinkLayerType = static_cast<LinkLayerType>(pcap_datalink(m_PcapDescriptor));
```

According to the definition of `pcap_datalink`, it will return the field `linktype` of the struct `pcap_t`, which is initialized in function `pcap_check_header`.  
```
/* in file pcap.c */
int
pcap_datalink(pcap_t *p)
{
        if (!p->activated)
                return (PCAP_ERROR_NOT_ACTIVATED);
        return (p->linktype);
}
```

The function `pcap_check_header` will return a value of `DLT_*` instead of `LINKTYPE_*`, which have different values in several cases ([tcpdump doc](http://www.tcpdump.org/linktypes.html)).  From this line in `pcap_check_header`, `linktype` is assigned by a `DLT_*` value (`linktype_to_dlt` is a function coverting a int from `LINKTYPE_*` to `DLT_*`).

```
        p->linktype = linktype_to_dlt(LT_LINKTYPE(hdr.linktype));
```

So, the conclusion is that the values of enum LinkLayerType should be `DLT_* `. The newly added header `DLTDef.h` is to provide these values. It's part of the header file `pcap/bpf.h` which provides all the values. I have to do this (copy some values from libpcap header) because of the difficulties of compiling on Windows when including `pcap/bpf.h` directly.


seladb:
Thanks a lot for this contribution!
Please see below few questions and comments:

1. Regarding addition of the file DLTDef.h: I'm not very keen in coping code from libpcap to PcapPlusPlus and would like to avoid that. When looking in this file I can see that most of the DLT_* values doesn't change between different platforms, only DLT_RAW changes between OpenBSD and other platforms. Since OpenBSD isn't officially supported in PcapPlusPlus I don't see a reason to add this file and DLT_* macros. In addition when looking at the linktypes in [http://www.tcpdump.org/linktypes.html](http://www.tcpdump.org/linktypes.html) it seems the LINKTYPE_RAW value should be 101, not 12 or 14 as defined in DLTDef.h , do you know the reason for this difference?

2. Regarding the changes in Packet++Test and Pcap++Test (both in main.cpp and in the Makefile) - did compilation fail on MacOS Sierra before you added `#elif defined(LINUX)` ? This is weird because Travis-CI contains MacOS Sierra configuration and the build process passes in this configuration. Do you know why?

Gnnng:
Thanks for the reply.

For your first questions, the real pain behind this is the usage of `LINKTYPE_*` names in the `enum LinkLayerType`, as the **Explanation of 2** saying the return of libpcap function `pcap_datalink(m_PcapDescriptor)` is `DLT_*`, which is defined in `pcap/bpf.h` (from which I copied the macros into `DLTDef.h`), instead of `LINKTYPE_*`. Luckily, the previous code was fine because these types (`LINKTYPE_ETHERNET`, `LINKTYPE_LINUX_SLL` & `LINKTYPE_NULL`) have exactly same values in `LINKTYPE_*` and `DLT_*` macros. When I blindly added `LINKTYPE_RAW` to add support of raw IP type, I failed because of the difference. Although http://www.tcpdump.org/linktypes.html mentioned in capture file `LINKTYPE_*` is used, it also said that `pcap_datalink` returns `DLT_*` (which you may miss or ignore). I have to admit my fix of copying macros is too straightforward, but what will you do when libpcap gives you `DLT_*` values instead of `LINKTYPE_*` values?

Unfortunately, I'm not aware of the reason for the second question. I can confirm that I am using Xcode 8.2.1 and in the previous code it didn't compile with an error saying that `<in.h>` can not be found. Then, I checked the code of Packet++Test and accidentally found that the macro `WIN32` is not defined in the Makefile. That's why I added the variable `DEPS` in the Makefile as you do in others.

seladb:
Regarding `LINKTYPE_*` vs. `DLT_*` - could you please give an example of the conflict you mention? I downloaded and opened several files in [https://www.caida.org/data/passive/](https://www.caida.org/data/passive/) and the link type is Cisco HDLC (LINKTYPE_C_HDLC) which value is 104. Could you please give a pcap example with `LINKTYPE_RAW` (value of 12 or 14)?
Like I wrote previously, the only `DLT_*` value that is changed between platforms is `DLT_RAW` so I think we can solve this specific case with an `ifdef` macro instead of coping code form libpcap.
If I miss something please let me know

Regarding missing `in.h` - this is really weird, the file should exist under `/usr/include/netinet` - could you please check? 
By the way - did you run `configure-mac_os_x.sh` before building the code? This script adds `/usr/include/netinet` to the include path. Without this compilation will surely fail

Gnnng:
Actually, `LINKTYPE_RAW` comes from this command, `editcap` from Wireshark.
```
editcap -r -v equinix-chicago.20150219-125911.UTC.anon.pcap first1000.pcap 1-1000
```
For example, the above line will export the first 1000 packets from a capture file from CAIDA to `first1000.pcap`, you can get the `LINKTYPE_RAW` from the exported file.

As for the header `in.h`. I've checked on my Mac, the header path should be `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/netinet`. I don't even have the path `/usr/include`.

seladb:
Could you please attach the first1000.pcap file to this thread?

Gnnng:
The file [first1000.pcap.zip](https://github.com/seladb/PcapPlusPlus/files/823110/first1000.pcap.zip). Please unzip it first.

And `xcode-select --install` works. Thanks.

seladb:
The linktype of the packets in the file you sent is `LINKTYPE_RAW` (value 101), but I found other pcap files with linktype `DLT_RAW` (value 14). 
So my suggestion is this: instead of adding a new file, let's add to the `LinkLayerType` enum a couple of new values: `LINKTYPE_DLT_RAW1` (value 12), and `LINKTYPE_DLT_RAW2` (value 14). In `PcapFileDevice` let's support all of them as raw linktype: `LINKTYPE_RAW`, `LINKTYPE_DLT_RAW1`, `LINKTYPE_DLT_RAW2`, `LINKTYPE_C_HDLC`. What do you think?

Regarding `xcode-select --install` - I'll add it to the Mac OS X prerequisite list

Gnnng:
Thanks for your careful investigation. From http://www.tcpdump.org/linktypes.html, I'm doubtful of the value 14 existing in the capture file because it said

> LINKTYPE_ value is the numerical value used in capture files

Neither value 12 nor 14 should be found in capture files because they are just DLT_ values corresponding to the DLT name `DLT_RAW` and they are returned from `pcap_datalink()` when the function is called on a capture file containing `LINKTYPE_RAW` (value 101). Maybe you could show some evidence about your findings?

seladb:
Here are 2 pcap files I found on the Internet that have linktype value of 12 (please unzip it):
[dlt_raw_examples.zip](https://github.com/seladb/PcapPlusPlus/files/825593/dlt_raw_examples.zip)
Also, if you open these files with HexEdit you can see that the linktype value written there is 12 (and not 101), so apparently `pcap_datalink()` can return all kinds of values (12, 14, 101, 104, etc.)

seladb:
So what do you think?

seladb:
I added support for Raw IP by adding the values LINKTYPE_DLT_RAW1 (12) and LINKTYPE_DLT_RAW2 (14) to LinkLayerType enum. It's now supported for both reading and writing pcap/pcapng files.

I also added to README.md running `xcode-select --install` as a prerequisite for building PcapPlusPlus on Mac OS X

Gnnng:
Sorry for the interrupt. I didn't manage to have time working on this. Glad to see it's been done.

PR #59, opened 2017-08-04
Fix boundary checking in getTcpOptionData(TcpOption), fixes #58

### LINKED ISSUE #58: Bad boundary check in TcpLayer.getTcpOptionData(TcpOption)
The `TcpLayer.getTcpOptionData(TcpOption)` method does boundary checking on the entire packet when parsing TCP options instead of the end of the TCP header.

This works fine when there is no payload in the packet (e.g. it's a SYN packet) and the end of the packet is the same as the end of the header, but when the packet has payload and the end of the entire packet is not the end of the header, it will either get stuck in an infinite loop when a 0 is read as option size somewhere in the payload, or return garbage as option data if the requested option code is found somewhere in the payload.

To fix it, the boundary check in the while loop should be changed from
```while ((curOptPtr - m_Data) < m_DataLen)```
 to:
```while ((curOptPtr - m_Data) < (((tcphdr *)m_Data)->dataOffset * 4))```

### PR DESCRIPTION
**Pull request by rpanah**

### CODE REVIEW

File: Packet++/src/TcpLayer.cpp [hunk 1]
```diff
@@ -29,7 +29,9 @@ TcpOptionData* TcpLayer::getTcpOptionData(TcpOption option)
 		return NULL;
 
 	uint8_t* curOptPtr = m_Data + sizeof(tcphdr);
-	while ((curOptPtr - m_Data) < m_DataLen)
+	uint16_t dataOffset = ((tcphdr *)m_Data)->dataOffset * 4;
+
+	while ((curOptPtr - m_Data) < dataOffset)
```
seladb: Thanks for the fix! I think it's worth checking both m_DataLen and dataOffset in case one of them is falsely set. What do you thin?

rpanah: Sure. I'll add that.

PR #60, opened 2017-08-09
For splitting by method client/server, PcapSplitter now includes the ip that file pertains to in its name

bpagon13:


Examples/PcapSplitter/main.cpp
```diff
@@ -184,6 +184,109 @@ std::string getFileNameWithoutExtension(const std::string& path)
 	return("");
 }
 
+std::string getIP(pcpp::Packet& packet, bool client)
+{
+	std::string packetStr = packet.printToString();
+	int srcPos = packetStr.find("IPv4") + 17;
+	int separator = packetStr.find(",", srcPos + 1);
+	std::string srcIP = packetStr.substr(srcPos, separator - srcPos);
```
seladb: Why extracting source and dest IP like this?
It's better to do something like this:

    pcpp::IPv4Layer* ipLayer = packet.getLayerOfType<IPv4Layer>();
    if (ipLayer == NULL)
        return "";
    std::string srcIP = ipLayer->getSrcIpAddress().toString();
    std::string dstIP = ipLayer->getDstIpAddress().toString();

Examples/PcapSplitter/main.cpp
```diff
@@ -346,6 +449,27 @@ int main(int argc, char* argv[])
 
 		// call the splitter to get the file number to write the current packet to
 		int fileNum = splitter->getFileNumber(parsedPacket, filesToClose);
+		std::string ipVal = "";
+		if (method == SPLIT_BY_IP_CLIENT) {
```
seladb: Maybe it's worth moving this logic to `ClientIPSplitter` class?

Examples/PcapSplitter/main.cpp
```diff
@@ -346,6 +449,27 @@ int main(int argc, char* argv[])
...
+			if (fileNum != 0) {
+				ipVal = "-" + getIP(parsedPacket, true);
+				hyphenIP(ipVal);
+			}
+			else {
+				// file number 0 is for various misceallaneous packets (ping, ARP, etc.)
+				ipVal = "-miscellaneous";
+			}
+		}
+		else if (method == SPLIT_BY_IP_SERVER) {
```
seladb: Maybe it's worth moving this logic to `ServerIPSplitter`?

Issue #56, opened 2017-06-27
Expose pcap_set_immediate_mode()


gpean:
It would be nice if this setting was exposed. Sometimes you need to get to packets in the fastest way possible (assuming there is not a lot of them).

gpean:
```
In immediate mode packets get delivered to the application as soon as they 
arrive, rather than after a timeout. This makes it very important for latency 
sensitive live captures. In my app this cuts the latency introduced by pcap 
from 2.5ms to 0.5ms.
```

(From https://github.com/google/gopacket/issues/36)

seladb:
The problem is this API isn't exposed in WinPcap and also not in libpcap < 1.5.0. What do you suggest?

gpean:
Hmm only expose if `-DHAS_PCAP_IMMEDIATE_MODE` i.e. if requested at
configure time?

seladb:
That's a great idea. Would you mind doing this change and create a pull request?

seladb:
Hi, I see you did a lot of changes in you branch. Do you have any plans to merge at least some of it back to master?

gpean:
I would need to clean them up - Right now I'm pressured to finish a project and as a result, I have to make quick changes in more places, which wouldn't necessary be fit for all the users of this library. I'll do my best to give back, if I find the time. If you want to copy paste some targeted changes, feel free to do it too. Thanks...

seladb:
Ok, I'll really appreciate if you find the time to merge some of the code back to master

seladb:
Fixed by @f-squirrel (many thanks!)

PR #68, opened 2017-10-16
ER #56 - Expose pcap_set_immediate_mode

f-squirrel:
- Adding HAS_PCAP_IMMEDIATE_MODE to make compilation flags enables pcap
immediate mode
- Before using it please check if your libpcap version supports the
feature

### DIFF
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -37,6 +37,8 @@
 #define LIBPCAP_OPEN_LIVE_TIMEOUT -1
 #endif
 
+static const int DEFAULT_SNAPLEN = 9000;
+
 namespace pcpp
 {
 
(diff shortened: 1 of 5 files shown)


seladb:
Thanks for this addition! You've used the macro `HAS_PCAP_IMMEDIATE_MODE`. Where do you get it from? How do you know during compile time whether the current installed libpcap/WinPcap version supports `pcap_set_immediate_mode()`?

f-squirrel:
Hi @seladb,
Actually, I did according to the discussion you have had in the original ticket:)
Thus, since the feature is optional, the one who wants to use it has to add `+DEPS += -DHAS_PCAP_IMMEDIATE_MODE` to Pcap++/Makefile.

Other way to bring the support is to do the following:
#ifndef WIN32
    pcap_version > 1.5.0
          pcap_set_immediate_mode(PCAP_IMMEDIATE_MODE)
#endif 
PCAP_IMMEDIATE_MODE is going to be passed through makefile with 0 as default.

WDYT?

seladb:
Actually I prefer the second way you suggested because it's easier to use and understand and doesn't involve further changes in the Makefile.
But I'm not sure how you can check libpcap version.
If specific libpcap version installed doesn't support `pcap_set_immediate_mode` then the code won't compile. I'm not sure there is an easy way in C/C++ to call a function only if it exists (unless you use a macro like you suggested in the first solution).
What do you think?

f-squirrel:
Hi,

1. As far as I know there is not portable way to know if function exists in *.so and *.dll and *.dynlib. We may invent some tricky way but we'll have to support it all the way down.
2. There is no way to know libpcap version at run-time.
3. `pcap_lib_version` returns version string in runtime. For instance, at MacOS it returns 
> libpcap version 1.8.1 -- Apple version 67.60.1
I suppose at Linux it returns something else. Thus, we need to parse the return value and of course it will happen only in realtime.

So, to  I propose to use the solution I've pulled to give the user a possibility to decide to use or not to use the feature(in any case if his libpcap does not support immediate mode the build will fail).

BR!

seladb:
I tend to agree with you. Let's go with your solution but I have one more suggestion to make this feature more accessible: why don't we add a switch to `configure-linux.sh` and `configure-macosx.sh` (something like `--use-immediate-mode`) which the user can set if he wants to enable this feature.

These scripts will add a `HAS_PCAP_IMMEDIATE_MODE=1` variable to `PcapPlusPlus.mk` which will be read by Pcap++ Makefile:
```
ifdef HAS_PCAP_IMMEDIATE_MODE
DEPS += -DHAS_PCAP_IMMEDIATE_MODE
endif

```
What do you think? Will you be able to add this to the pull request?

f-squirrel:
Hi,
It looks more elegant!
I'll try to find out how to add it to the configuration scripts and then update the pull request.

Thanks

seladb:
Did you manage to find out how to do it? Do you need any help?

f-squirrel:
Hi,
Yes, I did. Please, review the updated pull request.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -210,21 +212,66 @@ void* PcapLiveDevice::statsThreadMain(void *ptr)
 	return 0;
 }
 
+pcap_t* PcapLiveDevice::doOpen(DeviceMode mode)
+{
+	char errbuf[PCAP_ERRBUF_SIZE] = {'\0'};
+	pcap_t* pcap = pcap_create(m_Name, errbuf);
```
seladb: Why did you break `pcap_open_live` into all of these calls to libpcap?
Why not keeping `pcap_open_live` and then call `pcap_set_immediate_mode`?

f-squirrel: The reason is that `pcap_open_live` creates already activated pcap handle, but `pcap_set_immediate_mode` should be called for non-active instance. Otherwise, it returns `PCAP_ERROR_ACTIVATED ` error.
Please, refer to http://www.tcpdump.org/manpages/pcap_set_immediate_mode.3pcap.html

seladb: Oh ok I didn't know that. Thanks!

PR #70, opened 2017-10-24
Added gitignore file

f-squirrel:
Hi,
I've added git ignore to hide all auto-generated files like(binaries, compilation object, vim swap etc).


seladb:
For some reason all of my previous commits are included in this PR. Could you please create it again without them?

Issue #78, opened 2017-12-16
C++17 incompatibility in Packet++ lib


echo-Mike:
Library Packet++ can't be build with gcc/g++ version 7.0.* or higher if -std=c++17 flag is specified within build instructions for object files in next Makefiles:  
* Common++/Makefile:42
* Pacp++/Makefile:70
* Packet++/Makefile:36
* Pcap++Test/Makefile:39
* Packet++Test/Makefile:23  

My purpose was to use PcapPlusPlus library in project that written in terms of c++17 standard.  
Using flag -std=c++14 was found that in following files deprecated function (constructors) definitions was used:  
* Packet++/header/SipLayer.h:542 paired with  Packet++/src/SipLayer.c:133
* Packet++/header/HttpLayer.h:573 paired with  Packet++/src/HttpLayer.c:212  

See [P0003R0](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0003r0.html)  for details on deprecated definition.  
Pull request for closing this issue will be done in next couple of hours. The build process of library with changes was tested on Ubuntu 16.04.3 LTS with gcc 7.1.0.

seladb:
Resolved in PR #79

PR #79, opened 2017-12-16
C++17 compatibility issue resolved

echo-Mike:
Issue #78 closed.  
See [P0003R0](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0003r0.html) for details.

### DIFF
--- a/Packet++/src/HttpLayer.cpp
+++ b/Packet++/src/HttpLayer.cpp
@@ -5,7 +5,7 @@
 #include <string.h>
 #include <algorithm>
 #include <stdlib.h>
-
+#include <exception>
 
 namespace pcpp
 {
@@ -209,7 +209,7 @@ HttpRequestFirstLine::HttpRequestFirstLine(HttpRequestLayer* httpRequest) : m_Ht
 }
 
 HttpRequestFirstLine::HttpRequestFirstLine(HttpRequestLayer* httpRequest, HttpRequestLayer::HttpMethod method, HttpVersion version, std::string uri)
-		throw(HttpRequestFirstLineException)
+try		// throw(HttpRequestFirstLineException)
 {
 	if (method == HttpRequestLayer::HttpMethodUnknown)
 	{
@@ -240,7 +240,12 @@ HttpRequestFirstLine::HttpRequestFirstLine(HttpRequestLayer* httpRequest, HttpRe
 
 	m_IsComplete = true;
 }
-
+catch(const HttpRequestFirstLineException&) {
+	throw;
+}
+catch(...) {
+	std::terminate();
+}
 
 HttpRequestLayer::HttpMethod HttpRequestFirstLine::parseMethod(char* data, size_t dataLen)
 {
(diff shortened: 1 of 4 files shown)


Packet++/src/HttpLayer.cpp
```diff
@@ -209,7 +209,7 @@ HttpRequestFirstLine::HttpRequestFirstLine(HttpRequestLayer* httpRequest) : m_Ht
 }
 
 HttpRequestFirstLine::HttpRequestFirstLine(HttpRequestLayer* httpRequest, HttpRequestLayer::HttpMethod method, HttpVersion version, std::string uri)
-		throw(HttpRequestFirstLineException)
+try		// throw(HttpRequestFirstLineException)
```
seladb: Isn't it better to wrap the try-catch clause with curly brackets (in the beginning and end of the constructor?)

Packet++/src/SipLayer.cpp
```diff
@@ -129,7 +130,7 @@ SipRequestFirstLine::SipRequestFirstLine(SipRequestLayer* sipRequest) : m_SipReq
 }
 
 SipRequestFirstLine::SipRequestFirstLine(SipRequestLayer* sipRequest, SipRequestLayer::SipMethod method, std::string version, std::string uri)
-		throw(SipRequestFirstLineException)
+try		// throw(SipRequestFirstLineException)
```
seladb: Isn't it better to wrap the try-catch clause with curly brackets (in the beginning and end of the constructor?)

Packet++/src/HttpLayer.cpp
echo-Mike: @seladb Please take a look on this video https://youtu.be/vmtGCJbNKNE  
I personally think constructors of type:  
```
ClassName(args)
try 
{ /*...smth...*/}
catch(const KnownException& e)
{ /*...smth with exception and resources...*/ }
...
catch(...) 
{/*...smth with unknown exception..*/}
```
and
```
ClassName(args) 
{
    try 
    { /*...smth...*/}
    catch(const KnownException& e)
    { /*...smth with exception and resources...*/ }
    ...
    catch(...) 
    {/*...smth with unknown exception..*/}
}
```  
are not totally same from position of c++ run-time library and compiler. But i may be wrong.  
If You find a reason why any of this constructions are better please leave a response with links here.

seladb: I actually didn't know about function-try-blocks, thanks for the free c++ lesson :)
After reading about it a little and watching your video I think that function-try-blocks in c'tors are best for catching exceptions thrown by either by initializers or base-class c'tor. Both are not the case here. Also, function-try-blocks can't really catch the exception (as they have to end with `throw`) but they allow you to do some intermediate handling before the exception is thrown. In this case there is no intermediate handling in the catch clause - we just throw or terminate.
Having said all of that - there's nothing wrong in keeping that code, it may have some value in the future.
I leave it for you to decide. Either way I'll approve your PR. Please let me know what you think

echo-Mike: @seladb Ok, I am happy about that You can learn something new from this PR. By now I have no intention to change something so if You don't mind please approve merge.

PR #86, opened 2018-01-11
Fixed PcapLiveDevice timeout causing high CPU usage

### LINKED ISSUE #85: High CPU on windows machines
There is a problem on windows machines that causes high CPU usage.

The bug is that you pass `-1` ([here](https://github.com/seladb/PcapPlusPlus/blob/4b5e72ed882d76b804ccbd8c6b6e2800f100662e/Pcap%2B%2B/src/PcapLiveDevice.cpp#L37)) to pcap_set_timeout ([here](https://github.com/seladb/PcapPlusPlus/blob/master/Pcap%2B%2B/src/PcapLiveDevice.cpp#L234)) and that causes winpcap to not do context switches when you start the capture.

The fix should be to use a proper timeout (like 1000 ms).
I don't know what the timeout should be on Mac but I think it should also be 1000 ms.

Steps to reproduce:

1. Clone this repo on a windows machine
2. Build the PcapPlusPlus.sln and the PcapPlusPlus-Examples.sln
3. Run the DnsSpoofing.exe example
4. Observe the high CPU (25% in a 4 cores machine, meaning 1 core is at 100%) 

Image from Task Manager:
image

yosizelensky:
Fixes #85 

Before the fix:
image

After the fix:
image

### DIFF
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -29,14 +29,7 @@
 #include <net/if_dl.h>
 #endif
 
-// On Mac OS X timeout of -1 causes pcap_open_live to fail so value of 1ms is set here.
-// On Linux and Windows this is not the case so we keep the -1 value
-#ifdef MAC_OS_X
-#define LIBPCAP_OPEN_LIVE_TIMEOUT 1
-#else
-#define LIBPCAP_OPEN_LIVE_TIMEOUT -1
-#endif
-
+static const int DEFAULT_OPEN_LIVE_TIMEOUT = 1000;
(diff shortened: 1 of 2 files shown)


seladb:
Thanks for this PR.
There is a reason why the default timeout is set to -1: what this timeout means is how much time goes by until libpcap/WinPcap hands the captures packets back to user space. If the timeout is set to 1000 it means that when a packet arrives it may wait up to 1000ms (1 sec) before it is handed to the application. This reduces the number of system call and improves performance at heavy load. However at low load it causes packets to get "stuck" for a long time and overall performance goes down.
I suggest leaving it the way it is

PR #92, opened 2018-02-17
Support dpdk newer than 16.07

luigino:


seladb:
Thanks for this fix!
But please notice the Travis CI tests don't pass because you added `#include "rte_pci.h"` in `DpdkDevice.cpp` which doesn't exist in DPDK 16.11. Can you please move this include to inside of `#if (RTE_VER_YEAR >= 16) && (RTE_VER_MONTH > 07)`?

Pcap++/src/DpdkDevice.cpp
```diff
@@ -8,6 +8,11 @@
 #include "DpdkDevice.h"
 #include "DpdkDeviceList.h"
 #include "Logger.h"
+#include "rte_version.h"
+#if (RTE_VER_YEAR >= 16) && (RTE_VER_MONTH > 07)
+#include "rte_bus_pci.h"
+#endif
+#include "rte_pci.h"
```
seladb: This is the line which causes DPDK 16.11 build to fail. If you can add this line one line up, so it's inside the `#if (RTE_VER_YEAR >= 16) && (RTE_VER_MONTH > 07)` it'd great

Pcap++/src/DpdkDevice.cpp
```diff
@@ -8,6 +8,11 @@
 #include "DpdkDevice.h"
 #include "DpdkDeviceList.h"
 #include "Logger.h"
+#include "rte_version.h"
+#if (RTE_VER_YEAR >= 16) && (RTE_VER_MONTH > 07)
+#include "rte_bus_pci.h"
```
seladb: This file doesn't exist below DPDK 17.11 and is causing DPDK 16.11 to fail. Could you please fix this?

Pcap++/src/DpdkDeviceList.cpp
```diff
@@ -251,15 +251,26 @@ SystemCore DpdkDeviceList::getDpdkMasterCore()
...
 	if (logLevel == LoggerPP::Normal)
 		rte_set_log_level(RTE_LOG_NOTICE);
 	else // logLevel == LoggerPP::Debug
 		rte_set_log_level(RTE_LOG_DEBUG);
+#endif
 }
 
 LoggerPP::LogLevel DpdkDeviceList::getDpdkLogLevel()
 {
+#if (RTE_VER_YEAR >= 16) && (RTE_VER_MONTH > 07)
```
seladb: Same comment regarding the `#if`, in order for it to work on all DPDK version later than 17.11 you should change it to:
`#if (RTE_VER_YEAR > 17) || (RTE_VER_YEAR == 17 && RTE_VER_MONTH >= 11)`

Issue #119, opened 2018-06-22
"dpdk segmentation fault"  MBufRawPacket::~MBufRawPacket()


shidaoxuan:
My dpdk demo calls startCaptureMultiThreads() to capture packets without sending out.
this demo is similar to “PCAPP_TEST(TestDpdkMultiThread)”,  after few senconds, the demo crash because MBufRawPacket::~MBufRawPacket() called "rte_pktmbuf_free(m_MBuf);". i found m_Mbuf is an abnormal data. m_Mbuf.pool is 0. so rte_pktmbuf_free() cause segmentation fault.

I am confused about this, I do not know why this data has become abnormal.

shidaoxuan:
I found the reason.  In class DpdkDevice , a member named "struct rte_mbuf* m_mBufArray[256];", DpdkDevice.m_mBufArray will be shared by all capture threads, it's unsafe, this data member will be overwrite by multithreading.
My suggestion is as follows:
DpdkDevice.h:1018 ----> struct rte_mbuf* m_mBufArray[MAX_NUM_OF_CORES][256];

DpdkDevice.cpp:922 ----> uint32_t numOfPktsReceived = rte_eth_rx_burst(pThis->m_Id, queueId, pThis->m_mBufArray[coreId], MAX_BURST_SIZE);

DpdkDevice.cpp:935 ----> struct rte_mbuf* mBuf = pThis->m_mBufArray[coreId][index];

echo-Mike:
Hi, @shidaoxuan.
You have found a general flaw in callback based system in DpdkDevice class. The proposed solution is not bad but it implies false sharing if cache line alignment is not meet.
I do not see any easy way to fix this issue except to **not to use callback based system** by now.

echo-Mike:
@shidaoxuan, @seladb I possibly have found a solution, will make PR today near **22:00 CET**

seladb:
Thanks @shidaoxuan , this is indeed a bug and I'm not sure how I didn't find it until now so thanks for reporting it!

Your suggested solution seems reasonable. @echo-Mike - is this what you had in mind?

echo-Mike:
@seladb please see my commit above.
## Proposed solution
The **m_mBufArray** field is an opaque member field of ```DpdkDevice``` that is not used anywhere besides places mentioned above. There is no reason to store it inside ```DpdkDevice``` object and as we saw there is a reason to made this array **thread local** value that is located on thread stack.
[...]

PR #120, opened 2018-06-24
[Fix] Running fix for #119

echo-Mike:
See #119 description and conversation.

### DIFF
--- a/Pcap++/src/DpdkDevice.cpp
+++ b/Pcap++/src/DpdkDevice.cpp
@@ -905,6 +905,8 @@ void DpdkDevice::stopCapture()
 int DpdkDevice::dpdkCaptureThreadMain(void *ptr)
 {
 	DpdkDevice* pThis = (DpdkDevice*)ptr;
+	struct rte_mbuf* mBufArray[MAX_BURST_SIZE];
+	MBufRawPacket rawPackets[MAX_BURST_SIZE];
 
 	if (pThis == NULL)
 	{
@@ -919,7 +921,7 @@ int DpdkDevice::dpdkCaptureThreadMain(void *ptr)
 
 	while (likely(!pThis->m_StopThread))
 	{
-		uint32_t numOfPktsReceived = rte_eth_rx_burst(pThis->m_Id, queueId, pThis->m_mBufArray, MAX_BURST_SIZE);
+		uint32_t numOfPktsReceived = rte_eth_rx_burst(pThis->m_Id, queueId, mBufArray, MAX_BURST_SIZE);
 
 		if (unlikely(numOfPktsReceived == 0))
 			continue;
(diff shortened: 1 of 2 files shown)


Pcap++/src/DpdkDevice.cpp
```diff
@@ -929,11 +943,10 @@ int DpdkDevice::dpdkCaptureThreadMain(void *ptr)
 
 		if (likely(pThis->m_OnPacketsArriveCallback != NULL))
 		{
-			MBufRawPacket rawPackets[numOfPktsReceived];
+			MBufRawPacketArrayGuard sg(rawPackets, numOfPktsReceived);
```
seladb: I'm not sure why you need this new class `MBufRawPacketArrayGuard`. Couldn't you just put `MBufRawPacket rawPackets[MAX_BURST_SIZE]` instead?

If the reason is being more efficient (and avoid creating this array), maybe it's better to go over the elements of the array and call to `clear()` for each one, which is exactly what `MBufRawPacketArrayGuard` does in its d'tor?

shidaoxuan: Class “MBufRawPacketArrayGuard” is to enforce the RAII mechanism, but MBufRawPacket's  d'tor would have released resources(mbuf).
I'm not sure MBufRawPacketArrayGuard is necessary if put MBufRawPacket rawPackets[numOfPktsReceived] instead?

echo-Mike: @seladb @shidaoxuan There are two ways that execution may go from point of callback call:
1. Normal execution
2. Exception is thrown by user provided callback

In either of them resources of packets must be freed because they are allocated in some pages outside of process memory that will be automatically released by the system when process ends (a.e. systemwide memory leak is possible).  
Let us now count the amount of work to be done on every proposed solution:
1. ```MBufRawPacket rawPackets[numOfPktsReceived]``` inside ```if``` clause: UB in C++ but common sense tells that at least ```numOfPktsReceived``` constructors and destructors will be called;
2. ```MBufRawPacket rawPackets[MAX_BURST_SIZE]``` inside ```if``` clause: ```MAX_BURST_SIZE``` constructors and destructors will be called every time -> the overhead is ```MAX_BURST_SIZE - numOfPktsReceived``` on every burst received;
3. The RAII approach: ```MAX_BURST_SIZE``` constructors and destructors are called only once (when entering and leaving ```dpdkCaptureThreadMain```) + exactly ```numOfPktsReceived``` calls to ```MBufRawPacket::clear``` on either of execution paths described above. The small overhead of constructor and destructor of ```MBufRawPacketArrayGuard``` is negated by compiler on optimisation (code may be inlined).  

@seladb @shidaoxuan Please let me hear your opinion on my arguments.

seladb: @echo-Mike I agree that may be the most efficient, but the overhead of creating `MBufRawPacket` instances x`MAX_BURST_SIZE - numOfPktsReceived` times for every burst isn't that bad and makes the solution much cleaner and easier to understand. I'd go with option 2 (`MBufRawPacket rawPackets[MAX_BURST_SIZE]`)

seladb: Hi @echo-Mike , did you see my previous message? Please let me know if you agree so we can merge to PR to the main branch

PR #135, opened 2018-10-01
Add filter capabilities to PcapNgFileWriterDevice

vrobles-gee:
This patch adds support to specify BPF filters to in a PcapNgFileWriterDevice instance, in order to control what packets are written or not to the capture file.

### DIFF
--- a/Pcap++/header/PcapFileDevice.h
+++ b/Pcap++/header/PcapFileDevice.h
@@ -388,12 +388,17 @@ namespace pcpp
 	class PcapNgFileWriterDevice : public IFileWriterDevice
 	{
 	private:
-
 		void* m_LightPcapNg;
+		struct bpf_program m_Bpf;
+		bool m_BpfInitialized;
+		int m_BpfLinkType;
+		std::string m_CurFilter;
 
 		// private copy c'tor
 		PcapNgFileWriterDevice(const PcapFileWriterDevice& other);
 		PcapNgFileWriterDevice& operator=(const PcapNgFileWriterDevice& other);
+
+		bool matchPacketWithFilter(const uint8_t* packetData, size_t packetLen, timeval packetTimestamp, uint16_t linkType);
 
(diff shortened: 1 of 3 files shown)


vrobles-gee:
Hello
1. Notice I replicated the behavior of PcapNgFileReaderDevice, and it actually re-implements it. It also sets m_CurFilter and m_BpfLinkType wich are defined as class members. If you check both classes, you will notice I mostly mirrored the PcapNgFileReaderDevice code. Is there a reason for them to be different, or is PcapNgFileReaderDevice wrong and should be refactored too?
2. As in #1, I didn't use it because PcapNgFileReaderDevice doesn't use it. Logic is a little different too, and again sets and uses its own class members like m_BpfLinkType, m_BpfInitialized and m_CurFilter. So the same question as #1, is there a reason for them to be different, or should PcapNgFileReaderDevice be refactored too?. Also, notice IPcapDevice::matchPacketWithFilter() is mispelled as matchPakcetWithFilter.
3. Can you suggest or give a little advice on the test? Maybe we can add a writer inside TestPcapNgFileReadWriteAdv to write the 14 packets readed by readerDev5 using a filter, and validate only the ones matching the filter are written?

seladb:
Regarding 1 and 2 - you're right, your way is probably the best way to implement it, although it involves quite a lot of code duplication. Maybe in the future we can find a better way to merge the filtering code of `IPcapDevice`, `PcapNgFileReaderDevice` and `PcapNgFileWriterDevice` but for now let's leave it the way it is.

Regarding unit-tests - I agree with your suggestion, let's go for it.

I'll wait for you to add the unit-test code and then I'll merge the PR

vrobles-gee:
I added a test case for this feature as previously agreed.

vrobles-gee:
I am closing this PR, because I needed to update first commit that was done with wrong author information. It is replaced by #141.

PR #136, opened 2018-10-01
Add DPDK bridge example (analog to l2fwd)

vicenterb:
Adds a basic DPDK bridge example using PcapPlusPlus, analog to l2fwd app.


Examples/DpdkBridge/Common.h
```diff
@@ -0,0 +1,149 @@
...
+	AppWorkerConfig() : CoreId(MAX_NUM_OF_CORES+1), RxDevice(NULL), TxDevice(NULL)
+	{
+	}
+};
+
+
+/**
+ * Collect and analyze packet and flow statistics
+ */
+struct PacketStats
```
seladb: I'm not sure collecting packet stats are needed for this application, what do you think?
If we remove this, we can make the DPDK bridge example much simpler and easier to understand

Examples/DpdkBridge/main.cpp
```diff
@@ -0,0 +1,336 @@
...
+	{
+		EXIT_WITH_ERROR_AND_PRINT_USAGE("DPDK ports list is empty. Please use the -d switch");
+	}
+
+	// extract core vector from core mask
+	vector<SystemCore> coresToUse;
+	createCoreVectorFromCoreMask(coreMaskToUse, coresToUse);
+
+	// need minimum of 2 cores to start - 1 management core + 1 (or more) worker thread(s)
+	if (coresToUse.size() < 2)
```
seladb: Need to check a minimum of 3 cores - 1 for each of the two workers and 1 for management

Examples/DpdkBridge/main.cpp
```diff
@@ -0,0 +1,336 @@
...
+		{
+			EXIT_WITH_ERROR("DPDK device for port %d doesn't exist", *iter);
+		}
+		dpdkDevicesToUse.push_back(dev);
+	}
+
+	// go over all devices and open them
+	for (vector<DpdkDevice*>::iterator iter = dpdkDevicesToUse.begin(); iter != dpdkDevicesToUse.end(); iter++)
+	{
+		if (!(*iter)->openMultiQueues(1,1))
```
seladb: Why limiting for 1 RX queue only? We can have another command-line parameter that indicates how many RX queues to open (default can be 1)

Examples/DpdkBridge/main.cpp
```diff
@@ -0,0 +1,336 @@
...
+		{
+			EXIT_WITH_ERROR("Couldn't open DPDK device #%d, PMD '%s'", (*iter)->getDeviceId(), (*iter)->getPMDName().c_str());
+		}
+	}
+
+	// get DPDK device to send packets to (or NULL if doesn't exist)
+	DpdkDevice* sendPacketsTo = DpdkDeviceList::getInstance().getDeviceByPort(sendPacketsToPort);
```
seladb: This code is not used anywhere, please remove it

Examples/DpdkBridge/main.cpp
```diff
@@ -0,0 +1,336 @@
...
+ */
+int main(int argc, char* argv[])
+{
+	AppName::init(argc, argv);
+
+	std::vector<int> dpdkPortVec;
+
+	CoreMask coreMaskToUse = getCoreMaskForAllMachineCores();
+
+	int sendPacketsToPort = -1;
```
seladb: This is not used anywhere, please remove it

Examples/DpdkBridge/main.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+	// register the on app close event to print summary stats on app termination
+	DpdkBridgeArgs args;
+	args.workerThreadsVector = &workerThreadVec;
+	ApplicationEventHandler::getInstance().onApplicationInterrupted(onApplicationInterrupted, &args);
+
+	// infinite loop (until program is terminated)
+	while (!args.shouldStop)
+	{
+		sleep(5);
```
seladb: You may put `sleep(1)` instead of  `sleep(5)` for better reaction

seladb: I also would print here current traffic stats, consider a different alternative for the loop:

        uint64_t counter = 0;
        int statsCounter = 1;

        // Keep running while flag is on
        while (!args.shouldStop)
        {
                // Sleep for 1 second
                sleep(1);

                // Print stats every COLLECT_STATS_EVERY_SEC seconds
                if (counter % COLLECT_STATS_EVERY_SEC == 0)
                {
                        printf("\n\nStats #%d\n", statsCounter++);
                        printf("==========\n\n");

                        // Print stats of traffic going from Device1 to Device2
                        printf("\nDevice1->Device2 stats:\n\n");
                        printStats(device1, device2);

                        // Print stats of traffic going from Device2 to Device1
                        printf("\nDevice2->Device1 stats:\n\n");
                        printStats(device2, device1);
                }
                counter++;
        }

Here is the implementation of `printStats()`:

```
void printStats(pcpp::DpdkDevice* rxDevice, pcpp::DpdkDevice* txDevice)
{
        pcpp::DpdkDevice::DpdkDeviceStats rxStats;
        pcpp::DpdkDevice::DpdkDeviceStats txStats;
        rxDevice->getStatistics(rxStats);
        txDevice->getStatistics(txStats);

        std::vector<std::string> columnNames;
        columnNames.push_back(" ");
        columnNames.push_back("Total Packets");
        columnNames.push_back("Packets/sec");
        columnNames.push_back("Data");
        columnNames.push_back("Gbps");

        std::vector<int> columnLengths;
        columnLengths.push_back(10);
        columnLengths.push_back(15);
        columnLengths.push_back(15);
        columnLengths.push_back(15);
        columnLengths.push_back(15);

        pcpp::TablePrinter printer(columnNames, columnLengths);

        std::stringstream totalRx;
        totalRx << "rx" << "|" << rxStats.aggregatedRxStats.packets << "|" << rxStats.aggregatedRxStats.packetsPerSec << "|" << rxStats.aggregatedRxStats.bytes << "|" << rxStats.aggregatedRxSta$
        printer.printRow(totalRx.str(), '|');

        std::stringstream totalTx;
        totalTx << "tx" << "|" << txStats.aggregatedTxStats.packets << "|" << txStats.aggregatedTxStats.packetsPerSec << "|" << txStats.aggregatedTxStats.bytes << "|" << txStats.aggregatedTxSta$
        printer.printRow(totalTx.str(), '|');
}

```

Examples/DpdkBridge/AppWorkerThread.h
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		#define MAX_RECEIVE_BURST 64
+		pcpp::MBufRawPacket* packetArr[MAX_RECEIVE_BURST] = {};
+
+		// main loop, runs until be told to stop
+		while (!m_Stop)
+		{
+			// receive packets from network on the specified DPDK device
+			uint16_t packetsReceived = rxDevice->receivePackets(packetArr, MAX_RECEIVE_BURST, 0);
```
seladb: If we decide to let the user specify more than 1 RX queue, we may want to change the 0 value here

Examples/DpdkBridge/Common.h
vicenterb: Agreed, I removed it

Examples/DpdkBridge/main.cpp
vicenterb: Agreed, I fixed it

Examples/DpdkBridge/main.cpp
vicenterb: I limited the example to just 1 RX and TX queue for simplicity. I have added support to specify multiple RX queues per device as requested (and also total available queues information on --list command). Still, all the RX queues are polled by the same worker thread, to avoid making the example more complex (I think the queue distribution logic in the FilterTraffic example will make it harder for someone new to PcapPlusPlus/DPDK to understand the code).

Examples/DpdkBridge/main.cpp
vicenterb: Agreed, I removed it

Examples/DpdkBridge/main.cpp
vicenterb: Agreed, I removed it

Examples/DpdkBridge/main.cpp
vicenterb: Agreed, I added, with some small modifications, to make it more like DPDK's l2fwd example, like grouping stats per device and not per flow (I think actually it more generic, in case we add support for a third device in the future), or cleaning the screen between iterations (is much more easier to follow the execution that way).

Examples/DpdkBridge/AppWorkerThread.h
vicenterb: Agreed, I adjusted it based on the actual RX queue quantity. I also updated it to send the whole buffer in a single call to the TX device queue instead of iterating the packages (is no needed because we are not collecting stats anymore).

PR #142, opened 2018-10-25
Large vlan id bugfix

Lapshin:
FIX for VLAN IDs greater than 255. Little-endian system is under bug.
In old algorithm only 12bit was used for vlan_header instead 16bit.


seladb:
Thanks for this fix!
I went over the code and have some comments:

1. Can you move `EndianPortable.h` from `Packet++/header` to `3rdParty/EndianPortables`?
2. Since EndianPortables is a 3rd-party which may be used in other libraries, could you please move the code in `VlanLayer` that uses it from the H file to the CPP file? That way apps that use PcapPlusPlus will avoid conflicts between this H file and other parts of the app which may also use EndianPortables

Lapshin:
> 3. As you change `Tests/Packet++Test/PacketExamples/ArpRequestWithVlan.dat` to provide a better test case, could you also change `Tests/Packet++Test/PacketExamples/ArpRequestWithVlan.pcap` accordingly?

It is already accordingly changed (see commit a064f8c)
image

Test would not pass, if i could not do this.

seladb:
> 
> 
> > 1. As you change `Tests/Packet++Test/PacketExamples/ArpRequestWithVlan.dat` to provide a better test case, could you also change `Tests/Packet++Test/PacketExamples/ArpRequestWithVlan.pcap` accordingly?
> 
> It is already accordingly changed (see commit [a064f8c](https://github.com/seladb/PcapPlusPlus/commit/a064f8c68b7650193b40ff52beb83493c1819472))
> image
> 
> Test would not pass, if i could not do this.

You're right, I missed it, sorry

Lapshin:
I notice few points:

- Build system is not the best... I think cmake will be better
- Tests have bad organization, all in one file. It should be one *_test.cpp file for each class... And with google test framework it will be looks better.
- It's only the "functional" tests, there are no tests for a class methods..

So if i rewrite this, will you accept this pull requests?

seladb:
Hi, please notice your additional commits fail on VS2015 builds. I think it's because you need to add `3rdParty\EndianPortables` to the VS2015 IncludePath list, please see an example [here for debug-new](https://github.com/seladb/PcapPlusPlus/blob/master/mk/vs2015/Packet%2B%2BTest.vcxproj)

Regarding the points you brought up:

> 
> I notice few points:
> 
>  -   Build system is not the best... I think cmake will be better

I was told that already, but I'd like to keep PcapPlusPlus as common as possible, meaning void adding dependencies which may make it harder for projects to integrate. CMake is a good example: I'd like to provide easy integration for projects who use CMake as well as those who don't use CMake. The easiest way to achieve that is by not using CMake.

>  -   Tests have bad organization, all in one file. It should be one *_test.cpp file for each class... And with google test framework it will be looks better.

I know. It's still in my TODO list, I haven't had a chance to get to it. It'd be great if you can take it! Regarding google test - I'd like to avoid it from the same reason mentioned above - avoid unnecessary dependencies. I'd like to create an experience where people download the code and are able to build it with as less friction as possible. Gtest will definitely make this more complicated. Besides, I don't really need all the framework provided by Gtest, that's why I created this simple macro-based test infra which has all (and only what) PcapPlusPlus actually needs 
>  -   It's only the "functional" tests, there are no tests for a class methods..

I'm not sure what you mean by that. Could you please elaborate?

Lapshin:
> > * It's only the "functional" tests, there are no tests for a class methods..
> 
> I'm not sure what you mean by that. Could you please elaborate?

Tests now check only one case for each class. For example my problem with VLAN could be found early, if it were unit tests for methods getVlanID() and setVlanID().
I mean that tests aren't check all possible variants of input data.

seladb:
> Tests now check only one case for each class. For example my problem with VLAN could be found early, if it were unit tests for methods getVlanID() and setVlanID().
> I mean that tests aren't check all possible variants of input data.

I agree. Some tests aren't as comprehensive as I'd like them to be. You're more than welcome to add more tests as you see fit.

I have one more comment regarding the PR:
Since EndianPortables is a 3rd-party that may be used in other libraries, could you please move the code in `VlanLayer.h` that uses it from to `VlanLayer.cpp` and move `#include <EndianPortable.h>` to `VlanLayer.cpp` as well? That way apps that use PcapPlusPlus will avoid conflicts between this H file and other parts of the app which may also use EndianPortables

Lapshin:
> I have one more comment regarding the PR:

Sorry, I have got it wrong first time you told.. I think now all are done.

PR #148, opened 2018-11-19
Expose start and end times when reassembling a packet, move IPv6Exten…

Lnk2past:
Expose start and end times when reassembling a packet, move IPv6Extension::getNextHeader to public for better control over an instance (akin to pcpp::Layer::getNextLayer).

### DIFF
--- a/Packet++/header/IPv6Extensions.h
+++ b/Packet++/header/IPv6Extensions.h
@@ -59,6 +59,11 @@ namespace pcpp
 		 */
 		virtual ~IPv6Extension();
 
+		/**
+		 * @return A pointer to the next header or NULL if the extension is the last one
+		 */
+		inline IPv6Extension* getNextHeader() { return m_NextHeader; }
+
 	protected:
 
 		struct ipv6_ext_base_header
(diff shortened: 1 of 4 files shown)


Lnk2past:
@echo-Mike for the VS2015 builds complaining about the initialization of the new ```timeval startTime``` and ```timeval endTime``` members I have introduced is there a better way to initialize them in the initializer list or should I perhaps just not bother (e.g. in ```pcpp::RawPacket``` the member ```m_TimeStamp``` is not initialized explicitly)?

echo-Mike:
@Lnk2past You have initialized them in a pure C way. It is now hard for me to think in C++98, Bu I expect that 2 forms of initialization must be accepted by any compiler and done the job:  
1. startTime({}), endTime({})
2. startTime(), endTime()  
If You omit this members in member initializer list they are guaranteed to be default initialized (but only for POD types, timeval as a C structure is a POD type)

Lnk2past:
Ahh, I hadn't even tried the second syntax there. I initially had tried the first way when I was adding them but that syntax is C++11. I figured given that the C++11 features are optional that I would find a different way. Thanks!

Issue #29, opened 2016-05-20
'in.h' file not found #include <in.h>


tristancode:
Hello 
I am running on Mac OS X and getting this probleme 

And have an issue when running the DNSSpoof example file.. Please help
Full Error: 

main.cpp:16:10: fatal error: 'in.h' file not found
#include <in.h>

seladb:
Which Mac OS X version are you running?

tristancode:
This worked the info you gave me fixed my issue.!

seykuyinu:
Hi, 

I'm having the same issue when trying to compile PcapPlusPlus on OSX. I'm running on Yosemite, with Xcode 6.4. The error I get is:

> src/GreLayer.cpp:16:10: fatal error: 'in.h' file not found
> #include <in.h>

I've tried re-installing X-code developer tools, as well as using a different installation of g++, but no luck. 

Any help would be appreciated!

seladb:
Can you please send me the entire compilation output until the error occurs?
Please run "make clean" before you run "make all"

seykuyinu:
Here it is: [compilation.txt](https://github.com/seladb/PcapPlusPlus/files/380416/compilation.txt)

seladb:
Did you run:
`./configure-mac_os_x.sh`
before you started compiling? If not, please run this configuration script, then run "make clean" and "make all". 
If you did run it, please check if you have the following directory: /usr/include/netinet
If you don't have it, it means the XCode developer tools weren't installed correctly for some reason

seladb:
Fixed in [this commit](https://github.com/seladb/PcapPlusPlus/commit/2b8b254085bdaf8b9048461c81d470d94bfb9c9f)

PR #153, opened 2018-12-23
Resolving issues 'in.h" file not found for MacOS

Deanseo:
This is an extended fix from the '[in.h' file not found #include <in.h>](https://github.com/seladb/PcapPlusPlus/issues/29)'

No implementation is touched; this PR only contains modification against the order of `#include` and adding `DEPS` in a few `Makefile`s.

Tested in the following env:
- Windows 7 and Visual Studio 2017
- MacOS(10.14) with _apple-clang++_
- RHEL 7.5 with _g++ 4.8.5_

### DIFF
--- a/Examples/ArpSpoofing/main.cpp
+++ b/Examples/ArpSpoofing/main.cpp
@@ -2,6 +2,13 @@
 #include <stdlib.h>
 #include <fstream>
 #include <memory>
+#if defined(WIN32) || defined(WINx64) // for using ntohl, ntohs, etc.
+#include <winsock2.h>
+#elif LINUX
+#include <in.h>
+#elif MAC_OS_X
+#include <arpa/inet.h>
+#endif
 #include <MacAddress.h>
 #include <IpAddress.h>
 #include <PcapPlusPlusVersion.h>
@@ -12,13 +19,6 @@
 #include <EthLayer.h>
 #include <ArpLayer.h>
 #include <Logger.h>
-#ifdef WIN32 // for using ntohl, ntohs, etc.
-#include <winsock2.h>
-#elif LINUX
-#include <in.h>
-#elif MAC_OS_X
-#include <arpa/inet.h>
-#endif
 #include <getopt.h>
 
 using namespace std;
--- a/Examples/ArpSpoofing/main.cpp
+++ b/Examples/ArpSpoofing/main.cpp
@@ -2,9 +2,6 @@
 #include <stdlib.h>
 #include <fstream>
 #include <memory>
-#if defined(WIN32) || defined(WINx64)
-#include <winsock2.h>
-#endif
 #include <MacAddress.h>
 #include <IpAddress.h>
 #include <PcapPlusPlusVersion.h>
@@ -15,8 +12,12 @@
 #include <EthLayer.h>
 #include <ArpLayer.h>
 #include <Logger.h>
-#if !defined(WIN32) && !defined(WINx64) //for using ntohl, ntohs, etc.
+#ifdef WIN32 // for using ntohl, ntohs, etc.
+#include <winsock2.h>
+#elif LINUX
 #include <in.h>
+#elif MAC_OS_X
+#include <arpa/inet.h>
 #endif
 #include <getopt.h>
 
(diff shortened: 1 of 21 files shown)


seladb:
Thanks for your contribution. I reviewed the changes you've made and I'm not sure what made you do these changes. Could you please elaborate on the errors you got?

Deanseo:
> Thanks for your contribution. I reviewed the changes you've made and I'm not sure what made you do these changes. Could you please elaborate on the errors you got?

When I clone the current `origin/master` and build it in my MacOS whose version is mentioned in the initial comment, the following errors come and thus the building fails:

> fatal error: 'in.h' file not found

which is the same issue as #29.

I mentioned Linux and Windows because I wanted to make sure the change does not prevent building in those platforms from failing because of the changes I've made.

seladb:
You should be able to build PcapPlusPlus on MacOS without these changes.
Did you follow the instructions on "Building PcapPlusPlus - MacOS" in [this page](https://seladb.github.io/PcapPlusPlus-Doc/download.html)?

For example, do you have Xcode Command Line Tools installed?

seladb:
What is the MacOS and Xcode version you're using

Deanseo:
> What is the MacOS and Xcode version you're using

- Xcode version : Version 10.1 (10B61)
- MacOS(10.14, **Mojave**) with apple-clang++ 1000.10.44.4 (As stated in the initial comment)

Thanks in advance.

seladb:
I compile PcapPlusPlus with Xcode version10.1 (10B61) and it works just fine, although I have MacOS 10.13 and not 10.14. I wonder if the new 10.14 is the problem.
Could you please list the files in `/usr/include/netinet`? Do you see `in.h` there?

seladb:
> So then you must be able to build all tutorial examples under Examples/Tutorials also?
> For example, can you build Tutorial-PacketParsing without any errors?

Yes, I can build `Tutorial-PacketParsing`. But please note that before building it you need to successfully build PcapPlusPlus and then run `sudo make install`

> No, I don't think /usr/include/netinet is a valid path to look for headers at in MacOS anymore as updated with newer versions. Usually that path's just for most of Linux distributions only? which I'm not entirely sure of though.

You should have the folder `/usr/include/netinet` in your MacOS system. I'm not sure but I think that's part of the Xcode Command Line Tools package (which you install by running `xcode-select --install`)

Deanseo:
> You should have the folder /usr/include/netinet in your MacOS system. I'm not sure but I think that's part of the Xcode Command Line Tools package (which you install by running xcode-select --install)

I'll do a bit of researching later and let you know what causes me to fail to build in a certain environment, and let you know if I find something.

Otherwise I'll close the PR later by myself.

shihabawal:
I can confirm @Deanseo comment about the `/usr/include/netinet` not being a valid path any more.

I have also faced the same issue with the `in.h` library. But in my case I did find the library in 2 different locations

- `    /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/netinet`

- `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Kernel.framework/Versions/A/Headers/netinet`

Hope this helps

seladb:
I just found this post: https://apple.stackexchange.com/questions/337940/why-is-usr-include-missing-i-have-xcode-and-command-line-tools-installed-moja

It seems you need to run an additional step after installing the command line tools:
```
installer -pkg /Library/Developer/CommandLineTools/Packages /macOS_SDK_headers_for_macOS_10.14.pkg -target /
```

Could you please try that and let me know? If it works I'll update the documentation accordingly

seladb:
Also, could you please open a GitHub issue for better documentation of this issue?

Deanseo:
@seladb 

So I ended up reading [Xcode 10 Release Notes](https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes)
and found out that `/usr/include` is no longer the default path from the SDK in **Command Line Tools - New Features**

The second not-marked answer from [your link](https://apple.stackexchange.com/a/338150/223298) also says the same thing.

> could you please open a GitHub issue for better documentation of this issue?
I'll create a corresponding issue for you.

PR #155, opened 2019-01-12
Small fixes to HTTP printMethods

ezpz:
- Move stringstream to local scope to clear on each iteration
- Add missing printRow for POST method


seladb:
Good catch, thanks!

PR #161, opened 2019-02-22
enable override of deps for package managers by moving dep references…

solvingj:
Dependency management in visual studio projects is best managed through the use of props files, and by keeping all dependency-specific variables and details out of `vcxproj` files. This PR makes use of the existing `pcappluspluspropertysheet.props` by moving all dependency references from `.vcxproj` files into it. More accurately, it puts these references in the `props` file TEMPLATE, and they get copied into the actual `props` file when the `...configure.bat` script is run. *  

Crucially, now that all dependency information is out of the `vcxproj` file and injected via the `props` file, users can use package managers to inject the paths and other build variables for the dependencies into the project without any patching modification of the `vcxproj` files, and without having to run the `...configure.bat` script.   Users simply have to set or pass the following two variables for this to work: 

    PCAPPLUSPLUS_CONFIG_OVERRIDE=True
    PCAPPLUSPLUS_HOME=path_to_pcapplusplus_repo

 These can be set as environment variables or as MSBuild CLI properties such as: 

    /p:PCAPPLUSPLUS_CONFIG_OVERRIDE=True /p:PCAPPLUSPLUS_HOME=C:\PCapPlusPlus

On a minor note, the `pcappluspluspropertysheet.props` file which is actually committed in the repository is now a valid `.props` file, so that the Solution can be successfully opened in Visual Studio. It now simply raises an MSBuild "Error" which tells the user inside the Visual Studio GUI that they need to run the `...configure.bat` script, or use the alternative method described above in order to use the solution. 

* _Note: Only dependency references related to compiling and linking were moved.  Dependency references relating to post-build steps were not modified as part of this commit, and so most are presumably broken.  The next step would be discussing options for resolving those._

### DIFF
--- a/mk/vs2015/ArpSpoofing.vcxproj
+++ b/mk/vs2015/ArpSpoofing.vcxproj
@@ -100,7 +100,7 @@
       <Optimization>Disabled</Optimization>
       <PreprocessorDefinitions>WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <SDLCheck>true</SDLCheck>
-      <AdditionalIncludeDirectories>$(PcapPlusPlusHome)\Dist\header;$(PcapPlusPlusHome)\3rdParty\Getopt-for-Visual-Studio;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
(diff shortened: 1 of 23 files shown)


solvingj:
Looks like the bat script doesn't handle the new addition to the template.  Will fix shortly.

PR #184, opened 2019-04-23
Improvements in class Packet

gx740:
The only way to parse the packet is the using the class constructor. In many cases the total parsing is inappropriate. Currently there is no way to continue parsing the packet previously stopped on some layer. These new methods allow to have full control over parsing.
The approach with step-by-step parsing gives performance improvements.

### DIFF
--- a/Packet++/header/Packet.h
+++ b/Packet++/header/Packet.h
@@ -261,6 +261,21 @@ namespace pcpp
 		TLayer* getNextLayerOfType(Layer* after);
 
 		/**
+		 * A method makes attempt to parse next layer.
+		 * @return A pointer to the parsed layer, NULL if no more layer exists
+		 */
+		Layer* parseNextLayer();
+
+		/**
+		 * A method parses the next layers.
+		 * @param[in] parseUntil Parse the packet until it reaches this protocol. Can be useful for cases when you need to parse only up to a certain layer and want to avoid the
+		 * performance impact and memory consumption of parsing the whole packet. Default value is ::UnknownProtocol which means don't take this parameter into account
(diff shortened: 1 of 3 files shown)


seladb:
I'm sorry for the late response. I'm thinking maybe we should create a separate class for this gradually parsing of packets, please let me know what you think

seladb:
@gx740 I'm not sure if you saw my previous comment. Let's discuss and decide on the right approach

gx740:
I'm sorry for the late response. Do you have some idea of this class? What relationship should be between it and the class Packet? Should this class be an internal not included in interface?
Just I do not see how this class should be built in the current class hierarchy and what advantages it will give. Please describe how do you see this class and its place in class hierarchy.

seladb:
Thanks for your reply.

I'm thinking of one of 2 options:
1. A builder/iterator class that takes a `RawPacket` or `Packet` instance and emits (the same) `Packet` instance with more layers parsed
2. Instead of creating a separate class, just put them as static methods inside of `Packet` class

gx740:
I have no wide experience (about a year) in design of C++ classes but in my view the first approach is preferable. At least it will allow to simplify the class Packet by moving the parsing code into the new class (and files respectively).

seladb:
Yes, I agree. Although moving the parsing logic to a different class may create more complexity in the "standard" usage of `Packet` where you have a `RawPacket` and you want to create a parsed packet out of it.

That's why I'm suggesting having a builder/iterator class that only handles incremental parsing.

seladb:
I'd like to follow up on that. Any thoughts on how to move forward with this?

seladb:
Since I didn't see any response from you I'll close the PR for now. Please reopen it if needed

PR #186, opened 2019-04-24
Many improvements in class MacAddress

gx740:
- The implementation of many constructors and methods have been moved into header file, the implementation of method `operator==` is simplified and optimized.

- Two errors in method `init()` were fixed:
  1. The buffer overflow occures when the length of input argument is more than 6.
  2. For incorrect MAC address string that contains the wrong hexadecimal number the old version of this method is setting `m_IsValid` to true value.

- The qualifier `const` was added to many methods

- The operator of assignment and the copy constructor are unnecessary.

- The new constructor `MacAddress(std::initializer_list<uint8_t>)` was added for the compilers which are supporting C++11 language standard.

- Operator `'sizeof m_Address'` is now used in all cases instead of number 6 to avoid the potential mistakes with buffer overflow.

- Tests were added.

### DIFF
--- a/Common++/header/MacAddress.h
+++ b/Common++/header/MacAddress.h
@@ -67,7 +67,7 @@ namespace pcpp
 		MacAddress(std::initializer_list<uint8_t> octets) : m_IsValid { octets.size() == sizeof m_Address }
 		{
 			if(m_IsValid)
-				std::copy_n(std::cbegin(octets), sizeof m_Address, std::begin(m_Address));
+				std::copy(octets.begin(), octets.end(), std::begin(m_Address));
 		}
 #endif
 
--- a/Common++/header/MacAddress.h
+++ b/Common++/header/MacAddress.h
@@ -2,8 +2,13 @@
 #define PCAPPP_MACADDRESS
 
 #include <stdint.h>
+#include <string.h>
 #include <string>
-#include <memory>
+
+#if __cplusplus > 199711L
+#include <initializer_list>
+#include <algorithm>
+#endif
 
 /// @file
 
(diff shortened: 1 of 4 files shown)


Tests/Packet++Test/main.cpp
```diff
@@ -7113,6 +7113,50 @@ PACKETPP_TEST(RadiusLayerEditTest)
 	PACKETPP_TEST_PASSED;
 }
 
+PACKETPP_TEST(MacAddressTests)
```
seladb: There is already existing UT for `MacAddress` in Pcap++Test: https://github.com/seladb/PcapPlusPlus/blob/7bc25adc8f5e2c0dc59710a3fa94b6956498a298/Tests/Pcap%2B%2BTest/main.cpp#L663

Can you move these UTs to there?

Common++/header/MacAddress.h
```diff
@@ -52,61 +57,70 @@ namespace pcpp
...
 		 */
-		std::string toString();
+		std::string toString() const;
 
 		/**
 		 * Allocates a byte array of length 6 and copies address value into it. Array deallocation is user responsibility
 		 * @param[in] arr A pointer to where array will be allocated
 		 */
-		void copyTo(uint8_t** arr);
+		void copyTo(uint8_t** arr) const;
```
seladb: If moving most of the method implementations to the header file, why not moving the implementation of this method as well?

gx740: I'll do it

Tests/Packet++Test/main.cpp
gx740: Ok

Common++/header/MacAddress.h
gx740: Done

Tests/Packet++Test/main.cpp

PR #189, opened 2019-04-27
Minor changes

gx740:
- MacAddress: bugfix in assignment operator (thanks @eteran )
- Qualifier 'const' was added to many methods


gx740:
A `inline` specifier is unneccessary when the definition of the method is included into the class declaration because such behavior is by default.

seladb:
I agree. I'll merge it now

PR #195, opened 2019-05-17
Cmake

eteran:
update to add cmake support


cmake/Modules/FindDPDK.cmake
```diff
@@ -0,0 +1,59 @@
...
+# Variables defined by this module:
+#
+#  DPDK_FOUND                System has DPDK include and library dirs found
+#  DPDK_INCLUDE_DIRS         The DPDK include directories.
+#  DPDK_LIBRARIES            The DPDK library
+
+find_package(Threads)
+find_package(NUMA)
+
+find_library(DPDK_LIBRARIES dpdk
```
echo-Mike: I think that this is incorrect, this find_library function only adds libdpdk to build but DPDK have 20+ other libraries that must be added (most of them with -Wl whole_archive, and some system libraries too) to compile. I can send you an  FindDPDK file that I use for projects as an example

seladb:
Some of the unit-tests are failing but I'll merge the PR anyway so we can all work on it together

eteran:
I'll check on the failing unit tests (I'm assuming you tested with `make test`)?

I didn't test after syncing with upstream, so that's my bad.

cmake/Modules/FindDPDK.cmake
eteran: I had looked at the original makefile and saw a bunch of the libs, but it does seem that libsdpdk.a depends on pretty much everything else. So it seems like just pulling that in should be sufficient.

echo-Mike:
> I'll check on the failing unit tests (I'm assuming you tested with `make test`)?
> 
> I didn't test after syncing with upstream, so that's my bad.

This failings are reoccuring issue on Mac and Xcode, I have same troubles with my current work (even on  not changed master branch). So I think this fails are negligible.

eteran:
Oh, I see. 

I think I remember seeing a warning with gcc-9 about the possibility of a string being truncated and not properly nul terminated. Perhaps that could cause this fail:

    TestIPAddress                 : FAILED. assertion failed: IPv6 toString doesn't return the correct string

echo-Mike:
@eteran can you please report what you have seen (like warnings and such) because i tried gcc-7.3.0 with: 

> -Wextra -Wshadow -Wnon-virtual-dtor -Wold-style-cast \
>       -Wcast-align -Wunused -Woverloaded-virtual -Wconversion \
>       -Wsign-conversion -Wmisleading-indentation -Wduplicated-cond \
>       -Wduplicated-branches -Wlogical-op -Wnull-dereference \
>       -Wuseless-cast -Wdouble-promotion -Wformat=2 \
>       -Weffc++ -Wpedantic

and got nothing on this lines (641/655 in main.cpp)

eteran:
I only saw it with gcc-9+ when I'm back at my workstation, I'll report it.

IIRC, it involved a `strncpy` which did not guarantee null termination.

echo-Mike:
If someone want to help to find the issue there is a full [warning log](https://yadi.sk/d/6QAqnTvHrTZKiQ) 33k lines (on my personal Yandex.Disk)
A lot of useful warnings may lead to enhancements in PcapPlusPlus library
@eteran @seladb

seladb:
Thanks @echo-Mike , this is very helpful! Maybe we should open a new issue for tracking these warnings. I saw a lot of warning so it might take a while to take care of all of them.

I suggest we keep this thread focused on cmake. Please let me know your concerns.

@eteran could you please update this thread on which platforms you tested already using cmake and what is the next step?

One thing I will try to take care of is adding cmake builds into travis-ci (and later to appveyor as well). Please let me know what additional help is needed.

eteran:
@seladb  Tested on Linux, I'd like to improve flag detection for DPDK usage. Right now it simply hard codes `-msse3`, but it would be better if it tested settings to find what that particular machine's libdpdk install was built with.

Next steps, seem about building on macOS, it should nearly (if not already) "just work"... but I haven't tested yet.

eteran:
@echo-Mike 

Here are the warnings I see with gcc-9 that seem possibly relevant to regression tests failing on string routines:

```
In file included from /usr/include/string.h:494,
                 from /home/eteran/projects/PcapPlusPlus/Common++/src/IpAddress.cpp:6:
In function ‘char* strncpy(char*, const char*, size_t)’,
    inlined from ‘pcpp::IPv6Address::IPv6Address(const pcpp::IPv6Address&)’ at /home/eteran/projects/PcapPlusPlus/Common++/src/IpAddress.cpp:175:12:
/usr/include/bits/string_fortified.h:106:34: warning: ‘char* __builtin_strncpy(char*, const char*, long unsigned int)’ output may be truncated copying 39 bytes from a string of length 39 [-Wstringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

```
/home/eteran/projects/PcapPlusPlus/Packet++/src/Packet.cpp: In member function ‘std::string pcpp::Packet::printPacketInfo(bool)’:
/home/eteran/projects/PcapPlusPlus/Packet++/src/Packet.cpp:670:34: warning: ‘%06lu’ directive output may be truncated writing between 6 and 20 bytes into a region of size between 0 and 63 [-Wformat-truncation=]
  670 |   snprintf(buf, sizeof(buf), "%s.%06lu", tmbuf, timestamp.tv_usec);
      |                                  ^~~~~
/home/eteran/projects/PcapPlusPlus/Packet++/src/Packet.cpp:670:30: note: using the range [0, 18446744073709551615] for directive argument
  670 |   snprintf(buf, sizeof(buf), "%s.%06lu", tmbuf, timestamp.tv_usec);
      |                              ^~~~~~~~~~
In file included from /usr/include/stdio.h:867,
                 from /home/eteran/projects/PcapPlusPlus/Packet++/header/Layer.h:5,
                 from /home/eteran/projects/PcapPlusPlus/Packet++/header/Packet.h:5,
                 from /home/eteran/projects/PcapPlusPlus/Packet++/src/Packet.cpp:3:
/usr/include/bits/stdio2.h:67:35: note: ‘__builtin___snprintf_chk’ output between 8 and 85 bytes into a destination of size 64
   67 |   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   68 |        __bos (__s), __fmt, __va_arg_pack ());
      |        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

```
In file included from /usr/include/string.h:494,
                 from /home/eteran/projects/PcapPlusPlus/Common++/header/MacAddress.h:5,
                 from /home/eteran/projects/PcapPlusPlus/Packet++/header/ArpLayer.h:6,
                 from /home/eteran/projects/PcapPlusPlus/Pcap++/header/PcapFilter.h:8,
                 from /home/eteran/projects/PcapPlusPlus/Pcap++/header/Device.h:8,
                 from /home/eteran/projects/PcapPlusPlus/Pcap++/header/PcapDevice.h:4,
                 from /home/eteran/projects/PcapPlusPlus/Pcap++/header/PcapLiveDevice.h:5,
                 from /home/eteran/projects/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:4:
In function ‘char* strncpy(char*, const char*, size_t)’,
    inlined from ‘void pcpp::PcapLiveDevice::setDeviceMtu()’ at /home/eteran/projects/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:630:9:
/usr/include/bits/string_fortified.h:106:34: warning: ‘char* __builtin_strncpy(char*, const char*, long unsigned int)’ specified bound 16 equals destination size [-Wstringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function ‘char* strncpy(char*, const char*, size_t)’,
    inlined from ‘void pcpp::PcapLiveDevice::setDeviceMacAddress()’ at /home/eteran/projects/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:686:9:
/usr/include/bits/string_fortified.h:106:34: warning: ‘char* __builtin_strncpy(char*, const char*, long unsigned int)’ specified bound 16 equals destination size [-Wstringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |        
~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

eteran:
In particular, this line is **highly** suspect:

    strncpy(m_AddressAsString, other.m_AddressAsString, MAX_ADDR_STRING_LEN-1);

It really should be the following:

    strncpy(m_AddressAsString, other.m_AddressAsString, MAX_ADDR_STRING_LEN);
    m_AddressAsString[MAX_ADDR_STRING_LEN - 1] = '\0';

To ensure NUL termination.

echo-Mike:
@eteran I aggre with the source of the issue - by design the IP string is storred as not NUL-terminated.
So it can't be checked against NUL-terminated string by the means of `std::strcmp` as it done in tests as it done in next places:
https://github.com/seladb/PcapPlusPlus/blob/e8d79712d779dbeeb0887d8bd5221e9263a94407/Tests/Pcap%2B%2BTest/main.cpp#L641
and
https://github.com/seladb/PcapPlusPlus/blob/e8d79712d779dbeeb0887d8bd5221e9263a94407/Tests/Pcap%2B%2BTest/main.cpp#L655
And the source of the issue is:
https://github.com/seladb/PcapPlusPlus/blob/e8d79712d779dbeeb0887d8bd5221e9263a94407/Common%2B%2B/header/IpAddress.h#L68
It happens because as specified in [LINK](https://en.cppreference.com/w/cpp/string/basic_string/basic_string) the (5) constructor of `std::basic_string` is called but it expects a NUL-terminated C-string.
@seladb yep, I think this small issue must be directed to other thread(as an issue thread). Do You mind to do it?

seladb:
@echo-Mike @eteran let me take care of this small issue. I think the solution suggested by @eteran should be sufficient. Thanks guys for uncovering that bug!

@eteran regarding finding the DPDK library - there is no single library called `libdpdk`. DPDK is actually a bunch of libraries, all of them starts with `rte_*`. Currently PcapPlusPlus uses the following DPDK libs:
```
-lrte_pmd_bond -lrte_pmd_vmxnet3_uio -lrte_pmd_virtio -lrte_pmd_enic -lrte_pmd_i40e -lrte_pmd_fm10k -lrte_pmd_ixgbe -lrte_net -lrte_pmd_e1000 -lrte_pmd_ring -lrte_pmd_af_packet -lrte_ethdev -lrte_eal -lrte_mbuf -lrte_mempool -lrte_ring -lrte_kvargs -lrte_hash -lrte_cmdline
```

And for version >= 17.11 also:
```
-lrte_pci -lrte_bus_pci -lrte_bus_vdev -lrte_mempool_ring
```

echo-Mike:
@seladb regarding finding the DPDK library
Yep there is a single library called libdpdk and it's content is as follows:
image
I expect that it is a special static-library like file that must be integrated in build systems for easier builds (but i don't found any confiramtion of it in DPDK docs). So I think it is not enougth to simply link with this file, but the builds that @eteran have done telling the opposite.
I am not really good at linkers and such so i can't tell no more.

seladb:
Thanks @echo-Mike I wasn't aware of that. Is this something that was added at later versions?

If this single lib actually contains all of DPDK libs than I think it's fine linking with it.

PR #204, opened 2019-06-17
Add support for setting direction.

MrSiz:
I've already tested it, and it worked. Please review my code. Thanks.


Pcap++/header/PcapLiveDevice.h
```diff
@@ -174,6 +174,12 @@ namespace pcpp
 			 */
 			int packetBufferSize;
 
+            /**
+             * Set the direction for capture packets. You can read more here:
+             * https://www.tcpdump.org/manpages/pcap.3pcap.html.
+             */
+            pcap_direction_t direction;
```
seladb: can you please replace `pcap_direction_t` with a structure defined in this file? wherever possible I'd like to avoid using libpcap structs in header files to make it easier for apps/libs to link with PcapPlusPlus

seladb: Please use identation which is similar to the rest of the files (tabs in this case)

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -261,6 +261,17 @@ pcap_t* PcapLiveDevice::doOpen(const DeviceConfiguration& config)
 	}
 #endif
 
+#ifdef HAS_DIRECTION_SET
+    ret = pcap_setdirection(pcap, config.direction);
```
seladb: is the default value `PCAP_D_INOUT`? meaning, is not using `pcap_setdirection` similar to use it with `PCAP_D_INOUT`? I want to make sure everything is backward compatible

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -261,6 +261,17 @@ pcap_t* PcapLiveDevice::doOpen(const DeviceConfiguration& config)
 	}
 #endif
 
+#ifdef HAS_DIRECTION_SET
+    ret = pcap_setdirection(pcap, config.direction);
+	if (ret == 0)
+	{
+	    LOG_DEBUG("Direction is set");
```
seladb: maybe it's worth logging the direction itself also (IN, OUT, INOUT)

configure-linux.sh
```diff
@@ -16,7 +16,7 @@ function HELP {
    echo "  1) Without any switches. In this case the script will guide you through using wizards"
    echo "  2) With switches, as described below"
    echo ""
-   echo -e "Basic usage: $SCRIPT [-h] [--pf-ring] [--pf-ring-home] [--dpdk] [--dpdk-home] [--use-immediate-mode] [--install-dir] [--libpcap-include-dir] [--libpcap-lib-dir]"\\n
+   echo -e "Basic usage: $SCRIPT [-h] [--pf-ring] [--pf-ring-home] [--dpdk] [--dpdk-home] [--use-immediate-mode] [--direction-set-enable] [--install-dir] [--libpcap-include-dir] [--libpcap-lib-dir]"\\n
```
seladb: please add the same changes to `configure-mac-os-x.h` as well

Pcap++/src/PcapLiveDevice.cpp
MrSiz: Yes, it is.

Pcap++/header/PcapLiveDevice.h
MrSiz: I will do it.

MrSiz: So I should create a structure like
```
struct PcapThread
{
    pthread_t pthread;
};
```
So it can be
```
struct PcapDirection
{
    pcap_direction_t direction.
}
```
And could you please give me some advice on how to write a unit-tests for this? I just know I can test it with `tcpdump`, because I can compare the results got from `tcpdump` with it. But I don't  know how to do this in a test function.

Pcap++/src/PcapLiveDevice.cpp
MrSiz: Ok. I'll do it.

Pcap++/src/PcapLiveDevice.cpp
seladb: Awesome thanks for verifying that

Pcap++/header/PcapLiveDevice.h
seladb: Yes I think defining a new enum like `PcapDirection` is a better approach. You can then define a method in `PcapLiveDevice.cpp` to convert between this enum and `pcap_direction_t`.

Regarding unit-tests - maybe you can define a unit-tests like this: https://github.com/seladb/PcapPlusPlus/blob/bf438ab7a08a5aac995557f5579a4459adf7a990/Tests/Pcap%2B%2BTest/main.cpp#L1473

Set the configuration to "inflow" and check that all the packets captured are indeed "inflow"

MrSiz: Yes, that's good. But the second step was the key to my confusion. How to check that all the packets captured are "inflow"? I only know that I can compare the results with `tcpdump`, but I don't know how to do this in a test function.

seladb: Yes, the enum definition looks ok. Let's just make sure to clearly mention in the documentation of the enum and `DeviceConfiguration` that this feature is only supported in newer libpcap versions.

Regarding the unit-tests: is it possible to identify inflow packets by MAC address (meaning check that all captured packets have the NIC's MAC address in the dest MAC field)?

MrSiz: Thank you for your advice! 
I read the source file of `libpcap`. `pcap_setdirection` was supported since `libpcap-0.9.1`.
I got all the source code from this [website](https://www.tcpdump.org/release/), And `pcap_setdirection` is defined in the file `pcap.c`.

I will complete these modifications as soon as possible.

MrSiz: Regarding the unit-tests, I found we could't identify inflow packets by MAC address, because some incoming packets captured by the NIC have different MAC address from the NIC's MAC address, though the program works in normal mode rather than promiscuous mode.

seladb: Got it. So maybe you can just add a simple unit-test that sets "inflow" in `DeviceConfiguration` and starts capturing packets. It will not test the actual feature but at least it'll check that packet capture is still working.

By the way, I don't see the additional changes on the PR. Please make sure to commit them to the same branch you created the PR.

Also, could you please make these changes on a fork from [dev](https://github.com/seladb/PcapPlusPlus/tree/dev) branch instead of [master](https://github.com/seladb/PcapPlusPlus)?

MrSiz: Ok.  I  will do  it  as  soon  as possible.

seladb:
Merged duplicate PR #208

PR #215, opened 2019-07-22
Fixes possible memory leak in text-based protocols

JasperEm:


seladb:
Could you please explain the edge case you ran into that led you create this PR?

Basically the current code runs through all the fields until either `\n\n`, `\r\n\r\n` or the end of the packet are seen. Why in addition you need to limit it to 100 iterations?

JasperEm:
I analyze flow samples with only the first 120 bytes of the sampled packet and in one of millions of packets the ``TextBasedProtocolMessage::parseFields`` method allocates infinite memory. My short analysis resulted in the while codition not taking effect in case of a broken packet. I have set 100 iterations as limit, because this limit is "maybe" only reached when the packet is corrupted.

seladb:
Can you attach an example pcap file that show this behavior?

JasperEm:
unfortunately no they are productive data, which I cannot pass on. I am sorry.

Packet++/src/TextBasedProtocol.cpp
```diff
@@ -127,6 +128,11 @@ void TextBasedProtocolMessage::parseFields()
 			delete newField;
 			break;
 		}
+		if (iterations == 100)
+    {
```
seladb: can you please use tabs instead of spaces to align with the rest of the file?

Packet++/src/TextBasedProtocol.cpp
```diff
@@ -108,6 +108,7 @@ void TextBasedProtocolMessage::parseFields()
 	// last field can be one of:
 	// a.) \r\n\r\n or \n\n marking the end of the header
 	// b.) the end of the packet
+	int iterations = 0;
```
seladb: it'd be useful to add a comment which explains why we need to limit to 100 iterations

echo-Mike:
Hi @JasperEm.
I think that Your PR is an awful fixture that do not really solves the problem.
You have mentioned that the packet that causes the issue is an illformed one. What does _illformed_ actually means in Your case?
This is not a fix of some bug but just a crutch to get work done. It is not a good programming style anywhere in the world.
Please __identify__ the problem first. Write an __issue__ report. And then propose a __fix__ for the issue. This is the workflow that is appreciated.
If You provide the bindump of the packet caused an issue then we (the community) can help You with it.

Packet++/src/TextBasedProtocol.cpp
JasperEm: should be changed

Packet++/src/TextBasedProtocol.cpp
JasperEm: I have added a comment and a reference to the Apache documentation of the LimitRequestFields Directive.

seladb:
@JasperEm I do agree with @echo-Mike that it'd be beneficial to really understand what is causing this issue. I know you cannot provide pcap files, but can you explain exactly why this is happening?

@echo-Mike I think that using terms like "awful fixture" or "crutch to get work done" is a little too harsh. I'm sure there is a real issue here, let's work together with @JasperEm to get to the bottom of it.

echo-Mike:
I am sorry if it sound to harsh. I am just want to pinpoint that no one at current time knows the source of an issue. So the fix looks unargumented for me.
If I can help to resolve it then I'd like to do it.

JasperEm:
@echo-Mike it's okay you're right :smiley: , I don't like this fix either, but I didn't have enough time at the moment to go that deep and want to solve the problem quickly. 

I'll give a brief introduction on how I' m using your project. I analyze flow samples and create a ``pcpp::RawPacket`` instance for each sampled package and from this one ``pcpp::Packet```. The sampled packets have a fixed maximum size and it is possible that the application layer headers contained in the payload are not complete because they exceed the maximum size. At the moment i using the following code with my ugly fix and it works.
```cpp
  // create buffer
  auto buf = new uint8_t[MAX_PCAP_PKT_SIZE];
  // copy raw header into the buffer
  memcpy(buf, sample->s.header, sample->s.headerLen);

  pcpp::RawPacket p;
  // the raw data is set in this way, since the sample packet size 
  // may differ from the header length
  p.setRawData(buf, sample->s.headerLen, t,
               static_cast<pcpp::LinkLayerType>(pcpp::Ethernet), sample->s.sampledPacketSize);
  // without fix everthing is fine to this line
  pcpp::Packet parsedPacket(&p);
  // without Fix this line is in some cases never reached and the operating system
  // terminates the process with a out of memory message
```

seladb:
Thanks @JasperEm but unfortunately this code doesn't reveal the reason for this issue. Is there any option you can give us more information?

seladb:
Please let know if we should investigate this issue further or close the PR for now

JasperEm:
@seladb sorry for my slow reaction. i have a lot to do and will try to investigate the problem in more detail early next week and will report about it here.

JasperEm:
@seladb At the moment I haven't found time to analyze the problem more closely, so I will close the PR and maybe create a new one in the future. Thank you very much for your help and forgiveness for wasting time!

Issue #218, opened 2019-07-29
User macro and pre + post build events issues in VS. 


Dysl3xik:
Hello I recently submitted you a PR to fix an issue with your bat file to update the property sheets with correct paths to required libs.

I have now fully compiled your library and found a few other minor issues.

1. What is the purpose of the pre-build event fetch-git-info.bat? It does not seem to be properly configured and if I had to guess is something leftover from your own workflow?

2. The post build events again had an issue with spaces in the path, I added some more quotes to fix the issue and will submit a PR.

3.  Atleast for VS2017 I had to further modify the prop sheets to tell tell it the user macros exist. I will submit a PR for this but I just updated the file as follows:

```
  <PropertyGroup Label="UserMacros">
    <PcapPlusPlusHome>PUT_PCAPPLUSPLUS_HOME_HERE</PcapPlusPlusHome>
    <WinPcapHome>PUT_WIN_PCAP_HOME_HERE</WinPcapHome>
    <PThreadWin32Home>PUT_PTHREAD_HOME_HERE</PThreadWin32Home>
  </PropertyGroup>
  <!-- Tell VS to use these macros -->
  <ItemGroup>
    <BuildMacro Include="PcapPlusPlusHome">
      <Value>$(PcapPlusPlusHome)</Value>
    </BuildMacro>
	<BuildMacro Include="WinPcapHome">
      <Value>$(WinPcapHome)</Value>
    </BuildMacro>
	<BuildMacro Include="PThreadWin32Home">
      <Value>$(PThreadWin32Home)</Value>
    </BuildMacro>
  </ItemGroup>
```

seladb:
Thanks @Dysl3xik for making these changes! I just merged your PR. I tested it on VS2015 and it seems to be working.

Regarding `fetch-git-info.bat` - it's being used in the following APIs:

- https://github.com/seladb/PcapPlusPlus/blob/236e035be8d7109082e1b599abd435269f13032d/Common%2B%2B/header/PcapPlusPlusVersion.h#L39
- https://github.com/seladb/PcapPlusPlus/blob/236e035be8d7109082e1b599abd435269f13032d/Common%2B%2B/header/PcapPlusPlusVersion.h#L44

Their purpose of these APIs is to expose the Git info. For all platforms that have a Makefile (which is all of them except for MSVC) I put the code that fetches the Git info in `Common++` Makefile:

[...]

PR #219, opened 2019-07-29
Fixed issues with spaces in paths in post build event scripts for VS

Dysl3xik:
#218

### DIFF
--- a/Common++/header/IpAddress.h
+++ b/Common++/header/IpAddress.h
@@ -34,10 +34,15 @@ namespace pcpp
 		// protected c'tor
 		IPAddress() : m_IsValid(false) {}
 	public:
-#if __cplusplus > 199711L
-		typedef std::unique_ptr<IPAddress> Ptr_t; 
+		//Visual studio has always been stupid about returning something useful for __cplusplus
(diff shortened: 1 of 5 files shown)


Dysl3xik:
You should verify that the property sheet changes in the second commit don't cause any issues with your 2015 build, I image the tags will be ignored by VS, but I am not sure and don't have 2015 to test with.

Dysl3xik:
I also found an issue that you are still using auto_ptr in VS builds. I added another commit to fix this issue.

```
	//Visual studio has always been stupid about returning something useful for __cplusplus
	//Only recently was this fixed - and even then it requires a specific hack to the command line during build
	//Its easier/more consistent to test _MSC_VER in VS 
	//https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=vs-2017

	#if __cplusplus > 199711L || _MSC_VER >= 1800 //Maybe this can be 1600 for VS2010
	#define PCPP_SMART_PTR(T) std::unique_ptr<T>
	#else
	#define PCPP_SMART_PTR(T) std::auto_ptr<T>
	#endif
```

I just took the time to read your contributing guide. If I find any other issues I will switch to dev branch and start issuing PR from there. Sorry about that!

PR #222, opened 2019-08-14
File streaming and compression

Dysl3xik:
I have completed my updates to change the file IO for pcapng files to use streaming instead of memory copies and dumps. I also went ahead and integrated in support for compression - with a first implementation using zstd compression.

This was a decent chunk of work, it would be nice if you wouldn't mind marking me as a contributor.

This PR SHOULD fit into the existing code base without issue and simply turn on the streaming to file code. For the compression stuff to get enabled you have to build the library with specific macro enabled to build that compression stuff into Pcap++. For zstd support this macro is USE_Z_STD. If you wanted to sub in other compression types you would need to make a new macro, see where the USE_Z_STD is used to load in headers and setup pointers and so similar for your other compression library.

Some Notes:

I will submit a seperate PR with vs project changes, that will be where you need to spend the most time to decide exactly how you want the zstd integration to work. 

I did change the default behavior for reading/writing files from holding everything in memory to streaming, I did NOT add a switch to do this. So you may wish to update your version number as there may be some heathens out there that like eating up all their ram (or are severely limited by disk IO) that need everything to be kept in memory, and this could affect them. You could then build in a flag to revert to the old behavior if you/they find it necessary. The old functions are all still there.

At the top of light_compression.c you can see where a bunch of function pointers are subbed in to enable calling the specific compression implementation. I tried assigning these function pointers inside light_zstd_compression.c and marking the ones in light_compression.c as extern. The code would build OK, but when running in my test app it always defaulted back to the null compressor (no compression). So for now these assignments remain in the light_compression.c file. If you can find something that works by letting you assign them in the compression implementation file that would be better because then if someone wants to add in a new compression they don't need to touch light_compression.c they just do the assignment in their own C file.

For adding new compression you always have to update light_compression.h with a new if case to loop in the appropriate compression header file. If you wanted you could probably support multiple types of compression in a single build, but I didn't design as such for now.


seladb:
Closing this PR as it's replaced by #224

PR #223, opened 2019-08-14
File streaming compression vs project changes

Dysl3xik:
This PR includes the changes to VS project to build with compression enabled.

Basically to make this work you DL the latest release of zstd from here: https://github.com/facebook/zstd/releases 

There is a new macro I defined in the prop sheets called $(ZStdHome) this needs to point to the folder you download and unzipped from above.

Build the project and hopefully PROFIT!

PS:
I also added Platform\Configuration folders for your dist folder, this way you can build debug/release and have both outputs at the same time. In the existing code there was only a single dist folder. Technicially you would need a little more work to loop in the proper zstd version as there is 32 and 64 bit builds for that too, but most people will always make one or the other, not both.


Dysl3xik:
I will close this until the base changes are merged.

Dysl3xik:
This is still lingering to get merged in, I am going to re-open it to spark the conversation.

FWIW I have been using this in production steadily now without issues.

seladb:
Thanks @Dysl3xik for reopening this. Can you please rebase the code to the latest version of `dev`? 

Also this change is pretty huge in terms of number of files and lines of code changed, let's see how we can break this up so it'll be easier for me to review

Dysl3xik:
Actually I think almost all of this has already made it into your master, this should really just be the changes to vxproj files to toggle on zstd or not. I probably need to just update everything to your master and re-issue a new PR.

However I just want to check with you first, if I go through this effort will you move along to "officially" integrating the zstd stuff?

seladb:
I prefer not to have zstd as a must-have dependency for PcapPlusPlus. I'd rather have it as an optional dependency. Can we do that?

Dysl3xik:
That's how it is now, if you add a #define it turns everything on,
otherwise it's invisible. This is why I think most all the required code
changes already exists in the code base. It's just adding libs and project
config. I think the big issue last time was documentation and such.

seladb:
if this feature can be turned on and off via configuration, I'm all for it. Of course documentation is needed if we want people to know about this feature, but we can take care of it after merging the code.

One more thing that we need to take care of is tests. Tests and CI are the only way to make sure things don't break when we make changes. I try to keep test coverage as high as possible

Dysl3xik:
Superseded by PR #406

PR #224, opened 2019-08-15
File streaming and compression

Dysl3xik:
see PR #222 for details. This is same, but PR redirected to DEV branch


3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng_ext.c
```diff
@@ -35,8 +37,7 @@ struct _light_pcapng_t
 {
 	light_pcapng pcapng;
 	light_pcapng_file_info *file_info;
-	light_pcapng pcapng_iter;
```
seladb: I'm not entirely sure why you removed `pcapng_iter`. I saw it was being used extensively in the code, and instead you now use the `pcapng` member. Is `pcapng_iter` really redundant?

mk/vs2015/GitInfoPropertySheet.props
```diff
@@ -5,6 +5,14 @@
     <GitCommit>unavailable</GitCommit>
     <GitBranch>unavailable</GitBranch>
   </PropertyGroup>
+  <ItemGroup>
+  <BuildMacro Include="GitCommit">
```
seladb: please fix the indentation

seladb: I'm not sure what `BuildMacro` actually does. Could you please elaborate?

Common++/header/MacAddress.h
```diff
@@ -5,7 +5,7 @@
 #include <string.h>
 #include <string>
 
-#if __cplusplus > 199711L
+#if __cplusplus > 199711L || _MSC_VER >= 1800
```
seladb: so just to make sure: you added `_MSC_VER` because it's more reliable than using `__cplusplus`?

In that case, I'm not sure why `std::copy` fails

seladb:
hi @Dysl3xik , I went over the PR. It seems there are extensive changes in LightPcapNg which are not easy to track. Could you please describe the new design?

I also added a few comments in the code itself.

Also (even more important) - could you please describe how you tested these changes? Except for the unit-tests (`Packet++Test` and `Pcap++Test`) there are example apps that consume pcapng files, for example: [HttpAnalyzer](https://github.com/seladb/PcapPlusPlus/tree/master/Examples/HttpAnalyzer), [PcapPrinter ](https://github.com/seladb/PcapPlusPlus/tree/master/Examples/PcapPrinter) and a few more. Did you have a chance to check them while testing?

Let's make user we don't break anything while making this change. If you want please feel free to add more unit-test to `Pcap++Test`.

Dysl3xik:
I am copying these high level notes over from the original PR:

I have completed my updates to change the file IO for pcapng files to use streaming instead of memory copies and dumps. I also went ahead and integrated in support for compression - with a first implementation using zstd compression.

This was a decent chunk of work, it would be nice if you wouldn't mind marking me as a contributor.

This PR SHOULD fit into the existing code base without issue and simply turn on the streaming to file code. For the compression stuff to get enabled you have to build the library with specific macro enabled to build that compression stuff into Pcap++. For zstd support this macro is USE_Z_STD. If you wanted to sub in other compression types you would need to make a new macro, see where the USE_Z_STD is used to load in headers and setup pointers and so similar for your other compression library.

Some Notes:

I will submit a seperate PR with vs project changes, that will be where you need to spend the most time to decide exactly how you want the zstd integration to work.

I did change the default behavior for reading/writing files from holding everything in memory to streaming, I did NOT add a switch to do this. So you may wish to update your version number as there may be some heathens out there that like eating up all their ram (or are severely limited by disk IO) that need everything to be kept in memory, and this could affect them. You could then build in a flag to revert to the old behavior if you/they find it necessary. The old functions are all still there.

At the top of light_compression.c you can see where a bunch of function pointers are subbed in to enable calling the specific compression implementation. I tried assigning these function pointers inside light_zstd_compression.c and marking the ones in light_compression.c as extern. The code would build OK, but when running in my test app it always defaulted back to the null compressor (no compression). So for now these assignments remain in the light_compression.c file. If you can find something that works by letting you assign them in the compression implementation file that would be better because then if someone wants to add in a new compression they don't need to touch light_compression.c they just do the assignment in their own C file.

For adding new compression you always have to update light_compression.h with a new if case to loop in the appropriate compression header file. If you wanted you could probably support multiple types of compression in a single build, but I didn't design as such for now.

Common++/header/MacAddress.h
Dysl3xik: Yes __cplusplus reports 199711L on any version of VS up to 2017.3. Even in that version it lies unless you add a special compiler flag see this article:

https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/

The std::copy does not FAIL. I mentioned this on our other thread before but im copying it here:

When i flipped that switch it updated your code to use some newer C++ features specifically some code using std::copy. Apparently microsoft flagged this particular usage of std::copy as potentially risky because it can't check bounds in debug mode. See article discussing this kind of issue here https://devblogs.microsoft.com/cppblog/why-am-i-getting-these-_scl_secure_no_warnings-messages/ . You can either change your call to std::copy, or disable SDL, or similar . I will leave the remaining googling and decision to you.

mk/vs2015/GitInfoPropertySheet.props
Dysl3xik: Without that when I open the property sheet for viewing (at least in VS2017) the list of macros shows up empty:

image

Dysl3xik: Done, will update

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng_ext.c
Dysl3xik: Yes it was used in the old code all over the place because everything was in memory and this was the pointer to the "next" block or something if I recall.

With the new code there is never more than a few blocks in memory (usually only 1) so this kind of bookkeeping isn't necessary. Everything you are working with is contained in `pcapng` so only use that variable. I think this is way more clear/concise.

I could have left it there, but I figured it would be more confusing for the next person working on the code to leave a dead member variable behind.

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
```diff
@@ -256,6 +270,96 @@ static size_t __parse_mem_copy(struct _light_pcapng **iter, const uint32_t *memo
...
+/// </summary>
+/// <param name="fd">File to read from</param>
+/// <returns>The block read from the file - may contain sub blocks</returns>
+void light_read_record(light_file fd, light_pcapng *record)
+{
+	//FYI general block structure is like this
+
+	//0                   1                   2                   3
+	//0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+	//+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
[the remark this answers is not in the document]
Dysl3xik: Sorry VS probably prompted for this and I clicked through. I ran untabify which should fix it

Pcap++/header/PcapFileDevice.h
```diff
@@ -387,6 +387,7 @@ namespace pcpp
 	{
 	private:
 		void* m_LightPcapNg;
+		int m_compressionLevel;
```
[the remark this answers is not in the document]
Dysl3xik: Done

Pcap++/header/PcapFileDevice.h
```diff
@@ -404,8 +405,9 @@ namespace pcpp
 		 * A constructor for this class that gets the pcap-ng full path file name to open for writing or create. Notice that after calling this
 		 * constructor the file isn't opened yet, so writing packets will fail. For opening the file call open()
 		 * @param[in] fileName The full path of the file
+		 * @param[in] the compression level to use when writing the file, use 0 to disable compression or 10 for max compression
```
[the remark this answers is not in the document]
Dysl3xik: Done

mk/vs2015/fetch-git-info.bat
```diff
@@ -11,6 +11,14 @@ echo   ^<PropertyGroup Label="UserMacros"^>>> GitInfoPropertySheet.props
 echo     ^<GitCommit^>%GIT_COMMIT%^</GitCommit^>>> GitInfoPropertySheet.props
 echo     ^<GitBranch^>%GIT_BRANCH%^</GitBranch^>>> GitInfoPropertySheet.props
 echo   ^</PropertyGroup^>>> GitInfoPropertySheet.props
+echo   ^<ItemGroup^>>> GitInfoPropertySheet.props
```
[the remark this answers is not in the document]
Dysl3xik: Done

Dysl3xik:
For the testing I ran your test suite and confirmed it worked no issues. 

Additionally I had a little test app I made myself that ran a capture and wrote the output to two files simultaneously one compressed and one non compressed. It then read the files back and compared a few member variables for every packet, as well as the overall counts of packets read from each file, and finally compared the number of packets read from both files to the number captured.

My test code works with pcpp compiled with both ztd turned on, and with it turned off. If you request compression when its NOT built into the library it just ignores the request and continues without compression.

I can give you that test cpp file and you can integrate that into your test suite if you want.

3rdParty/LightPcapNg/LightPcapNg/src/light_null_compression.c
```diff
@@ -0,0 +1,36 @@
+// light_null_compression.c
```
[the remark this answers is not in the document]
Dysl3xik: You can read my old note about wanting to use externs below. I am not as fluent in old school C as I am at modern C++. Anyways I read up on externs in C and was able to do what I wanted all along.

Now the light_compression defines so function which you plug your compression implementation into. If there is no compression built with the lib OR if user requests no compression in their code these pointers are all NULL and all the compression stuff gets skipped. 

If you build with compression, then inside the C file for your implementation you need to assign your function implementations into the global function pointers. This way the generic light_compression implementation can dispatch the work to your implementation.

For examples of this see the updated light_null_compression.c and light_zstd_compression.c.

Dysl3xik:
Here is the test code I mentioned

```
#include <tuple>
#include <iostream>
#include <stdio.h>     // printf
#include <stdlib.h>    // free
#include <string.h>    // memset, strcat, strlen
#include "PcapLiveDeviceList.h"
#include "PlatformSpecificUtils.h"
#include "PcapFileDevice.h"
#include <chrono>
#include "NetworkTrafficCapture.h"

using namespace std::chrono_literals;

pcpp::PcapNgFileWriterDevice *writer = nullptr, *writer2 = nullptr;
static constexpr const char *outFileName = "outputCompress.pcapng.zstd";
static constexpr const char *outFileName2 = "outputCompress.pcapng";
static constexpr int compressionLevel = 5;
static constexpr int minCompressSize = 20000;
static constexpr int captureTimeSec = 20;
static constexpr int capturePort = 4141;
static constexpr const char *captureIP = "192.168.1.100";

/**
 * A struct for collecting packet statistics
 */
struct PacketStats
{
	int ethPacketCount;
	int ipv4PacketCount;
	int ipv6PacketCount;
	int tcpPacketCount;
	int udpPacketCount;
	int dnsPacketCount;
	int httpPacketCount;
	int sslPacketCount;

	/**
	 * Clear all stats
	 */
	void clear() { ethPacketCount = 0; ipv4PacketCount = 0; ipv6PacketCount = 0; tcpPacketCount = 0; udpPacketCount = 0; tcpPacketCount = 0; dnsPacketCount = 0; httpPacketCount = 0; sslPacketCount = 0; }

	/**
	 * C'tor
	 */
	PacketStats() { clear(); }

	/**
	 * Collect stats from a packet
	 */
	void consumePacket(pcpp::Packet& packet)
	{
		if (packet.isPacketOfType(pcpp::Ethernet))
			ethPacketCount++;
		if (packet.isPacketOfType(pcpp::IPv4))
			ipv4PacketCount++;
		if (packet.isPacketOfType(pcpp::IPv6))
			ipv6PacketCount++;
		if (packet.isPacketOfType(pcpp::TCP))
			tcpPacketCount++;
		if (packet.isPacketOfType(pcpp::UDP))
			udpPacketCount++;
		if (packet.isPacketOfType(pcpp::HTTP))
			httpPacketCount++;
		if (packet.isPacketOfType(pcpp::SSL))
			sslPacketCount++;
	}

	/**
	 * Print stats to console
	 */
	void printToConsole()
	{
		printf("Ethernet packet count: %d\n", ethPacketCount);
		printf("IPv4 packet count:     %d\n", ipv4PacketCount);
		printf("IPv6 packet count:     %d\n", ipv6PacketCount);
		printf("TCP packet count:      %d\n", tcpPacketCount);
		printf("UDP packet count:      %d\n", udpPacketCount);
		printf("DNS packet count:      %d\n", dnsPacketCount);
		printf("HTTP packet count:     %d\n", httpPacketCount);
		printf("SSL packet count:      %d\n", sslPacketCount);
	}
};

/**
 * A callback function for the async capture which is called each time a packet is captured
 */
static void onPacketArrives(pcpp::RawPacket* packet, pcpp::PcapLiveDevice* dev, void* cookie)
{
	// extract the stats object form the cookie
	PacketStats* stats = (PacketStats*)cookie;

	// parsed the raw packet
	pcpp::Packet parsedPacket(packet);

	// collect stats from packet
	stats->consumePacket(parsedPacket);

	if (writer && writer->isOpened())
		writer->writePacket(*packet);

	if (writer2 && writer2->isOpened())
		writer2->writePacket(*packet);
}

/**
 * main method of the application
 */
int main(int argc,const char* argv[])
{
	SetDllDirectory("C:\\Windows\\System32\\Npcap\\");

	//return main2();

	// IPv4 address of the interface we want to sniff
	std::string interfaceIPAddr = captureIP;

	// create the stats object
	PacketStats stats;

	// find the interface by IP address
	pcpp::PcapLiveDevice* dev = pcpp::PcapLiveDeviceList::getInstance().getPcapLiveDeviceByIp(interfaceIPAddr.c_str());
	if (dev == NULL)
	{
		printf("Cannot find interface with IPv4 address of '%s'\n", interfaceIPAddr.c_str());
		exit(1);
	}

	// Get device info
	// ~~~~~~~~~~~~~~~

	// before capturing packets let's print some info about this interface
	printf("Interface info:\n");
	// get interface name
	printf("   Interface name:        %s\n", dev->getName());
	// get interface description
	printf("   Interface description: %s\n", dev->getDesc());
	// get interface MAC address
	printf("   MAC address:           %s\n", dev->getMacAddress().toString().c_str());
	// get default gateway for interface
	printf("   Default gateway:       %s\n", dev->getDefaultGateway().toString().c_str());
	// get interface MTU
	printf("   Interface MTU:         %d\n", dev->getMtu());

	// get DNS server if defined for this interface
	if (dev->getDnsServers().size() > 0)
		printf("   DNS server:            %s\n", dev->getDnsServers().at(0).toString().c_str());

	// open the device before start capturing/sending packets
	if (!dev->open())
	{
		printf("Cannot open device\n");
		exit(1);
	}

	// Using filters
	// ~~~~~~~~~~~~~

	// create a filter instance to capture only traffic on designated port
	pcpp::PortFilter portFilter(capturePort, pcpp::SRC_OR_DST);

	// create a filter instance to capture only TCP traffic
	pcpp::ProtoFilter protocolFilter(pcpp::TCP);

	// create an AND filter to combine both filters - capture only TCP traffic on designated port
	pcpp::AndFilter andFilter;
	andFilter.addFilter(&portFilter);
	andFilter.addFilter(&protocolFilter);

	// set the filter on the device
	dev->setFilter(andFilter);

	// Write packets to PCAP file
	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
	// create a pcap-ng file writer. Specify file name. Link type is not necessary because
	// pcap-ng files can store multiple link types in the same file
	pcpp::PcapNgFileWriterDevice pcapNgWriter(outFileName, compressionLevel);
	pcpp::PcapNgFileWriterDevice pcapNgWriter2(outFileName2, 0);

	writer = &pcapNgWriter;
	writer2 = &pcapNgWriter2;

	// try to open the file for writing
	if (!pcapNgWriter.open() || !pcapNgWriter2.open())
	{
		printf("Cannot open output file for writing\n");
		exit(1);
	}

	printf("\nStarting packet capture with a filter in place...\n");

	// start capture in async mode. Give a callback function to call to whenever a packet is captured and the stats object as the cookie
	dev->startCapture(onPacketArrives, &stats);

	PCAP_SLEEP(captureTimeSec);

	// stop capturing packets
	dev->stopCapture();

	// stop the file capture
	pcapNgWriter.close();
	pcapNgWriter2.close();

	// print results - should capture only packets which match the filter (which is TCP port 80)
	printf("Results:\n");
	stats.printToConsole();

	pcap_stat nicStats;
	dev->getStatistics(nicStats);
	printf("Dropped %d or %d Packets!\n\n", nicStats.ps_drop, nicStats.ps_ifdrop);

	// close the device before application ends
	dev->close();

	std::chrono::time_point start = std::chrono::steady_clock::now();
	
	pcpp::PcapNgFileReaderDevice reader(outFileName);
	reader.open();

	pcpp::PcapNgFileReaderDevice reader2(outFileName2);
	reader2.open();

	pcpp::RawPacket packet, packet2;

	int packetCount = 0;
	while (true)
	{
		bool res1 = reader.getNextPacket(packet); 
		bool res2 = reader2.getNextPacket(packet2);
		_ASSERT(res1 == res2);
		if (!res2 && !res2)
			break;
		_ASSERT(packet.getObjectType() == packet2.getObjectType());
		_ASSERT(packet.getRawDataLen() == packet2.getRawDataLen());
		++packetCount;
	}

	_ASSERT(packetCount == stats.ethPacketCount);

	printf("Comparison took %d miliseconds!\n\n", std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start));

};
```

Common++/header/MacAddress.h
seladb: sure, I'll build the code and see what's the best option

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
```diff
@@ -34,518 +35,660 @@
...
+/// <param name="fd">File to read from</param>
+/// <returns>The block read from the file - may contain sub blocks</returns>
+void light_read_record(light_file fd, light_pcapng *record)
+{
+   //FYI general block structure is like this
+
+   //0                   1                   2                   3
+   //0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+   //+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   //|								Block Type										|
```
seladb: can you please fix the indentation here?

seladb:
thanks for the thorough explanation and also thanks for referring to my comments.

I had another look at the current unit-tests for pcapng ([here](https://github.com/seladb/PcapPlusPlus/blob/898519e67f95ba2c4c5871554e51e3c08ee0f346/Tests/Pcap%2B%2BTest/main.cpp#L1032) and [here](https://github.com/seladb/PcapPlusPlus/blob/898519e67f95ba2c4c5871554e51e3c08ee0f346/Tests/Pcap%2B%2BTest/main.cpp#L1108)) and they look sufficient to me. The only things that are not covered are large files and compression, which are the reason you made this change. If you think we should add tests for them please let me know and we'll figure this out together.

Before merging the PR I'd like to take a look at `std::copy` and decide how to resolve it in a way it will build on MSVC debug. Please give me some time to do so.

One last (but very important) comment - you've made significant contributions to PcapPlusPlus and you should be added as one of the most significant [contributors](https://github.com/seladb/PcapPlusPlus/graphs/contributors) to this project, however it seems your commits come from a Git user which is not your GitHub user, that's probably why GitHub doesn't add you to the contributors list. I'm not sure how to fix that, especially after you already made all the changes, but if you can find a way to fix that please let me know.

Dysl3xik:
I added my work email to my github account. I THINK that has now tagged me as the writer of the commits, maybe you can verify that on your side. Hopefully this will fix the issue. 
I am not SUPER familiar with github, and thought you had to tag contributors. However, its clear now that was a difference in configuration on my side, I used my work email at work, but my github is tied to my personal email as primary.

Anyways hopefully its all fixed up now.

seladb:
I think it's ok now:

image

I hope GitHub will add you to the list of contributors once I merge the code to `master`.

I'll take a look at the `std::copy` issue soon and will keep you posted.

seladb:
After merging your PR, you're now in the list of contributors :-)

image

PR #228, opened 2019-08-23
Minor adjustments for pcapng and new minor helper function

Dysl3xik:
This PR has a minor adjustment for file streaming PR you took before. I found in some cases the compiler would complain about duplicate definitions (atleast in VS). I also found that depending upon compilation order you could end up NOT Having USE_NULL_COMPRESSION defined in the C file so I added an include to make sure that can't happen.

I added a new function to pcapLiveDevice to tell if the capture thread is currently running for that device. 

I added a new class called FilterTester that duplicates the functionality of IPcapDevice::verifyFilter and IPcapDevice::matchPacketWithFilter but it avoids repeated building and discarding of the filter. I have an application where I expect to playback a file with captures from a variety of sensors and I need to pipe the packets out to the appropriate handlers based upon passing the filter criteria or not. This way I can make a filter for each source and test each packet in the file against that filter and either pass it or reject it. With the current offering there is no way I could find to have a reusable filter object to test packets with.

### DIFF
--- a/Pcap++/header/PcapFilter.h
+++ b/Pcap++/header/PcapFilter.h
@@ -3,9 +3,13 @@
 
 #include <string>
 #include <vector>
+#include <memory>
 #include "ProtocolType.h"
 #include <stdint.h>
 #include "ArpLayer.h"
+
+//Required for FilterTester
+struct bpf_program;
 
 /**
  * @file
@@ -82,6 +86,52 @@ namespace pcpp
 		 * Virtual destructor, does nothing for this class
 		 */
 		virtual ~GeneralFilter();
+	};
+
+	//Used in FilterTester
+	class RawPacket;
+
+	/**
+	 * @class FilterTester
+	 * This class can be loaded with a filter string and then can be used to verify the string is valid and check if a packet matches it<BR>
+	 */
+	class FilterTester : public GeneralFilter
+	{
+#if __cplusplus > 199711L || _MSC_VER >= 1800 //Maybe this can be 1600 for VS2010
+		typedef std::unique_ptr<bpf_program> Ptr_t;
+#else
+		typedef std::auto_ptr<bpf_program> Ptr_t;
+#endif
+
+	private:
+		bool verified, verifiedResult;
+		const std::string filterStr;
+		Ptr_t prog;
+
+	public:
+		FilterTester(const std::string& filterStr);
+
+		virtual ~FilterTester();
+
+		/**
+		 * A method that parses the class instance into BPF string format
+		 * @param[out] result An empty string that the parsing will be written into. If the string isn't empty, its content will be overridden
+		 */
+		virtual void parseToString(std::string& result);
+
+		/**
+		* Verify the filter is valid
+		* @return True if the filter is valid or false otherwise
+		*/
+		bool verifyFilter();
+
+		/**
+		 * Match a raw packet with a given BPF filter.
+		 * @param[in] rawPacket A pointer to the raw packet to match the BPF filter with
+		 * @return True if a raw packet matches the BPF filter or false otherwise
+		 */
+		bool matchPacketWithFilter(RawPacket* rawPacket);
+
 	};
 
 
--- a/Pcap++/src/PcapFilter.cpp
+++ b/Pcap++/src/PcapFilter.cpp
@@ -9,9 +9,61 @@
 #elif LINUX
 #include <in.h>
 #endif
+#include <pcap.h>
+#include "RawPacket.h"
 
 namespace pcpp
 {
+
+FilterTester::FilterTester(const std::string &filterStr) : filterStr(filterStr), prog(nullptr)
+{}
+
+FilterTester::~FilterTester()
+{
+	if(prog)
+		pcap_freecode(prog.get());
+}
+
+void FilterTester::parseToString(std::string& result)
+{
+	if (verifyFilter())
+		result = filterStr;
+}
+
+bool FilterTester::verifyFilter()
+{
+	if (verified)
+		return verifiedResult;
+
+	verified = true;
+	prog = Ptr_t(new bpf_program());
+	LOG_DEBUG("Compiling the filter '%s'", filterStr.c_str());
+	if (pcap_compile_nopcap(9000, pcpp::LINKTYPE_ETHERNET, prog.get(), filterStr.c_str(), 1, 0) < 0)
+	{
+		verifiedResult = false;
+	}
+	else
+	{
+		verifiedResult = true;
+	}
+
+	return verifiedResult;
+}
+
+bool FilterTester::matchPacketWithFilter(RawPacket* rawPacket)
+{
+	if (!verifyFilter())
+		return false;
+
+	struct pcap_pkthdr pktHdr;
+	pktHdr.caplen = rawPacket->getRawDataLen();
+	pktHdr.len = rawPacket->getRawDataLen();
+	pktHdr.ts = rawPacket->getPacketTimeStamp();
+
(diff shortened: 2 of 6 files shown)


Pcap++/header/PcapFilter.h
```diff
@@ -84,6 +88,52 @@ namespace pcpp
...
+	//Used in FilterTester
+	class RawPacket;
+
+	/**
+	 * @class FilterTester
+	 * This class can be loaded with a filter string and then can be used to verify the string is valid and check if a packet matches it<BR>
+	 */
+	class FilterTester : public GeneralFilter
+	{
+#if __cplusplus > 199711L || _MSC_VER >= 1800 //Maybe this can be 1600 for VS2010
```
seladb: Is using smart pointers really necessary here? I try to avoid using them wherever possible because of the different implementations in different versions of C++

Pcap++/header/PcapFilter.h
```diff
@@ -84,6 +88,52 @@ namespace pcpp
...
+		typedef std::unique_ptr<bpf_program> Ptr_t;
+#else
+		typedef std::auto_ptr<bpf_program> Ptr_t;
+#endif
+
+	private:
+		bool verified, verifiedResult;
```
seladb: please follow the coding convention of private members' names starting with `m_` followed by a capital letter

Pcap++/header/PcapFilter.h
```diff
@@ -3,10 +3,14 @@
 
 #include <string>
 #include <vector>
+#include <memory>
 #include "ProtocolType.h"
 #include <stdint.h>
 #include "ArpLayer.h"
 
+//Required for FilterTester
+struct bpf_program;
```
seladb: can we hide this structure and move it to the implementation file?

instead you can have the `prog` private member be of type `void *` and do the casting in the implementation file

Pcap++/src/PcapFilter.cpp
```diff
@@ -9,10 +9,62 @@
 #elif LINUX
 #include <in.h>
 #endif
+#include <pcap.h>
+#include "RawPacket.h"
 
 namespace pcpp
 {
 
+FilterTester::FilterTester(const std::string &filterStr) : filterStr(filterStr), prog(nullptr)
```
seladb: setting a smart pointer with `nullptr` doesn't work in majority of the platform, please see the CI builds that fail on most platforms

Pcap++/header/PcapFilter.h
Dysl3xik: Easy fix, will do.

Pcap++/header/PcapFilter.h
Dysl3xik: I can do this, I am just accustomed to keeping type safety, but in this case its all internal so it doesn't TECHNICALLY void the safety as external users can't touch this.

Pcap++/src/PcapFilter.cpp
Dysl3xik: DIdn't know this, I can remove the smart pointer and use old school ptr here and just de-allocate it in destructor. 

Is there a cross platform compatible way to handle this?

Pcap++/header/PcapFilter.h
Dysl3xik: Not required. I am just spoiled and accustomed to using latest C++. dumb pointer will suffice and I can allocate when used and kill with destructor.

Dysl3xik:
My understanding of the current filtering is you provide a filter and then anything that passes the filter is passed on, but if it fails the filter its ignored. I have 1 capture file, but a variety of captured packets within. I need to filter the captured packets out to the proper destinations. 

 If I have for example 5 different sources contributing to a capture file i need filters to check which destination to "replay" them back to.  I can use the same filter that passed for the capture to filter the replay. Anyways in the current form if I need to check 5 filters (5 sources) for every packet, it will allocate and release resources for every one of those packets as I check each filter. This is wasteful.

Pcap++/src/PcapFilter.cpp
seladb: I'm not sure if there's a cross platform way to do it with smart pointers. But if you're changing it anyway to standard pointers that shouldn't be an issue anymore

seladb:
I think I understand your use case now. Let me suggest a different approach:

We can add a method to `GeneralFilter` that will compile the filter and generate an internal `bpf_program` object . In addition we can add on overload method for `IPcapDevice::matchPacketWithFilter` that takes a pointer to `GeneralFilter` instead of `std::string filterAsString` and will use the internal `bpf_program` object.

In addition, could you please create the PR from `dev` instead of `master`?

Dysl3xik:
**Well I think baking the bpf_program into the `GeneralFilter` is a solid plan that opens up future uses both externally and internally.**

**If the program is moved into the `GeneralFilter` I would also suggest migrating `IPcapDevice::matchPacketWithFilter` into `GeneralFilter` as well.** The existing implementation can probably be left there for compatibility, and maybe use a free function to power both versions so you don't duplicate code. Free function would probably look like `bool needsAGoodName(bpf_program* prog, RawPacket* rawPacket)`. The `GeneralFilter` will have its own `bpf_program` it can pass in, and your existing implementation can keep its program in the static function member as it does today.

I don't think adding a new filter type that parses from a string violates your idea of trying to make building filters easier. You are not FORCING someone to use that particular class its just an option. They are still free to use the building block classes. 

**So I think I would just go with a new class maybe `BPFStringFilter` (not sure about casing for abbreviation)  that inherits from `GeneralFilter` and simply takes BPF string as arg for the constructor.** 

I also don't see any harm in dropping the abstract requirement - aside from making it possible for people to make empty and pointless filters, or opening up the possibility of object slicing. 

The only other way I can think would be to add some static factory function to `GeneralFilter` that takes a BPF string and returns a `std::unique_ptr<GeneralFilter>` type. This way you could hide this particular filter implementation. But this seems like HUGE overkill and I don't get the impression that you like using unique_ptr too much in your API. If you don't use unique ptr then you can leak easily. IMO this whole approach is a non option.

Let me know if you agree to the propositions in **bold**, and what your preferred names are, and I will issue a new PR with this into the dev branch.

Dysl3xik:
* Bake `bpf_program` into `GeneralFilter` in a way that enables creation of `bpf_program` from a filter object. Let's make sure it is a lazy creation and is done only upon user request
**Agree**
* Create a `BPFStringFilter` class which inherits `GeneralFilter` and takes a BPF string as a c'tor. This should fulfill your requirement. (BTW `BPFStringFilter` is ok in terms of naming conventions)
**Agree**
* Because `BPFStringFilter` is added it seems there's no need to change `GeneralFilter` abstraction, am I wrong here?
**Yes, I just added options in my last post to foster discussion. If we add this there is no need to touch existing abstract base.**
* I think we can remove `IPcapDevice::verifyFilter()` as it will be part of the new `BPFStringFilter` class
**Agree - do you want to mark it deprecated first instead of just pulling it out?**
* Keep `IPcapDevice::matchPacketWithFilter()` and add another overload to it that takes a `GeneralFilter` instead of  `std::string filterAsString`
**Ok, would you like the string variant marked as deprecated as well? The goal being to hopefully push people to use the GeneralFilter option instead? If they are only ever calling this with the same filter its no difference, but if checking different filters there is a performance gain in switching.**

Another question. You expressed concerns over me forward declaring `bpf_program` in the PcapFilter.h previously: https://github.com/seladb/PcapPlusPlus/pull/228#discussion_r317379330 with the suggestion to use a `void*` in header and cast behind the scenes. Should I assume you want me to do that in this implementation as well since the `bpf_program*` will now live inside the `GeneralFilter`?

Dysl3xik:
Closing this current PR and will re-issue with discussed changes.

PR #229, opened 2019-08-28
Add BPFStringFilter class add matchPackWithFilter to GeneralFilter

Dysl3xik:
Changes as discussed in #228 

New Notes:

I added test cases for this PR in the test code!

Should `verifyFIlter` be placed in GeneralFilter? This is easy copy paste change if we decide to do it. I don't know if its possible to build an invalid filter using your helper classes, if it is this would enable catching it.

I realized just before submitting this that your filter helper classes allow mutations. Because of this I had to update `matchPacketWithFilter(...)` to always call `parseToString` and compare it to the cached string used to build the `bpf_program`. Without this the first time you call match the filter is locked in, so if someone mutated the filter and called it again it would not actually use the new filter. I don't like having to do this check as it must rebuild the string EVERY call which is a lot when parsing all the packets in a file, but its the "safer" way to do this. 

The only ways around this last item I can think of are:

1. Make the user provide an extra boolean in `matchPacketWithFilter` to force the rebuild.
2. Add a bool into the base class something like "m_isDirty" which must be set any time a mutation is called or when class constructed. This is alot of work, and probably prone to error over time though....
3. Comapre the strings every time as I did in current code.

### DIFF
--- a/3rdParty/Getopt-for-Visual-Studio/getopt.h
+++ b/3rdParty/Getopt-for-Visual-Studio/getopt.h
@@ -56,7 +56,8 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#pragma warning(disable:4996);
+#pragma warning(push)
+#pragma warning(disable: 4996)	//Disable this warning for this file
 
 #define __GETOPT_H__
 
@@ -649,5 +650,7 @@ getopt_long_only(int nargc, char * const *nargv, const char *options,
 #ifdef __cplusplus
 }
 #endif
+
+#pragma warning(pop)
 
 #endif /* !defined(__UNISTD_H_SOURCED__) && !defined(__GETOPT_LONG_H__) */
--- a/3rdParty/LightPcapNg/LightPcapNg/include/light_null_compression.h
+++ b/3rdParty/LightPcapNg/LightPcapNg/include/light_null_compression.h
@@ -24,10 +24,15 @@
 #ifndef INCLUDE_LIGHT_NULL_COMPRESSION_H_
 #define INCLUDE_LIGHT_NULL_COMPRESSION_H_
 
+#if defined(USE_NULL_COMPRESSION)
+
 #include <stdlib.h>
 
 typedef void _compression_t;
 typedef void _decompression_t;
 
 struct light_file_t;
+
+#endif
+
 #endif /* INCLUDE_LIGHT_NULL_COMPRESSION_H_ */
--- a/3rdParty/LightPcapNg/LightPcapNg/src/light_null_compression.c
+++ b/3rdParty/LightPcapNg/LightPcapNg/src/light_null_compression.c
@@ -21,6 +21,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+#include "light_compression.h"
 #include "light_null_compression.h"
 #include "light_compression_functions.h"
 #include "light_file.h"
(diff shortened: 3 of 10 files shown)


Dysl3xik:
Some of the builds are failing because cmake isn't right on this branch. In this case all of the tests actually passed though.

Some of the builds are failing because C++14 support isn't turned on (or available) so [[deprecated]] attribute causes error. Do you want a macro for this that swaps between versions or just #disable it if c++ version is old? Until C++14 this was non standard so its going to vary with compiler and can potentially be a pain to deal with... 

Some of the windows builds are failing because it says PCAPP_ASSERT ins't defined!?? I am not sure if this is a dev branch issue or not. I can certainly build in VS on my box and the additional tests pass as well.

Dysl3xik:
Looks like tests are finally starting to pass now!

Notes:

I suppose you renamed PCAPP_ASSERT to PTF_ASSERT in this branch I modified push for this.

I removed the deprecated attribute in versions of C++ < 14 or VS version < 2015 (minimum version to support the attribute).

You globally disable warning 4996 in your test code. This actually hides the fact that I marked two functions as deprecated. I removed the global disable and made it scoped. This ultimately made the test break. As a short term fix I re added it to the test code - but still scoped to the main.cpp file. I suggest you follow the compilers advice and  #define _CRT_SECURE_NO_WARNINGS or similar instead of just disabling this warning, otherwise you GLOBALLY cut off those deprecation warnings which could be baked into 3rd party libs you use and you will never know they are going to go away.

The existing `IPcapDevice::matchPacketWithFilter(filterAsString, *iter)` was flagged for a memory leak due to its static allocation of bpf_filter I assume. This is probably not a REAL leak, but i removed it from test code to get a pass - you can see this in my very last commit.

Dysl3xik:
Looks like tests are finally starting to pass now!

Notes:

I suppose you renamed PCAPP_ASSERT to PTF_ASSERT in this branch I modified push for this.

I removed the deprecated attribute in versions of C++ < 14 or VS version < 2015 (minimum version to support the attribute).

You globally disable warning 4996 in your test code. This actually hides the fact that I marked two functions as deprecated. I removed the global disable and made it scoped. This ultimately made the test break. As a short term fix I re added it to the test code - but still scoped to the main.cpp file. I suggest you follow the compilers advice and  #define _CRT_SECURE_NO_WARNINGS or similar instead of just disabling this warning, otherwise you GLOBALLY cut off those deprecation warnings which could be baked into 3rd party libs you use and you will never know they are going to go away.

The existing `IPcapDevice::matchPacketWithFilter(filterAsString, *iter)` was flagged for a memory leak due to its static allocation of bpf_filter I assume. This is probably not a REAL leak, but i removed it from test code to get a pass - you can see this in my very last commit.

Some VS builds are still causing grief because WARNING C4996 (the deprecated function one) is getting promoted to ERROR. I will let you decide how to handle this. It will pass barring this issue.

seladb:
Please rebase this PR to the latest version of `dev`. All tests are passing there now

seladb:
@Dysl3xik please let me know what is the status of this

Dysl3xik:
Ill see if I can figure this out tonight. The original work was done on master. I just issued the PR onto dev. I would assume github would play my commit changes onto the head of dev, but maybe not?

seladb:
Actually I don't think GitHub will do that automatically. You need to fetch the latest changes from `dev` to your branch manually

Dysl3xik:
Ok I merged my new work into master, but I can't update the source branch for this PR. So I need to close this an re-issue with my dev based branch.

PR #233, opened 2019-09-04
Add BPFStringFilter class add matchPackWithFilter to GeneralFilter

Dysl3xik:
Reissue of PR #229 using dev as base branch merging into dev


Dysl3xik:
Most checks are passing. I am copying these relevant notes from last thread. If you want me to undo my "unleashing" of warning 4996 (use of deprecated stuff) let me know.

============= OLD NOTES =================

I removed the deprecated attribute in versions of C++ < 14 or VS version < 2015 (minimum version to support the attribute).

You globally disable warning 4996 in your test code. This actually hides the fact that I marked two functions as deprecated. I removed the global disable and made it scoped. This ultimately made the test break. As a short term fix I re added it to the test code - but still scoped to the main.cpp file. I suggest you follow the compilers advice and #define _CRT_SECURE_NO_WARNINGS or similar instead of just disabling this warning, otherwise you GLOBALLY cut off those deprecation warnings which could be baked into 3rd party libs you use and you will never know they are going to go away.

The existing IPcapDevice::matchPacketWithFilter(filterAsString, *iter) was flagged for a memory leak due to its static allocation of bpf_filter I assume. This is probably not a REAL leak, but i removed it from test code to get a pass - you can see this in my very last commit.

Some VS builds are still causing grief because WARNING C4996 (the deprecated function one) is getting promoted to ERROR. I will let you decide how to handle this. It will pass barring this issue.

seladb:
Thanks for moving the PR to the `dev` branch.

Unfortunately the CI checks are still failing. In VS it's because of the `deprecated` annotation and in MacOS there is a segmentation fault on Pcap++Test run. Could you please look into it?

Dysl3xik:
For VS with the deprecated issue if you want me to fix it, I am just going to revert to your old code which globally disables warning 4996.

The Mac issue is only showing up in the newer compilers for some reason, its actually passing for 3 mac builds. There is nothing substantive in the error coming out that lets me debug it, and I don't have a mac or xcode. The only option I have would be adding a bunch of print statements to find where it is dying, and then hopefully being able to see the exact cause. Then I have to commit and push and wait for ALL of the tests to run, and then repeat until this issue is fixed. Also this is passing on every other platform..... Is there an easier way to test issues with that particular build/platform?

seladb:
Regarding the 4996 warning - we can either revert back to the old code or completely remove it and change all the places that depend on the old code to go to the new code.

Regarding the MacOS issue - let me take a look, I'll get back to you.

Dysl3xik:
Sorry I just saw you got back to me on this.

Are you sure about this change? In your original implementation bpf_program was a struct and code looked like this:

```
struct bpf_program prog;
pcap_compile_nopcap(9000, pcpp::LINKTYPE_ETHERNET, &prog, filterAsString.c_str(), 1, 0)
```

vs mine

```
m_program = new bpf_program();
pcap_compile_nopcap(9000, pcpp::LINKTYPE_ETHERNET, m_program, m_filterStr.c_str(), 1, 0);
```

So I am not newing a whole filter, just the same struct which you allocated on the stack before. I also tried doing this and passing in a nullptr to `pcap_compile_nopcap` and it explodes as expected

seladb:
Yes, I verified that: if you remove the line `m_program = new bpf_program();` everything works as expected on macOS

Dysl3xik:
If you remove that line in windows it throws an exception...

I also don't understand how this makes any sense, why would you pass a NULL ptr (default value) to that function?

seladb:
Let me check that again and get back to you

seladb:
Well, this explains the issue: https://stackoverflow.com/questions/11448369/pcap-compile-expression

We cannot store a pointer to `bpf_program` without allocating its memory, and allocating memory also doesn't seem to work. On the other hand we  cannot store `bpf_program` as a member without adding `#include "pcap.h"` to `PcapFilter.h`. I'm not sure what good option we have here...

I browsed through GitHub and indeed people either define the `bpf_program` as a local variable or have it as a member.

Dysl3xik:
Well not allocating the pointer caused every build to fail as I expected.

I made an edit where I wrapped the `bpf_program` into another struct as a member, and accessed it that way. In this latest push its essentially passing tests in the same way as the previous version with `bpf_program` as a ptr where everything is passing except for XCode 9+. 

I think something else is going on here, its failing with a pointer and now failing with a struct too.

Do you have XCode on a real machine you can test with live? I am unsure how to proceed here.

Dysl3xik:
I tried one last ditch effort to see if I can allocate a `bpf_program` at all on the mac builds.

It seems you can allocate the wrapper class which holds a `bpf_program` fine on the mac builds. It appears you cannot new any class/struct which even contains a `bpf_program`.

I cant say what is causing this behavior, but I think this is an issue with the libraries behind the scene. Perhaps its some strange memory alignment requirement or something. The only other thing I can think of is if your mem plumber stuff is somehow causing issues, but I wouldn't expect that to interfere with a the already compiled lib files....???

seladb:
You were right! MemPlumber is somehow causing the issue. When I disabled it things seem to be working on MacOS.

So my suggestion would be to separate these tests from the other tests in`TestPcapFiltersOffline`, and add a tag to the new test case called `skip_mem_leak_check` to disables mem leak check for this test case. Please refer to `TestPcapLiveDeviceList` as an example where mem leak check is already disabled:

https://github.com/seladb/PcapPlusPlus/blob/24e31a84d74555bf00ad2e17848e755563a3532c/Tests/Pcap%2B%2BTest/main.cpp#L6567

In addition - please remove the prints you put in the code and in addition please take care of the failures in MSVC due to the deprecated flag.

Thank you so much for your hard work and efforts on this!

Dysl3xik:
Your last commit on dev does not pass units tests on mac, this is causing failure for the combined version now too:

GtpLayerParsingTest           : FAILED (line: 7221). assert equal failed: actual: 2001:507::1:200:8600:0:2 != expected: 2001:507:0:1:200:8600:0:2

seladb:
Yes I'm aware of that. I'll take care of the fix.

I just noticed that the PR is somehow forking from `master` again. Could you please change is to fork from `dev`? 

image

Dysl3xik:
This may show some content being merged in from master because me
maintaining my custom build, your main, and dev, but my dev branch is based
on yours. It is not straight from master. I'm not at PC to review it at the
moment.

seladb:
I just checked the previous commit and it clearly said the merge was going to `seladb:dev`:

image

I'm pretty sure that if I merged this PR it'll go directly to master. Can you please check from your side?

Dysl3xik:
You are right, I don't know how it got changed along the line, but apparently it did.

I just modified it back to dev again. Good catch!

Dysl3xik:
Thanks for working through all this with me!

seladb:
Sure :)

Thanks for all your efforts on this! I really appreciate your continuous contribution to this project!

seladb:
FYI I've just merged your changes to `master` branch.

Thanks again for your great contribution to this project! Looking forward to your next PR :)

Issue #242, opened 2019-09-25
Incorrect RADIUS packet is decoded as correct


gx740:
Hello.

During decoding of RADIUS packets (refer to `UdpLayer::parseNextLayer` method) the check of correctness of header is not carried out. I think, at least the length of header must be checked (It is better to check the correctness of whole packet with all attributes).

In the attached file is represented the incorrect RADIUS packet for which the `RadiusLayer` is created successfully. The processing of attributes of this packet leads to infinite loop when using following code:
```
for(auto attr = radiusLayer->getFirstAttribute(); attr.isNotNull(); attr = radiusLayer->getNextAttribute(attr))
{
// some code
}

```
In this case I can add the checks to eliminate this issue:
1. Checking the RADIUS header length
2. Checking the length of attribute. It must be at least 2 bytes (`attr.getTotalLength() >= 2`).

But it is not the correct way to fix it.

[radius.zip](https://github.com/seladb/PcapPlusPlus/files/3653587/radius.zip)

gx740:
Moreover I think that the following methods of base class should be overridden in RadiusLayer class:
getFirstAttribute, getNextAttribute, getAttribute, getAttributeCount.
The additional check of attribute length should be added in all methods. If the length less than 2 then the packet is incorrect and method must return TLVRecordType(NULL).
Otherwise the client's code that handles the attributes must have such checks.

seladb:
I agree with you, however these checks are not performed for any of the layers, mostly for performance reasons: parsing the whole layer well in advance will definitely cause performance degradation.

Having said that, we can add basic checks like making sure the layer length is greater than the RADIUS header. Also, we should prevent infinite loop if packet is malformed.

Would you mind making these fixes and create a PR?

gx740:
Yes, I will make changes today

seladb:
Your fix is merged into `master`. Thank you so much for your continuous contribution to this project!

PR #243, opened 2019-09-27
Radius fix

gx740:
1. Bugfix of #242.
2. Minor improvements and optimizations

I have defined the new preprocessor constant `PCPP_NOEXCEPT` in `RadiusLayer.h` but the best way is to create the new header file which will contain such constants.

I did not add the tests for changes in `getNextTLVRecord` and `getFirstTLVRecord` because I have no wrong RADIUS packets for which the method `RadiusLayer::isDataValid` returns true. However I have successfully done the tests of these methods on many wrong packets (several thousands). For this purpose I have temporarily turned off the checks of length in `UdpLayer::parseNextLayer`.


gx740:
=== Off topic ===

One more constant will be useful:

```
#if __cplusplus > 199711L || _MSC_VER >= 1800
  #define PCPP_CPP11
#endif

```

Packet++/header/TLVData.h
```diff
@@ -249,18 +260,24 @@ namespace pcpp
 		 */
 		TLVRecordType getNextTLVRecord(TLVRecordType& record, uint8_t* tlvDataBasePtr, size_t tlvDataLen)
 		{
-			if (record.isNull())
-				return TLVRecordType(NULL);
+      TLVRecordType resRec(NULL); // for NRVO optimization
```
seladb: please use the same indentation as the rest of the file (tabs instead of spaces)

Tests/Packet++Test/main.cpp
```diff
@@ -6937,7 +6937,16 @@ PTF_TEST_CASE(RadiusLayerParsingTest)
 		radiusAttr = radiusLayer->getNextAttribute(radiusAttr);
 	}
 
-
+  // incorrect RADIUS packet
```
seladb: please use the same indentation as the rest of the file (tabs instead of spaces)

Packet++/header/RadiusLayer.h
```diff
@@ -6,6 +6,14 @@
 
 /// @file
 
+#ifndef PCPP_NOEXCEPT
```
seladb: is there a particular reason to use `noexcept` here? Usually I try not to use C++11 features unless absolutely necessary

Packet++/header/TLVData.h
gx740: Done

Tests/Packet++Test/main.cpp

Packet++/header/RadiusLayer.h
gx740: The compiler can use this information to optimize calls to the function. In old C++ the throw() is used for this purpose. For inlined functions noexcept is unnecessary I guess.

gx740: I think it's better to make `isDataValid` inline. But following code should be moved into the header file:
```
#if defined(WIN32) || defined(WINx64) //for using ntohl, ntohs, etc.
#include <winsock2.h>
#elif LINUX
#include <in.h> //for using ntohl, ntohs, etc.
#elif MAC_OS_X
#include <arpa/inet.h> //for using ntohl, ntohs, etc.
#endif

```
 What You think?

seladb: I prefer not to move this code to the header file, it creates a lot of issues when linking app to PcapPlusPlus library.

Regarding `noexcept` - I think we should either keep only `throw()`, or remove `PCPP_NOEXCEPT` altogether. I don't see a point here in keeping both `throw()` (for older C++ versions) and `noexcept`.

BTW, I just noticed that there is no NULL check in `isDataValid()`, meaning if `udpData` is NULL, you might end up with seg fault. Am I missing anything?

gx740: The function isDataValid is corrected.

throw()/noexcept specification is excluded from sources because throw() is removed in C++20.

PR #244, opened 2019-09-27
New useful methods in MacAddress class

gx740:
Two new methods were added into `MacAddress` class: `rawDataSize` and `getRawData`.

`rawDataSize` allows to know needed size of buffer before calling the method `copyTo(uint8_t* arr)`.

`getRawData` is useful when we need to put MAC into the function that takes the pointer to data. Without this method the copying into temporary buffer is necessary.

### DIFF
--- a/Common++/header/MacAddress.h
+++ b/Common++/header/MacAddress.h
@@ -111,21 +111,8 @@ namespace pcpp
 				std::copy(octets.begin(), octets.end(), std::begin(m_Address));
 				#endif
 			}
-
 			return *this;
 		}
-
-		/**
-		 * Returns the size of raw data
-		 * @return The size of raw data
-		 */
-		static constexpr size_t rawDataSize() { return 6; }
-#else
-		/**
-		 * Returns the size of raw data
-		 * @return The size of raw data
-		 */
-		static size_t rawDataSize() { return 6; }
 #endif
 
 		/**
--- a/Common++/header/MacAddress.h
+++ b/Common++/header/MacAddress.h
@@ -81,7 +81,6 @@ namespace pcpp
 				std::copy(octets.begin(), octets.end(), std::begin(m_Address));
 				#endif
 			}
-				
 		}
 #endif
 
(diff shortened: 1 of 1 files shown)


Common++/header/MacAddress.h
```diff
@@ -107,16 +106,34 @@ namespace pcpp
...
 			}
 
 			return *this;
 		}
+
+		/**
+		 * Returns the size of raw data
+		 * @return The size of raw data
+		 */
+		static constexpr size_t rawDataSize() { return 6; }
```
seladb: why is it needed to implement the same method twice (one with `constexpr` and one without it)?

Common++/header/MacAddress.h
```diff
@@ -107,16 +106,34 @@ namespace pcpp
...
+#else
+		/**
+		 * Returns the size of raw data
+		 * @return The size of raw data
+		 */
+		static size_t rawDataSize() { return 6; }
```
seladb: are you using these new methods anywhere in the code?

Common++/header/MacAddress.h
gx740: `constexpr` allows to use this method in expressions calculated on compile time. The simplest example: `std::array<uint8_t, MacAddress::rawDataSize()> raw`.

https://docs.microsoft.com/en-us/cpp/cpp/constexpr-cpp?view=vs-2019

Common++/header/MacAddress.h
gx740: Yes, I will use this methods. In my project the MAC addresses are saved into the binary file queue and later are loaded into database. My current code:

```
uint8_t mac[6];
macData->value.mac.copyTo(mac);
mFieldMac.assign(mac, sizeof mac);

```
1. It is not a good idea to use directly defined size.
2. I am forced to copy MAC's bytes into temporary array to pass them into the function.

The class has `void copyTo(uint8_t* arr)` method therefore it is good to know the size of the buffer.

Other way is to create a constant instead of a method: enum { RawDataSize = 6 }

gx740: Other way is to create a constant instead of a method: enum { RawDataSize = 6 }

Common++/header/MacAddress.h
seladb: Since this method returns a const number (with the value of 6), why is it have to be a method?

Why not creating a const public member of even a macro for that?

Common++/header/MacAddress.h
seladb: As I replied earlier, we should consider converting this from a method to something else, preferably a static const member or a macro

gx740: I have removed rawDataSize method and let's not create a new constant.

PR #251, opened 2019-10-01
Fixing an error with buffer overrun

gx740:
If `m_Name` will contain a pointer to C-string with length more than 16 bytes then buffer overrun will occur.

From the MAN page for `strncpy`: 

> Warning: If there is no null byte among the first n bytes of src, the string placed in dest will not be null-terminated

### DIFF
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -671,7 +671,7 @@ void PcapLiveDevice::setDeviceMtu()
 	struct ifreq ifr;
 
 	memset(&ifr, 0, sizeof(ifr));
-	strncpy(ifr.ifr_name, m_Name, sizeof(ifr.ifr_name));
+	strncpy(ifr.ifr_name, m_Name, sizeof(ifr.ifr_name) - 1);
 
 	int socketfd = socket(AF_INET, SOCK_DGRAM, IPPROTO_IP);
 	if (ioctl(socketfd, SIOCGIFMTU, &ifr) == -1)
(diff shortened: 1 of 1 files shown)


gx740:
g++ 8.3.1 shows next warnings:

```
Building file: src/WinPcapLiveDevice.cpp
src/PcapLiveDevice.cpp: In member function ‘void pcpp::PcapLiveDevice::setDeviceMtu()’:
src/PcapLiveDevice.cpp:674:9: warning: ‘char* strncpy(char*, const char*, size_t)’ specified bound 16 equals destination size [-Wstringop-truncation]
  strncpy(ifr.ifr_name, m_Name, sizeof(ifr.ifr_name));
  ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/PcapLiveDevice.cpp: In member function ‘void pcpp::PcapLiveDevice::setDeviceMacAddress()’:
src/PcapLiveDevice.cpp:730:9: warning: ‘char* strncpy(char*, const char*, size_t)’ specified bound 16 equals destination size [-Wstringop-truncation]
  strncpy(ifr.ifr_name, m_Name, sizeof(ifr.ifr_name));
  ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -671,7 +671,7 @@ void PcapLiveDevice::setDeviceMtu()
 	struct ifreq ifr;
 
 	memset(&ifr, 0, sizeof(ifr));
-	strncpy(ifr.ifr_name, m_Name, sizeof(ifr.ifr_name));
+	strncpy(ifr.ifr_name, m_Name, sizeof(ifr.ifr_name) - 1);
```
seladb: why is it required to reduce the size by 1? My understanding is that `sizeof(ifr.ifr_name)` should give the size of the name

gx740: `Sizeof` returns the max size of buffer including trailing '\0'. In case the length of `m_Name` is more than or equal to `sizeof(ifr.ifr_name)` then `strncpy` will fill all bytes of the buffer but the last byte will not be '\0' so the destination string is not null-terminated. Look into man on strncpy. Reducing the size by 1 we guarantee the destination string will be null-terminated.

seladb: ok got it, thanks for the explanation. I'll merge the PR

Issue #250, opened 2019-10-01
TcpReassembly, memory leakage


gx740:
When the session is removed the `flowKey` is put into `m_ClosedConnectionList` map.
Unfortunately this map never clears.

seladb:
Yes, this is probably an issue in high scale.

On one hand we need to keep a list of closed connections in case we see a packet that comes after the FIN packet (which happens sometimes due to network issues). On the other hand we probably don't need to keep it forever. 

When I built this feature I was thinking whether to build a cleanup mechanism or not and I've decided to postpone it until the demand comes. This mechanism should be carefully designed to avoid performance overhead.

Another approach might be moving the cleanup responsibility to the user by exposing a method to do so. But this will complicate the API and will make TcpReassembly more difficult to use.

Please let me know what you think.

seladb:
@gx740 PR is merged and pushed to `master`, please let me know if we can close this issue

gx740:
You can close this PR.

Issue #248, opened 2019-09-30
TcpReassembly (question)


gx740:
I have found that the methods `closeConnectionInternal` and `closeAllConnections` have different behavior in relation to `m_ConnectionInfo` vector. The `closeAllConnections` clears this vector while closeConnectionInternal does not remove the data from it.

Should `closeConnectionInternal` remove data from vector or not?

gx740:
I think this is a bug.

seladb:
Yes I think you're right, good job on catching it!

Since I wrote this code a long back, I'd suggest testing the change carefully as I'm not sure if I had a good reason for not removing the `ConnectionData` from `m_ConnectionInfo`

gx740:
I have looked into the code and found that there is one reason not to remove connection info: if the client's code wants to use this info after session removing. There is the method getConnectionInformation to get a const reference to m_ConnectInfo at any time. But I don't see the reasonable use case for such idea. We just get a memory leakage.

gx740:
It's a little about getConnectionInformation method. It seems to me that this method is quite useless. Why? It returns an unsorted vector and we have no way of quick access by flowKey.

seladb:
Thanks for checking, I think that was probably the reason I decided not to remove the connection info. I do agree that in large scale this will cause a memory leak, but maybe there are use cases where you need to query this data. 

I also agree that for a large number of connection using `getConnectionInformation()` will be quite useless. Maybe we should consider replacing the vector with a map?

To summarize, maybe we can add the following features:

- Replace the connection info vector with a map (which its key is flow key)
- Add a flag indicating whether to remove the connection data after a connection is closed
- Add a method for cleaning connection data upon user request

gx740:
Yes we can add these features.
Should the flag be a property of class or a parameter of function?

I think additional optimizations can be done (maybe in another PR):
1. Removing m_ClosedConnectionList at all
2. m_ConnectionList will store a pointer to data or null. Null means the connection is closed.

seladb:
As per my answer in #250 we cannot simply remove `m_ClosedConnectionList`, but we can create a mechanism to clean it a short time after a connection is closed (maybe a few seconds?). 

When I come to think of it maybe the same mechanism should apply for `m_ConnectionInfo` as well to provide a good reflection of the flows currently being handled in TcpReassebmly.

So maybe we can divide the work into 2 phases:

1. Convert `m_ConnectionInfo` into a map
[...]

PR #256, opened 2019-10-09
Fixing memory leak in TcpReassembly

gx740:
Bugfix #248 and #250.

### DIFF
--- a/Packet++/header/TcpReassembly.h
+++ b/Packet++/header/TcpReassembly.h
@@ -213,12 +213,6 @@ public:
 	 */
 	const ConnectionData& getConnectionDataRef() const { return m_Connection; }
 
-	/**
-	 * A getter for the connection data
-	 * @return The const reference to connection data
-	 */
-	const ConnectionData& getConnectionDataRef() { return m_Connection; }
-
 private:
 	uint8_t* m_Data;
 	size_t m_DataLen;
@@ -237,17 +231,22 @@ private:
 struct TcpReassemblyConfiguration
 {
 	/** The flag indicating whether to remove the connection data after a connection is closed */
-	bool doNotRemoveConnInfo;
+	bool removeConnInfo;
+
 	/** How long the closed connections will not be cleaned up. The value is expressed in seconds. If the value is set to 0 then TcpReassembly should use the default value */
 	uint32_t closedConnectionDelay;
+
 	/** The maximum number of items to be cleaned up per one call of purgeClosedConnections. If the value is set to 0 then TcpReassembly should use the default value */
 	uint32_t maxNumToClean;
 
 	/**
 	 * A c'tor for this struct
+	 * @param[in] removeConnInfo The flag indicating whether to remove the connection data after a connection is closed. The default is true
+	 * @param[in] closedConnectionDelay How long the closed connections will not be cleaned up. The value is expressed in seconds. If it's set to 0 the default value will be used. The default is 5.
+	 * @param[in] maxNumToClean The maximum number of items to be cleaned up per one call of purgeClosedConnections. If it's set to 0 the default value will be used. The default is 30.
 	 */
-	TcpReassemblyConfiguration(bool doNotRemoveConnInfo = false, uint32_t closedConnectionDelay = 5, uint32_t maxNumToClean = 5000) :
-		doNotRemoveConnInfo(doNotRemoveConnInfo), closedConnectionDelay(closedConnectionDelay), maxNumToClean(maxNumToClean)
+	TcpReassemblyConfiguration(bool removeConnInfo = true, uint32_t closedConnectionDelay = 5, uint32_t maxNumToClean = 30) :
+		removeConnInfo(removeConnInfo), closedConnectionDelay(closedConnectionDelay), maxNumToClean(maxNumToClean)
 	{
 	}
 };
@@ -347,8 +346,8 @@ public:
 	void closeAllConnections();
 
 	/**
-	 * Get a list of all connections managed by this TcpReassembly instance (both connections that are open and those that are already closed)
-	 * @return A list of all connections managed. Notice this list is constant and cannot be changed by the user
+	 * Get a map of all connections managed by this TcpReassembly instance (both connections that are open and those that are already closed)
+	 * @return A map of all connections managed. Notice this map is constant and cannot be changed by the user
 	 */
 	const ConnectionInfoList &getConnectionInformation() const { return m_ConnectionInfo; }
 
@@ -363,8 +362,9 @@ public:
 	 * Clean up the closed connections from the memory
 	 * @param[in] currentTime Current time
 	 * @param[in] maxNumToClean The maximum number of items to be cleaned up per one call. This parameter, when its value is not zero, overrides the value that was set by the constructor.
+	 * @return The number of cleared items
 	 */
-	void purgeClosedConnections(time_t currentTime = time(NULL), uint32_t maxNumToClean = 0);
+	uint32_t purgeClosedConnections(time_t currentTime = time(NULL), uint32_t maxNumToClean = 0);
 
 private:
 	struct TcpFragment
@@ -414,7 +414,7 @@ private:
 	ClosedConnectionList m_ClosedConnectionList;
 	ConnectionInfoList m_ConnectionInfo;
 	CleanupList m_CleanupList;
-	bool m_DoNotRemoveConnInfo;
+	bool m_RemoveConnInfo;
 	uint32_t m_ClosedConnectionDelay;
 	uint32_t m_MaxNumToClean;
 	time_t m_PurgeTimepoint;
@@ -427,7 +427,7 @@ private:
 
 	void closeConnectionInternal(uint32_t flowKey, ConnectionEndReason reason);
 
-	void insertIntoCleanupList(uint32_t flowKey, time_t currentTime = time(NULL));
+	void insertIntoCleanupList(uint32_t flowKey);
 };
 
 }
--- a/Packet++/header/TcpReassembly.h
+++ b/Packet++/header/TcpReassembly.h
@@ -5,6 +5,9 @@
 #include "IpAddress.h"
 #include "PointerVector.h"
 #include <map>
+#include <list>
+#include <time.h>
+
 
 /**
  * @file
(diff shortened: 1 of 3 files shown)


Packet++/header/TcpReassembly.h
```diff
@@ -208,6 +230,29 @@ class TcpStreamData
...
+	bool doNotRemoveConnInfo;
+	/** How long the closed connections will not be cleaned up. The value is expressed in seconds. If the value is set to 0 then TcpReassembly should use the default value */
+	uint32_t closedConnectionDelay;
+	/** The maximum number of items to be cleaned up per one call of purgeClosedConnections. If the value is set to 0 then TcpReassembly should use the default value */
+	uint32_t maxNumToClean;
+
+	/**
+	 * A c'tor for this struct
+	 */
+	TcpReassemblyConfiguration(bool doNotRemoveConnInfo = false, uint32_t closedConnectionDelay = 5, uint32_t maxNumToClean = 5000) :
```
seladb: do we really want to set `maxNumToClean` default value to 5000? I was thinking more in the area of 10-50.

In a live traffic environment `reassemblePacket()` will be called a lot so if there are peaks of many connections being closed I don't want that to slow down the TCP reassembly process

Packet++/header/TcpReassembly.h
```diff
@@ -299,7 +350,7 @@ class TcpReassembly
 	 * Get a list of all connections managed by this TcpReassembly instance (both connections that are open and those that are already closed)
 	 * @return A list of all connections managed. Notice this list is constant and cannot be changed by the user
 	 */
```
seladb: Please change from "list" to "map" where the key is flowkey

Packet++/header/TcpReassembly.h
```diff
@@ -308,6 +359,13 @@ class TcpReassembly
 	 */
 	int isConnectionOpen(const ConnectionData& connection) const;
 
+	/**
+	 * Clean up the closed connections from the memory
+	 * @param[in] currentTime Current time
+	 * @param[in] maxNumToClean The maximum number of items to be cleaned up per one call. This parameter, when its value is not zero, overrides the value that was set by the constructor.
+	 */
+	void purgeClosedConnections(time_t currentTime = time(NULL), uint32_t maxNumToClean = 0);
```
seladb: do you really need the param `currentTime`? It is always set to `time(NULL)`

Packet++/header/TcpReassembly.h
```diff
@@ -359,6 +426,8 @@ class TcpReassembly
 	void handleFinOrRst(TcpReassemblyData* tcpReassemblyData, int sideIndex, uint32_t flowKey);
 
 	void closeConnectionInternal(uint32_t flowKey, ConnectionEndReason reason);
+
+	void insertIntoCleanupList(uint32_t flowKey, time_t currentTime = time(NULL));
```
seladb: do you really need the param `currentTime`? the default value `time(NULL)` is always being used

Packet++/src/TcpReassembly.cpp
```diff
@@ -800,4 +810,38 @@ int TcpReassembly::isConnectionOpen(const ConnectionData& connection) const
...
+void TcpReassembly::insertIntoCleanupList(uint32_t flowKey, time_t currentTime)
+{
+	std::pair<CleanupList::iterator, bool> pair = m_CleanupList.insert(std::make_pair(currentTime + m_ClosedConnectionDelay, CleanupList::mapped_type()));
+	CleanupList::mapped_type &keysList = pair.first->second;
+	keysList.push_back(flowKey);
+}
+
+void TcpReassembly::purgeClosedConnections(time_t currentTime, uint32_t maxNumToClean)
+{
+	uint32_t count = 0;
```
seladb: this method implementation is pretty complex in its nature. It'd beneficial if you can add comments to make it easier for people going over the code to understand its flow

Packet++/header/TcpReassembly.h
```diff
@@ -197,6 +213,12 @@ class TcpStreamData
 	 */
 	const ConnectionData& getConnectionDataRef() const { return m_Connection; }
 
+	/**
+	 * A getter for the connection data
+	 * @return The const reference to connection data
+	 */
+	const ConnectionData& getConnectionDataRef() { return m_Connection; }
```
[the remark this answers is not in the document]
gx740: Fixed.

Packet++/header/TcpReassembly.h
```diff
@@ -208,6 +230,29 @@ class TcpStreamData
...
+ * A structure for configuring the TcpReassembly class
+ */
+struct TcpReassemblyConfiguration
+{
+	/** The flag indicating whether to remove the connection data after a connection is closed */
+	bool doNotRemoveConnInfo;
+	/** How long the closed connections will not be cleaned up. The value is expressed in seconds. If the value is set to 0 then TcpReassembly should use the default value */
+	uint32_t closedConnectionDelay;
+	/** The maximum number of items to be cleaned up per one call of purgeClosedConnections. If the value is set to 0 then TcpReassembly should use the default value */
+	uint32_t maxNumToClean;
```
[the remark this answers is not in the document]
gx740: Done

Packet++/header/TcpReassembly.h
```diff
@@ -208,6 +230,29 @@ class TcpStreamData
...
 
 
+/**
+ * @struct TcpReassemblyConfiguration
+ * A structure for configuring the TcpReassembly class
+ */
+struct TcpReassemblyConfiguration
+{
+	/** The flag indicating whether to remove the connection data after a connection is closed */
+	bool doNotRemoveConnInfo;
```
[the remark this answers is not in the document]
gx740: Ok

Packet++/header/TcpReassembly.h
```diff
@@ -208,6 +230,29 @@ class TcpStreamData
...
+	/** How long the closed connections will not be cleaned up. The value is expressed in seconds. If the value is set to 0 then TcpReassembly should use the default value */
+	uint32_t closedConnectionDelay;
+	/** The maximum number of items to be cleaned up per one call of purgeClosedConnections. If the value is set to 0 then TcpReassembly should use the default value */
+	uint32_t maxNumToClean;
+
+	/**
+	 * A c'tor for this struct
```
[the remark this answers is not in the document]
gx740: Done

Packet++/header/TcpReassembly.h
gx740: The default has been changed to 30

Packet++/header/TcpReassembly.h
gx740: Done

Packet++/header/TcpReassembly.h
gx740: Param has been removed.

Packet++/header/TcpReassembly.h
gx740: Param has been removed

Packet++/header/TcpReassembly.h
gx740: This parameter allows to use the value that is got earlier and saved in local/global variable. It's a small optimization. For example: I have  a global variable of type `volatile time_t` in my code. This variable is being updated by certain thread. Therefore I just use this variable instead of calling the function `time`.

gx740: I am planning to change the signature of this method as:

`uint32_t purgeClosedConnections(time_t currentTime = time(NULL), uint32_t maxNumToClean = 0);`

The return value can be useful for performance measurement.

Packet++/src/TcpReassembly.cpp
gx740: Comments were added

Packet++/header/TcpReassembly.h
gx740: Following code is used in `reassemblyPacket`:

```
  // automatic cleanup
  if(m_RemoveConnInfo == true)
  {
    time_t currentTime = time(NULL);
    if(currentTime >= m_PurgeTimepoint)
    {
      purgeClosedConnections(currentTime, 0);
      m_PurgeTimepoint = time(NULL) + PURGE_FREQ_SECS;
    }
  }
```
Without this parameter time() will be called twice.

Packet++/header/TcpReassembly.h
```diff
@@ -208,6 +224,34 @@ class TcpStreamData
...
+ * @struct TcpReassemblyConfiguration
+ * A structure for configuring the TcpReassembly class
+ */
+struct TcpReassemblyConfiguration
+{
+	/** The flag indicating whether to remove the connection data after a connection is closed */
+	bool removeConnInfo;
+
+	/** How long the closed connections will not be cleaned up. The value is expressed in seconds. If the value is set to 0 then TcpReassembly should use the default value */
+	uint32_t closedConnectionDelay;
```
seladb: It'd be useful to mention that this parameter is only relevant if `removeConnInfo` is equal to `true`

Packet++/header/TcpReassembly.h
```diff
@@ -208,6 +224,34 @@ class TcpStreamData
...
+
+	/** The maximum number of items to be cleaned up per one call of purgeClosedConnections. If the value is set to 0 then TcpReassembly should use the default value */
+	uint32_t maxNumToClean;
```
seladb: It'd be useful to mention that this parameter is only relevant if `removeConnInfo` is equal to `true`

Tests/Pcap++Test/main.cpp
```diff
@@ -5259,37 +5293,97 @@ PTF_TEST_CASE(TestTcpReassemblyIPv6_OOO)
...
+	iterConn1 = managedConnections.find(flowKeys[0]);
+	iterConn2 = managedConnections.find(flowKeys[1]);
+	iterConn3 = managedConnections.find(flowKeys[2]);
+	PTF_ASSERT(iterConn1 != managedConnections.end(), "Connection #1 not found in flow keys list");
+	PTF_ASSERT(iterConn2 != managedConnections.end(), "Connection #2 not found in flow keys list");
+	PTF_ASSERT(iterConn3 != managedConnections.end(), "Connection #3 not found in flow keys list");
+	PTF_ASSERT(tcpReassembly.isConnectionOpen(iterConn1->second) == -1, "Connection #1 still exists");
+	PTF_ASSERT(tcpReassembly.isConnectionOpen(iterConn2->second) == -1, "Connection #2 still exists");
+	PTF_ASSERT(tcpReassembly.isConnectionOpen(iterConn3->second) == -1, "Connection #3 still exists");
 }
```
seladb: Since this is a new test case, you can use the new assertion macros I recently added: `PTF_ASSERT_EQUAL`, `PTF_ASSERT_TRUE`, `PTF_ASSERT_FALSE`, `PTF_ASSERT_NULL`, `PTF_ASSERT_NOT_NULL` and `PTF_ASSERT_BUF_COMPARE`. 

Please take a look at the new test cases that are already using them: https://github.com/seladb/PcapPlusPlus/blob/220434db398e68b5fc5ae473bbb439a61f485dda/Tests/Packet%2B%2BTest/main.cpp#L7093

Eventually I'd like to migrate all test cases to use them, but it will probably be a long process. In the meantime, let's try to use them in the new test cases we're writing

Packet++/header/TcpReassembly.h
seladb: I do understand the optimization, however we should consider what is the overhead of `time()` (which is something I'm not sure about). If the overhead is small then it's probably better and cleaner to call it one more time. If the overhead is big then I agree with your approach.

gx740: The overhead is small. I will remove this parameter.

Tests/Pcap++Test/main.cpp
seladb: Unfortunately not yet. You can leave this one as it is

PR #259, opened 2019-10-18
IpAddress: bugfix and improvements

gx740:


Common++/src/IpAddress.cpp
```diff
@@ -180,7 +178,7 @@ IPv6Address::IPv6Address(const IPv6Address& other)
 
 IPv6Address::~IPv6Address()
 {
-	delete m_pInAddr;
+	delete[] m_pInAddr;
```
seladb: why did you add `[]` here?

Common++/src/IpAddress.cpp
```diff
@@ -246,10 +244,8 @@ bool IPv6Address::operator!=(const IPv6Address& other)
 
 IPv6Address& IPv6Address::operator=(const IPv6Address& other)
 {
-	if (m_pInAddr != NULL)
-		delete m_pInAddr;
-
-	m_pInAddr = (in6_addr*)new uint8_t[sizeof(in6_addr)];
+	if (m_pInAddr == NULL)
+		m_pInAddr = (in6_addr*)new uint8_t[sizeof(in6_addr)];
```
seladb: I'm not sure why you did this change. Could you please explain?

Common++/src/IpAddress.cpp
```diff
@@ -131,10 +131,8 @@ uint32_t IPv4Address::toInt() const
 
 IPv4Address& IPv4Address::operator=(const IPv4Address& other)
 {
-	if (m_pInAddr != NULL)
-		delete m_pInAddr;
-
-	m_pInAddr = new in_addr();
+	if (m_pInAddr == NULL)
+		m_pInAddr = new in_addr();
```
seladb: I'm not sure why you did this change. Could you please explain?

Common++/src/IpAddress.cpp
gx740: Because in constructor of IPv6Address the memory is allocated by by `new[]`. Refer to line 171.

`m_pInAddr = (in6_addr*)new uint8_t[sizeof(in6_addr)];`

The current implementation leads to memory leakage.

gx740: When I fixed this issue, the memory consumption decreased from 1.5Gbytes downto 490 Mbytes in my application.

seladb: Got it, that makes sense. Thanks for discovering this bug!

Common++/src/IpAddress.cpp
gx740: The buffer `m_pInAddr` have the constant size `sizeof(in6_addr)`. The copy operation is made with the same size: `memcpy(m_pInAddr, other.m_pInAddr, sizeof(in6_addr))`. So if a pointer to buffer is not null we can just copy the data into it. It eliminates, at least, delete operator.
Besides, we can remove the condition `if (m_pInAddr == NULL)` because the buffer is always allocated by constructors.

Common++/src/IpAddress.cpp
gx740: The same as for IPv6Address.

Common++/src/IpAddress.cpp
gx740: More precisely, we can remove following code at all:
```
if (m_pInAddr == NULL)
  m_pInAddr = (in6_addr*)new uint8_t[sizeof(in6_addr)];

```

PR #264, opened 2019-10-24
Feature: let the user add custom LDFLAGS to the main Makefile

rom1sqr:
Generally speaking, but especially with [archlinux](https://www.archlinux.org/) [makepkg](https://www.archlinux.org/pacman/makepkg.8.html), it is good practice to compile binaries with full [RELRO](http://tk-blog.blogspot.com/2009/02/relro-not-so-well-known-memory.html) support (`-Wl,-z,relro,-z,now`). Otherwise, [namcap](https://wiki.archlinux.org/index.php/Namcap) complains about it.

> pcapplusplus W: ELF file ('usr/bin/TcpReassembly') lacks FULL RELRO, check LDFLAGS.
> [...]

You can check the files produced by the linux compilation process with [checkrelro.sh](http://www.trapkit.de/tools/checkrelro.sh).

> $ ./checkrelro.sh --file Dist/examples/TcpReassembly
> Dist/examples/TcpReassembly - partial RELRO

This commit enables the linker flags to be added, either via the user environment, or through building tools. Especially with makepkg, who gives the standard `LDFLAGS` variable via `/etc/makepkg.conf`.

> $ grep LDFLAGS /etc/makepkg.conf
> LDFLAGS="-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now"

Feel free to ask if you have any further question.

Thanks

### DIFF
--- a/mk/PcapPlusPlus.mk.linux
+++ b/mk/PcapPlusPlus.mk.linux
@@ -6,4 +6,6 @@ PCAPPP_INCLUDES += -I/usr/include/netinet
 # libs
 PCAPPP_LIBS += -lpcap -lpthread
 
+# allow user to add custom LDFLAGS
+PCAPPP_BUILD_FLAGS += $(LDFLAGS)
 


seladb:
Thanks for providing this info!

So just to make sure: in [the link](http://tk-blog.blogspot.com/2009/02/relro-not-so-well-known-memory.html) you sent (which is pretty old) it is mentioned that RELRO is enabled by default in Ubuntu 8.10 and openSUSE 11.1. I'm guessing that by now it is enabled by default on all modern Linux distros. In that case, why is `checkrelro.sh` showing that `TcpReassembly` supports RELRO only partially?

rom1sqr:
Unless i'm mistaken, the article says that **partial RERLO** is enabled by default. Not the full one :)

seladb:
Yes I'm sorry, you're right. 

Actually `LDFLAGS` can be used for other purposes besides RELRO so I think it's a good addition to PcapPlusPlus. I'll merge the PR

PR #283, opened 2019-11-04
Improvements in IPReassembly and LRUList

### LINKED ISSUE #281: LRUList: the complexity of method "put": is not O(1)
This method has the following code:
```
if (m_CacheItemsList.size() > m_MaxSize)
{
  ListIterator lruIter = m_CacheItemsList.end();
  lruIter--;
  T* deletedValue = new T(*lruIter);
  m_CacheItemsMap.erase(*lruIter);
  m_CacheItemsList.erase(lruIter);

  return deletedValue;
}
```
`m_CacheItemsList` is of type `std::list`. Unfortunately, the complexity of method `size` of linked list can be O(n) (refer to https://en.cppreference.com/w/cpp/container/list/size). The C++ standard guarantees the constant complexity for C++ 11 and newer. We need to rework this method because it is the significant impact on performance (LRUList is used by IPReassembly).

I have made the tests under CentOS-7 with g++ 8.3.1 and got that one call of `size()` is performed 2.4 milliseconds (the size of list was 500000). The same result will be for the default compiler (g++ 4.8).

I suggest to use an internal counter of elements stored in `m_CacheItemsList`.

gx740:
1. Fix for #281
2. Code of `IPReassembly` was simplified: the type of private member `m_PacketLRU` was changed from `LRUList<uint32_t>*` to `LRUList<uint32_t>`
3. Performance improvements in both classes

I have prepared the new optimized method `put` in `LRUList` so I suggest to mark the old one as deprecated.

### DIFF
--- a/Common++/header/LRUList.h
+++ b/Common++/header/LRUList.h
@@ -40,7 +40,7 @@ namespace pcpp
 		/**
 		 * Puts an element in the list. This element will be inserted (or advanced if it already exists) to the head of the
 		 * list as the most recently used element. If the list already reached its max size and the element is new this method
-		 * will remove the least recently used element and return a pointer to it. Method complexity is O(log(size()))
+		 * will remove the least recently used element and return a pointer to it. Method complexity is O(log(getSize()))
 		 * @param[in] element The element to insert or to advance to the head of the list (if already exists)
 		 * @return A pointer to the element that was removed from the list in case the list already reached its max size.
 		 * If the list didn't reach its max size NULL will be returned. Notice it's the responsibility of the user to free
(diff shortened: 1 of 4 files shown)


gx740:
The current implementation creates a new dynamically allocated object by copying the existed one. The main goal of the new method `put` is to avoid memory allocation. Another benefit is that the copying can be avoided if the user passes NULL pointer as the last argument.

gx740:
The new method is based on the old one.

seladb:
Regarding this PR - I'm still waiting for PR #285 to be merged (which also touches IPReassemly), then I'll review and merge this PR

gx740:
The changes in these PRs do not cross. I have checked.

gx740:
Something went wrong. This PR should be closed because it was branched from master not dev.

seladb:
Yes. Please reopen it from `dev`

PR #287, opened 2019-11-05
Improvements in LRUList and IPReassembly

gx740:
1. Fix for #281
2. Code of `IPReassembly` was simplified: the type of private member `m_PacketLRU` was changed from `LRUList<uint32_t>*` to `LRUList<uint32_t>`
3. Performance improvements in both classes

The current implementation of `LRUList::put` creates a new dynamically allocated object by copying the existed one. The main goal of the new method is to avoid memory allocation. Another benefit is that the copying can be avoided if the user passes `NULL` as the last argument. The new method is based on the old one.

I suggest to mark the old method `put` as deprecated or to remove at all (this changes the interface of LRUList). If old method will be removed we can change the signature of new method from `int put(const T& element, T* deletedValue` to `int put(const T& element, T* deletedValue = NULL)`

### DIFF
--- a/Common++/header/LRUList.h
+++ b/Common++/header/LRUList.h
@@ -44,49 +44,14 @@ namespace pcpp
 		/**
 		 * Puts an element in the list. This element will be inserted (or advanced if it already exists) to the head of the
 		 * list as the most recently used element. If the list already reached its max size and the element is new this method
-		 * will remove the least recently used element and return a pointer to it. Method complexity is O(log(getSize()))
-		 * @param[in] element The element to insert or to advance to the head of the list (if already exists)
-		 * @return A pointer to the element that was removed from the list in case the list already reached its max size.
-		 * If the list didn't reach its max size NULL will be returned. Notice it's the responsibility of the user to free
-		 * this pointer's memory when done using it
-		 */
-		T* put(const T& element)
-		{
-			m_CacheItemsList.push_front(element);
-
-			// Inserting a new element. If an element with an equivalent key already exists the method returns an iterator to the element that prevented the insertion
-			std::pair<MapIterator, bool> pair = m_CacheItemsMap.insert(std::make_pair(element, m_CacheItemsList.begin()));
-			if(pair.second == false) // already exists
-			{
-				m_CacheItemsList.erase(pair.first->second);
-				pair.first->second = m_CacheItemsList.begin();
-			}
-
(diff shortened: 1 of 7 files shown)


Common++/header/LRUList.h
```diff
@@ -58,7 +62,7 @@ namespace pcpp
 				pair.first->second = m_CacheItemsList.begin();
 			}
 
-			if (m_CacheItemsList.size() > m_MaxSize)
+			if (m_CacheItemsMap.size() > m_MaxSize)
```
gx740: Fix for #281

seladb:
Could you please rebase this PR to the latest changes in `dev`?

gx740:
Done

Common++/header/LRUList.h
```diff
@@ -72,11 +76,52 @@ namespace pcpp
...
+			{
+				ListIterator lruIter = m_CacheItemsList.end();
+				lruIter--;
+
+				if (deletedValue != NULL)
+#if __cplusplus > 199711L || _MSC_VER >= 1800
+					*deletedValue = std::move(*lruIter);
+#else
+					*deletedValue = *lruIter;
+#endif
```
seladb: This is a nice optimization! However I don't think we need 2 versions of `put()`, we should choose between the two. This version is more optimized but the API is a bit less readable IMO. What do you think?

gx740: I suggest to remove the old method and to change the new one to following:

`int put(const T& element, T* deletedValue = NULL)`

The user should only make changes to process the new type of return value (`deletedValue` has the default value).

seladb: Yes I agree. Please notice `LRUList` is also used in 2 examples: `PcapSplitter` and `TcpReassembly`. Please make sure that you make the changes there too. In addition, I'd really appreciate if you can run some tests on these examples after the change to make sure they behave correctly

gx740: The changes have been made. I have run the tests with following parameters and had the same results for different versions of certain application:

- `TcpReassembly -r three_http_streams.pcap -f 2`

- `PcapSplitter -f sll.pcap -o old -m client-ip -p 5`

In this case I have changed the `MAX_NUMBER_OF_CONCURRENT_OPEN_FILES` constant to 5 before compilation.

In addition, `valgrind` has found the memory leakages in examples.

PR #288, opened 2019-11-05
Performance improvement of SSLLayer::IsSSLMessage()

gx740:
1. Performance of SSLLayer::IsSSLMessage method was significantly improved
2. New SSL ports were added (I am not sure that it is necessary)

### DIFF
--- a/Packet++/header/SSLLayer.h
+++ b/Packet++/header/SSLLayer.h
@@ -248,7 +248,7 @@ namespace pcpp
 		 */
 		void parseNextLayer();
 
-    OsiModelLayer getOsiModelLayer() const { return OsiModelPresentationLayer; }
+		OsiModelLayer getOsiModelLayer() const { return OsiModelPresentationLayer; }
 
 	protected:
 		SSLLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet) : Layer(data, dataLen, prevLayer, packet) { m_Protocol = SSL; }
--- a/Packet++/header/SSLLayer.h
+++ b/Packet++/header/SSLLayer.h
@@ -1,8 +1,6 @@
 #ifndef PACKETPP_SSL_LAYER
 #define PACKETPP_SSL_LAYER
 
-#include <map>
-#include <vector>
 #include "PointerVector.h"
 #include "Layer.h"
 #include "SSLCommon.h"
@@ -175,6 +173,12 @@ namespace pcpp
 	public:
 
 		/**
+		 * A static method that checks whether the port is considered as SSL/TLS
+		 * @param[in] port The port number to be checked
+		 */
+		static bool isSSLPort(uint16_t port);
+
+		/**
 		 * A static methods that gets raw data of a layer and checks whether this data is a SSL/TLS record or not. This check is
 		 * done using the source/dest port and matching of a legal record type in the raw data. The list of ports identified
 		 * as SSL/TLS is hard-coded and includes the following ports:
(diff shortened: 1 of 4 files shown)


gx740:
> In this PR you're replacing the port map with a simple switch case statement, but at the same time keeping the port map (which is now being used only in `SSLLayer::getSSLPortMap()`.

Unfortunately, `SSLLayer::getSSLPortMap()` had been made public so I could not simply remove this map.

> I think there's no point in this duplicate. Why is the switch case statement faster than the map? Is it somehow related to the benchmark you made in #290 ?

Operator `switch` will faster than the lookup in map (or sorted vector) any time:

1. Its complexity is constant
2. Generally, the compilers are optimized it very well

> Instead of keeping both map and switch case statement, what we can do is:
> 
> * Keep the map as is
> * Add this to optimize for most cases:
>   ```
>   if (port == 443) // HTTPS, this is likely case
>        return true;
>   ```

It's better to remove `SSLLayer::getSSLPortMap()` at all because it's useless for end user. Why?
- The SSL ports are well known and the new ones are rarely added
- Lookup in map (or in vector as described in #290) is very slow comparing to switch operator
- The logic of SSLLayer does not depend on certain port number

I've made the tests for the worst case (lookup of 995):
```
Measurement of lookup performance: container size is 14
Operator switch: duration(ms): 15, iterations: 1000000
Lookup in map: duration(ms): 1045, iterations: 1000000
```

Source code of tests:
```
#include <vector>
#include <map>
#include <cstdint>
#include <chrono>
#include <iostream>
#include <algorithm>

static std::map<uint16_t, bool> createSSLPortMap()
{
  std::map<uint16_t, bool> result;

  result[0] = true; //default
  result[261] = true; //NSIIOPS
  result[443] = true; //HTTPS
  result[448] = true; //DDM-SSL
  result[465] = true; //SMTPS
  result[563] = true; //NNTPS
  result[614] = true; //SSHELL
  result[636] = true; //LDAPS
  result[989] = true; //FTPS - data
  result[990] = true; //FTPS - control
  result[992] = true; //Telnet over TLS/SSL
  result[993] = true; //IMAPS
  result[994] = true; //IRCS
  result[995] = true; //POP3S

  return result;
}

static const std::map<uint16_t, bool> SSLPortMap = createSSLPortMap();

static bool isSSLPort(uint16_t port)
{
  if(port == 443) // HTTPS, this is likely case
    return true;

  switch(port)
  {
    case 0:   // default
    case 261: // NSIIOPS
    case 448: // DDM-SSL
    case 465: // SMTPS
    case 563: // NNTPS
    case 614: // SSHELL
    case 636: // LDAPS
    case 989: // FTPS - data
    case 990: // FTPS - control
    case 992: // Telnet over TLS/SSL
    case 993: // IMAPS
    case 994: // IRCS
    case 995: // POP3S
      return true;
    default:
      return false;
  }
}

{
  constexpr auto tryCount = 1000000U;
  constexpr uint16_t portToSearch = 995;
  auto count = 0U;

  std::cout << "Measurement of lookup performance: container size is " << SSLPortMap.size() << '\n';

  // function
  auto startTime = std::chrono::steady_clock::now();

  for(auto i = 0U; i < tryCount; ++i)
    if(isSSLPort(portToSearch))
      ++count;

  auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - startTime);
  std::cout << "Operator switch: duration(ms): " << duration.count() << ", iterations: " << count << '\n';

  // map
  count = 0;
  startTime = std::chrono::steady_clock::now();

  for(auto i = 0U; i < tryCount; ++i)
    if(SSLPortMap.find(portToSearch) != SSLPortMap.cend())
      ++count;

  duration = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - startTime);
  std::cout << "Lookup in map: duration(ms): " << duration.count() << ", iterations: " << count << '\n';

  return 0;
}
```

gx740:
We can append a new public method `isSSLPort` instead of `getSSLPortMap`.

seladb:
> We can append a new public method `isSSLPort` instead of `getSSLPortMap`

Yes, that makes sense to me. It'll break the current API but I don't think it's a major concern because as you mentioned not a lot of people use it. You will have to change the `SSLAnalyzer` example app though

gx740:
The changes have been made

Packet++/header/SSLLayer.h
```diff
@@ -249,7 +248,7 @@ namespace pcpp
 		 */
 		void parseNextLayer();
 
-        OsiModelLayer getOsiModelLayer() const { return OsiModelPresentationLayer; }
+    OsiModelLayer getOsiModelLayer() const { return OsiModelPresentationLayer; }
```
seladb: Please fix the indentation here

gx740: Done

PR #293, opened 2019-11-07
Fixed indentations in GtpLayer.h

gx740:


Packet++/header/GtpLayer.h
```diff
@@ -32,7 +32,7 @@ namespace pcpp
 		/** A 1-bit value that differentiates GTP (value 1) from GTP' (value 0) */
 				protocolType:1,
 		/** GTP version */
-		        version:3;
+				version:3;
 #else
 		/** GTP version */
 		uint8_t version:3,
```
seladb: Can you fix the indentation for all of these fields?

gx740: I thought these indentations had been made consciously, in order to group bit fields that belong to one byte.

seladb: No, it's probably a mistake. If you look at the file you'll see that `uint8_t npduNumberFlag:1` has a different indentation comparing to the rest of the fields. I'd prefer to align all fields to the same indentation as `uint8_t npduNumberFlag:1`

gx740: Done

PR #296, opened 2019-11-08
Added 'const' to a number of methods

gx740:
I hope this PR is final in which the qualifier `const` is added to methods.
In addition a lot of changes which improve readability.

### DIFF
--- a/Packet++/header/IPv4Layer.h
+++ b/Packet++/header/IPv4Layer.h
@@ -580,7 +580,7 @@ namespace pcpp
 		/**
 		 * @return Size of IPv4 header (including IPv4 options if exist)
 		 */
-		size_t getHeaderLen() { return (size_t)(getIPv4Header()->internetHeaderLength*4) + m_TempHeaderExtension; }
+		size_t getHeaderLen() const { return (size_t)(getIPv4Header()->internetHeaderLength) * 4 + m_TempHeaderExtension; }
 
 		/**
 		 * Calculate the following fields:
@@ -591,7 +591,7 @@ namespace pcpp
 		 */
 		void computeCalculateFields();
 
-		std::string toString();
+		std::string toString() const;
 
 		OsiModelLayer getOsiModelLayer() const { return OsiModelNetworkLayer; }
 
--- a/Packet++/header/IPv4Layer.h
+++ b/Packet++/header/IPv4Layer.h
@@ -239,7 +239,7 @@ namespace pcpp
 			if (dataSize < 2)
 				return res;
 
-			uint8_t valueOffset = 1;
+			uint8_t valueOffset = (uint8_t)(1);
 
 			while (valueOffset < dataSize)
 			{
@@ -250,7 +250,7 @@ namespace pcpp
 
 				res.push_back(IPv4Address(curValue));
 
-				valueOffset += 4;
+				valueOffset += (uint8_t)(4);
 			}
 
 			return res;
@@ -280,7 +280,7 @@ namespace pcpp
 
 			res.type = (IPv4TimestampOptionValue::TimestampType)m_Data->recordValue[1];
 
-			uint8_t valueOffset = 2;
+			uint8_t valueOffset = (uint8_t)(2);
 			bool readIPAddr = (res.type == IPv4TimestampOptionValue::TimestampAndIP);
 
 			while (valueOffset < dataSize)
@@ -298,7 +298,7 @@ namespace pcpp
 				if (res.type == IPv4TimestampOptionValue::TimestampAndIP)
 					readIPAddr = !readIPAddr;
 
-				valueOffset += 4;
+				valueOffset += (uint8_t)(4);
 			}
 
 			return res;
@@ -580,7 +580,7 @@ namespace pcpp
 		/**
 		 * @return Size of IPv4 header (including IPv4 options if exist)
 		 */
-		size_t getHeaderLen() const { return (size_t)(getIPv4Header()->internetHeaderLength) * 4 + m_TempHeaderExtension; }
+		size_t getHeaderLen() const { return (size_t)((uint16_t)(getIPv4Header()->internetHeaderLength) * 4) + m_TempHeaderExtension; }
 
 		/**
 		 * Calculate the following fields:
--- a/Packet++/header/IPv4Layer.h
+++ b/Packet++/header/IPv4Layer.h
@@ -26,12 +26,12 @@ namespace pcpp
 		/** IP header length, has the value of 5 for IPv4 */
 		uint8_t internetHeaderLength:4,
 		/** IP version number, has the value of 4 for IPv4 */
-				ipVersion:4;
+		ipVersion:4;
 #else
 		/** IP version number, has the value of 4 for IPv4 */
 		uint8_t ipVersion:4,
 		/** IP header length, has the value of 5 for IPv4 */
-				internetHeaderLength:4;
+		internetHeaderLength:4;
 #endif
 		/** type of service, same as Differentiated Services Code Point (DSCP)*/
 		uint8_t typeOfService;
@@ -228,18 +228,18 @@ namespace pcpp
 		 * (meaning zeroed addresses - 0.0.0.0) will not be added to the returned list. If some error occurs during the parsing or the value is invalid an empty vector is returned
 		 * @return A vector of IPv4 addresses parsed from the IPv4 option value
 		 */
-		std::vector<IPv4Address> getValueAsIpList()
+		std::vector<IPv4Address> getValueAsIpList() const
 		{
 			std::vector<IPv4Address> res;
 
 			if (m_Data == NULL)
 				return res;
 
-			size_t dataSize =  getDataSize();
+			size_t dataSize = getDataSize();
 			if (dataSize < 2)
 				return res;
 
-			uint8_t valueOffset = (uint8_t)(1);
+			uint8_t valueOffset = 1;
 
 			while (valueOffset < dataSize)
 			{
@@ -250,7 +250,7 @@ namespace pcpp
 
 				res.push_back(IPv4Address(curValue));
 
-				valueOffset += (uint8_t)(4);
+				valueOffset += 4;
 			}
 
 			return res;
@@ -263,7 +263,7 @@ namespace pcpp
 		 * the lists. If some error occurs during the parsing or the value is invalid an empty result is returned
 		 * @return A structured containing the IPv4 timestamp value
 		 */
-		IPv4TimestampOptionValue getTimestampOptionValue()
+		IPv4TimestampOptionValue getTimestampOptionValue() const
 		{
 			IPv4TimestampOptionValue res;
 			res.clear();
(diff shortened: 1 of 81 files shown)


Packet++/header/IPv4Layer.h
```diff
@@ -228,18 +228,18 @@ namespace pcpp
...
 			if (m_Data == NULL)
 				return res;
 
-			size_t dataSize =  getDataSize();
+			size_t dataSize = getDataSize();
 			if (dataSize < 2)
 				return res;
 
-			uint8_t valueOffset = (uint8_t)(1);
+			uint8_t valueOffset = 1;
```
seladb: I think removing the casting might cause a compilation warning on some platfroms

Packet++/header/IPv4Layer.h
```diff
@@ -250,7 +250,7 @@ namespace pcpp
 
 				res.push_back(IPv4Address(curValue));
 
-				valueOffset += (uint8_t)(4);
+				valueOffset += 4;
```
seladb: I think removing the casting might cause a compilation warning on some platfroms

Packet++/header/IPv4Layer.h
```diff
@@ -280,7 +280,7 @@ namespace pcpp
 
 			res.type = (IPv4TimestampOptionValue::TimestampType)m_Data->recordValue[1];
 
-			uint8_t valueOffset = (uint8_t)(2);
+			uint8_t valueOffset = 2;
```
seladb: I think removing the casting might cause a compilation warning on some platfroms

Packet++/header/IPv4Layer.h
```diff
@@ -298,7 +298,7 @@ namespace pcpp
 				if (res.type == IPv4TimestampOptionValue::TimestampAndIP)
 					readIPAddr = !readIPAddr;
 
-				valueOffset += (uint8_t)(4);
+				valueOffset += 4;
```
seladb: I think removing the casting might cause a compilation warning on some platfroms

Packet++/header/IPv4Layer.h
```diff
@@ -580,7 +580,7 @@ namespace pcpp
 		/**
 		 * @return Size of IPv4 header (including IPv4 options if exist)
 		 */
-		size_t getHeaderLen() { return (size_t)(getIPv4Header()->internetHeaderLength*4) + m_TempHeaderExtension; }
+		size_t getHeaderLen() const { return (size_t)(getIPv4Header()->internetHeaderLength) * 4 + m_TempHeaderExtension; }
```
seladb: The idea is to cast the while `getIPv4Header()->internetHeaderLength*4` expression to `size_t` to avoid compilation warning

Packet++/header/Layer.h
```diff
@@ -124,12 +124,12 @@ namespace pcpp
...
 		 * @param[out] toArr The destination byte array
 		 */
-		void copyData(uint8_t* toArr);
+		void copyData(uint8_t* toArr) const;
 
 
 		// implement abstract methods
 
-		uint8_t* getDataPtr(size_t offset = 0) { return (uint8_t*)(m_Data + offset); }
+		uint8_t* getDataPtr(size_t offset = 0) const { return m_Data + offset; }
```
seladb: this might cause compilation warning on some platforms

Packet++/src/DnsResource.cpp
```diff
@@ -248,48 +248,30 @@ size_t DnsResource::getDataLength() const
...
 	case DNS_TYPE_MX:
-	{
-		return DnsResourceDataPtr(new MxDnsResourceData(resourceRawData, dataLength, this));
-		break;
-	}
+		return DnsResourceDataPtr(new MxDnsResourceData(resourceRawData, dataLength, const_cast<IDnsResource*>(static_cast<const IDnsResource*>(this))));
 
 	default:
-	{
 		return DnsResourceDataPtr(new GenericDnsResourceData(resourceRawData, dataLength));
```
seladb: I think that adding the curly braces back will make the code look more readable

Packet++/src/GtpLayer.cpp
```diff
@@ -587,22 +587,25 @@ void GtpV1Layer::parseNextLayer()
 
 	// GTP-U message, try to parse the next layer
 
-	uint8_t subProto = *(uint8_t*)(m_Data + headerLen);
+	uint8_t* payload = m_Data + headerLen;
```
seladb: this might cause compilation warning on some platforms (casting may be needed here)

Packet++/src/IPv6Extensions.cpp
```diff
@@ -174,7 +174,7 @@ IPv6TLVOptionHeader::IPv6TLVOptionHeader(const std::vector<IPv6TLVOptionBuilder>
 	for (std::vector<IPv6TLVOptionBuilder>::const_iterator iter = options.begin(); iter != options.end(); iter++)
 	{
 		IPv6Option option = iter->build();
-		memcpy((uint8_t*)(getDataPtr() + offset), option.getRecordBasePtr(), option.getTotalSize());
+		memcpy(getDataPtr() + offset, option.getRecordBasePtr(), option.getTotalSize());
```
seladb: why did you remove the casting?

Packet++/src/IPv6Extensions.cpp
```diff
@@ -207,15 +207,15 @@ IPv6RoutingHeader::IPv6RoutingHeader(uint8_t routingType, uint8_t segmentsLeft,
 
 	if (additionalRoutingDataLen > 0 && additionalRoutingData != NULL)
 	{
-		uint8_t* additionalDataPtr = (uint8_t*)(getDataPtr() + sizeof(ipv6_routing_header));
+		uint8_t* additionalDataPtr = getDataPtr() + sizeof(ipv6_routing_header);
```
seladb: why did you remove the casting?

Packet++/src/IPv6Extensions.cpp
```diff
@@ -207,15 +207,15 @@ IPv6RoutingHeader::IPv6RoutingHeader(uint8_t routingType, uint8_t segmentsLeft,
...
 		memcpy(additionalDataPtr, additionalRoutingData, additionalRoutingDataLen);
 	}
 }
 
 uint8_t* IPv6RoutingHeader::getRoutingAdditionalData() const
 {
 	if (getExtensionLen() > sizeof(ipv6_routing_header))
-		return (uint8_t*)(getDataPtr() + sizeof(ipv6_routing_header));
+		return getDataPtr() + sizeof(ipv6_routing_header);
```
seladb: why did you remove the casting?

Packet++/src/IPv6Extensions.cpp
```diff
@@ -263,15 +263,15 @@ IPv6AuthenticationHeader::IPv6AuthenticationHeader(uint32_t securityParametersIn
 
 	if (integrityCheckValueLen > 0 && integrityCheckValue != NULL)
 	{
-		uint8_t* icvPtr = (uint8_t*)(getDataPtr() + sizeof(ipv6_authentication_header));
+		uint8_t* icvPtr = getDataPtr() + sizeof(ipv6_authentication_header);
```
seladb: why did you remove the casting?

Packet++/src/IPv6Extensions.cpp
```diff
@@ -263,15 +263,15 @@ IPv6AuthenticationHeader::IPv6AuthenticationHeader(uint32_t securityParametersIn
...
 		memcpy(icvPtr, integrityCheckValue, integrityCheckValueLen);
 	}
 }
 
 uint8_t* IPv6AuthenticationHeader::getIntegrityCheckValue() const
 {
 	if (getExtensionLen() > sizeof(ipv6_authentication_header))
-		return (uint8_t*)(getDataPtr() + sizeof(ipv6_authentication_header));
+		return getDataPtr() + sizeof(ipv6_authentication_header);
```
seladb: why did you remove the casting?

Packet++/header/IPv4Layer.h
gx740: It's just a positive constant. There is no any reason for warning. But I can roll back the changes.

I have looked into the build logs and not found any warnings that concern the changes made in this PR.

seladb: I think I added that casting for a reason, probably a compilation warning. But it was a long time ago so I don't quite remember on which platform it was...

Packet++/header/IPv4Layer.h
gx740: I understand. I had moved the parenthesis to avoid following warning in Visual Studio:
`Warning	C26451	Arithmetic overflow: Using operator '*' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '*' to avoid overflow (io.2)`

Packet++/header/IPv4Layer.h
gx740: Ok. I will return back the old code.

gx740: Done

Packet++/header/IPv4Layer.h
seladb: Got it. But in the fix you've made, multiplying `(size_t)(getIPv4Header()->internetHeaderLength)` by 4 might make the compiler think the result will be larger that `size_t`, maybe we should add 2 castings there?

Packet++/header/Layer.h
gx740: The changes have been undone

Packet++/header/IPv4Layer.h
gx740: Like this?
`return (size_t)((uint16_t)(getIPv4Header()->internetHeaderLength) * 4) + m_TempHeaderExtension`

Packet++/src/DnsResource.cpp
gx740: The braces will be returned back.

Packet++/src/GtpLayer.cpp
gx740: The casting has been added

Packet++/src/IPv6Extensions.cpp
gx740: It is unnecessary because `getDataPtr()` returns `uint8_t*`

gx740: And `memcpy` gets a `void*` in the first parameter

Packet++/src/IPv6Extensions.cpp
gx740: It is unnecessary because getDataPtr() returns uint8_t*

gx740: When a constant is added to a pointer the expression result will have the same type as pointer has.

Packet++/src/IPv6Extensions.cpp
gx740: When a constant is added to a pointer the expression result will have the same type as pointer has. The `getDataPtr()` returns `uint8_t*`

Packet++/src/IPv6Extensions.cpp
gx740: When a constant is added to a pointer the expression result will have the same type as pointer has.

Packet++/src/IPv6Extensions.cpp
gx740: When a constant is added to a pointer the expression result will have the same type as pointer has.

Packet++/src/IPv6Extensions.cpp
seladb: ok

Packet++/header/IPv4Layer.h
seladb: yes, that looks ok. Thanks!

Issue #275, opened 2019-10-29
httpstats only support port 80 & 8080?


ugenehan:
httpstats only support port 80 & 8080?can not support other port?

seladb:
yes, as of now the are the only supported ports:

https://github.com/seladb/PcapPlusPlus/blob/2eee2ad31ae37ce6bf3c70d21d85961aaf787d69/Packet%2B%2B/src/HttpLayer.cpp#L13-L19

Would you like to add a new port to this map?

ugenehan:
> yes, as of now the are the only supported ports:
> 
> https://github.com/seladb/PcapPlusPlus/blob/2eee2ad31ae37ce6bf3c70d21d85961aaf787d69/Packet%2B%2B/src/HttpLayer.cpp#L13-L19
> 
> Would you like to add a new port to this map?

Many enterprise systems have custom ports.I have defined a static function in class 'HttpMessage',it can add port to HTTPPortMap.

seladb:
I'm not sure adding custom specific port will be useful for many people. Do you have a suggestion to make it more generic?

seladb:
I'm closing the issue for now, please reopen if needed

gx740:
I can give an idea how to implement the support of any port number if needed.

seladb:
thanks @gx740 , please feel free to suggest your ideas

gx740:
The idea is ready for discussion in PR #298.

PR #298, opened 2019-11-11
Customizing the port numbers

gx740:
This PR demonstrates the idea how to implement the feature of customizing the port numbers supported by certain layer as had been requested in issue #275. Currently, there are 6 classes that depends from the port numbers: `DNSLayer, HTTPLayer, RadiusLayer, SSLLayer, SIPLayer, GTVv1Layer`. We can implement this feature in all of these.

I have prepared two versions of classes that manage the port numbers and have selected the SSLLayer class for tests.

The first and the fastest implementation is `PortList`. The performance of lookup operation is the same as `isSSLPort` (for Windows) with hard-coded ports and even better (for Linux). The only disadvantage is that the class has the following member: `bool m_Ports[65536]` so the size of object will be 64K. I think it's not a problem for static class members.

The second class `PortList2` is slower than `PortList` in every sense but have a small size. Lookup is slower, at least, in 3 times than for `PortList` (for port 443). In the worst case (for port 995) the speed is decreasing in 10 times.

Test results:

1. Windows 10, Core i7 9700K, VS 2019, Release (x64):

```
=== Measurement of lookup performance for port number 443 ===

isSSLPort() with operator switch: duration(ms): 432, iterations: 1000000000

PortList::contains(): duration(ms): 435, iterations: 1000000000

PortList2::contains(): duration(ms): 1303, iterations: 1000000000
```

```
=== Measurement of lookup performance for port number 995 ===

isSSLPort() with operator switch: duration(ms): 429, iterations: 1000000000

PortList::contains(): duration(ms): 427, iterations: 1000000000

PortList2::contains(): duration(ms): 4517, iterations: 1000000000
```

2. Xeon E5-2630 v4, CentOS 7.7.1908 x64, g++ 8.3.1:

```
=== Measurement of lookup performance for port number 443 ===

isSSLPort() with operator switch: duration(ms): 1949, iterations: 1000000000

PortList::contains(): duration(ms): 340, iterations: 1000000000

PortList2::contains(): duration(ms): 973, iterations: 1000000000
```

```
=== Measurement of lookup performance for port number 995 ===

isSSLPort() with operator switch: duration(ms): 2594, iterations: 1000000000

PortList::contains(): duration(ms): 339, iterations: 1000000000

PortList2::contains(): duration(ms): 5186, iterations: 1000000000
```

### DIFF
--- a/Common++/header/PortList.h
+++ b/Common++/header/PortList.h
@@ -26,7 +26,7 @@ namespace pcpp
 		/**
 		 * The default constructor that creates an empty list
 		 */
-		PortList() : m_Ports() {}
+		PortList() : m_Ports(65536, false) {}
 
 		/**
 		 * A constructor that creates an object by the port number
@@ -360,7 +360,7 @@ namespace pcpp
 		}
 
 	protected:
-		bool m_Ports[65536];
+		std::vector<bool> m_Ports;
 	}; // class PortList
 
 
@@ -369,20 +369,20 @@ namespace pcpp
 	// implementation of inline methods
 
 	PortList::PortList(uint16_t p1)
-		: m_Ports()
+		: m_Ports(65536, false)
 	{
 		m_Ports[p1] = true;
 	}
 
 	PortList::PortList(uint16_t p1, uint16_t p2)
-		: m_Ports()
+		: m_Ports(65536, false)
 	{
 		m_Ports[p1] =
 			m_Ports[p2] = true;
 	}
 
 	PortList::PortList(uint16_t p1, uint16_t p2, uint16_t p3)
-		: m_Ports()
+		: m_Ports(65536, false)
 	{
 		m_Ports[p1] =
 			m_Ports[p2] =
@@ -390,7 +390,7 @@ namespace pcpp
 	}
 
 	PortList::PortList(uint16_t p1, uint16_t p2, uint16_t p3, uint16_t p4)
-		: m_Ports()
+		: m_Ports(65536, false)
 	{
 		m_Ports[p1] =
 			m_Ports[p2] =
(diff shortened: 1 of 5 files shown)


gx740:
Pay attention how `isSSLPort` slows down under Linux comparing with Windows. Its body should be moved into header file.

seladb:
@gx740 I'm trying to understand the high-level design you propose:

You created 2 options for a class that holds a port list (`PortList` and `PortList2`, we'll eventually choose one of them). As of now you simply replace the hard-coded port implementation in `SSLLayer` with a private static member that you initialize with hard-coded values in `SSLLayer.cpp`. So far the new implementation and the old one are practiaclly the same, aren't they?

Are you also planning to add functionality that will let the users change this port list without making changes to the library code itself? For example, let the users pass a port list for each layer during PcapPlusPlus initialization?

gx740:
Yes, the new implementation is similar to the old one with one exception: it supports the changing of ports. Ordinary user does not need in this functionality so the hard-coded well known ports can be assigned in cpp file of certain layer. When some user wants to reassign the ports he/she just takes a reference to a static object and does the changes, for example on initialization phase of application.

As the first implementation is very fast we can replace the hard-coded port numbers in `parseNextLayer` methods of all layers, where it does matter, without lost in speed.

gx740:
Implementing the `PortList` we get following benefits:

1. All ports supported by certain layer will defined in its CPP file
2. The user has the ability to redefine a port list

gx740:
I have changed the type of `m_Ports` in `PortList` to `std::vector<bool>` and tested again. The results does not differ in speed from the previous version. So we can use `std::vector<bool>` instead of C-array.

seladb:
if the user needs to change the layer code itself I'm not sure I see the real benefit here... Maybe we can think of an API to let the user change the default port list of a certain layer.

But to be honest, I'm not sure how critical this feature is. Only a few users would want to change the default ports, so maybe we should invest our energy somewhere else. Please let me know what you think

gx740:
I don't insist on implementation of this feature, it's just suggestion.

PR #299, opened 2019-11-12
Performance improvements

gx740:
Significantly improved performance of detecting of HTTP and DNS layers. The static methods `getHTTPPortMap` and `getDNSPortMap` were removed (it was made in the same way as in `SSLLayer`, PR #288 ).

Improved readability: created static inline methods in layer classes that checks the port numbers. The only exception is the `DhcpLayer`.

A small improvement in `IPv4Layer::parseNextLayer()`.


Packet++/header/SSLLayer.h
```diff
@@ -520,7 +522,36 @@ namespace pcpp
...
+	// implementation of inline methods
+
+	bool SSLLayer::isSSLPort(uint16_t port)
+	{
+		if (port == 443) // HTTPS, this is likely case
+			return true;
+
+		switch (port)
+		{
+		case 0:   // default
```
gx740: I am not sure that port 0 should stay in this method, it's not a SSL port.

gx740:
Don't merge this PR. I will split it to several PR's for easy understanding

PR #301, opened 2019-11-14
Added additional length check to decodeName() to prevent buffer overflow

Minilulatsch:
The additional length check simply checks, whether the next word of the domain name would result in a total domain name size bigger than 255 bytes. If this is the case, the word should not be copied anymore and the functions returns.

### DIFF
--- a/Packet++/src/DnsResource.cpp
+++ b/Packet++/src/DnsResource.cpp
@@ -55,6 +55,10 @@ size_t IDnsResource::decodeName(const char* encodedName, char* result, int itera
 	// A string to parse
 	while (wordLength != 0)
 	{
+		// Return if next word makes total domain name size > 255
+		if (encodedNameLength + wordLength > 255)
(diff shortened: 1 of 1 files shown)


seladb:
please note that the `TestDnsParsing` test-case is failing:

https://github.com/seladb/PcapPlusPlus/blob/3b9d687ac2e9f34ccbde62a05d82172c291a5d87/Tests/Pcap%2B%2BTest/main.cpp#L3068

seladb:
Could you fork your branch from `dev` branch instead of `master` so the PR will be merge into `dev`? I usually don't merge PRs directly to `master` (you can read more about it in [CONTRIBUTING.md](https://github.com/seladb/PcapPlusPlus/blob/master/CONTRIBUTING.md))

Also, it'd be good to add the example pcap you provided to the test-cases. If you don't know how to do it or don't have time for it please leave it to me

PR #304, opened 2019-11-20
Performance improvements in TcpReassembly

gx740:
1. `getLayerPayload` and `getLayerPayloadSize` have been made virtual in class `Layer`. In some cases it is useful to override these methods in derived class, because an implementation in base class uses the calling of virtual methods of derived class.

2. The new inline methods `getLayerPayload` and `getLayerPayloadSize` have been added in `IPv4Layer`, `IPv6Layer` and `TcpLayer`.  These override corresponding methods of the base class. It improves the performance of methods of `TcpReassembly` class.

### DIFF
--- a/Packet++/header/IPv4Layer.h
+++ b/Packet++/header/IPv4Layer.h
@@ -569,6 +569,18 @@ namespace pcpp
 		 */
 		bool removeAllOptions();
 
+		/**
+		 * This method overrides the method of base class (Layer) to improve performance in some cases
+		 * @return A pointer for the layer payload, meaning the first byte after the header
+		 */
+		virtual uint8_t* getLayerPayload() const { return getData() + getHeaderLen(); }
+
+		/**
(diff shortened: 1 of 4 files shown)


gx740:
Different way is to make the methods `getLayerPayload` and `getLayerPayloadSize` as pure virtual and implement them in all derived classes.

seladb:
Have you measured the performance gain for this change?

I know that using virtual methods has some cost but it make the code cleaner and more readable. I don't feel the improvement is significant enough comparing to the other computation / data processing that is being done in TCP reassembly. Please let me know what you think

gx740:
No, I didn't. Let's look to these methods from the other point of view. The class `Layer` has an abstract method `getHeaderLen` because only the derived class, that implements a certain layer, knows this. From my point of view the `getLayerPayload` and `getLayerPayloadSize` are similar to the `getHeaderLen`.

Ok, let's close this PR.

PR #305, opened 2019-11-20
Improvements in TcpReassembly (performance increased)

gx740:
1. The signatures of callback function have been changed because passing the arguments by value is useless and gives performance overhead.
2. The type of `TcpStreamData::m_Connection` has been changed from `ConnectionData` to `const ConnectionData*` to avoid data copying
3. The private method `TcpReassembly::prepareMissingDataMessage` has been replaced by the new static function which is optimized on speed (no dynamic memory allocation).

### DIFF
--- a/Examples/TcpReassembly/main.cpp
+++ b/Examples/TcpReassembly/main.cpp
@@ -330,7 +330,7 @@ void listInterfaces()
 /**
  * The callback being called by the TCP reassembly module whenever new data arrives on a certain connection
  */
-static void tcpReassemblyMsgReadyCallback(int sideIndex, TcpStreamData tcpData, void* userCookie)
+static void tcpReassemblyMsgReadyCallback(int sideIndex, const TcpStreamData& tcpData, void* userCookie)
 {
 	// extract the connection manager from the user cookie
 	TcpReassemblyConnMgr* connMgr = (TcpReassemblyConnMgr*)userCookie;
@@ -412,7 +412,7 @@ static void tcpReassemblyMsgReadyCallback(int sideIndex, TcpStreamData tcpData, 
 /**
  * The callback being called by the TCP reassembly module whenever a new connection is found. This method adds the connection to the connection manager
  */
-static void tcpReassemblyConnectionStartCallback(ConnectionData connectionData, void* userCookie)
+static void tcpReassemblyConnectionStartCallback(const ConnectionData& connectionData, void* userCookie)
 {
 	// get a pointer to the connection manager
 	TcpReassemblyConnMgr* connMgr = (TcpReassemblyConnMgr*)userCookie;
@@ -433,7 +433,7 @@ static void tcpReassemblyConnectionStartCallback(ConnectionData connectionData, 
  * The callback being called by the TCP reassembly module whenever a connection is ending. This method removes the connection from the connection manager and writes the metadata file if requested
  * by the user
  */
-static void tcpReassemblyConnectionEndCallback(ConnectionData connectionData, TcpReassembly::ConnectionEndReason reason, void* userCookie)
+static void tcpReassemblyConnectionEndCallback(const ConnectionData& connectionData, TcpReassembly::ConnectionEndReason reason, void* userCookie)
 {
 	// get a pointer to the connection manager
 	TcpReassemblyConnMgr* connMgr = (TcpReassemblyConnMgr*)userCookie;
--- a/Packet++/header/TcpReassembly.h
+++ b/Packet++/header/TcpReassembly.h
@@ -205,12 +205,12 @@ public:
 	 * A getter for the connection data
 	 * @return The const reference to connection data
 	 */
-	const ConnectionData& getConnectionData() const { return m_Connection; }
+	const ConnectionData& getConnectionData() const { return *m_Connection; }
 
 private:
 	uint8_t* m_Data;
 	size_t m_DataLen;
-	ConnectionData m_Connection;
+	const ConnectionData* m_Connection;
 	bool m_DeleteDataOnDestruction;
 
 	void setDeleteDataOnDestruction(bool flag) { m_DeleteDataOnDestruction = flag; }
@@ -281,7 +281,7 @@ public:
 	 * @param[in] tcpData The TCP data itself + connection information
 	 * @param[in] userCookie A pointer to the cookie provided by the user in TcpReassembly c'tor (or NULL if no cookie provided)
 	 */
-	typedef void (*OnTcpMessageReady)(int side, TcpStreamData tcpData, void* userCookie);
+	typedef void (*OnTcpMessageReady)(int side, const TcpStreamData& tcpData, void* userCookie);
 
 	/**
 	 * @typedef OnTcpConnectionStart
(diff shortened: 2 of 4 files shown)


seladb:
Have you measured the performance improvements of these changes? How significant are they?

gx740:
No, I didn't. The changes made in this PR are easy to understand.

First, there are classes `IPv4Address` and `IPv6Address` of 58 and 68 bytes long accordingly and their constructors allocate the memory. So the copy operation for object of one of these classes is heavyweight.

We have the classes `ConnectionData` (its size is 40 bytes) and `TcpStreamData` (32 bytes long) which are being passed to the callback functions by value. Both of these use two pointers to `IPAddress` which, in fact, point to the objects of types `IPv4Address/IPv6Address`.
In other words, passing an object by value forces to:
- copy almost 200 bytes
- make two memory allocations

Changing the type of `TcpStreamData::m_Connection` pursues the same goal: avoid object's copying.

In fact, everything is much worse. Let's look into source code where the callback is called:
```
TcpStreamData streamData(tcpLayer->getLayerPayload() + newLength, tcpPayloadSize - newLength, tcpReassemblyData->connData);
streamData.setDeleteDataOnDestruction(false);
m_OnMessageReadyCallback(sideIndex, streamData, m_UserCookie);
```
We can see an object of type `ConnectionData` is being copied twice.

seladb:
Thanks for your answers/

Regarding passing `TcpStreamData` by value - I did it on purpose so users can use their own copy of the object without being worried that its memory will get freed or object will be changed. Of course coping objects has a performance impact but in the same time it simplifies the interaction with the users. So it's a trade-off between performance and usability...

Regarding the change in `prepareMissingDataMessage()` - your solution is definitely more efficient, but a little bit more complex. I'm also not sure how much performance improvement this change contributes to the overall performance of the TCP reassembly module

gx740:
> Regarding passing `TcpStreamData` by value - I did it on purpose so users can use their own copy of the object without being worried that its memory will get freed or object will be changed. Of course coping objects has a performance impact but in the same time it simplifies the interaction with the users. So it's a trade-off between performance and usability...

Even having the copy of `ConnectionData` in a callback function, the user is forced to copy it anyway. So the const reference is more appropriated in this case.

The same is true for `TcpStreamData`. In the most cases the member `m_Data` points to data in packet or fragment. The only case it points to newly allocated memory is when missing data is detected. What the user can do with it? Nothing except copying because the user should stick together the parts of certain stream to parse the next layer.

gx740:
> Regarding the change in `prepareMissingDataMessage()` - your solution is definitely more efficient, but a little bit more complex. I'm also not sure how much performance improvement this change contributes to the overall performance of the TCP reassembly module

Changes have a little impact on overall performance and can be omitted.

gx740:
If to implement my proposal we can simplify `class TcpStreamData` to the next structure:
```
struct TcpStreamData
{
  uint8_t* m_Data;
  size_t m_DataLen;
  ConnectionData& m_Connection;
}
```

Many source code lines will disappeared.

And following code:
```
// send new data to callback
if (m_OnMessageReadyCallback != NULL)
{
  // prepare missing data text
  std::string missingDataTextStr = prepareMissingDataMessage(missingDataLen);

  // add missing data text to the data that will be sent to the callback. This means that the data will look something like:
  // "[xx bytes missing]<original_data>"
  size_t dataWithMissingDataTextLen = missingDataTextStr.length() + curTcpFrag->dataLength;
  uint8_t* dataWithMissingDataText = new uint8_t[dataWithMissingDataTextLen];
  memcpy(dataWithMissingDataText, missingDataTextStr.c_str(), missingDataTextStr.length());
  memcpy(dataWithMissingDataText + missingDataTextStr.length(), curTcpFrag->data, curTcpFrag->dataLength);

  //TcpStreamData streamData(curTcpFrag->data, curTcpFrag->dataLength, tcpReassemblyData->connData);
  //streamData.setDeleteDataOnDestruction(false);
  TcpStreamData streamData(dataWithMissingDataText, dataWithMissingDataTextLen, tcpReassemblyData->connData);
  m_OnMessageReadyCallback(sideIndex, streamData, m_UserCookie);

  LOG_DEBUG("Found missing data on side %d: %d byte are missing. Sending the closest fragment which is in size %d + missing text message which size is %d",
    sideIndex, missingDataLen, (int)curTcpFrag->dataLength, (int)missingDataTextStr.length());
}
```
can be changed to:
```
// send new data to callback
if (m_OnMessageReadyCallback != NULL)
{
  // prepare missing data text
  std::string missingDataTextStr = prepareMissingDataMessage(missingDataLen);

  // add missing data text to the data that will be sent to the callback. This means that the data will look something like:
  // "[xx bytes missing]<original_data>"
  size_t dataWithMissingDataTextLen = missingDataTextStr.length() + curTcpFrag->dataLength;

//  === START OF CHANGES ===
  std::vector<uint8_t> dataWithMissingDataText;
  dataWithMissingDataText.reserve(dataWithMissingDataTextLen);

  const char *begin = missingDataTextStr.c_str(), *end = missingDataTextStr.c_str() + missingDataTextStr.length();
  std::copy(begin, end, std::back_inserter(dataWithMissingDataText));
  
  begin = curTcpFrag->data, end = curTcpFrag->data + curTcpFrag->dataLength;
  std::copy(begin, end, std::back_inserter(dataWithMissingDataText));
//  === END OF CHANGES ===

  //TcpStreamData streamData(curTcpFrag->data, curTcpFrag->dataLength, tcpReassemblyData->connData);
  //streamData.setDeleteDataOnDestruction(false);
  TcpStreamData streamData(&dataWithMissingDataText[0], dataWithMissingDataTextLen, tcpReassemblyData->connData);
  m_OnMessageReadyCallback(sideIndex, streamData, m_UserCookie);

  LOG_DEBUG("Found missing data on side %d: %d byte are missing. Sending the closest fragment which is in size %d + missing text message which size is %d",
    sideIndex, missingDataLen, (int)curTcpFrag->dataLength, (int)missingDataTextStr.length());
}
```

gx740:
Or just to remove the allocated memory after `m_OnMessageReadyCallback` is completed.

gx740:
An implementation with using of vector is exception safe

seladb:
> Even having the copy of `ConnectionData` in a callback function, the user is forced to copy it anyway. So the const reference is more appropriated in this case.
> 
> The same is true for `TcpStreamData`. In the most cases the member m_Data points to data in packet or fragment. The only case it points to newly allocated memory is when missing data is detected. What the user can do with it? Nothing except copying because the user should stick together the parts of certain stream to parse the next layer.

Yes, I agree with you. Let's keep those changes

> >Regarding the change in prepareMissingDataMessage() - your solution is definitely more efficient, but a little bit more complex. I'm also not sure how much performance improvement this change contributes to the overall performance of the TCP reassembly module
> 
> Changes have a little impact on overall performance and can be omitted.

Can you please omit those changes?

> If to implement my proposal we can simplify class TcpStreamData to the next structure:
> 
> ```cpp
> struct TcpStreamData
> {
>   uint8_t* m_Data;
>   size_t m_DataLen;
>   ConnectionData& m_Connection;
> }
> ```
> 
> Many source code lines will disappeared.

Why do you think we should remove the part that deals with freeing the data memory?

gx740:
The changes in `prepareMissingDataMessage` has been undone.

> Why do you think we should remove the part that deals with freeing the data memory?

The only reason is to reduce the source code and improve readability. Current implementation is exception-safe but has much unnecessary code. We have a standard container (`std::vector`) that can manage the memory. I have prepared the changes that use this approach in the latest commit. We should decide whether the old class stay unchanged or apply my changes.

PR #309, opened 2019-11-26
Supporting the user defined options when DPDK is initialized

gx740:
DpdkDeviceList::initDpdk has been enhanced by supporting the user defined options. Is this feature useful or not? I had planned to use some DPDK init options (for example: to override the number of memory channels which is hard-coded) but I'm currently not sure I will use it.

Should I test these changes in real application?


seladb:
I'm not sure if this is useful. So far no one has asked for it

PR #310, opened 2019-11-27
PcapLiveDevice: fix compilation on Alpine Linux (musl libc)

russkel:
I had to fix the path to sysctl.h on Alpine Linux 3.10 in order to get it to compile.

### DIFF
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -22,7 +22,7 @@
 #else
 #include <arpa/inet.h>
 #include <sys/ioctl.h>
-#include <sys/sysctl.h>
+#include <linux/sysctl.h>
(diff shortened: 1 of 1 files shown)


russkel:
Okay, so that doesn't work. I will speak to the alpine linux devs and find the way to do it.

russkel:
I found this: https://bugzilla.mozilla.org/show_bug.cgi?id=1261414

> The author of musl libc says:
> > sysctl does not work, and NEVER worked. using it is bogus.
> > it was a bogus experimental syscall that was deprecated before it was ever used
> > (basically, a broken binary version of /proc/sys, without any stability between kernel
> > versions for what the binary constants meant).

Is sysctl required?

russkel:
https://github.com/seladb/PcapPlusPlus/pull/311

PR #319, opened 2019-12-09
Gradually parsing of packets (new feature)

gx740:
In many cases the total parsing is inappropriate but, currently, there is no way to parse the packets gradually. In this PR I tried to implement what was discussed in #184.

Any suggestions and comments are welcome.


Packet++/header/Packet.h
```diff
@@ -330,6 +333,40 @@ namespace pcpp
...
+		 * @return A pointer to the parsed layer or NULL if there are no more layers
+		 */
+		static Layer* parseNextLayer(Packet* pkt);
+
+		/**
+		 * A method decodes the next layers in a packet
+		 * @param[in] parseUntil Parse the packet until it reaches this protocol. The default value is ::UnknownProtocol which means don't take this parameter into account
+		 * @param[in] parseUntilLayer Parse the packet until certain layer in OSI model. The default value is ::OsiModelLayerUnknown which means don't take this parameter into account
+		 */
+		static void parseLayers(Packet* pkt, ProtocolType parseUntil = UnknownProtocol, OsiModelLayer parseUntilLayer = OsiModelLayerUnknown);
```
seladb: I'm not sure I understand this API. Why putting both `parseUntil` and `parseUntilLayer` in the same method? why not separating them to two different methods? I think it'll be more clear that way

gx740: 1. There are `setRawPacket` method and a constructor in `Packet` that take both of these parameters
2. Having two parameters we can parse until a certain OSI model layer and stop if an inappropriate protocol has been reached. The simplest example is:
```
parseLayers(pkt, ProtocolType::ICMP, OsiModelLayer::OsiModelTransportLayer)
```
This call allows to parse until `TCP/UDP` layer and stops if an `ICMP` protocol is present.

I can append two methods:
```
static void parseLayers(Packet* pkt, ProtocolType parseUntil);
static void parseLayers(Packet* pkt, OsiModelLayer parseUntilLayer);
```

seladb: Yes, you're right. I forgot I added it myself. My bad 😄

PR #320, opened 2019-12-09
Fix a memory leak in example

gx740:


seladb:
All you added is `m_RecentConnsWithActivity` in `GlobalConfig` but you didn't use it anywhere... was it intentional?

gx740:
I have added a destructor in which an object is destroyed

gx740:
`m_RecentConnsWithActivity` is a pointer to dynamically allocated object that is never deleted therefore I have added a destructor.

seladb:
Oops you're right, my bad... I'll merge the PR now

PR #328, opened 2019-12-15
Adding RPCAP support on linux.

tyunkinilya:
Works with the latest version of libpcap from [libpcap repository](https://github.com/the-tcpdump-group/libpcap/). TODO: integrate propely ( check libpcap version before building?), add proper error management.


seladb:
The unit-tests don't pass now. Can you please check?

seladb:
it seems that you removed `#if defined(WIN32) || defined(WINx64) || defined(PCAPPP_MINGW_ENV)` and now the MacOS and FreeBSD builds obviously tests fail, but surprisingly the Linux builds fails as well, because the method `pcap_findalldevs_ex` isn't found. Maybe this is because this method only exists in the latest libpcap build

PR #334, opened 2019-12-18
Format PcapPrinter

NanXiao:


gx740:
You should use std::auto_ptr instead std::unique_ptr for compilers that don't support C++11

gx740:
You can use following code:
```c++
#if __cplusplus > 199711L || _MSC_VER >= 1800 //Maybe this can be 1600 for VS2010
#define PCPP_SMART_PTR(T) std::unique_ptr<T>
#else
#define PCPP_SMART_PTR(T) std::auto_ptr<T>
#endif
```

gx740:
Notice the `std::auto_ptr` cannot be used in standard container.

seladb:
What is the reason for using smart pointers in this case? why not keep it as simple pointers?

NanXiao:
@gx740 @seladb  Using smart pointers just avoids forgetting to delete `reader` and make code simpler. IMHO, most compilers should support `C++11` now, maybe you can consider using more modern `C++` features. Just my personal opinion, for reference only, thanks!

seladb:
Ideally I would like PcapPlusPlus to support devices who don't run `C++11`. That's why there is no `C++11` specific code in this repo. Regardless, in this specific case I don't think smart pointers are needed because `reader` is created in `main()` and deleted in the end of the same method

NanXiao:
@seladb Got it, thanks very much for your reply! I updated the PR and removed the smart pointer related code. Could you check it again? Thanks!

PR #308, opened 2019-11-22
The new version of hash5Tuple function

gx740:
I have created the new function `hash5Tuple` that works faster than the old one about 5.7/6.7 times and can be used in `TcpReassembly::reassemblePacket`. Now there are two functions:

1. `uint32_t hash5Tuple(const Packet* packet)`
2. `uint32_t hash5Tuple(const Packet* packet, const Layer* networkLayer, const Layer* transportLayer)`

Performance of version 1 comparing to the old function with the same signature is almost the same.
The result values of all versions don't differ.

Test results:

=== Centos7, Intel Xeon(R) CPU E5-2630 v4 @ 2.20GHz ===
```
Old func: duration(ms) = 3528, iterations = 28000000, calls per sec = 7936000
New func: duration(ms) = 3454, iterations = 28000000, calls per sec = 8106000
Optimized func: duration(ms) = 609, iterations = 28000000, calls per sec = 45977000

Old func: duration(ms) = 3433, iterations = 28000000, calls per sec = 8156000
New func: duration(ms) = 3463, iterations = 28000000, calls per sec = 8085000
Optimized func: duration(ms) = 592, iterations = 28000000, calls per sec = 47297000

Old func: duration(ms) = 3434, iterations = 28000000, calls per sec = 8153000
New func: duration(ms) = 3182, iterations = 28000000, calls per sec = 8799000
Optimized func: duration(ms) = 593, iterations = 28000000, calls per sec = 47217000
```

=== Windows 10, Core i7 9700K ===
```
Old func: duration(ms) = 2046, iterations = 28000000, calls per sec = 13685000
New func: duration(ms) = 2035, iterations = 28000000, calls per sec = 13759000
Optimized func: duration(ms) = 292, iterations = 28000000, calls per sec = 95890000

Old func: duration(ms) = 2010, iterations = 28000000, calls per sec = 13930000
New func: duration(ms) = 2001, iterations = 28000000, calls per sec = 13993000
Optimized func: duration(ms) = 290, iterations = 28000000, calls per sec = 96551000

Old func: duration(ms) = 1961, iterations = 28000000, calls per sec = 14278000
New func: duration(ms) = 2025, iterations = 28000000, calls per sec = 13827000
Optimized func: duration(ms) = 291, iterations = 28000000, calls per sec = 96219000
```

Source code of tests:
``` C++
PTF_TEST_CASE(TestHashPerfomance)
{
  // reading
  std::string errMsg;
  std::vector<RawPacket> rawPackets;
[...]

Issue #100, opened 2018-03-22
Create a TCP reordering module


seladb:
In some cases, TCP packets get scrambled on the network and don't arrive to the other side in the same order that they were sent. The purpose of this module is to re-order the in the same way they were sent, meaning according to the TCP stream order.
This sounds a little like TCP reassembly, but the difference is that this module should only re-order packets, nothing more. It should be simpler and more lightweight than TCP reassembly

wtdcode:
I have checked #82 , so is there any progress?

seladb:
I still haven't got to it. But if you can find time to build it it'd be great

echo-Mike:
Hi, @seladb !
I have thought about this task and want to share some thoughts with You.
The general reordering module must solve the next problem:
> Given some data __G__ represented as ordered sequence __S(K)__ = __{{p0, g0}, ..., {pK, gK}}__ of parts __{pi, gi}__ with distinct incrementing part numbers __pi__ (as __pi__ < __pi+1__ in mathematical sense) assigneg to each one let the __S'(K)__ be the _permutation_ of original set __S(K)__ and the __s(N)__ the subset of __S'(K)__ with cardinality of _N_ <= _K_ that _keeps_ the  relative order of elements of __S'(K)__ and the __q(M)__ the subset of __s(N)__ with cardinality of _M_ <= _N_ that do _not keeps_ relative order of elements of __s(N)__ the __general reorderer(GR)__ must while __"being fed"__ with sequence of parts selected randomly from __s(N)__ or __q(M)__ (with the respect of relative element ordering in each set -> "as-if" the _currrent_ pointer is assigned to each set and which one is advanced and dereferenced is chosen randomly) reconstruct the original ordered set __S(K)__ (stream all known parts in the same order as in __S(K)__) as soon as it will be possible and report any failures to do so.

This representation of the __GR__ will suffer from HOL(Head-Of-the-Line blocking) problem.
My questions:
1. Do I correctly understand the purpose and definition of __GR__ facility?
2. The HOL problem is the general problem for any _streaming_ channels (like TCP). In the TCP it is soleved by retransmissions but We are working with already finished (probably long time ago sessions) so I am afraid that perfect recreation of stream or the reasanoble reacreation of it without explicit requests (inform GR about session end) or fixed threshold buffering (flush after buffer of some size is filled with more than threshold value of entries) is not possibli. What are Your thoughts about it?  What kind of the HLD (high-level-design) should __GR__ follow?

[...]

PR #337, opened 2019-12-20
Add TcpSorter to support sorting TCP segments.

rickyzhang82:
I have performed online capturing test and offline pcap file test. The result looks good.

Also, I ran valgrind to check memory leak for hours. The shared pointer makes my life easier. I did find pcap library itself has memory leak when lists NIC. I didn't look deeper into the issue since I can't fix libpcap here.

Regarding to the TCP sorter logic, the TCP packet is flushed to the user once another side sends ACK. You can find more technical detail in the Doxygen header.

### DIFF
--- a/Packet++/src/TcpSorter.cpp
+++ b/Packet++/src/TcpSorter.cpp
@@ -511,7 +511,7 @@ void TcpSorter::flushPacket(SPTcpSorterData tcpSorterData, uint32_t ack, int rcv
 		// found missing packets
 		else
 		{
-			m_OnPacketMissingCallback(rcvIdx, tcpSorterData->connData, seq, seq - expSeq, m_UserCookie);
+			m_OnPacketMissingCallback(rcvIdx, tcpSorterData->connData, expSeq, seq - expSeq, m_UserCookie);
 			// update expected sequence number with packet sequence number
 			expSeq = seq;
 			// add back packet to the front
--- a/Packet++/src/TcpSorter.cpp
+++ b/Packet++/src/TcpSorter.cpp
@@ -1,0 +1,585 @@
+#define LOG_MODULE PacketLogModuleTcpSorter
+
+#include "TcpSorter.h"
+#include "TcpLayer.h"
+#include "IPv4Layer.h"
+#include "IPv6Layer.h"
+#include "PacketUtils.h"
+#include "IpAddress.h"
+#include "Logger.h"
+#include <sstream>
+#include <random>
+#include <vector>
+#include <algorithm>
+#if defined(WIN32) || defined(PCAPPP_MINGW_ENV) //for using ntohl, ntohs, etc.
+#include <winsock2.h>
+#elif LINUX
+#include <in.h> //for using ntohl, ntohs, etc.
+#elif MAC_OS_X || FREEBSD
+#include <arpa/inet.h> //for using ntohl, ntohs, etc.
+#endif
+
+#define PURGE_FREQ_SECS 1
+
+namespace pcpp
+{
+
+TcpSorter::TcpSorter(
+							OnTcpPacketReady onPacketReadyCallback,
+							OnTcpPacketMissing onPacketMissingCallback,
+							void* userCookie,
+							const TcpSorterConfiguration &config)
+{
+	m_isClosed = false;
+	m_OnPacketReadyCallback = onPacketReadyCallback;
+	m_OnPacketMissingCallback = onPacketMissingCallback;
+	m_UserCookie = userCookie;
+	// configuration value
+	m_MaxNumCapturedPacket = config.maxNumCapturedPacket;
+	m_MaxIdleTimeout = config.maxIdleTimeout;
+	m_MaxNumInactiveConnScan = config.maxNumInactiveConnScan;
+	m_MaxSegmentLifeTime = config.maxSegmentLifeTime;
+	m_CleanUpInactiveConnPeriod = config.cleanUpInactiveConnPeriod;
+	m_ShouldIncludeEmptySegments = config.shouldIncludeEmptySegments;
+	// reset clean up timer
+	m_LastCleanupTime = time(NULL);
+}
+
+TcpSorter::~TcpSorter()
+{
+	closeAllConnections();
+}
+
+void TcpSorter::sortPacket(SPRawPacket spRawPacket)
+{
+	// stop sorting if it is closed
+	if(m_isClosed)
+	{
+		return;
+	}
+	// parse the packet
+	Packet packet(spRawPacket.get(), false);
+
+	time_t now = time(NULL);
+
+	// clean up any expired inactive TCP connection
+	if (0 != m_MaxIdleTimeout && m_LastCleanupTime + m_CleanUpInactiveConnPeriod < now)
+	{
+		cleanUpInactiveTcpConnection(now);
+		m_LastCleanupTime = now;
+	}
+
+	// get IP layer
+	Layer* ipLayer = nullptr;
+	if (packet.isPacketOfType(IPv4))
+		ipLayer = dynamic_cast<Layer*>(packet.getLayerOfType<IPv4Layer>());
+	else if (packet.isPacketOfType(IPv6))
+		ipLayer = dynamic_cast<Layer*>(packet.getLayerOfType<IPv6Layer>());
+
+	if (ipLayer == nullptr)
+		return;
+
+	// Ignore non-TCP packets
+	TcpLayer* tcpLayer = packet.getLayerOfType<TcpLayer>();
+	if (tcpLayer == nullptr)
+		return;
+
+	// Ignore the packet if it's an ICMP packet that has a TCP layer
+	// Several ICMP messages (like "destination unreachable") have TCP data as part of the ICMP message.
+	// This is not real TCP data and packet can be ignored
+	if (packet.isPacketOfType(ICMP))
+	{
+		LOG_DEBUG("Packet is of type ICMP so TCP data is probably  part of the ICMP message. Ignoring this packet");
+		return;
+	}
+
+	// set the TCP payload size
+	uint32_t tcpPayloadSize =static_cast<uint32_t>(tcpLayer->getLayerPayloadSize());
+
+	// get TCP flag
+	bool isAck = (tcpLayer->getTcpHeader()->ackFlag == 1);
+	bool isSyn = (tcpLayer->getTcpHeader()->synFlag == 1);
+	bool isFin = (tcpLayer->getTcpHeader()->finFlag == 1);
+	bool isRst = (tcpLayer->getTcpHeader()->rstFlag == 1);
+
+	// TCP connection for captured packet
+	SPTcpSorterData tcpSorterData = nullptr;
+
+	// calculate flow key for this packet
+	uint32_t flowKey = hash5Tuple(&packet);
+
+	// find the connection in the connection map
+	ConnectionList::iterator iter = m_ConnectionList.find(flowKey);
+
+	// get packet's source and dest IP address
+	IPAddress* srcIP = nullptr;
+	IPAddress* dstIP = nullptr;
+	IPv4Address srcIP4Addr = IPv4Address::Zero;
+	IPv6Address srcIP6Addr = IPv6Address::Zero;
+	IPv4Address dstIP4Addr = IPv4Address::Zero;
+	IPv6Address dstIP6Addr = IPv6Address::Zero;
+	if (ipLayer->getProtocol() == IPv4)
+	{
+		srcIP4Addr = (dynamic_cast<IPv4Layer*>(ipLayer))->getSrcIpAddress();
+		srcIP = &srcIP4Addr;
+		dstIP4Addr = (dynamic_cast<IPv4Layer*>(ipLayer))->getDstIpAddress();
+		dstIP = &dstIP4Addr;
+	}
+	else if (ipLayer->getProtocol() == IPv6)
+	{
+		srcIP6Addr = (dynamic_cast<IPv6Layer*>(ipLayer))->getSrcIpAddress();
+		srcIP = &srcIP6Addr;
+		dstIP6Addr = (dynamic_cast<IPv6Layer*>(ipLayer))->getDstIpAddress();
+		dstIP = &dstIP6Addr;
+	}
+
+	// if TCP connection doesn't exist, create one
+	if (iter == m_ConnectionList.end())
+	{
+		// The SYN flag in the first packet must be set.
+		if (! isSyn)
+		{
+			LOG_DEBUG("Ignore packet of a flow [0x%X] that doesn't start with 3 way handshake.", flowKey);
+			return;
+		}
+		// create a TcpSorterData object and add it to the active connection list
+		tcpSorterData = std::make_shared<TcpSorterData>();
+		tcpSorterData->connData.setSrcIpAddress(srcIP);
+		tcpSorterData->connData.setDstIpAddress(dstIP);
+		tcpSorterData->connData.srcPort = ntohs(tcpLayer->getTcpHeader()->portSrc);
+		tcpSorterData->connData.dstPort = ntohs(tcpLayer->getTcpHeader()->portDst);
+		tcpSorterData->connData.flowKey = flowKey;
+		timeval ts = packet.getRawPacket()->getPacketTimeStamp();
+		tcpSorterData->connData.setStartTime(ts);
+		m_ConnectionList[flowKey] = tcpSorterData;
+	}
+	else // connection already exists
+	{
+		tcpSorterData = iter->second;
+	} // end of connection exists
+
+	// sender's side index
+	int sndIdx = -1;
+
+	// get packet's source port
+	uint16_t srcPort = tcpLayer->getTcpHeader()->portSrc;
+
+	// if this is a new connection and it's the first packet we see on that connection
+	if (tcpSorterData->numOfSides == 0)
+	{
+		LOG_DEBUG("Setting side for new connection");
+
+		// open the first side of the connection, side index is 0
+		sndIdx = 0;
+		tcpSorterData->twoSides[sndIdx].setSrcIP(srcIP);
+		tcpSorterData->twoSides[sndIdx].srcPort = srcPort;
+		tcpSorterData->numOfSides++;
+	}
+	// if there is already one side in this connection (which will be at side index 0)
+	else if (tcpSorterData->numOfSides == 1)
+	{
+		// check if packet belongs to that side
+		if (tcpSorterData->twoSides[0].srcIP->equals(srcIP) && tcpSorterData->twoSides[0].srcPort == srcPort)
+		{
+			sndIdx = 0;
+		}
+		else
+		{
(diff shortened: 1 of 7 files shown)


seladb:
Thanks for working on this!

Could you please check the CI failures? I had a quick look and it seems it's failing because you're using C++11 features. PcapPlusPlus is intended to support older C++ versions as well

rickyzhang82:
C++11 is fully supported by [GCC 4.8.1](https://gcc.gnu.org/projects/cxx-status.html#cxx11). Any Linux distribution and FreeBSD which are **NOT** in end of life provide GCC version above 4.8.1.

I won't take away C++11 feature in this PR. It is up to you if you want to add compiler flag `-std=c++11` or not. That can address most (if not all) compile problem.

seladb:
Unfortunately I don't think I'll add the `-std=c++11` flag for now for the reason I mentioned before. Can I please understand which parts of your code depend on C++11 features? Maybe we can change those pieces of code to make it compatible with older versions of C++

rickyzhang82:
It is just two weeks away that we will start our next decade. I'm surprised that if we still say no to C++11 standard. Anyway, it is a choice you made. I respect that. But I won't patch it by myself. So please drop the PR if nobody is going to make it C++11 free.

I used the following C++11 features to save me from typing and troubles:

1. Use static_cast/dynamic_cast in casting. (This might not be not C++11, I'd guess)
1. Use nullptr whenever possible to replace NULL pointer in C++.
1. Use auto keyword to avoid typing a long string of actual type.
1. Use smart pointer like shared_ptr whenever I allocate memory from heap.

From 1-3, you can replace it mechanically. The 4th bullet point involves more work. I consider it is a crime in 2020 to ask any programmers to free up memory manually.

seladb:
Thanks @rickyzhang82 for all your work and efforts. Since the rest of the library is not C++11 compatible I prefer not to use C++11 in just specific modules. I'll try to make those changes you suggest to make it C++98 compatible. 

Thanks again for all of your contribution to this project

rickyzhang82:
I patched one minor bug I found. The missing packet callback should pass in expected sequence number, instead.

It is a pity that you don't enable C++11 for older GCC. I have to fork your repository to maintain mine now.

gx740:
@rickyzhang82: let me to propose you to use a `std::unordered_map` instead of `std::map` in your local repository. The lookup in an `unordered_map` works in 10 times faster than in `std::map`.

rickyzhang82:
But I need the ordering of (seq + TCP payload size) when flushing the packets by other’s side ACK.

Does RB tree that slow? I’m curious about this. Because each TCP connection should not hold that many unacknowledged TCP packets at any given time.

I did find that the longer callback function execution, the higher probability the libpcap loses the packet in capturing. I’m thinking about using multithreaded solution. See my discussion issue #342

gx740:
The `m_ConnectionList` is a `std::map`. It can be replaced by a `std::unordered_map`.

rickyzhang82:
That’s true. Will change it.

gx740:
`ConnectionData` has two pointers to IPAddress instances that are heavyweight. I have prepared the new classes but migration process has not completed.

Change the signature of callback from
```c++
typedef void (*OnTcpPacketReady)(int side, ConnectionData connData, SPRawPacket spRawPacket, void* userCookie);
```
to
```c++
typedef void (*OnTcpPacketReady)(int side, const ConnectionData& connData, SPRawPacket spRawPacket, void* userCookie);
```

gx740:
The following code is also heavyweight:
```c++
// get packet's source and dest IP address
IPAddress* srcIP = nullptr;
IPAddress* dstIP = nullptr;
IPv4Address srcIP4Addr = IPv4Address::Zero; // !!! mem alloc
IPv6Address srcIP6Addr = IPv6Address::Zero; // !!! mem alloc
IPv4Address dstIP4Addr = IPv4Address::Zero; // !!! mem alloc
IPv6Address dstIP6Addr = IPv6Address::Zero; // !!! mem alloc
if (ipLayer->getProtocol() == IPv4)
{
  srcIP4Addr = (dynamic_cast<IPv4Layer*>(ipLayer))->getSrcIpAddress(); // !!! mem alloc
  srcIP = &srcIP4Addr;
  dstIP4Addr = (dynamic_cast<IPv4Layer*>(ipLayer))->getDstIpAddress(); // !!! mem alloc
  dstIP = &dstIP4Addr;
}
else if (ipLayer->getProtocol() == IPv6)
{
  srcIP6Addr = (dynamic_cast<IPv6Layer*>(ipLayer))->getSrcIpAddress();
  srcIP = &srcIP6Addr;
  dstIP6Addr = (dynamic_cast<IPv6Layer*>(ipLayer))->getDstIpAddress();
  dstIP = &dstIP6Addr;
}
```

gx740:
Look into PR #308 in which I have proposed the performance improvements. You can use it in your code  in place where a hash is calculated.

seladb:
@rickyzhang82 I'll review the code soon, but one thing that is missing is UTs. Could you please add some to either `Packet++Test` or `Pcap++Test`?

You can take a look at the UTs for TcpReassembly:

https://github.com/seladb/PcapPlusPlus/blob/ed30c808bec1d42c1a9d706ef5f793aade636402/Tests/Pcap%2B%2BTest/main.cpp#L4692-L5370

rickyzhang82:
Please let me add UT and send another PR later.

rickyzhang82:
@gx740 

Do you have a empirical test to verify your claim that  

> The lookup in an unordered_map works in 10 times faster than in std::map

Also, how about insertion and deletion?

rickyzhang82:
@gx740 

Regarding to the copy cost, it makes sense to squeeze the performance by passing as const reference. I also change all methods that pass in shared pointer into pass in as const reference as well.

gx740:
> @gx740
> 
> Do you have a empirical test to verify your claim that
> 
> > The lookup in an unordered_map works in 10 times faster than in std::map
> 
> Also, how about insertion and deletion?

Yes, I did the tests but did not test the deletion. You can make the tests in your environment.

Test results: Win10x64, VS2019, Core i7-9700K
```
Fill the map: duration(ms): 22, count: 100.000
Lookup in the map: duration(ms): 2469, iterations: 15.000.000
Insuccessfull lookup in the map: duration(ms): 2622, iterations: 15.000.000

Fill the unordered map: duration(ms): 7, count: 100.000
Lookup in the unordered map: duration(ms): 133, iterations: 15.000.000
Insuccessfull lookup in the unordered map: duration(ms): 154, iterations: 15.000.000
```

Memory usage (valgrind result): CentOS 7, g++ 8.3.1

**key**: uint32_t, **value**: uint64_t, **size**: 500 000

**map:**
500,000 allocs, 500,000 frees, 24,000,000 bytes allocated

**unordered_map:**
500,018 allocs, 500,018 frees, 20,204,296 bytes allocated
500,001 allocs, 500,001 frees, 16,161,928 bytes allocated (pre-alloc)



```c++
#include <vector>
#include <map>
#include <cstdint>
#include <chrono>
#include <iostream>
#include <algorithm>
#include <cstdlib>
#include <ctime>
#include <unordered_map>
#include <random>


void testHashMapAndMap()
{
  constexpr auto tryCount = 15000000U;
  constexpr uint32_t numToSearch = 1U;
  constexpr auto maxSize = 100000U;

  // create a shuffled unique set
  std::unordered_map<uint32_t, uint32_t> shuffledMap, shuffledErrMap;
  shuffledMap.reserve(maxSize);
  shuffledErrMap.reserve(maxSize);

  using RandomGenerator = std::mt19937;
  std::random_device dev;
  RandomGenerator randomGenerator(dev());
  std::uniform_int_distribution<RandomGenerator::result_type> dist(1, maxSize * 3);

  while(shuffledMap.size() < maxSize)
  {
    auto rndValue = dist(randomGenerator);
    if(shuffledMap.find(rndValue) == shuffledMap.cend())
      shuffledMap.insert(std::make_pair(rndValue, rndValue));
  }
  
  while(shuffledErrMap.size() < maxSize)
  {
    auto rndValue = dist(randomGenerator);
    if(shuffledMap.find(rndValue) == shuffledMap.cend() && shuffledErrMap.find(rndValue) == shuffledErrMap.cend())
      shuffledErrMap.insert(std::make_pair(rndValue, rndValue));
  }

  std::vector<uint32_t> keys, errKeys;
  keys.reserve(shuffledMap.size());
  errKeys.reserve(shuffledErrMap.size());

  for(auto const &pair : shuffledMap)
    keys.push_back(pair.first);
  
  for(auto const &pair : shuffledErrMap)
    errKeys.push_back(pair.first);

  shuffledMap.clear();
  shuffledErrMap.clear();

  {
    // fill the map
    std::map<uint32_t, uint64_t> map;

    auto startTime = std::chrono::steady_clock::now();

    for(auto key : keys)
      map.insert(std::make_pair(key, key));

    auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - startTime);
    std::cout << "Fill the map: duration(ms): " << duration.count() << ", count: " << map.size() << '\n';

    // lookup in the map
    auto count = 0U;

    startTime = std::chrono::steady_clock::now();

    for(auto i = 0U; i < tryCount; ++i)
      if(map.find(keys[i % maxSize]) != map.cend())
        ++count;

    duration = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - startTime);
    std::cout << "Lookup in the map: duration(ms): " << duration.count() << ", iterations: " << count << '\n';

    // insuccessfull lookup in the map
    count = 0U;

    startTime = std::chrono::steady_clock::now();

    for(auto i = 0U; i < tryCount; ++i)
      if(map.find(errKeys[i % maxSize]) == map.cend())
        ++count;

    duration = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - startTime);
    std::cout << "Insuccessfull lookup in the map: duration(ms): " << duration.count() << ", iterations: " << count << '\n';

    std::cout << '\n';
  }

  {
    // fill the unordered map

    std::unordered_map<uint32_t, uint64_t> unorderedMap;
    //hashMap.reserve(keys.size());

    auto startTime = std::chrono::steady_clock::now();

    for(auto key : keys)
      unorderedMap.insert(std::make_pair(key, key));

    auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - startTime);
    std::cout << "Fill the unordered map: duration(ms): " << duration.count() << ", count: " << unorderedMap.size() << '\n';

    // lookup in the hash map
    auto count = 0U;

    startTime = std::chrono::steady_clock::now();

    for(auto i = 0U; i < tryCount; ++i)
      if(unorderedMap.find(keys[i % maxSize]) != unorderedMap.cend())
        ++count;

    duration = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - startTime);
    std::cout << "Lookup in the unordered map: duration(ms): " << duration.count() << ", iterations: " << count << '\n';

    // insuccessfull lookup
    count = 0U;

    startTime = std::chrono::steady_clock::now();

    for(auto i = 0U; i < tryCount; ++i)
      if(unorderedMap.find(errKeys[i % maxSize]) == unorderedMap.cend())
        ++count;

    duration = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - startTime);
    std::cout << "Insuccessfull lookup in the unordered map: duration(ms): " << duration.count() << ", iterations: " << count << '\n';

    std::cout << '\n';
  }
}


constexpr auto maxSize = 500000U;

void mapTestMem()
{
  std::map<uint32_t, uint64_t> map;

  for(auto i = 0U; i < maxSize; ++i)
    map.emplace(i, i);
}

void unorderedMapTestMem()
{
  std::unordered_map<uint32_t, uint64_t> map;
  map.reserve(maxSize);

  for(auto i = 0U; i < maxSize; ++i)
    map.emplace(i, i);
}

{
  testHashMapAndMap();
  return 0;
}
```

Packet++/src/TcpSorter.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+void TcpSorter::sortPacket(SPRawPacket spRawPacket)
+{
+	// stop sorting if it is closed
+	if(m_isClosed)
+	{
+		return;
+	}
+	// parse the packet
+	Packet packet(spRawPacket.get(), false);
```
gx740: You can make a small optimization if it is applicable: `Packet packet(spRawPacket.get(), false, TCP)`

Packet++/src/TcpSorter.cpp
```diff
@@ -0,0 +1,585 @@
...
+	else if (packet.isPacketOfType(IPv6))
+		ipLayer = dynamic_cast<Layer*>(packet.getLayerOfType<IPv6Layer>());
+
+	if (ipLayer == nullptr)
+		return;
+
+	// Ignore non-TCP packets
+	TcpLayer* tcpLayer = packet.getLayerOfType<TcpLayer>();
+	if (tcpLayer == nullptr)
+		return;
```
gx740: Possible optimizations:

```c++
  // get IP layer
  Layer* ipLayer;
  if (packet.isPacketOfType(IPv4))
    ipLayer = dynamic_cast<Layer*>(packet.getLayerOfType<IPv4Layer>());
  else if (packet.isPacketOfType(IPv6))
    ipLayer = dynamic_cast<Layer*>(packet.getLayerOfType<IPv6Layer>());
  else
    return;

  // Ignore non-TCP packets
  TcpLayer* tcpLayer = packet.getNextLayerOfType<TcpLayer>(ipLayer);
  if (tcpLayer == nullptr)
    return;
```

If you are sure that the packet is parsed until TCP the following code can be used:

```c++
  // get IP layer
  Layer* ipLayer;
  if (packet.isPacketOfType(IPv4))
    ipLayer = dynamic_cast<Layer*>(packet.getLayerOfType<IPv4Layer>());
  else if (packet.isPacketOfType(IPv6))
    ipLayer = dynamic_cast<Layer*>(packet.getLayerOfType<IPv6Layer>());
  else
    return;

  // Ignore non-TCP packets
  if (packet.isPacketOfType(TCP))
    return;
	
  TcpLayer* tcpLayer = packet.getLastLayer();
```

Packet++/src/TcpSorter.cpp
```diff
@@ -0,0 +1,585 @@
...
+	if (iter == m_ConnectionList.end())
+	{
+		// The SYN flag in the first packet must be set.
+		if (! isSyn)
+		{
+			LOG_DEBUG("Ignore packet of a flow [0x%X] that doesn't start with 3 way handshake.", flowKey);
+			return;
+		}
+		// create a TcpSorterData object and add it to the active connection list
+		tcpSorterData = std::make_shared<TcpSorterData>();
```
gx740: If a `unique_ptr` is acceptiable in this case you should use it instead of `shared_ptr`

Packet++/src/TcpSorter.cpp
```diff
@@ -0,0 +1,585 @@
...
+			tcpSorterData->twoSides[sndIdx].setSrcIP(srcIP);
+			tcpSorterData->twoSides[sndIdx].srcPort = srcPort;
+			tcpSorterData->numOfSides++;
+		}
+	}
+	// if there are already 2 sides open for this connection
+	else if (tcpSorterData->numOfSides == 2)
+	{
+		// check if packet matches side 0
+		if (tcpSorterData->twoSides[0].srcIP->equals(srcIP) && tcpSorterData->twoSides[0].srcPort == srcPort)
```
gx740: Small optimization: change the order of the checks

Packet++/src/TcpSorter.cpp
```diff
@@ -0,0 +1,585 @@
...
+		{
+			sndIdx = 0;
+		}
+		// check if packet matches side 1
+		else if (tcpSorterData->twoSides[1].srcIP->equals(srcIP) && tcpSorterData->twoSides[1].srcPort == srcPort)
```
gx740: Small optimization: change the order of the checks

Packet++/src/TcpSorter.cpp
```diff
@@ -0,0 +1,585 @@
...
+{
+	// if disable clean up
+	if (0 == m_MaxIdleTimeout)
+	{
+		return;
+	}
+
+	// collect flowKeys from TCP connection list
+	auto iterKey = m_ConnectionList.begin(), iterKeyEnd = m_ConnectionList.end();
+	std::vector<uint32_t> accessKeys;
```
gx740: Append the following: `accessKeys.reserve(m_ConnectionList.size());`

seladb:
hi @rickyzhang82 I'm wondering if we can revive this PR. I think you almost completed the work, why shouldn't we complete the remaining work and merge it?

rickyzhang82:
There are 3 road blocks:

1. No unit test
1. C++ 11
1. missing capturing packet at rate `0.73%` by libpcap

If you can live with all 3, I will send a PR again.

seladb:
with number 3 I can live and we can work together on numbers 1 and 2.

Specifically for number 2 - I can help if you'd like me to.

rickyzhang82:
C++ 11 is a deal breaker to me. The app `PacketSorter` has ran for almost two months in my router. I don't want to mess with memory leak in Valgrind because of taking away smart pointer.

seladb:
You can still run your own branch of `TcpSorter` with the C++11 features and at the same merge to `master` a version that doesn't include these feature

rickyzhang82:
I prefer to have one single code base. Otherwise, neither of us share the benefit of the merging.

I may find bugs and patch it in future. But the fix may not be in your repository and vice versa.

seladb:
the issue itself is still open (#100) so at some point someone else might want to implement the TCP reorder/sort functionality. That will be a pity because you already implemented it, and we'll end up having 2 implementations of the same thing.

Can we find any way to bridge this gap and get your code merged?

If I remember correctly there aren't a lot of places you're using C++11 specific features so maybe we can work something out

rickyzhang82:
I'm not convinced that removing smart pointer is a good idea in year 2020. So I can't compromise on C++ 11.

One of the callback function makes a copy of packet and passes it down to the end user application. It is better leave it to the smart pointer to do clean up. I have no interest to exercise memory leak detection again in Valgrind.

seladb:
Ok I give up. If you think of any way we can bridge the gaps and get your PR merged please let me know. Otherwise let's keep it as is

rickyzhang82:
Please let me know when you allow C++ 11 in PcapPlusPlus.

All latest Linux and FreeBSD comes with GCC and Clang supports C++ 11.

Thanks!

seladb:
hi @rickyzhang82 do you mind if I use your code and make some necessary changes to support older versions of C++?

rickyzhang82:
@seladb Please feel free to use it from my my-master branch: 

https://github.com/rickyzhang82/PcapPlusPlus

Thanks!

seladb:
Thanks! since the branch you created the PR from is gone, the only way I see is to copy the code into my branch and then do the necessary changes. However that way your name won't be in it. Are you ok with this or do you want to create another PR that I'll later modify?

seladb:
@rickyzhang82 did you get my previous comment?

rickyzhang82:
No problem. Feel free to do whatever you want!

Thanks!

seladb:
Thank you @rickyzhang82 , I'll keep you posted

PR #345, opened 2019-12-23
TcpReassembly: decreasing the memory consumption (for C++11 compilers)

gx740:
For C++11 compilers we can use `std::forward_list` container instead of `std::list` which is using less memory than the last one (minus 4 or 8 bytes per node).


seladb:
I don't think we should add C++11-specific features. It makes the code more complex and less readable

gx740:
Let's change the code of two methods. It will allow replacing the container in a local user repository and easy sync with the parent one.

seladb:
Did you remove the C++11 specific code? I can't see it anymore

gx740:
Yes, I did it

seladb:
Got it. I'm trying to understand the change this PR is making. I guess the headline is a little misleading...

gx740:
I have just rearrange the code and it is applicable for both containers, nothing else.

gx740:
The order in which the connections are cleaned up is not important

seladb:
Ok, I'll merge it later

PR #346, opened 2019-12-25
Lookup a layer in reverse order (enhancement)

gx740:
In many cases the lookup in reverse order in packet is faster than in the direct one.

This PR adds such feature and improvements in a number of modules.

### DIFF
--- a/Tests/Packet++Test/main.cpp
+++ b/Tests/Packet++Test/main.cpp
@@ -7561,6 +7561,32 @@ PTF_TEST_CASE(EthDot3LayerCreateEditTest)
 
 
 
+PTF_TEST_CASE(PacketLayerLookupTest)
+{
+	// reverse lookup
+	int bufferLength = 0;
+	uint8_t* buffer = readFileIntoBuffer("PacketExamples/radius_1.dat", bufferLength);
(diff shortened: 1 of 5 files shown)


gx740:
For example: when a packet is being passed into IP reassembly logic the last layer, usually, is a IPv4/v6 so the searching has two operations: `dynamic_cast` and comparing to `NULL`.

gx740:
I was wrong. An example above is incorrect. Currently the user is forced to make parsing until the layer he/she interested in. So the latest layer will not be the IPv4/v6.

Tests/Packet++Test/main.cpp
```diff
@@ -7561,6 +7561,32 @@ PTF_TEST_CASE(EthDot3LayerCreateEditTest)
 
 
 
+PTF_TEST_CASE(PacketLayerLookupTest)
```
seladb: I'd suggest to add a few tests:

- Packet that has a few layers of the same type (you can use `Vxlan1.dat`)
- Try to get prev layer of the first layer (result should be `NULL`)
- Try `getLayerOfType<...>(true)` with a layer type that doesn't exist in the packet

gx740: I will do it today

PR #351, opened 2019-12-27
Performance and readability improvements in DhcpOption

gx740:


Packet++/header/DhcpLayer.h
```diff
@@ -457,11 +457,11 @@ namespace pcpp
 		 */
 		void setValueString(const std::string& stringValue, int valueOffset = 0)
 		{
-			std::string val = stringValue;
-			if (stringValue.length() > (size_t)m_Data->recordLen - (size_t)valueOffset)
-				val = stringValue.substr(0, (size_t)m_Data->recordLen - valueOffset);
+			size_t len = (size_t)m_Data->recordLen - (size_t)valueOffset;
+			if (stringValue.length() < len)
```
seladb: I think the `>` sign should remain (you replaced it with `<`), am I wrong?

gx740: The `>` should be. I have added the comments for easy understanding

PR #352, opened 2019-12-27
Performance and readability improvements in IFileReaderDevice::getReader

gx740:


seladb:
I don't think this change significantly improves the readability of the code. And performance is also not a very big deal here

gx740:
I'm letting you decide what to with this PR

gx740:
I have removed the creation of temporary string objects. So it's the simplest solution.

PR #354, opened 2019-12-27
[IP addresses migration] TCP reassembly and corresponding unit tests

gx740:


seladb:
In my opinion, it'd be more beneficial to first replace the old IP addresses implementation with the new one and then make the changes in the other parts of the project. That way will require less changes , for example you won't have to add `pcpp::experimental::` everywhere...

gx740:
Ok, let try this approach

PR #373, opened 2020-01-10
Pr bug fix

rickyzhang82:


Pcap++/src/PcapFileDevice.cpp
```diff
@@ -560,6 +560,9 @@ void PcapFileWriterDevice::flush()
 
 void PcapFileWriterDevice::close()
 {
+	if (!m_DeviceOpened)
+		return;
+
```
seladb: In addition to this change, maybe we also want to add a check before calling `fclose(m_File)` to verify that `m_File != NULL`?

rickyzhang82: Done.

PR #375, opened 2020-01-10
[IP addresses] improvements in IPReassembly

gx740:


Packet++/header/IPReassembly.h
```diff
@@ -92,13 +92,6 @@ namespace pcpp
 			 * @return A pointer to a new instance which is a clone of the current instance
 			 */
 			virtual PacketKey* clone() const = 0;
-
-		protected:
-			// private c'tor
-			PacketKey() {}
-
-			// private copy c'tor
-			PacketKey(const PacketKey& other);
```
seladb: Why did you remove these methods?

Packet++/src/IPReassembly.cpp
```diff
@@ -58,13 +58,10 @@ class IPFragmentWrapper
 	virtual uint8_t* getIPLayerPayload() = 0;
 	virtual size_t getIPLayerPayloadSize() = 0;
 
-	virtual ~IPFragmentWrapper() { }
-
-protected:
-
-	IPFragmentWrapper() {}
```
seladb: why did you remove the protected c'tor?

Packet++/src/IPReassembly.cpp
```diff
@@ -290,7 +280,7 @@ Packet* IPReassembly::processPacket(Packet* fragment, ReassemblyStatus& status,
 	// create fragment wrapper
 	IPv4FragmentWrapper ipv4Wrapper(fragment);
 	IPv6FragmentWrapper ipv6Wrapper(fragment);
-	IPFragmentWrapper* fragWrapper = NULL;
+	IPFragmentWrapper* fragWrapper;
```
seladb: I suggest keeping this `NULL` assignment, just in case

seladb: Did you have a chance to see my comment?

PR #377, opened 2020-01-10
[IP addresses] improvements in PcapLiveDevice

gx740:


Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -926,10 +920,8 @@ ThreadStart PcapLiveDevice::getCaptureThreadStart()
 
 PcapLiveDevice::~PcapLiveDevice()
 {
-	if (m_Name != NULL)
-		delete [] m_Name;
-	if (m_Description != NULL)
-		delete [] m_Description;
+	delete [] m_Name;
+	delete [] m_Description;
```
seladb: why removing the `NULL` checks?

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -819,9 +818,9 @@ void PcapLiveDevice::setDefaultGateway()
 	if (retVal == NO_ERROR)
 	{
 		PIP_ADAPTER_INFO curAdapterInfo = adapterInfo;
+		std::string name(m_Name);
```
seladb: I'm not sure why you moved this line to here

Pcap++/src/PcapLiveDevice.cpp
gx740: C++ guarantees correct behavior when a `NULL` pointer is passed to operator `delete`. So the check is unnecessary.

Pcap++/src/PcapLiveDevice.cpp
gx740: I had placed this line out of scope of the loop because the value is not changed there.

Pcap++/src/PcapLiveDevice.cpp
seladb: Are you sure this is the case in all platforms and toolchains (including MinGW, Visual Studio, etc.)?

gx740: It's included into language standard: https://en.cppreference.com/w/cpp/language/delete
```
For the first (non-array) form, expression must be a pointer to an object type or a class type contextually implicitly convertible to such pointer, and its value must be either null or pointer to a non-array object created by a new-expression...
```

**(until C++14)**
```
If expression evaluates to a null pointer value, no destructors are called, and the deallocation function may or may not be called (it's implementation-defined), but the default deallocation functions are guaranteed to do nothing when handed a null pointer.
```
**(since C++14)**
```
If expression evaluates to a null pointer value, no destructors are called, and the deallocation function is not called.
```

PR #378, opened 2020-01-10
[IP addresses] improvements in PcapFilter, PcapRemoteDevice and PcapRemoteDeviceList

gx740:


Pcap++/src/PcapFilter.cpp
```diff
@@ -170,17 +169,15 @@ void IPFilter::convertToIPAddressWithLen(std::string& ipAddrmodified, int& len)
 	}
 	else if (ipAddr.isIPv4())
 	{
-		const IPv4Address& ip4Addr = ipAddr.getIPv4();
-		uint32_t addrAsInt = ip4Addr.toUInt();
-		uint32_t mask = ((uint32_t)-1) >> ((sizeof(uint32_t)*8)-m_Len);
+		uint32_t addrAsInt = ipAddr.getIPv4().toUInt();
+		uint32_t mask = (uint32_t)(-1) >> ((sizeof(uint32_t) * 8) - m_Len);
```
gx740: Very hard to understand what does this code

Pcap++/src/PcapFilter.cpp
```diff
@@ -170,17 +169,15 @@ void IPFilter::convertToIPAddressWithLen(std::string& ipAddrmodified, int& len)
...
 		ipAddrmodified = IPv4Address(addrAsInt).toString();
 	}
 	else // IPv6
 	{
-		const IPv6Address& ip6Addr = ipAddr.getIPv6();
-		uint8_t* addrAsArr; size_t addrLen;
-		ip6Addr.copyTo(&addrAsArr, addrLen);
+		uint8_t addrAsArr[16];
+		ipAddr.getIPv6().copyTo(addrAsArr);
 		uint64_t addrLowerBytes = (long)addrAsArr;
```
gx740: Is this code right? A pointer is being converted to `uint64_t`...

gx740: Is this code right? A pointer is being converted to uint64_t...

gx740: I think, this method has an error for IPv6 addresses because `addrLowerBytes` and `addrHigherBytes` variables are not copied back to `addrAsArr`

Pcap++/src/PcapFilter.cpp
seladb: yes I agree. Any suggestion on how to make that more clear?

Pcap++/src/PcapFilter.cpp
seladb: yes I think you're right, this is probably not covered in the unit-test. Would you mind fixing it?

Pcap++/src/PcapFilter.cpp
gx740: I have made the changes in the latest commit

Pcap++/src/PcapFilter.cpp
gx740: I guess `IPFilter` does not support `IPv6`. I have found the following `@todo Add IPv6 filtering support` in header.

seladb: I think you're right 😃 I wrote this code a long time ago, I can't remember why I added this piece of code. Could you please remove it?

seladb:
Can you please run all test cases that involve `IPFilter` to make sure nothing got broken?

Pcap++/src/PcapFilter.cpp
gx740: Code has been removed

seladb:
Thanks for making the changes. Did you have a chance to run all `IPFilter` test-cases to make sure nothing got broken?

gx740:
I have added the tests for `parseToString` which uses the `convertToIPAddressWithLen`. Besides, there are the unit-tests in `TestPcapFiltersOffline` which use the `IPFilter` and `parseToString` and these tests are executed every time. So more checks are not necessary.

PR #384, opened 2020-01-10
[IP addresses] improvements in DhcpLayer

gx740:


Packet++/header/DhcpLayer.h
```diff
@@ -416,8 +416,7 @@ namespace pcpp
 		 */
 		IPv4Address getValueAsIpAddr() const
 		{
-			uint32_t addrAsInt = getValueAs<uint32_t>();
-			return IPv4Address(addrAsInt);
+			return getValueAs<uint32_t>();
```
seladb: why did you remove calling `IPv4Address` c'tor? Is the compiler adding that automatically?

gx740: Yes. `IPv4Address` has a constructor which takes a `uint32_t` as argument and it will be called automatically.

PR #395, opened 2020-01-25
[IP addresses] Improvements in TcpReassemby

gx740:
- changed the type of `ConnectionData::srcIP(dstIP)` from `IPAddress*` to `IPAddress`
- the static function `prepareMissingDataMessage` replaced the private method with the same name
- other performance and readability improvements


gx740:
Can you sync `iprefactor` branch with `dev` before merging?

Packet++/src/TcpReassembly.cpp
```diff
@@ -125,15 +67,42 @@ void TcpReassembly::reassemblePacket(Packet& tcpData)
...
+	else
+		return;
 
-	if (ipLayer == NULL)
+	// in real traffic the IP addresses cannot be an unspecified
+	if(srcIP.isUnspecified() || dstIP.isUnspecified())
+	{
+		LOG_ERROR("Some IP address is unspecified: srcIP [%s], dstIP [%s]. Ignoring this packet", srcIP.toString().c_str(), dstIP.toString().c_str());
 		return;
+	}
```
seladb: but maybe the input packet is not an IP packet. In this case this error message is somewhat confusing. That's why I didn't return an error message in the first place

gx740: Calling the LOG_ERROR has been removed

gx740:
Let's push `ipaddr_refactor` into `dev`. The next changes I will prepare in `dev`.

seladb:
what more changes are you planning?

at some point we should stop and do more more thorough testing to make sure nothing got broken. The CI unfortunately doesn't cover everything because it doesn't run with live traffic and it doesn't test the examples

gx740:
I will prepare two PRs. Other PR's that I am planning to prepare concern the unit tests and can be merged into `dev` instead of `ipaddr_refactor`

seladb:
sure. When you done with the changes on `ipaddr_refactor` let me know and we can run some tests

PR #397, opened 2020-01-28
TCP payload data validation before parsing

gx740:
The new static method `TcpLayer::isDataValid` allows to make the minimal checks before parsing the payload data. It is now used in `IPv4Layer::parseNextLayer` and `IPv6Layer::parseNextLayer`.

There are two malformed packets that lead to `bad_alloc` in `TcpReassembly`. The new method avoids this behavior.


Tests/Packet++Test/main.cpp
```diff
@@ -7716,6 +7735,7 @@ int main(int argc, char* argv[]) {
 	PTF_RUN_TEST(TcpPacketWithOptionsParsing2, "tcp");
 	PTF_RUN_TEST(TcpPacketCreation, "tcp");
 	PTF_RUN_TEST(TcpPacketCreation2, "tcp");
+	PTF_RUN_TEST(TcpPacketParsing, "tcp");
```
seladb: I'd suggest changing the test-case name to `TcpMalformedPacketParsing` or something similar because there are already test-cases that test TCP packet parsing: `TcpPacketNoOptionsParsing`, `TcpPacketWithOptionsParsing`, `TcpPacketWithOptionsParsing2`

PR #406, opened 2020-02-28
Integrate streaming ZStd compression for pcapng files

Dysl3xik:
Ok I finally got all of this updated and 🤞 100% ready to merge.

Whats Included:
0% Code changes - All of the code changes required for this to function were already in the codebase they were just idle until the #DEFINE USE_Z_STD preprocessor was added YAY

I updated the VS builder bat file to make activating this 100% painless, it will ask if you want to enable zstd, if you select yes it will prompt you for the path for the SDK - it also notifies the user where they can download it from, just like pthreads & pcap. 

I updated all the vs proj template files to add/remove all the include paths, lib paths, lib files, and #defines. I also modified the files a tad to put the output libs into folders based upon Platform/Configuration otherwise its too hard to track x64 vs x32 and debug release. Now you can build everything and have copies of all those files in easy to identify folders. Previously you can't tell if a dll or lib was a debug build or not. Same issue for x32 x64

I added UNIT tests to verity functionality. I copied your existing tests which read & write npcap files and check them against each other to run a second time but with compression. If the library is built without compression the results of the compression test pcapng file are binary the same as the non compressed file (this is because if compression isn't built, and the user requests it it just ignores that request and makes a non compressed file). I also verified that if you uncompress the zstd file they are binary the same. This was redundant because the unit test already reads the compressed file back in and compares it to the uncompressed but eh why not.

You can swap in some other compression library by implementing the interface I created, adding your own #define USE_MY_COMPRESSION and doing your own work 👷 

To enable compression you simply build a pcapng writer with a second arg, an int for compression level from 0-10. This kicks off the compressor to get used if available or else is ignored if compression was not built.

Whats Not Included:
The bat file does not (yet) support cmdline arguments for setting zstd up like you did with the other libs. You can't skip the wizard for now.

Appending to a compressed pcapng file is not implemented!!!! This may actually just "work" due to the way zstd framing is implemented but I didn't need it and I think it would be niche anyways? If it becomes a popular request it has a TODO stub in the code. Maybe you just open the file with the append flag and it will work, but, until someone changes that it will fail to open.

To read a compressed pcapng file it MUST have a .zstd extension! Otherwise it doesn't know its compressed and won't run it through decompression.

I don't work on Linux or MAC systems so I will not be able to configure their project settings, linkers, etc. But you or someone should be able to copy the work done here over without too much effort. You can get the zstd libs for all those platforms with ease.

Questions?


Dysl3xik:
Well something always has to break.... 

I realized I did NOT change the paths for the examples projects after I moved the libs (remember I added the Platform/Configuration directories). I can go and adjust these projects if you are OK with me moving those files in general. Let me know. Otherwise I think everything is OK.

Dysl3xik:
bump

seladb:
I'm sorry for the delayed response. Let me go over the PR and provide feedback. That might take me a few more days. I'm so sorry for the delay

seladb:
One thing even before I go over the code: the Windows CI doesn't pass, probably because when building the examples it cannot find the PcapPlusPlus lib files. Can you please fix that?

Dysl3xik:
Tests are passing EXCEPT the ones based on npcap. I think the DLL is not getting loaded for those test cases and the test can't launch successfully.

Aside from that, everything else seems to be in order now.

configure-windows-visual-studio.bat
```diff
@@ -24,11 +28,14 @@ if "%ERRORLEVEL%" NEQ "0" exit /B 1
 if "%VS_VERSION%"=="" echo Visual studio version was not provided. Exiting & exit /B 1
 if "%PTHREAD_HOME%"=="" echo pthread-win32 directory was not provided. Exiting & exit /B 1
 if "%PCAP_SDK_HOME%"=="" echo WinPcap/Npcap SDK directory was not provided. Exiting & exit /B 1
+if "%ZSTD_HOME%"=="" echo WinPcap/Npcap SDK directory was not provided. ZSTD Support will be skipped!
```
seladb: did you mean to say "ZStd SDK directory was not provided"?

configure-windows-visual-studio.bat
```diff
@@ -261,8 +272,41 @@ set /p PTHREAD_HOME=    Please specify pthreads-win32 path: %=%
...
+:while3
+set /p USE_ZSTD=     To enable ZStd support type use_zstd otherwise type no: %=%
+if "%USE_ZSTD%" NEQ "use_zstd" if "%USE_ZSTD%" NEQ "no" (echo Please choose one of "use_zstd" or "no" && goto while3)
+
+
+if "%USE_ZSTD%" EQU "use_zstd" (
+::Can't use set staement inside an if so we have to do a little trickery here
+goto while4
+)
+goto skipWhile4
```
seladb: instead of having this, maybe we can just ask for zstd SDK path and let the user know they can type "no" or simply press "return" if they want to skip it

configure-windows-visual-studio.bat
```diff
@@ -261,8 +272,41 @@ set /p PTHREAD_HOME=    Please specify pthreads-win32 path: %=%
...
+set "ZSTD_INCLUDE_PATH=$(ZStdHome)\include;"
+set ZSTD_LIB_NAME=libzstd.lib;
+
+:skipWhile4
+
+if "%USE_ZSTD%" EQU "no" set USE_ZSTD=
+
+echo.
+echo.
+
```
seladb: we should add zstd to the command-line option too (lines 165-234)

mk/vs/PcapPlusPlusPropertySheet.props.template
```diff
@@ -4,6 +4,7 @@
     <PcapPlusPlusHome>PUT_PCAPPLUSPLUS_HOME_HERE</PcapPlusPlusHome>
     <WinPcapHome>PUT_PCAP_SDK_HOME_HERE</WinPcapHome>
     <PThreadWin32Home>PUT_PTHREAD_HOME_HERE</PThreadWin32Home>
+	<ZStdHome>PUT_ZSTD_HOME_HERE</ZStdHome>
```
seladb: please change indentation from tab to spaces

Tests/Pcap++Test/main.cpp
```diff
@@ -1067,6 +1069,81 @@ PTF_TEST_CASE(TestPcapNgFileReadWrite)
 
 }
 
+PTF_TEST_CASE(TestPcapNgFileReadWriteCompress)
```
seladb: instead of duplicating the existing test-cases, maybe we can add another writer to the original test-case that will use compression and then check the two writers behave the same

Tests/Pcap++Test/main.cpp
```diff
@@ -1320,6 +1397,262 @@ PTF_TEST_CASE(TestPcapNgFileReadWriteAdv)
 	writerDev2.close();
 }
 
+PTF_TEST_CASE(TestPcapNgFileReadWriteAdvCompress)
```
seladb: instead of duplicating the existing test-cases, maybe we can add another writer to the original test-case that will use compression and then check the two writers behave the same

mk/vs/GitInfoPropertySheet.props
```diff
@@ -2,8 +2,8 @@
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ImportGroup Label="PropertySheets" />
   <PropertyGroup Label="UserMacros">
-    <GitCommit>unavailable</GitCommit>
-    <GitBranch>unavailable</GitBranch>
+    <GitCommit>a9fe051678ee9e05fefecbf0cf7d29452ce8792f</GitCommit>
+    <GitBranch>master</GitBranch>
```
seladb: please revert back to the original value

mk/vs/LightPcapNg.vcxproj.template
```diff
@@ -106,42 +106,54 @@
...
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <OutDir>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\Lib</OutDir>
     <IntDir>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\Obj</IntDir>
+    <LibraryPath>$(ZStdHome)\dll;$(VC_LibraryPath_x64);$(WindowsSDK_LibraryPath_x64);$(NETFXKitsDir)Lib\um\x64</LibraryPath>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <OutDir>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\Lib</OutDir>
     <IntDir>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\Obj</IntDir>
+    <LibraryPath>$(ZStdHome)\dll;$(VC_LibraryPath_x64);$(WindowsSDK_LibraryPath_x64);$(NETFXKitsDir)Lib\um\x64</LibraryPath>
```
seladb: why is it necessary to add `<LibraryPath>`?

mk/vs/LightPcapNg.vcxproj.template
```diff
@@ -106,42 +106,54 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <OutDir>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\Lib</OutDir>
     <IntDir>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\Obj</IntDir>
+	<LibraryPath>$(ZStdHome)\dll;$(VC_LibraryPath_x86);$(WindowsSDK_LibraryPath_x86);$(NETFXKitsDir)Lib\um\x86</LibraryPath>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
     <OutDir>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\Lib</OutDir>
     <IntDir>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\Obj</IntDir>
+    <LibraryPath>$(ZStdHome)\dll;$(VC_LibraryPath_x86);$(WindowsSDK_LibraryPath_x86);$(NETFXKitsDir)Lib\um\x86</LibraryPath>
```
seladb: please fix the indentation

mk/vs/LightPcapNg.vcxproj.template
```diff
@@ -106,42 +106,54 @@
...
+      <PreprocessorDefinitions>PUT_USE_ZSTD_HERE;WIN32;_DEBUG;_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <AdditionalIncludeDirectories>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\include;PUT_ZSTD_INCLUDE_PATH_HERE;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
     </ClCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
     </Link>
+    <Lib>
+      <AdditionalDependencies>PUT_ZSTD_LIB_NAME_HERE;%(AdditionalDependencies)</AdditionalDependencies>
+	  <AdditionalOptions>/delayload:libzstd.dll %(AdditionalOptions)</AdditionalOptions>
+    </Lib>
```
seladb: why is this part necessary?

mk/vs/LightPcapNg.vcxproj.template
```diff
@@ -106,42 +106,54 @@
...
+      <PreprocessorDefinitions>PUT_USE_ZSTD_HERE;WINx64;_DEBUG;_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <AdditionalIncludeDirectories>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\include;PUT_ZSTD_INCLUDE_PATH_HERE;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
     </ClCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
     </Link>
+    <Lib>
+      <AdditionalDependencies>PUT_ZSTD_LIB_NAME_HERE;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalOptions>/delayload:libzstd.dll %(AdditionalOptions)</AdditionalOptions>
+    </Lib>
```
seladb: why is this part necessary?

mk/vs/LightPcapNg.vcxproj.template
```diff
@@ -150,14 +162,18 @@
...
     </ClCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
       <EnableCOMDATFolding>true</EnableCOMDATFolding>
       <OptimizeReferences>true</OptimizeReferences>
     </Link>
+    <Lib>
+      <AdditionalDependencies>PUT_ZSTD_LIB_NAME_HERE;%(AdditionalDependencies)</AdditionalDependencies>
+	  <AdditionalOptions>/delayload:libzstd.dll %(AdditionalOptions)</AdditionalOptions>
+    </Lib>
```
seladb: why is this part necessary?

mk/vs/LightPcapNg.vcxproj.template
```diff
@@ -166,14 +182,18 @@
...
```
seladb: why is this part necessary?

mk/vs/Pcap++Test.vcxproj.template
```diff
@@ -101,7 +101,8 @@
       <SubSystem>Console</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <AdditionalDependencies>%(AdditionalDependencies)</AdditionalDependencies>
-      <AdditionalLibraryDirectories>$(PcapPlusPlusHome)\Dist;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <AdditionalLibraryDirectories>$(PcapPlusPlusHome)\Dist\$(Platform)\$(Configuration);%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <DelayLoadDLLs>libzstd.dll</DelayLoadDLLs>
```
seladb: what is `DelayLoadDLLs`?

configure-windows-visual-studio.bat
Dysl3xik: Yes, fixed

mk/vs/PcapPlusPlusPropertySheet.props.template
Dysl3xik: The never ending programmer argument rears its ugly head again 👿 . Fixed

Tests/Pcap++Test/main.cpp
Dysl3xik: Since this is optional to build I think it may be wise to keep the test isolated? Otherwise you will have to interweave some pragmas into the existing test. I could still edit the new test to run dual writers and compare results.

Let me know what you think.

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: The delay load sets up the exe so it does not look for the DLL on startup, only if a function which needs the dll actually gets invoked. This means if someone builds  pcap++ with Zstd support, but does not actually utilize it, they don't have to ship the DLL. Otherwise when the application starts up it will throw an error saying can't find zstd dll and exit, even if zstd is not actually used. 

If zstd is totally missing in the build it will leave a little warning message noting that no symbols are used from that dll, but does not hurt anything.

mk/vs/Pcap++Test.vcxproj.template
Dysl3xik: The delay load sets up the exe so it does not look for the DLL on startup, only if a function which needs the dll actually gets invoked. This means if someone builds  pcap++ with Zstd support, but does not actually utilize it, they don't have to ship the DLL. Otherwise when the application starts up it will throw an error saying can't find zstd dll and exit, even if zstd is not actually used. 

If zstd is totally missing in the build it will leave a little warning message noting that no symbols are used from that dll, but does not hurt anything.

Dysl3xik:
I agree with your two points:

In order to fully support zstd we need to add it to the other configuration files and makefiles. It doesn't make a lot of sense if it's supported in VS only
**I am not a linux developer so although I agree with your point I would not be best suited to do this. I think what I have provided is sufficient for someone familiar with the other build processes to implement this in those other builds**

We need to add zstd scenarios to the CI
**Again I am not familiar with setting up these automated builds and checks, so I am not the best candidate for this**

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: The delay load sets up the exe so it does not look for the DLL on startup, only if a function which needs the dll actually gets invoked. This means if someone builds  pcap++ with Zstd support, but does not actually utilize it, they don't have to ship the DLL. Otherwise when the application starts up it will throw an error saying can't find zstd dll and exit, even if zstd is not actually used. 

If zstd is totally missing in the build it will leave a little warning message noting that no symbols are used from that dll, but does not hurt anything.

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: This is telling VS where to find the lib files for zstd for linking.

Tests/Pcap++Test/main.cpp
Dysl3xik: Same as above, if you want everything integrated I can do it, it will just make the code a bit messy due to disabled sections woven into the existing test code (if user does not build with zstd)

mk/vs/GitInfoPropertySheet.props
Dysl3xik: done

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: Done

configure-windows-visual-studio.bat
Dysl3xik: When I originally skipped this I didn't realize you were using it for your automated builds, so I agree its necessary. I will do it.

configure-windows-visual-studio.bat
Dysl3xik: Done

Tests/Pcap++Test/main.cpp
Dysl3xik: I just realized what I said here is not entirely true, since it falls back to NON compression mode when zstd isn't built you don't need jumper out any code. 

Anyways if you still feel strongly about re-working this I can do that. Just let me know.

configure-windows-visual-studio.bat
Dysl3xik: done

Tests/Pcap++Test/main.cpp
seladb: Yes, I think it'd be better to integrate this into the existing tests. If zstd is not enabled it will fall back to NON compression

mk/vs/PcapPlusPlusPropertySheet.props.template
seladb: Hehe 😄 Even the library itself is not fully consistent. In some places I have tabs and in others I have spaces... but I try to keep each file consistent

mk/vs/LightPcapNg.vcxproj.template
seladb: For the other dependencies (like pthreads, WinPcap) I used the property sheet file: https://github.com/seladb/PcapPlusPlus/blob/master/mk/vs/PcapPlusPlusPropertySheet.props.template

I think we should follow the same convention for zstd

mk/vs/LightPcapNg.vcxproj.template
seladb: can we also move this to the property sheet file https://github.com/seladb/PcapPlusPlus/blob/master/mk/vs/PcapPlusPlusPropertySheet.props.template ?

seladb:
> **In order to fully support zstd we need to add it to the other configuration files and makefiles. It doesn't make a lot of sense if it's supported in VS only**
I am not a linux developer so although I agree with your point I would not be best suited to do this. I think what I have provided is sufficient for someone familiar with the other build processes to implement this in those other builds

I can work on the Linux/MacOS/FreeBSD part. Can you apply the same changes for the MinGW/MinGW-w64 config file (https://github.com/seladb/PcapPlusPlus/blob/master/configure-windows-mingw.bat)?

Since I'm not really familiar with zstd, can you please provide some instructions on where to download it from and how to install on Windows?

> **We need to add zstd scenarios to the CI**
> Again I am not familiar with setting up these automated builds and checks, so I am not the best candidate for this

I can add this to the CI

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: Done

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: Moved

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: moved

mk/vs/Pcap++Test.vcxproj.template
Dysl3xik: moved

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: moved

Tests/Pcap++Test/main.cpp
Dysl3xik: Unified

Dysl3xik:
I will look over the MinGW stuff and see if I can work some magic there. I have not worked with that compiler in the past, but hopefully I can figure it out using the existing files.

Since this other work is beyond the scope of this existing PR, if you are happy with the test changes and the work done for the VS projects can we merge/close this PR and start a new branch/PR for the other work?

I can post a new issue and put instructions for getting ZStd in there and use that as a springboard for this new branch, assuming you are happy with the final work here.

P.S. TestMain had mixed tabs and spaces (not from me haha), I tabified the file as a single commit. So if you want to sensibly review the changes to the tests go back one commit and review there!

Tests/Pcap++Test/main.cpp
Dysl3xik: merged

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: moved

seladb:
Can you undo commit https://github.com/seladb/PcapPlusPlus/pull/406/commits/d01eaa832c2951f77e682fe50d6db58467a2dcd0? It makes it very difficult to review the changes

Dysl3xik:
I rolled back. Please fix tabs/spaces after merge.

mk/vs/PcapPlusPlusPropertySheet.props.template
seladb: Can you please fix indentation?

configure-windows-visual-studio.bat
seladb: In second thought, maybe we should not add zstd to the step-by-step configuration because most user will probably not use it. I suggest adding it only as a command-line option. Please let me know what you think

mk/vs/LightPcapNg.vcxproj.template
```diff
@@ -166,14 +179,17 @@
       <Optimization>Full</Optimization>
       <FunctionLevelLinking>true</FunctionLevelLinking>
       <IntrinsicFunctions>true</IntrinsicFunctions>
-      <PreprocessorDefinitions>WINx64;NDEBUG;_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <AdditionalIncludeDirectories>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <PreprocessorDefinitions>PUT_USE_ZSTD_HERE;WINx64;NDEBUG;_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <AdditionalIncludeDirectories>$(PcapPlusPlusHome)\3rdParty\LightPcapNg\LightPcapNg\include;PUT_ZSTD_INCLUDE_PATH_HERE;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
seladb: why do we need those here? Can't we use the definitions in `PcapPlusPlusPropertySheet.props`?

mk/vs/LightPcapNg.vcxproj.template
```diff
@@ -166,14 +179,17 @@
...
     </ClCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
       <EnableCOMDATFolding>true</EnableCOMDATFolding>
       <OptimizeReferences>true</OptimizeReferences>
     </Link>
+    <Lib>
+      <AdditionalDependencies>PUT_ZSTD_LIB_NAME_HERE;%(AdditionalDependencies)</AdditionalDependencies>
+    </Lib>
```
seladb: why do we need those here? Can't we use the definitions in `PcapPlusPlusPropertySheet.props`?

configure-windows-visual-studio.bat
Dysl3xik: I think its presumptuous to assume most people won't use it.

1. I think this is (will be) a unique + powerful feature of this library and you should advertise it as such.
2. On average I am seeing about 4x space savings I have 140MB pcap files compressed down to 35MB! This is a HUGE savings and comes at nearly no overhead as it is done live during capture not post processing in bulk which is slow. I chose ZStd because it is well suited for streaming compression.
3. Integration is VERY easy they don't have to build ZStd. There are pre-built binaries just download them and point the parameter at the correct path.
4. The compression is over the whole file NOT per packet. Therefore you do NOT need to use pcap++ do decode it. Any ZStd compatible decompressor can dump out the original pcapng file. So users are not losing compatibility with any pcapng reader. 

Overall I think if you unveil this in an official release and advertise it well in your capabilities I think a lot of people will consider trying it and will be pleasantly surprised at the huge space savings.

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: You can, I just defines it for all the projects then instead of only this one.

Anyways I did it.

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: This is ONLY a dependency for lightpcapng library not any other library.

I don't think this should be moved to the prop sheets as it will then show up as a dependency for all the projects in the solution.

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: Well after making my comment below, please re-affirm this is what you want. Defining this in every project is less of a big deal than marking ZStd as a dependency for every project.

I have the code changes ready I just have not pushed them yet.

mk/vs/PcapPlusPlusPropertySheet.props.template
Dysl3xik: Oops I think I fixed this last time NOT in the template file. I think I got it this time.

configure-windows-visual-studio.bat
seladb: I agree with you that it's a very powerful feature that people might find useful. Thank you for the efforts you're putting into this and for your great contribution to this project.

However based on my experience, different users use this library for different purposes. Not all of them use pcapng, and in fact many of them don't read or write to files at all.

One thing that is very important to me as the library maintainer (and as a user of other open source libraries) is the simplicity and ease of use of the library. That includes installation, configuration, straight-forward design, and more. To put it in simple words, I'm looking to optimize the time it takes for an average user to have a simple code example up and running, assuming he/she starts from scratch, on any platform.

That's why I want to have the simplest installation and configuration process with minimum dependencies. Zstd is definitely a feature for more advanced users and that's why my suggestion would be to put it in the command-line configuration only.

 BTW, I've already done so for many advance features in Linux and MacOS, for example: `--use-immediate-mode`, `--set-direction-enabled`, `--install-dir`, `--libpcap-include-dir`, `--libpcap-lib-dir`.

mk/vs/LightPcapNg.vcxproj.template
seladb: Got it, that makes sense. for `PUT_USE_ZSTD_HERE` I agree, but for `PUT_ZSTD_INCLUDE_PATH_HERE` I'm not sure we need that, we might put `$(ZStdHome)\include` instead, same as you did in `<LibraryPath>$(ZStdHome)\dll;...`

mk/vs/LightPcapNg.vcxproj.template
seladb: please see my comment above

configure-windows-visual-studio.bat
Dysl3xik: I totally understand your opinion and I agree this is a more advanced feature, and it is only applicable for people writing pcapng files.  However, IMO I don't think it overly onerous to put NO when prompted. I can also add notes in the prompt that indicate it will only benefit pcapng files in addition to the existing notes that it is optional.

Anyways at the end of the day the decision is in your hands as you are the chief maintainer/owner. I will make the change if requested, I was just sharing my opinion as originally asked.

mk/vs/LightPcapNg.vcxproj.template
Dysl3xik: I didn't even notice you had highlighted the include part part as it was off the side of the window.  Ill see about cleaning that up on Monday.

seladb:
I've rebased your PR with the recent changes from mainstream `dev`.

I'm trying to make it work on my Linux machine but I can find how to download/install pre-compiled version of Zstd. Can you please help with that?

Dysl3xik:
Bottom of the page here: https://github.com/facebook/zstd/releases/tag/v1.4.4

seladb:
@Dysl3xik I've completed the work for this. I've add Linux and MacOS support for Zstd and also added Linux, MacOS and Visual Studio to PcapPlusPlus CI.
I couldn't get MinGW to work, probably because Zstd doesn't support this toolchain. Also we can leave FreeBSD for later if someone asks for it.

I think the PR is ready to be merged.

One last thing we need to do is to add documentation to [PcapPlusPlus web-site](https://pcapplusplus.github.io/). Can you help with this?
You can either add it here on the PR, or fork https://github.com/PcapPlusPlus/pcapplusplus.github.io and add a markdown page under [here](https://github.com/PcapPlusPlus/pcapplusplus.github.io/tree/master/docs)

Dysl3xik:
Now I get to tell you thanks for all the hard work!!

I have been a little inactive these days due to abnormal work conditions and work overload with the virus and all, so sorry for not getting MinGW stuff going myself earlier.

Anyways I will make a markdown page for the CODE related stuff like where to get the lib, the #define which must be added, the function calls to use it, and other things I can think of to be useful. I can complete this Monday for sure.

For the website changes do you have something in particular in mind for this change?
I can assume we should modify:
Features -> Read and write packets from/to files
Build From Source on Windows - Visual Studio
Homepage -> Read and write packets to files : is it OK to add a brief sentence here about streaming compression?

I will do all this in a new fork & PR

seladb:
thanks @Dysl3xik for the kind words, I really appreciate your continuous efforts on this!

As for the documentation - in second thought maybe we can add this to the [Feature Overview](https://pcapplusplus.github.io/docs/features) page. This will make it easier for you. There you can describe the feature and what are the APIs to use it.

I can take care of the rest of the changes you mentioned.  I want to make it as easy as possible for you to contribute this documentation and avoid the boilerplate of building the web-site and so on. 

Please let me know if that would work for you

seladb:
@Dysl3xik I've merged the PR to `dev`. I'll merge it to `master` after we finish the documentation changes

seladb:
@Dysl3xik can you please create the documentation for Zstd so I can merge this to `master`?

Dysl3xik:
PR issued, sorry for the delay!

PR #411, opened 2020-03-11
Add RawPacket timestamp public setters

v0l0dia:
This PR allows to change RawPacket timestamp value after object creation. It maybe very helpful in case of imprecise timestamps, gathered from underlying capture engines.

Two setter functions were added: (legacy) timeval setter, and timespec setter.
Both setters are virtual, allowing some special processing in RawPacket subclasses.

Actually as the change is very small and straightforward, I don't know if the test-related commit is really needed (maybe cherry-pick RawPacket-relater only).

### DIFF
--- a/Packet++/header/RawPacket.h
+++ b/Packet++/header/RawPacket.h
@@ -344,6 +344,20 @@ namespace pcpp
 		timespec getPacketTimeStamp() const { return m_TimeStamp; }
 
 		/**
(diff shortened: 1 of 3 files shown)


v0l0dia:
Close my PR due to possible break of backward compatibility (adding virtual functions to base class changes vtable).
Maybe just make those setters non-virtual.

v0l0dia:
Reopened

PR #416, opened 2020-04-02
add http dst port filter

makefriend8:
Http port can be specified in HttpAnalyzer

### DIFF
--- a/Examples/HttpAnalyzer/main.cpp
+++ b/Examples/HttpAnalyzer/main.cpp
@@ -31,6 +31,8 @@
 #include "SystemUtils.h"
 #include "PcapPlusPlusVersion.h"
 #include <getopt.h>
+#include <iostream>
+#include <sstream>
 
 #define EXIT_WITH_ERROR(reason, ...) do { \
 	printf("\nError: " reason "\n\n", ## __VA_ARGS__); \
@@ -70,7 +72,6 @@ static struct option HttpAnalyzerOptions[] =
 };
 
 
-
 struct HttpPacketArrivedData
 {
 	HttpStatsCollector* statsCollector;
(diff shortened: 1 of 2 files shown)


Examples/HttpAnalyzer/main.cpp
```diff
@@ -434,10 +444,18 @@ void analyzeHttpFromLiveTraffic(PcapLiveDevice* dev, bool printRatesPeriodicaly,
 	if (!dev->open())
 		EXIT_WITH_ERROR("Could not open the device");
 
-	// set a port 80 filter on the live device to capture only HTTP packets
-	PortFilter httpPortFilter(80, SRC_OR_DST);
-	if (!dev->setFilter(httpPortFilter))
-		EXIT_WITH_ERROR("Could not set up filter on device");
+	if (GlobalConfig::getInstance().setDstFilterPort)
```
seladb, line 447: instead of add a `GlobalConfig` object, why can't we simply add the port as additional parameter to `analyzeHttpFromLiveTraffic()`?

Examples/HttpAnalyzer/main.cpp
```diff
@@ -398,9 +399,18 @@ void analyzeHttpFromPcapFile(std::string pcapFileName)
 		EXIT_WITH_ERROR("Could not open input pcap file");
 
 	// set a port 80 filter on the reader device to process only HTTP packets
-	PortFilter httpPortFilter(80, SRC_OR_DST);
-	if (!reader->setFilter(httpPortFilter))
-		EXIT_WITH_ERROR("Could not set up filter on file");
+	if (GlobalConfig::getInstance().setDstFilterPort)
```
seladb, line 402: instead of add a `GlobalConfig` object, why can't we simply add the port as additional parameter to `analyzeHttpFromPcapFile()`?

This parameter can have a default value of `80`

Examples/HttpAnalyzer/main.cpp
```diff
@@ -94,7 +95,7 @@ void printUsage()
 			"-------------------------\n"
 			"%s [-hvld] [-o output_file] [-r calc_period] -i interface\n"
 			"\nOptions:\n\n"
-			"    -i interface   : Use the specified interface. Can be interface name (e.g eth0) or interface IPv4 address\n"
+			"    -i interface   : Use the specified interface. Can be interface name (e.g eth0) or interface IPv4 address with port (e.g 192.168.1.1 or 192.168.1.1:80)\n"
```
seladb, line 98: adding `:<port_number>` to the IP address can be less intuitive to many users. Why not add an additional command line parameter (`-p` maybe?) that will be optional and will contain the port number?

This parameter can have a default value of `80`

makefriend8: how about sp  and dp? (src port number and dest port number)?
May I add a function about "http wrtier filter"  ?   or  create another example is better?

seladb: I'd strongly recommend to keep with only dst port. 

You can add an additional command-line parameter for filtering the traffic, same as the `-i` parameter in `PcapPrinter`:

https://github.com/seladb/PcapPlusPlus/tree/master/Examples/PcapPrinter

PR #422, opened 2020-04-22
Initial fuzzing support

zlowram:
I've been auditing this project and, as part of the process, I wrote a fuzzer using libFuzzer for it. It works and it has already discovered some bugs that I'll be reporting to you as soon as I send this PR.

The idea behind this PR is to provide a fuzzer for this project to allow continuous fuzzing in order to improve its security. I will also apply to [Google's OSS-Fuzz](https://github.com/google/oss-fuzz/) initiative. If the project gets accepted it will be fuzzed 24/7 in their cluster and the bugs will be automatically reported to you.

I tried to match your style for builds as much as I could, but feel free to adapt it as you wish.

Also, I would highly recommend to integrate regression tests with the samples that trigger bugs in your testing pipelines to avoid uncovering previously fixed bugs. You can do this by doing the linux build with the "--asan" option and calling the fuzzer built (located at Test/Fuzzers/Bin/fuzz_target, once built) passing as argument one of the samples at a time.

Thanks!


seladb:
thanks @zlowram for taking the initiative to integrate PcapPlusPlus into [OSS-Fuzz](https://github.com/google/oss-fuzz/).

I think it's a very important initiative that will definitely improve the security, stability and robustness of this project. Before merging the PR I have a few topics to discuss with you:

- When running Fuzzer, what is the corpus you used?
- In the PR you didn't include running Fuzzer as part of the CI. Should we do that? If yes, how do file new bugs that will be discovered during CI run?
- The fuzzing code includes only simple IPv4 checks.
  - Do you think we should add more protocol checks? Maybe the basic ones first (Eth, IPv6, TCP, UDP, DNS, HTTP, SSL)?
  - Do you think we should also include code that manipulate packets?
- As you can imagine no one has ever run fuzzer in PcapPlusPlus and I know that some parts in the code (especially the older code) won't handle edge cases very well, so I'm expecting a large number of bugs, at least in the beginning. I'm not sure I'll have the capacity to handle all of them. Will you be able to assist in resolving those bugs? I will obviously help and guide you through the process

zlowram:
Replying inline:

> thanks @zlowram for taking the initiative to integrate PcapPlusPlus into [OSS-Fuzz](https://github.com/google/oss-fuzz/).
> 
> I think it's a very important initiative that will definitely improve the security, stability and robustness of this project. Before merging the PR I have a few topics to discuss with you:
> 
> * When running Fuzzer, what is the corpus you used?

I used the corpus from the libpcap fuzzer.

> * In the PR you didn't include running Fuzzer as part of the CI. Should we do that? If yes, how do file new bugs that will be discovered during CI run?

I wouldn't run it aiming to discover bugs as part of the CI. What I would include in the CI, though, is running the fuzzer with the ASAN build against samples that are known to trigger already fixed bugs to avoid regressions. E.g. when one of the reported bugs is fixed, the CI should run the fuzzer with the ASAN build of PcapPlusPlus passing that single sample as parameter (if there were multiple samples, one time for each of them).

> * The fuzzing code includes only simple IPv4 checks.
>   * Do you think we should add more protocol checks? Maybe the basic ones first (Eth, IPv6, TCP, UDP, DNS, HTTP, SSL)?
>   * Do you think we should also include code that manipulate packets?

Maybe I would just start with this simple one as it already handles parsing of any kind of packet (e.g. one of the bugs reported is DNS related). As time goes and bugs start appearing and being fixed, I would definitely add more checks to allow more coverage within the codebase.

> * As you can imagine no one has ever run fuzzer in PcapPlusPlus and I know that some parts in the code (especially the older code) won't handle edge cases very well, so I'm expecting a large number of bugs, at least in the beginning. I'm not sure I'll have the capacity to handle all of them. Will you be able to assist in resolving those bugs? I will obviously help and guide you through the process

Unfortunately I don't have the time to help you with fixing them. The OSSFuzz initiative will let you know about the bugs privately so they can be fixed prior to disclosure. The good part is that I think that thanks to oss-fuzz the project will have more visibility to atract other devs that can help.

zlowram:
Good news! [It has been accepted](https://github.com/google/oss-fuzz/pull/3700)

I will work on the integration there, and will try to help with the integration in the CI here. Do you want me to send the commits to this same PR?

zlowram:
I have created Regression tests with the samples for #423 and #425, added them to both travis-ci and cirrus-ci on a separate task.

Also, I've done a rebase from dev branch so the regression tests are green already and the commit history for the PR looks a little bit better.

Makefile
```diff
@@ -9,6 +9,7 @@ PACKETPP_HOME        := Packet++
 PCAPPP_HOME          := Pcap++
 PACKETPP_TEST        := Tests/Packet++Test
 PCAPPP_TEST          := Tests/Pcap++Test
+FUZZERS_HOME		 := Tests/Fuzzers
```
seladb, line 12: Please fix the indentation (few extra spaces before `:= Tests/Fuzzers`)

Tests/RegressionTests/run_tests.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
seladb, line 1: Since these tests depends on the Fuzzer, I'd suggest to move the `RegessionTests` directory under `Fuzzers`

configure-linux.sh
```diff
@@ -30,6 +30,10 @@ function HELP {
    echo ""
    echo "--set-direction-enabled  --Set direction for capturing incoming or outgoing packets (supported on libpcap>=0.9.1)"
    echo ""
+   echo "--asan                   --Build PcapPlusPlus with Address Sanitizer enabled"
+   echo ""
+   echo "--fuzzing                --Build PcapPlusPlus with Clang enabling Coverage for fuzzing and compile existing fuzzers"
```
seladb, line 35: I'm guessing many people don't know what is fuzzing. I'd suggest to add a link that gives more information.

Also, it'd be helpful to explain the difference between this flag and the `--asan` flag

configure-linux.sh
```diff
@@ -30,6 +30,10 @@ function HELP {
    echo ""
    echo "--set-direction-enabled  --Set direction for capturing incoming or outgoing packets (supported on libpcap>=0.9.1)"
    echo ""
+   echo "--asan                   --Build PcapPlusPlus with Address Sanitizer enabled"
```
seladb, line 33: it'd be helpful to explain that this switch is related to fuzzing and what's the difference between this flag and the `--fuzzing` flag

configure-linux.sh
```diff
@@ -30,6 +30,10 @@ function HELP {
...
+   echo ""
+   echo "--fuzzing                --Build PcapPlusPlus with Clang enabling Coverage for fuzzing and compile existing fuzzers"
+   echo ""
```
seladb, line 36: please add both `--fuzzing` and `--asan` to the `Basic usage...` section

.travis.yml
```diff
@@ -39,6 +39,9 @@ matrix:
       env:
```
seladb, line 39: is there a reason to run regression in both Cirrus CI and Travis CI?

I'm thinking that maybe running them in Cirrus CI should be enough. What do you think?

.cirrus.yml
```diff
@@ -27,6 +27,18 @@ task:
     - cd Tests/Packet++Test && Bin/Packet++Test
     - cd ../Pcap++Test && Bin/Pcap++Test -n
 
+task:
+  name: Regression Tests Linux
+  container:
+    image: zlowram/ubuntu1910-clang:latest
```
seladb, line 33: can you check if the regression can run on the `seladb/ubuntu1804` container?

I can add `clang` to this image if needed

zlowram: It shouldn't be a problem to run there if there's `clang`. 

Please, make sure to install the latest version of `clang` available in the repositories (it should be `clang-9`) and that the `clang` command is available. Installing `clang-9` on Ubuntu 18.04 does not create `clang`, but `clang-9`. A symbolic link should fix that.

.travis.yml
zlowram: That's my fault actually for not looking before acting. I went straight to add it to Travis CI and then I realized that you were using Cirrus CI. I will revert the changes for Travis CI then!

.cirrus.yml
zlowram: Let me know when you update the image and I'll change it. Or feel free to merge it and change it yourself if you want.

zlowram:
Please, do not merge yet. I'm fixing some stuff to the build system to make it easier and compatible with OSS-Fuzz.

.cirrus.yml
seladb: does it have to be `clang-9`? will it work if I install the default `clang` version that is installed with `apt-get install clang`?

zlowram:
I have slightly modified the build system to honor the `make` variables: `CC`, `CXX`, `CFLAGS`, and `CXXFLAGS`.

This is usually a good idea to allow modifying then in the environment variables, and [OSS-Fuzz requires it](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#Requirements).

zlowram:
I have added a configure flag for linux to allow linking against libpcap statically. Fuzz targets in OSSFuzz can't have any other dependencies than those installed in the base system, and I think it's also a nice feature to have.

.cirrus.yml
seladb: @zlowram before I make changes to the image please let me know if `clang-9` is specifically required or any `clang` version will do

mk/PcapPlusPlus.mk.linux
```diff
@@ -3,8 +3,18 @@
...
+
 # libs
-PCAPPP_LIBS += -lpcap -lpthread
+PCAPPP_LIBS += -lpthread
+
+ifndef BUILD_WITH_STATIC_LIBPCAP
+PCAPPP_LIBS += -lpcap
+else
+PCAPPP_LIBS += $(STATIC_LIBPCAP_LIB)
+endif
```
seladb, line 17: generally speaking, I try to avoid `ifdef` statements in `PcapPlusPlus.mk.*` files because they make the build system even more complicated. Instead those `ifdef` statements usually reside in the `configure-*` files. Since I'm not sure there is an easy way to replace `-lpcap` with the static libpcap lib without using these `ifdefs` my suggestion would be to remove the feature of building statically with libpcap.

Is there a way to avoid static linking in OSSFuzz?

Otherwise we can think of a way to create a separate `PcapPlusPlus.mk.linux.fuzzing` and maybe create a command line flag in `configure-linux.sh` that will require the user to provide static libpcap location only if he uses fuzzing

Tests/Fuzzers/RegressionTests/run_tests.sh
```diff
@@ -0,0 +1,20 @@
...
+for sample in $(ls ${SAMPLES}); do
+	echo -n "Running sample $sample..."
+	$BINARY $SAMPLES/$sample &> /dev/null && echo -e "${GREEN}[OK]${NC}" || { FAILED=True && echo -e "${RED}[FAIL]${NC}"; }
+done
+
+if [[ ! -z $FAILED ]]; then
+	ERR_CODE=1
+fi
+
+exit $ERR_CODE
```
seladb, line 20: I fix bug #424 also, can you add it to the regression tests also?

.cirrus.yml
zlowram: Oh sorry, I missed this one. In OSSFuzz they use the Clang version closest to trunk. Right now it's Clang10 so my recommendation would be to use the latest version available in the distro repositories.

mk/platform.mk.macosx
```diff
@@ -6,9 +6,13 @@ LIB_PREFIX := lib
 
 LIB_EXT := .a
 
+ifeq ($(origin CXX),default)
```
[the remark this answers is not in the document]
zlowram: `CC` and `CXX` variables have a default value, which is `gcc` and `g++` respectively. This check is necessary because we need to know if the value assigned to that is the default, or has been user provided by overriding the environment variables.

mk/PcapPlusPlus.mk.linux
zlowram: Sadly, there's no way to avoid static linking in OSSFuzz, as explained [here](https://github.com/google/oss-fuzz/issues/25#issuecomment-254320632).

I thought it would be nice to have it across all the library, tests, and examples, but if you prefer to keep it only for the fuzzers I'm fine with it too. I will try to do this today.

Tests/Fuzzers/RegressionTests/run_tests.sh
zlowram: Nice, ofc.

mk/PcapPlusPlus.mk.common
```diff
@@ -18,3 +18,12 @@ PCAPPP_BUILD_FLAGS :=
...
 	PCAPPP_BUILD_FLAGS += -DPCAPPP_CPP_FEATURE_DETECTION -std=c++11 
 endif
+
+ifndef CFLAGS
+CFLAGS := -O2 -g -Wall -c
+endif
+
+ifndef CXXFLAGS
+CXXFLAGS := -O2 -g -Wall -c
+endif
```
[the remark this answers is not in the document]
zlowram: I left the examples with their own build flags because I saw that they were slightly different (no optimization level instad of 2). If you'd like to have it there, no problem I will propagate it too.

Pcap++/Makefile
```diff
@@ -63,18 +63,21 @@ endif
...
+INCLUDES += -I"$(STATIC_LIBPCAP_INCLUDE)"
+endif
 
 ifdef HAS_PCAP_IMMEDIATE_MODE
 DEPS += -DHAS_PCAP_IMMEDIATE_MODE
 endif
 
 ifdef USE_DPDK
-FLAGS := -msse -msse2 -msse3 -mssse3
+CXXFLAGS := -msse -msse2 -msse3 -mssse3
```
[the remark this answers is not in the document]
zlowram: Yup, my bad. Good catch!

mk/PcapPlusPlus.mk.common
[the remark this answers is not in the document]
seladb: I don't think the optimization level is different on purpose 😃 

I think we can have the same optimization level across the lib and examples

mk/platform.mk.macosx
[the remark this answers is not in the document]
seladb: ok that makes sesne

Common++/Makefile
```diff
@@ -33,6 +33,10 @@ DEPS += -DGIT_COMMIT=\"$(GIT_CUR_COMMIT)\" -DGIT_BRANCH=\"$(GIT_CUR_BRANCH)\"
 
 INCLUDES := -I"./src" \
 			-I"./header"
+
+ifdef BUILD_WITH_STATIC_LIBPCAP
+INCLUDES += -I"$(STATIC_LIBPCAP_INCLUDE)"
+endif
```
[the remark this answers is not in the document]
seladb: if we plan to remove this feature this comment is irrelevant

.cirrus.yml
seladb: sure, I'll install `clang-9`. But please note the executable name will probably be `clang-9` instead of `clang`. Is this ok?

Common++/Makefile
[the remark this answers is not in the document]
zlowram: If we decide to keep it only for fuzzing then yes, only for Linux.

mk/PcapPlusPlus.mk.common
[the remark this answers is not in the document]
zlowram: Good, I will apply those changes to the examples too.

.cirrus.yml
zlowram: The value I'm giving to `CC`  and `CXX` for the fuzzing build when they're not overriden by the user is `clang` and `clang++`, so I'd would be good to have them.

A symbolic link for both `clang` and `clang++` would do the trick for this.

seladb: this is done. You can use the image `seladb/ubuntu1804`. It has `clang-9` and I added the following symbolic links:

- `clang-9` --> `clang`
- `clang++-9` --> `clang++`

mk/PcapPlusPlus.mk.linux
zlowram: I pushed a different version that removes the `ifdef` statements from `PcapPlusPlus.mk.*` platform specific files. I created alternate versions for each platform file such as `PcapPlusPlus.mk.linux.static_libpcap` with a check on the `configure-*.sh` files to write one or another to the final `PcapPlusPlus.mk` file.

With this I'm preserving the functionality of allowing to build everything with static libpcap while keeping the `PcapPlusPlus.mk.{linux,macosx,freebsd}` files clean of `ifdef` statements.

.cirrus.yml
zlowram: Awesome, I already changed it and [works like a charm](https://cirrus-ci.com/task/6696600401608704).

mk/PcapPlusPlus.mk.common
```diff
@@ -16,5 +16,14 @@ PCAPPP_POST_BUILD :=
...
+	PCAPPP_BUILD_FLAGS += -DPCAPPP_CPP_FEATURE_DETECTION -std=c++11
 endif
+
+ifndef CFLAGS
+CFLAGS := -O2 -g -Wall -c
+endif
+
+ifndef CXXFLAGS
+CXXFLAGS := -O2 -g -Wall -c
+endif
```
seladb, line 28: I just recalled that we already have `PCAPPP_BUILD_FLAGS`. Do we really need `CFLAGS` and `CXXFLAGS` as well?

mk/PcapPlusPlus.mk.linux
```diff
@@ -4,7 +4,7 @@
 PCAPPP_INCLUDES += -I/usr/include/netinet
 
 # libs
-PCAPPP_LIBS += -lpcap -lpthread
+PCAPPP_LIBS += -lpthread -lpcap
```
seladb, line 7: why changing the order?

mk/PcapPlusPlus.mk.linux
seladb: while that's a valid solution, it add some complexity to the build system because from now on each parameter/flag we add to `PcapPlusPlus.mk.*` we'll need to add it also in the corresponding `PcapPlusPlus.mk.*.static_libpcap`. I'm afraid this will be hard to maintain.

My suggestion would be to create a new `configure-fuzzing.sh` script which will incorporate everything that is needed to compile PcapPlusPlus with fuzzing on Linux and create separate `PcapPlusPlus.mk.fuzzing` that will contain the build flags etc.

This will decouple fuzzing from everything else in PcapPlusPlus and will be easier to maintain. Please let me know what you think

mk/PcapPlusPlus.mk.linux
zlowram: That was a mistake when manually reverting the changes. Will put it back as it was before.

mk/PcapPlusPlus.mk.common
zlowram: `CFLAGS` and `CXXFLAGS` are `make` variables and are commonly used. Also, OSSFuzz [requires them](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#Requirements) to be used.

If you prefer using only `PCAPPP_BUILD_FLAGS` internally in the `Makefiles` and such, I can assign the value of `CXXFLAGS` to `PCAPPP_BUILD_FLAGS` after their assignment here.

mk/PcapPlusPlus.mk.linux
zlowram: I'm good with that too if you don't want to keep the ability of linking PcapPlusPlus against static libpcap for all other builds.

Will do the decoupling today.

mk/PcapPlusPlus.mk.common
seladb: Yes, I think that would be a better option

configure-fuzzing.sh
```diff
@@ -0,0 +1,374 @@
+#!/bin/bash
```
seladb, line 1: I think the configuration file can be much simpler in this case. No need to have all of these command-line parameter etc.

It can be pretty basic and just add the necessary flags to compile PcapPlusPlus with fuzzing. What do you think?

Pcap++/Makefile
```diff
@@ -63,18 +63,21 @@ endif
...
+INCLUDES += -I"$(STATIC_LIBPCAP_INCLUDE)"
+endif
 
 ifdef HAS_PCAP_IMMEDIATE_MODE
 DEPS += -DHAS_PCAP_IMMEDIATE_MODE
 endif
 
 ifdef USE_DPDK
-FLAGS := -msse -msse2 -msse3 -mssse3
+CXXFLAGS += -msse -msse2 -msse3 -mssse3
```
seladb, line 75: I'm not sure why you changed `FLAGS` to `CXXFLAGS`

Makefile
```diff
@@ -33,7 +34,16 @@ UNAME := $(shell uname)
...
 .SILENT:
 
-all: libs
+ifndef COMPILE_FOR_FUZZING
+ALL_TARGET := examples
+else
+ALL_TARGET := fuzzers
+endif
+
+all: $(ALL_TARGET)
```
seladb, line 43: I don't think this is needed. We can keep it the way it was since you added `fuzzing:` below anyway

mk/PcapPlusPlus.mk.fuzzing
```diff
@@ -0,0 +1,12 @@
...
+# includes
+PCAPPP_INCLUDES += -I/usr/include/netinet
+
+# libs
+PCAPPP_LIBS += -lpthread
+
+# allow user to add custom LDFLAGS
+PCAPPP_BUILD_FLAGS += $(LDFLAGS)
+
+COMPILE_FOR_FUZZING := 1
```
seladb, line 12: we can probably remove this variable if we decide not to use it in the main Makefile

configure-fuzzing.sh
zlowram: I would keep all the flags that the other `configure-*.sh` files have, because this first fuzzer might not need it but in the future it might be good to implement new fuzzers to get more coverage on all the library code base.

Pcap++/Makefile
zlowram: `FLAGS` was only used in the call to the compiler below. I replaced it with `CXXFLAGS` but now it doesn't make sense anymore. Now it should be `PCAPPP_BUILD_FLAGS`. Fixing it.

Makefile
zlowram: I used this becasue with the fuzzing build flags the examples do not build correctly. I will leave it as it was and add a warning message on the `configure-fuzzing.sh` script to tell the user to run only `make fuzzers` and not `make all` or `make`.

configure-fuzzing.sh
seladb: I usually try to keep things simple in PcapPlusPlus. My suggestion would be to keep the minimal options required for fuzzing and add options later as needed

zlowram: As you wish then. I've stripped down the `configure-fuzzing.sh` file to have the minimal options. It should be ready for merge now.

PR #429, opened 2020-05-05
Fix fuzzing build

zlowram:
Hi!

I saw a bug on the fuzzing build system that wasn't allowing OSFuzz to build the fuzzers with the different fuzzing engines and sanitizers. This PR should fix it. Also, I saw a flag that is not being used anymore, so I removed it.

As soon as commit gets to master I'll push the other changes to the OSSFuzz PR.

Thanks!

### DIFF
--- a/mk/platform.mk.fuzzing
+++ b/mk/platform.mk.fuzzing
@@ -14,7 +14,9 @@ ifeq ($(origin CC),default)
 CC := clang
 endif
 
+ifndef CXXFLAGS
 CXXFLAGS := -g -O0 -fsanitize=address,fuzzer-no-link
+endif
 
 AR := ar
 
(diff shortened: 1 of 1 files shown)


seladb:
The PR is merged to `master`. Can you please guide me on how to see the status in OSSFuzz?

zlowram:
I saw that, thanks! By now the [PR](https://github.com/google/oss-fuzz/pull/3759) is being reviewed. Travis is green so I don't think it will take long until they merge it.

Once merged, their infrastructure will start building and runing the fuzzer. When new bugs are found, they will notify you via email at pcapplusplus@gmail.com (this is the contact email I saw in your Readme) with all the information needed to reproduce it (i.e. ASAN stack trace, sample that triggers the bug, etc.), very much like I did when I reported those 3 bugs.

I would also recommend being familiar with their [disclosure guidelines](https://google.github.io/oss-fuzz/getting-started/bug-disclosure-guidelines/).

PR #432, opened 2020-05-07
Regression Tests for each Sanitizer

zlowram:
Today I noticed that I forgot to add regression tests for all the different Sanitizers that are run in OSS-Fuzz (`address`, `memory` and `undefiend`), so I added the option of selecting the sanitizer on the `configure-fuzzing.sh` file and created two more regression tests for `memory` and `undefined.

The workflow after fixing a bug related to fuzzing is exactly the same, just put the sample that triggers the bug in the regression samples dir: `Tests/Fuzzers/RegressionTests/regression_samples/`

This PR does not affect the fuzzers already running on OSS-Fuzz, it is purely to add robustness to the regression tests.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -28,11 +28,23 @@ task:
     - cd ../Pcap++Test && Bin/Pcap++Test -n
 
 task:
-  name: Regression Tests Linux
+  name: Regression Tests Linux (Address Sanitizer)
   container:
     image: seladb/ubuntu1804:latest
   configure_script:
-    - ./configure-fuzzing.sh --default
+    - ./configure-fuzzing.sh --sanitizer address
+    - ldconfig
+  build_script:
+    - make fuzzers
+  test_script:
+    - cd Tests/Fuzzers/RegressionTests && ./run_tests.sh
+
+task:
+  name: Regression Tests Linux (Undefined Behavior Sanitizer)
+  container:
+    image: seladb/ubuntu1804:latest
+  configure_script:
+    - ./configure-fuzzing.sh --sanitizer undefined
     - ldconfig
   build_script:
     - make fuzzers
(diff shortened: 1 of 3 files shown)


zlowram:
Regression Tests with the Memory Sanitizer are failing because those samples trigger an `use-of-uninitialized-value` report on them. Once that is fixed they should be green again.

seladb:
As discussed in #433 let's disable MemorySanitizer for now

.cirrus.yml
```diff
@@ -28,11 +28,23 @@ task:
...
+    - make fuzzers
+  test_script:
+    - cd Tests/Fuzzers/RegressionTests && ./run_tests.sh
+
+task:
+  name: Regression Tests Linux (Undefined Behavior Sanitizer)
+  container:
+    image: seladb/ubuntu1804:latest
+  configure_script:
+    - ./configure-fuzzing.sh --sanitizer undefined
```
seladb, line 47: the `undefined` option also runs it with address sanitizer. Why do we need 2 tests with the same configuration

zlowram: What do you mean that it also runs it with ASan? The `--sanitizer undefined` option of the `configure-fuzzing.sh` script only builds PcapPlusPlus and the fuzzers with the Undefined Behavior Sanitizer, but not with Address Sanitizer. Thus the idea of having 2 separate tests (also OSSFuzz runs each sanitizer on a different fuzzer build).

seladb: oh you're right, my bad. What does `undefined` sanitizer mean btw?

zlowram: No problem! That's the name for the [Undefined Behavior Sanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)

seladb: got it thanks!

PR #435, opened 2020-05-14
feat(TcpReassembly): add `status` parameter for TcpReassembly

starliiit:
I defined 3 categories for tcp reassembly status:

1. Success: in-order or out-of-order packets with payload, or bare FIN/RST packet.
2. Ignore: ICMP packet, bare ACK packet, packet of closed flow, retransmission packet.
3. Error: non-ip-packet, non-tcp-packet and packet not matching to any flow.

I'm a little bit of confused about what to do with bare ACK packet though...In a way they are valid packets indicating tcp state transition, while in other way they have no contribution to actual stream data. Any better idea?

### DIFF
--- a/Packet++/header/TcpReassembly.h
+++ b/Packet++/header/TcpReassembly.h
@@ -350,17 +350,17 @@ public:
 	 * The most important method of this class which gets a packet from the user and processes it. If this packet opens a new connection, ends a connection or contains new data on an
 	 * existing connection, the relevant callback will be called (TcpReassembly#OnTcpMessageReady, TcpReassembly#OnTcpConnectionStart, TcpReassembly#OnTcpConnectionEnd)
 	 * @param[in] tcpData A reference to the packet to process
-	 * @param[in] status A mutable reference which returns the reassembly status
+	 * @return A enum of `TcpReassembly::ReassemblyStatus`, indicating status of TCP reassembly
 	 */
-	void reassemblePacket(Packet& tcpData, ReassemblyStatus &status);
+	ReassemblyStatus reassemblePacket(Packet& tcpData);
 
 	/**
 	 * The most important method of this class which gets a raw packet from the user and processes it. If this packet opens a new connection, ends a connection or contains new data on an
 	 * existing connection, the relevant callback will be invoked (TcpReassembly#OnTcpMessageReady, TcpReassembly#OnTcpConnectionStart, TcpReassembly#OnTcpConnectionEnd)
 	 * @param[in] tcpRawData A reference to the raw packet to process
-	 * @param[in] status A mutable reference which returns the reassembly status
+	 * @return A enum of `TcpReassembly::ReassemblyStatus`, indicating status of TCP reassembly
 	 */
-	void reassemblePacket(RawPacket* tcpRawData, ReassemblyStatus &status);
+	ReassemblyStatus reassemblePacket(RawPacket* tcpRawData);
 
 	/**
 	 * Close a connection manually. If the connection doesn't exist or already closed an error log is printed. This method will cause the TcpReassembly#OnTcpConnectionEnd to be invoked with
--- a/Packet++/header/TcpReassembly.h
+++ b/Packet++/header/TcpReassembly.h
@@ -240,6 +240,20 @@ public:
 		TcpReassemblyConnectionClosedManually
 	};
 
+	enum ReassemblyStatus 
+	{
+		Success_TcpMessageHandled,
+		Success_OutOfOrderTcpMessageBuffered,
+		Success_FIN_RSTWithNoData,
+		Ignore_IcmpPacket,
+		Ignore_PacketWithNoData,
+		Ignore_PacketOfClosedFlow,
+		Ignore_Retransimission,
+		Error_NonIpPacket,
+		Error_NonTcpPacket,
+		Error_PacketDoesNotMatchFlow,
+	};
+
 	/**
 	 * The type for storing the connection information
 	 */
@@ -300,6 +314,9 @@ public:
 	 * @param[in] tcpRawData A reference to the raw packet to process
 	 */
 	void reassemblePacket(RawPacket* tcpRawData);
+
+	void reassemblePacket(RawPacket* tcpRawData, ReassemblyStatus &status);
+	void reassemblePacket(Packet& tcpData, ReassemblyStatus &status);
 
 	/**
 	 * Close a connection manually. If the connection doesn't exist or already closed an error log is printed. This method will cause the TcpReassembly#OnTcpConnectionEnd to be invoked with
--- a/Packet++/header/TcpReassembly.h
+++ b/Packet++/header/TcpReassembly.h
@@ -242,15 +242,60 @@ public:
 
 	enum ReassemblyStatus 
 	{
-		Success_TcpMessageHandled,
-		Success_OutOfOrderTcpMessageBuffered,
-		Success_FIN_RSTWithNoData,
-		Ignore_IcmpPacket,
+		/** 
+		 * The processed packet contains valid TCP payload, and its payload is processed by `OnMessageReadyCallback` callback function.
+		 * The packet may be:
+		 * 1. An in-order TCP packet, meaning `packet_sequence == sequence_expected`. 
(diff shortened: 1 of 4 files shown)


Packet++/header/TcpReassembly.h
```diff
@@ -301,6 +315,9 @@ class TcpReassembly
 	 */
 	void reassemblePacket(RawPacket* tcpRawData);
 
+	void reassemblePacket(RawPacket* tcpRawData, ReassemblyStatus &status);
+	void reassemblePacket(Packet& tcpData, ReassemblyStatus &status);
```
seladb, line 319: I'm wondering whether we should provide a different API with the status or just add this to the current API.

On one hand if we add it to the current API it won't be backward compatible, on the other hand it's quite confusing to have so many methods doing the same thing.

I'm leaning towards adding the status to the existing API, please let me know what you think

Packet++/src/TcpReassembly.cpp
```diff
@@ -132,20 +132,28 @@ void TcpReassembly::reassemblePacket(Packet& tcpData)
 	else if (tcpData.isPacketOfType(IPv6))
 		ipLayer = (Layer*)tcpData.getLayerOfType<IPv6Layer>();
 
-	if (ipLayer == NULL)
+	if (ipLayer == NULL) 
+	{
+		status = Error_NonIpPacket;
```
seladb, line 137: I'm not sure we should call it an "error", it's just a simple case where the packet is not of type IPv4 or IPv6

Packet++/src/TcpReassembly.cpp
```diff
@@ -132,20 +132,28 @@ void TcpReassembly::reassemblePacket(Packet& tcpData)
...
 		return;
+	}
+
 
 	// Ignore non-TCP packets
 	TcpLayer* tcpLayer = tcpData.getLayerOfType<TcpLayer>(true); // lookup in reverse order
 	if (tcpLayer == NULL)
+	{
+		status = Error_NonTcpPacket;
```
seladb, line 146: same comment here - I don't think this is an "error"

Packet++/src/TcpReassembly.cpp
```diff
@@ -132,20 +132,28 @@ void TcpReassembly::reassemblePacket(Packet& tcpData)
...
 		return;
+	}
 
 	// Ignore the packet if it's an ICMP packet that has a TCP layer
 	// Several ICMP messages (like "destination unreachable") have TCP data as part of the ICMP message.
 	// This is not real TCP data and packet can be ignored
 	if (tcpData.isPacketOfType(ICMP))
 	{
 		LOG_DEBUG("Packet is of type ICMP so TCP data is probably  part of the ICMP message. Ignoring this packet");
+		status = Ignore_IcmpPacket;
```
seladb, line 156: I don't think we need a special enum for that. It's just another case of non-TCP packet

Packet++/header/TcpReassembly.h
```diff
@@ -240,6 +240,20 @@ class TcpReassembly
...
+		Success_OutOfOrderTcpMessageBuffered,
+		Success_FIN_RSTWithNoData,
+		Ignore_IcmpPacket,
+		Ignore_PacketWithNoData,
+		Ignore_PacketOfClosedFlow,
+		Ignore_Retransimission,
+		Error_NonIpPacket,
+		Error_NonTcpPacket,
+		Error_PacketDoesNotMatchFlow,
+	};
```
seladb, line 255: Can you please add doxygen documentation to the enum and enum values?

Packet++/src/TcpReassembly.cpp
```diff
@@ -312,6 +327,7 @@ void TcpReassembly::reassemblePacket(Packet& tcpData)
 		LOG_DEBUG("Got FIN or RST packet without data on side %d", sideIndex);
 
 		handleFinOrRst(tcpReassemblyData, sideIndex, flowKey);
+		status = Success_FIN_RSTWithNoData;
```
seladb, line 330: I'm not sure we need the word "Success" here. The status can simply be: `FIN_RSTWithNoData`

Packet++/src/TcpReassembly.cpp
```diff
@@ -352,11 +368,12 @@ void TcpReassembly::reassemblePacket(Packet& tcpData)
 			TcpStreamData streamData(tcpLayer->getLayerPayload(), tcpPayloadSize, tcpReassemblyData->connData);
 			m_OnMessageReadyCallback(sideIndex, streamData, m_UserCookie);
 		}
+		status = Success_TcpMessageHandled;
```
seladb, line 371: same here - I'm not sure the word "success" is needed

Packet++/src/TcpReassembly.cpp
```diff
@@ -113,7 +113,7 @@ TcpReassembly::~TcpReassembly()
 	}
 }
 
-void TcpReassembly::reassemblePacket(Packet& tcpData)
+void TcpReassembly::reassemblePacket(Packet& tcpData, ReassemblyStatus &status)
```
seladb, line 116: Is there an initial value we'd like to set?

Packet++/src/TcpReassembly.cpp
```diff
@@ -465,6 +501,7 @@ void TcpReassembly::reassemblePacket(Packet& tcpData)
 		tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.pushBack(newTcpFrag);
 
 		LOG_DEBUG("Found out-of-order packet and added a new TCP fragment with size %d to the out-of-order list of side %d", (int)tcpPayloadSize, sideIndex);
+		status = Success_OutOfOrderTcpMessageBuffered;
```
seladb, line 504: we probably don't need the word "success" here too

Packet++/header/TcpReassembly.h
starliiit: I agree with you. I removed the redundant method declarations in the new commit. I also touched some test and example code to make them compile.

Packet++/header/TcpReassembly.h
starliiit: I added some comments to the enum values, but I haven't used doxygen before and I didn't figure it out how to generate api-doc html on my machine. Hope it works...

Packet++/src/TcpReassembly.cpp
starliiit: It's a non-const reference so I don't think we can provide a default value...

Packet++/header/TcpReassembly.h
seladb: Thank you!

I have another idea that I'd like to share with you: what if instead of adding the status as a second parameter to `reassemblePacket()` we can just let it return this enum value?

That way the API change will be backward compatible and we still have this feature added.

Also, it'd be great to add a few test cases in `Pcap++Test` but I can do it for you after the PR is merged.

Packet++/header/TcpReassembly.h
seladb: Thanks for adding the comments. Generating the html code isn't necessary, there is a CI job that does that automatically:

https://github.com/seladb/PcapPlusPlus/blob/5ef0c6da8c34b5d4155f865e94bdf83cb18fcbad/.cirrus.yml#L68

Packet++/src/TcpReassembly.cpp
seladb: If we change the API as I suggested here: https://github.com/seladb/PcapPlusPlus/pull/435#discussion_r428446802 this won't be necessary

Packet++/header/TcpReassembly.h
starliiit: Way better than my approach. Implemented in [1953f80](https://github.com/seladb/PcapPlusPlus/pull/435/commits/1953f803ddadb935bdd248d92f1d2a67d117cea6)

seladb:
Thanks for making these changes, just one last comment: I noticed you forked your repo from `master` branch which means the PR will be merged into `master`. 

I usually ask contributors to fork from `dev` so the PR is merged into `dev` first. Can you please do that?

I'm really sorry for all the trouble

starliiit:
Oh, my bad... Let me see if I can change base

starliiit:
I'm closing this and making a new PR.

Issue #437, opened 2020-05-18
Provide pre-allocated and pre-filled buffer to Packet and assign PayloadLayer


maruu:
Hi,

it's me again with a new proposal :-)

I will try to give you a background of what I am trying to achieve and I have already thought about the steps that are required to implement this. I wanted to hear your thoughts on this - maybe I have overlooked something or this is not a direction that you want PcapPlusPlus to evolve to.

**I have the following use-case:**
I need to provide the user of my library with a data buffer in which he can write its payload, that I then want to encapsulate into an Ethernet frame and send out. The data provided by the user is solely payload data. At the time of the user's request, I do not know which transport or routing protocol I am going to use, so I cannot build the Ethernet frame beforehand and use the already implemented SetPayload function of a PayloadLayer that I could add to the packet. Also, this would involve copying the data provided by the user to the PayloadLayer. I do not want that, since the payload involves complex data structures that the user could already construct in the provided data buffer.

I have visualized this here:
![image](https://user-images.githubusercontent.com/2047922/82227484-1cd9d180-9928-11ea-83fa-6de9e2360ef7.png)

As you can see, the buffer reserves memory in front of the payload location so a later addition of transport and routing protocol layers is possible.

You may ask yourself: How do you know how large the buffer needs to be so you can avoid costly reallocations?
Well, I know this very well for my use-case. For all other, I would implement the same reallocation strategy used in the available insertLayer method. And also, since the initial size of the buffer is not something that PcapPlusPlus has to mandate, but is actually provided by the user of the library, you don't limit the usability of this feature.

**What I need:**
A way to pass a pre-allocated and pre-filled buffer to the Packet. The payload which I can identify by pointer and length then needs to be encapsulated in a PayloadLayer.

This would involve adding the following constructor to the Packet:
Packet(uint8_t* buffer, size_t dataLen, uint8_t* payload, size_t payloadLen)

The buffer can also already be passed as a RawPacket.

The packet would then proceed to assign the buffer as its own Packet buffer and create a PayloadLayer with the given data pointer and length.

You may ask yourself: How can you add a layer to this packet, if the PayloadLayer is not at the start of the packet (i.e. not at the start of the linked list)?

For this, I propose a method called addLayerBefore() which does the reverse operation of the addLayer() method. 
The API looks like this: addLayerBefore(Layer* addBeforeThis, Layer* addThis)

The final packet will look like this, having added all needed layers:
[...]

PR #438, opened 2020-05-20
437 provide pre allocated packet

maruu:
Hi,

I have implemented the proposed changes (see [here](https://github.com/seladb/PcapPlusPlus/issues/437)).

I have added the following concept: The RawPacket does not start at m_RawData but at m_StartOfBuffer. m_RawData is always pointing at the location where the raw data (e.g. protocol data, or frame) starts. The m_RawDataLen always denotes the length of the buffer, as there seems to already be a variable called m_FrameLength which denotes the length of the frame. I think that this could be a source of a problem if the two methods of adding layers (before and after the previous layers) are used together, since the addAfterPreviousLayer method (i.e. addLayer) uses getRawDataLen instead of getFrameLength to get the data length. I think this needs adaption if we continue with this way of implementing this feature (this is exactly why I am creating this PR, to get your thoughts on this..).

I am creating this pull request as a draft since I think that there are still some changes required. I am not fully satisfied with how this turned out and with the API changes I had to make to the RawPacket. 
Please let me know your thoughts on this. 

Br,
maruu


seladb:
Thanks for sending the PR. I'll review the request and the PR and will get back to you soon

seladb:
@maruu can we close this PR as well?

PR #450, opened 2020-06-08
change flow key to uint64 for tcp reassembly

### LINKED ISSUE #447: flowKey of TcpReassembly should be uint64
According to birthday paradox: [https://en.wikipedia.org/wiki/Birthday_problem](url)  
If we use uint32 for flowKey, when there is 10k tcp connections, it's 75% probability at least one hash collision.  
Now I develop a software will listen 100k tcp connections depend on pcapplusplus, uint32 flowKey  always trigger hash collision.

DinoStray:
resolve #447

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,7 @@ mk/vs2019/**
 #Tutorial outputs
 Examples/Tutorials/**/Makefile
 Examples/Tutorials/**/*.exe
+
+#self
+.idea/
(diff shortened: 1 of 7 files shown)


seladb:
As we discussed in the issue itself, the solution would probably be to handle collisions better rather than changing the flow key to `uint64_t`.

I'm closing the PR for now, let's discuss this more in the issue comments

PR #452, opened 2020-06-08
Small improvements in IcmpLayer and examples

gx740:
The same changes as in PR #382

### DIFF
--- a/Packet++/header/IcmpLayer.h
+++ b/Packet++/header/IcmpLayer.h
@@ -3,6 +3,7 @@
 
 #include "Layer.h"
 #include "IPv4Layer.h"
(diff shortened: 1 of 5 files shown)


Packet++/header/IcmpLayer.h
```diff
@@ -3,6 +3,7 @@
 
 #include "Layer.h"
 #include "IPv4Layer.h"
+#include "IpAddress.h"
```
seladb, line 6: is this include actually needed? 

`IPv4Layer.h` is already including `IpAddress.h`

gx740: An include has been removed

PR #460, opened 2020-06-09
Performance and readability improvements in TcpReassembly

gx740:
These changes were partially implemented in PR #395


gx740:
The main purpose is to improve readability. Currently, the `ipLayer` pointer is defined in one place but is used much lines below. In the new version the detection of source/destination IPs is performed in one place.

gx740:
The iterator `iter` is also defined and used in one place.

PR #461, opened 2020-06-09
Decreased memory consumption in TcpReassembly

gx740:
Current changes save 4 bytes per connection and give the way for next optimization (in a different PR).
Notice that the signature of callback function 'OnTcpMessageReady' has changed.


seladb:
There are conflicts between this PR and the previous one that got merged. Can you please resolve them?

gx740:
The conflict has been resolved

PR #463, opened 2020-06-15
Update DhcpLayer.cpp

6923403:
modify case sorting
seeing uncomfortable and tired


Packet++/src/DhcpLayer.cpp
```diff
@@ -125,51 +125,50 @@ std::string DhcpLayer::toString() const
...
+			break;
+		}
+		case DHCP_INFORM:
+		{
+			msgType = "Inform";
+			break;
+		}
+		default:
+			break;
+
```
seladb, line 170: The existing `switch` indentation is the same throughout the project, I prefer not to change it

PR #467, opened 2020-06-20
Fix IPcapDevice::matchPacketWithFilter when checking empty filter.

skorup101:
Bug:
bool IPcapDevice::matchPacketWithFilter(std::string filterAsString, RawPacket* rawPacket)
"pcap_compile_nopcap" would not be called untill first, non empty filterAsString.

### DIFF
--- a/Pcap++/src/PcapDevice.cpp
+++ b/Pcap++/src/PcapDevice.cpp
@@ -74,8 +74,7 @@ bool IPcapDevice::matchPacketWithFilter(std::string filterAsString, RawPacket* r
 {
 	static std::string curFilter = "";
 	static struct bpf_program prog;
-	static bool isProgCompiled = false;
-	if ( (curFilter != filterAsString) || !isProgCompiled )
+	if ( (curFilter != filterAsString) || (prog.bf_insns == nullptr) )
 	{
 		LOG_DEBUG("Compiling the filter '%s'", filterAsString.c_str());
 		pcap_freecode(&prog);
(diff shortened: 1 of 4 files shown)


Pcap++/src/PcapDevice.cpp
```diff
@@ -74,15 +74,16 @@ bool IPcapDevice::matchPacketWithFilter(std::string filterAsString, RawPacket* r
 {
 	static std::string curFilter = "";
 	static struct bpf_program prog;
-	if (curFilter != filterAsString)
+	static bool isProgCompiled = false;
+	if ( (curFilter != filterAsString) || !isProgCompiled )
```
seladb, line 78: Maybe instead of adding a static boolean flag we can give `curFilter` a different initial value, for example: `"--"` and then an empty value will be compiled. Please let me know what you think

skorup101: Hey, I propose another solution, keeping in mind your comment and also resolving another bug I've found. Here's the buggy scenario:
PROPER_FILTER = "";
matchPacketWithFilter(PROPER_FILTER , rawPacket); // filter gets compiled, packet gets checked against it and now curFilter == PROPER_FILTER. Function returns true - ok.
matchPacketWithFilter("some wrong filter", rawPacket); // prog gets freed and the new filter compiled but with error, so prog is freed and **curFilter remains as it was curFilter == PROPER_FILTER**. Function returns false - ok but the state of static objects is buggy.
matchPacketWithFilter(PROPER_FILTER , rawPacket); // because curFilter == PROPER_FILTER  prog will not be compiled **while it was freed before**. Function returns false - **not ok.**

seladb: That's indeed a bug, thanks for providing a solution to it!

PR #487, opened 2020-07-03
no packet CPU 100% problem  for function PfRingDevice::captureThreadMain

jos666:
Fix pf_ring multi-thread packet capture. When there is no packet, the CPU is fully loaded

### DIFF
--- a/Pcap++/src/PfRingDevice.cpp
+++ b/Pcap++/src/PfRingDevice.cpp
@@ -9,6 +9,7 @@
 #include "PlatformSpecificUtils.h"
 #include <errno.h>
 #include <pfring.h>
+#include <unistd.h>
 
 
 #define DEFAULT_PF_RING_SNAPLEN 1600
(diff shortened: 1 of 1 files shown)


seladb:
CPU is at 100% to achieve maximum throughput. The app should always look for enqueued packets so it can process them as fast as possible. This is not a bug, it's on purpose.

Let me close this PR now, we can discuss this more if you want.

jos666:
Hello, if you always see 100% performance of this program? This can't realize the alarm

DinoStray:
100% CPU usage is a little rough

seladb:
@jos666 @DinoStray 
Usually apps who are using PF_RING or DPDK (which bypass the kernel network processing) do that in order to achieve the maximum throughput possible. Otherwise they would just use libpcap which doesn't bypass any kernel processing.

So in these kind of apps CPU reaching to 100% is not usually an issue.
Having said that, if you think we should still add an option to lower the CPU in exchange of lower throughput we can do that. But let's keep the amount of `sleep()` time configurable, which means we should add an object similar to [`DeviceConfiguration`](https://github.com/seladb/PcapPlusPlus/blob/96fba43565e46496959d1b321163a1696eb49303/Pcap%2B%2B/header/PcapLiveDevice.h#L170) and pass it to `PfRingDevice`'s `open` methods.

russkel:
This surely can't be right. Wireshark reports nowhere near as high CPU usage as a very basic Pcap++ tool. My laptop fan starts spinning and everything.

PR #516, opened 2020-08-17
Enable HAS_SET_DIRECTION_ENABLED for regular build

fquinner:
It looks like the option to enable pcap direction doesn't get compiled
in during a manual build (unless I have missed something). The option
is a gmake variable and is indeed promoted in Tests/Pcap++Test/Makefile,
but never seemed to get promoted to a compiler macro argument in the
standard Makefile. This change should correct this.


seladb:
Yes, I missed it somehow during the code review for the original PR. Thanks for the fix!

PR #523, opened 2020-09-06
Add missing bytes field to TcpStreamData

### LINKED ISSUE #522: Missing bytes field in TcpStreamData
Per discussion on support mail, adding this issue to propose adding a `bytesMissing` field to `TcpStreamData` as obtaining this currently requires parsing reassembly callback data.

cryptocode:
Closes #522 to add missing byte count due to packet loss, as this currently needs to be parsed out. I've kept the original data string to avoid breaking existing API consumers.

I'm not very familiar with PcapPlusPlus internals/dev process, so this is just the most minimal implementation I could think of. I did a quick test by modifying the reassembly example, and it reports the same missing byte count as the existing string.

### DIFF
--- a/Packet++/header/TcpReassembly.h
+++ b/Packet++/header/TcpReassembly.h
@@ -126,10 +126,11 @@ public:
 	 * A c'tor for this class that get data from outside and set the internal members
 	 * @param[in] tcpData A pointer to buffer containing the TCP data piece
 	 * @param[in] tcpDataLength The length of the buffer
+	 * @param[in] missingPacketBytes The number of missing bytes due to packet loss.
 	 * @param[in] connData TCP connection information for this TCP data
 	 */
-	TcpStreamData(const uint8_t* tcpData, size_t tcpDataLength, const ConnectionData& connData)
-		: m_Data(tcpData), m_DataLen(tcpDataLength), m_Connection(connData)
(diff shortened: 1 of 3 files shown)


seladb (commented):
Thank for adding this functionality to PcapPlusPlus!

Few quick comments:
- Maybe we should consider adding a getter method `isBytesMissing()` to indicate whether there are missing bytes or not (which is equivalent to `getMissingByteCount() == 0`)
- Can you please add this to [TCP reassembly tests](https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Pcap%2B%2BTest/Tests/TcpReassemblyTests.cpp)?
- Please create your fork from `dev` branch so the PR will be merged to `dev`. I usually don't merge PRs directly to `master`

cryptocode:
Ah yeah, I'll rebase to dev. The repos I've used in the past with develop branches have it as the default branch on GitHub.

I'll try to find time to address your points, appreciate the feedback.

cryptocode:
By the way, I got a segfault on master when running `sudo Bin/Pcap++Test -n` on macOS, not sure if this is a known issue?

```
...
TestDnsParsing                : PASSED
[1]    61497 segmentation fault  Bin/Pcap++Test -n
```

cryptocode:
@seladb there was already `TestTcpReassemblyMissingData` which I updated to also check against expected loss, which means `getMissingByteCount` gets covered by tests.

PR #526, opened 2020-09-12
Dev

Minilulatsch:
I fixed several bugs caused by missing size checks, which were found during fuzzing. To verify these bugs check the following example files, which will trigger the different bugs and crash the PcapPrinter example: [crash_files.zip](https://github.com/seladb/PcapPlusPlus/files/5212651/crash_files.zip)


seladb:
Thanks for spending time on finding and fixing those issues! 

Just curious - did you find them in a fuzzing project you were running on PcapPlusPlus or is it somehow related to [Google's OSS Fuzz project](https://google.github.io/oss-fuzz/) that PcapPlusPlus is part of?

Minilulatsch:
The fuzzing project during which I found those bugs are actually not related to OSS Fuzz

Minilulatsch:
Fixed an OOM bug caused by an infinite allocation loop. See the following crash file for verification: 
[ssl_handshake_oom.zip](https://github.com/seladb/PcapPlusPlus/files/5215153/ssl_handshake_oom.zip)

Minilulatsch:
Fixed another OOM bug. The problem here is, that a malformed packet may cause an infinite loop, in which next layers of a packet are parsed. Note that the fix is rather ugly, as it simply limits the number of next layers for a single packet. But the fix should highlight the problem, which might help finding a better fix. See the following crash file for verification of the bug: 
[packet_parse_layer_oom.zip](https://github.com/seladb/PcapPlusPlus/files/5215250/packet_parse_layer_oom.zip)

Packet++/src/Packet.cpp
```diff
@@ -57,14 +57,17 @@ void Packet::setRawPacket(RawPacket* rawPacket, bool freeRawPacket, ProtocolType
...
+	while (curLayer != NULL && (curLayer->getProtocol() & parseUntil) == 0 && curLayer->getOsiModelLayer() <= parseUntilLayer && counter < 64)
 	{
 		m_ProtocolTypes |= curLayer->getProtocol();
 		curLayer->parseNextLayer();
 		curLayer->m_IsAllocatedInPacket = true;
 		curLayer = curLayer->getNextLayer();
+		
 		if (curLayer != NULL)
 			m_LastLayer = curLayer;
+		counter++;
```
seladb, line 70: I didn't have time to get into the packet example you provided. Can you explain what causes the infinite loop?

If it's a specific protocol maybe we can fix there...

Packet++/src/SipLayer.cpp
```diff
@@ -307,8 +307,14 @@ void SipRequestFirstLine::parseVersion()
 	verPos++;
 
 	int endOfVerPos = 0;
-	while (((verPos+endOfVerPos)[0] != '\r') && ((verPos+endOfVerPos)[0] != '\n'))
+	while (((verPos+endOfVerPos)[0] != '\r') && ((verPos+endOfVerPos)[0] != '\n')) {
+		if ((verPos + endOfVerPos) >= (char *) (m_SipRequest->m_Data + m_SipRequest->m_DataLen)) {
```
seladb, line 311: maybe this condition can be added to the while loop?

also, can you please fix the indentation and make sure the opening bracket `{` is in a separate line?

seladb (commented):
it's very exciting to know that there is another fuzzing project testing PcapPlusPlus! 

Is it a private fuzzing project or can you expose it?

Anyway, currently the OSS Fuzz regression tests are [here](https://github.com/seladb/PcapPlusPlus/tree/f0f5a613411798d75804ae7815c847a405c6f460/Tests/Fuzzers/RegressionTests) and run as part of [PcapPlusPlus CI](https://github.com/seladb/PcapPlusPlus/blob/f0f5a613411798d75804ae7815c847a405c6f460/.cirrus.yml#L34-L56). Maybe it'd be a good idea to add the packet samples you provided to these regression tests, what do you think?

Minilulatsch:
These fuzzing results are more a less a side effect of some development/research on fuzzing, so it's more a sporadic than a continuous fuzzing project.
I will recheck the commit and include your recommendation asap.

seladb:
> 
> 
> These fuzzing results are more a less a side effect of some development/research on fuzzing, so it's more a sporadic than a continuous fuzzing project.
> I will recheck the commit and include your recommendation asap.

@Minilulatsch do you know when you will have time to address these comments?

seladb:
@Minilulatsch any update on this?

Minilulatsch:
Sorry for the delay, but I just came back to these problems...I will post my updates as answers to your questions above

Packet++/src/Packet.cpp
Minilulatsch: The underlying problem seems to be here: https://github.com/seladb/PcapPlusPlus/blob/f30de1387acc1f823e0d4779e5565f4cc32dd4a6/Packet%2B%2B/src/IPv4Layer.cpp#L240

For the provided packet example, `hdrLen` seems to be 0, so that the payload of the new packet, is at the same position as the payload of the current packet, when calculating the position of the payload here: https://github.com/seladb/PcapPlusPlus/blob/f0f5a613411798d75804ae7815c847a405c6f460/Packet%2B%2B/src/IPv4Layer.cpp#L251

I guess that means, that the exact same data is parsed all the time when `hdrLen` is 0, which results in an infinite loop. I reverted the bug fix from above and commited a new one, which checks whether `hdrLen` is not 0. I hope that fix is okay, as a header with length 0 is malformed anyway, right? Also, this bug might also occur for other layers, if they also assume that `hdrLen` is always valid.

Packet++/src/SipLayer.cpp
Minilulatsch: Yep, did that.

Packet++/src/Packet.cpp
Minilulatsch: I also added the example crash files to the folder containing the regression tests

Packet++/src/SSLHandshake.cpp
```diff
@@ -1346,6 +1346,8 @@ SSLServerHelloMessage::SSLServerHelloMessage(uint8_t* data, size_t dataLen, SSLH
 		}
 
 		m_ExtensionList.pushBack(newExt);
+		if (newExt->getTotalLength() <= 0)
```
seladb, line 1349: I already fixed that in `master`: https://github.com/seladb/PcapPlusPlus/blob/f30de1387acc1f823e0d4779e5565f4cc32dd4a6/Packet%2B%2B/src/SSLHandshake.cpp#L1258

Packet++/src/IPv4Layer.cpp
```diff
@@ -237,7 +237,7 @@ IPv4Layer& IPv4Layer::operator=(const IPv4Layer& other)
 void IPv4Layer::parseNextLayer()
 {
 	size_t hdrLen = getHeaderLen();
-	if (m_DataLen <= hdrLen)
+	if (m_DataLen <= hdrLen || hdrLen <= 0)
```
seladb, line 240: `hdrLen` can't be lower than 0 because it's of type `size_t` which is unsigned.

So you can change this to:
```cpp
if (m_DataLen <= hdrLen || hdrLen == 0)
```

Packet++/src/SipLayer.cpp
```diff
@@ -307,7 +307,7 @@ void SipRequestFirstLine::parseVersion()
 	verPos++;
 
 	int endOfVerPos = 0;
-	while (((verPos+endOfVerPos)[0] != '\r') && ((verPos+endOfVerPos)[0] != '\n'))
+	while (((verPos+endOfVerPos)[0] != '\r') && ((verPos+endOfVerPos)[0] != '\n') && ((verPos + endOfVerPos) < (char *) (m_SipRequest->m_Data + m_SipRequest->m_DataLen)))
```
seladb, line 310: I suggest we switch the order and check `((verPos + endOfVerPos) < (char *) (m_SipRequest->m_Data + m_SipRequest->m_DataLen))` first (before we check for `\r` and `\n`). 

The reason is that if `verPos + endOfVerPos` is already at the end of the packet, then reading the byte in index zero: `(verPos+endOfVerPos)[0]` is a buffer overflow. I actually discovered that from one of the files you added as examples: `sip_layer_parse_version.bug`.

seladb (commented):
Please notice that the files you added fail the fuzzers test: https://cirrus-ci.com/task/4534050868166656

I tried to reproduce it in my local machine (Ubuntu 20.04) but all of the files actually pass except for `sip_layer_parse_version.bug` which I explain why here: https://github.com/seladb/PcapPlusPlus/pull/526/files#r511332084

Maybe you should test it on Ubuntu 18.04 where the CI runs on: https://github.com/Minilulatsch/PcapPlusPlus/blob/master/.cirrus.yml#L34-L44

In addition, please update your branch from `master`, at least one issue was already fixed there

seladb:
@Minilulatsch this PR is open for a long time. Do you think you can find the time to complete the remaining work so we can merge it?

Packet++/src/SSLHandshake.cpp
seladb: I just realized I only fixed it for the client-hello message and this fix is for server-hello

Packet++/src/SipLayer.cpp
seladb: I fixed that

seladb:
I already made the changes. I also remove the example files because the fuzzing test application currently [looks only at the first packet in the file](https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Fuzzers/FuzzTarget.cpp#L45-L49) so there is no point in providing it a pcap file with multiple packets, unless the first is the one we want to test (which I believe is not the case in those files). If you want to provide other examples with 1 packet in each file you're more than welcome to do it.

PR #527, opened 2020-09-12
Added support for ICMP in hash5Tuple. (with unittest)

miketruman:
Added support for ICMP in hash5Tuple. (with unittest).

### DIFF
--- a/Packet++/src/PacketUtils.cpp
+++ b/Packet++/src/PacketUtils.cpp
@@ -5,6 +5,7 @@
 #include "IPv6Layer.h"
 #include "TcpLayer.h"
 #include "UdpLayer.h"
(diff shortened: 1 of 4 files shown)


seladb (commented):
ICMP is very different from TCP and UDP, simply because of the fact it doesn't have ports. So the term "5-tuple" (which usually means IP src/dst, Port src/dst and protocol) doesn't apply to it.

I'd consider creating a separate method for ICMP hashing.

seladb:
Please also notice that `PcapSplitter` tests fails

seladb:
@miketruman have you had a chance to look at my comments?

seladb:
Closing this PR for new, please reopen if relevant

PR #538, opened 2020-10-27
Add configurable out-of-order packets per flow limit

LeonBuikstraFox:
This optional limit can prevent unidirectional flows from building up
an unlimited amount of out-of-order packets after a single packet was
lost.


Packet++/src/TcpReassembly.cpp
```diff
@@ -41,6 +42,8 @@ TcpReassembly::TcpReassembly(OnTcpMessageReady onMessageReadyCallback, void* use
 	m_ClosedConnectionDelay = (config.closedConnectionDelay > 0) ? config.closedConnectionDelay : 5;
 	m_RemoveConnInfo = config.removeConnInfo;
 	m_MaxNumToClean = (config.removeConnInfo == true && config.maxNumToClean == 0) ? 30 : config.maxNumToClean;
+	// if m_MaxOutOfOrderFragments is set to numeric_limits<size_t>::max() there is no limit, since size() cannot return a value bigger than that
+	m_MaxOutOfOrderFragments = (config.maxOutOfOrderFragments ? config.maxOutOfOrderFragments : std::numeric_limits<size_t>::max());
```
seladb, line 46: let's not use `<limits>` and `std::numeric_limits<size_t>::max()` because I don't think it's supported on all platforms. Instead, we have a few options:
- Set the max possible value to some very high number, let's say `0xffffffff`
- Convert `m_MaxOutOfOrderFragments ` to `int` and set its default value to `-1` which means "unlimited". Then add a check whether this number is negative and treat it as unlimited

Tests/Pcap++Test/TestDefinition.h
```diff
@@ -57,6 +57,7 @@ PTF_TEST_CASE(TestTcpReassemblyIPv6);
 PTF_TEST_CASE(TestTcpReassemblyIPv6MultConns);
 PTF_TEST_CASE(TestTcpReassemblyIPv6_OOO);
 PTF_TEST_CASE(TestTcpReassemblyCleanup);
+PTF_TEST_CASE(TestTcpReassemblyMaxOutOfOrderFrags);
```
seladb, line 60: Can you shorted the test-case name to fit into 30 chars? Otherwise it looks kind of weird when running the tests:

image

Issue #556, opened 2020-12-28
Issue with TestDpdkDeviceWorkerThreads & TestDpdkMbufRawPacket test in unit tests of Pcap++Test


cartman-156:
I have installed pcap++(v20.08) & dpdk(v18.11.10) in my CentOS7 VM. 
Installation was done as follows.
**DPDK Installation**
`make config T=x86_64-native-linuxapp-gcc` and `make install`
**PCAP++ Installation**
Using `install-with-dpdk.sh` with DPDK path as input. 
Port was bound to DPDK driver using `dpdk-setup.sh` present inside DPDK installer.

Traffic generation is handled manually using `tcpreply` via another VM which is connected to the port which is linked with DPDK-compatible driver. 

With this setup I am trying to run the` Pcap++Test` which comes with pcap++ installation with input `-i <my_vm_ip>` and `-d 0` since there only one DPDK port. With this inputs TestDpdkDeviceWorkerThreads & TestDpdkMbufRawPacket keeps on failing with 
`(Tests/DpdkTests.cpp:645). assert lower than failed: actual: 20 >= expected: 20` & 
`(Tests/DpdkTests.cpp:867). assert lower than failed: actual: 30 >= expected: 30` respectively. 
When port is detached from dpdk the traffic is successfully received on the same via same `tcpreply` command.

PFB the complete output from `Pcap++test`

```
EAL: Detected 4 lcore(s)
EAL: Detected 1 NUMA nodes
EAL: Multi-process socket /var/run/dpdk/rte/mp_socket
EAL: No free hugepages reported in hugepages-1048576kB
EAL: Probing VFIO support...
EAL: PCI device 0000:0b:00.0 on NUMA socket -1
EAL:   Invalid NUMA socket, default to 0
EAL:   probe driver: 15ad:7b0 net_vmxnet3
EAL: PCI device 0000:13:00.0 on NUMA socket -1
EAL:   Invalid NUMA socket, default to 0
EAL:   probe driver: 15ad:7b0 net_vmxnet3
EAL: PCI device 0000:1b:00.0 on NUMA socket -1
EAL:   Invalid NUMA socket, default to 0
EAL:   probe driver: 15ad:7b0 net_vmxnet3
subnet prefixLength '0' illegal
TestIPAddress                 : PASSED
TestMacAddress                : PASSED
TestLRUList                   : PASSED
TestGeneralUtils              : PASSED
TestGetMacAddress             : PASSED
[...]

PR #561, opened 2021-01-07
=Fix for issue #556. Added logic for traversing all the available Rx queues

cartman-156:


cartman-156:
Fix for #556  Added logic for traversing all the available Rx queues in TestDpdkMbufRawPacket and TestDpdkDeviceWorkerThreads

Issue #564, opened 2021-01-09
Error while using IPReassembly.h with -Wextra and -Werror flags


cartman-156:
I'm trying to use IPReassembly method from Pcap++ in the project to perform the defragmentation of received packets. However, when I am trying to compile the code with IPReassembly.h it's failing with below errors.

```
PcapPlusPlus/Dist/header/IPReassembly.h:222:4: error: base class ‘class pcpp::IPReassembly::PacketKey’ should be explicitly initialized in the copy constructor [-Werror=extra]
    IPv6PacketKey(const IPv6PacketKey& other) : m_FragmentID(other.m_FragmentID), m_SrcIP(other.m_SrcIP), m_DstIP(other.m_DstIP) { }
```

```
PcapPlusPlus/Dist/header/IPReassembly.h:130:4: error: base class ‘class pcpp::IPReassembly::PacketKey’ should be explicitly initialized in the copy constructor [-Werror=extra]
    IPv4PacketKey(const IPv4PacketKey& other) : m_IpID(other.m_IpID), m_SrcIP(other.m_SrcIP), m_DstIP(other.m_DstIP) { }
```

As I understand this is because `PacketKey` copy constructor not being called in case of `IPv4PacketKey` & `IPv6PacketKey` copy constructor. This doesn't get caught in `IPFragUtil` & `IPDefragUtil` as `-Wextra` and `-Werror` flags are not present in compilation of main.cpp file. When I tried to include the same in Makefile, I got similar error there as well.

I've fixed the same in local PcapPlusPlus project, however I would prefer if you could fix the same in your code as well. 

PFB the change which I did in local PcapPlusPlus project(`Dist/header dir`) which allowed me to proceed for now.

```
[root@localhost Dist/header]# diff -wb IPReassembly.h ../../Packet++/header/
130c130
<                       IPv4PacketKey(const IPv4PacketKey& other) : PacketKey(other), m_IpID(other.m_IpID), m_SrcIP(other.m_SrcIP), m_DstIP(other.m_DstIP) { }
---
>                       IPv4PacketKey(const IPv4PacketKey& other) : m_IpID(other.m_IpID), m_SrcIP(other.m_SrcIP), m_DstIP(other.m_DstIP) { }
222c222
<                       IPv6PacketKey(const IPv6PacketKey& other) : PacketKey(other), m_FragmentID(other.m_FragmentID), m_SrcIP(other.m_SrcIP), m_DstIP(other.m_DstIP) { }
---
[...]

PR #566, opened 2021-01-11
Added call to PacketKey copy constructor

cartman-156:
Fix for #564. Call to base class(PacketKey) copy constructor is added from IPv4PacketKey & IPv6PacketKey copy constructors.


cartman-156:
@seladb I am not sure why Travis CI build is failing for this. In total I can see 4 builds being fired for this PR of which 1 is successful but others are failed. Please let me know if there is anything I can do to fix the same

PR #589, opened 2021-02-18
Missing dnsPacketCount increment

ramazantokay:
Hello,
consumePacket() function is missing the dnsPacketCount increment so I added.

Best


seladb:
Good catch! Thanks for the fix, I really appreciate your contribution!

Issue #602, opened 2021-03-09
Makefile dependencies


amirgon:
This continues the discussion on #599 :

When a header file changes the build system does not detect it and does not build the required files as expected.
(Cmake handles this automatically, but [you can achieve this with pure Makefiles as well](http://make.mad-scientist.net/papers/advanced-auto-dependency-generation/).  )

It looks like the dep files are created (with `-MF ` gcc option) but not used or not used correctly.

To reproduce:

- Run `make libs`
- Change a header file (for example `Packet++/header/EthLayer.h`)
- Run `make libs`

The second run of `make libs` does not detect the changed header and does not build the project with the modified header fille.

amirgon:
> Just to make sure I understand: what do you mean by "does not build the project with the modified header fille"? Do you mean that the object files aren't rebuilt or the header file isn't copied to `Dist/`?

Object files aren't rebuilt.

> BTW, did you find any clues on how to solve it?

It's explained very well on http://make.mad-scientist.net/papers/advanced-auto-dependency-generation/
I recommend reading this.

What's basically missing is an include in the Makefile to the dep files.
But there are are some subtle details, like the include position in the Makefile, preventing Make from failing when deps are missing etc., everything is explained in that link.

If you want to avoid this headache - you probably need to switch to CMake.

seladb:
Thanks for your help @amirgon ! I think I figured out the issue - please see PR #603 .
Please let me know what you think

seladb:
Thanks @amirgon for reporting this issue and for your help in solving it, I really appreciate it!
The fix is now pushed to `master` so I'll close this ticket. Please reopen if any issue.

PR #603, opened 2021-03-10
Issue #602 - Fix makefile dependencies

seladb:


seladb:
thanks @amirgon for the feedback!
Unfortunately I couldn't reproduce the issue you mentioned.
When I run `touch Packet++/header/EthLayer.h` and then `make libs` I get the same results as you, but if I run `make libs` again nothing is being re-built. Here is the output I see (running on Ubuntu 20.04, but I got similar results on Windows/MinGW):
```shell
seladb@ubunu2004:~/PcapPlusPlus$ touch Packet++/header/EthLayer.h 
seladb@ubunu2004:~/PcapPlusPlus$ make libs
==== Building target: Common++ ====
Finished successfully building: Common++.lib
 
Finished successfully building: Common++.debug
 
==== Building target: Packet++ ====
Building file: src/ArpLayer.cpp
Building file: src/VlanLayer.cpp
Building file: src/SllLayer.cpp
Building file: src/GreLayer.cpp
Building file: src/VxlanLayer.cpp
Building file: src/EthLayer.cpp
Building file: src/Packet.cpp
Building file: src/NullLoopbackLayer.cpp
Finished successfully building: Packet++
 
==== Building target: Pcap++ ====
Building file: src/NetworkUtils.cpp
Building file: src/RawSocketDevice.cpp
Finished successfully building: Pcap++
 
Finished successfully building PcapPlusPlus libs
 
seladb@ubunu2004:~/PcapPlusPlus$ make libs
==== Building target: Common++ ====
Finished successfully building: Common++.lib
 
Finished successfully building: Common++.debug
 
==== Building target: Packet++ ====
Finished successfully building: Packet++
 
==== Building target: Pcap++ ====
Finished successfully building: Pcap++
 
Finished successfully building PcapPlusPlus libs

```

amirgon:
Debuggin this with `make -nd libs` I can see that:

```
...
   Considering target file `Obj/Packet.o'.
...
     Prerequisite `Obj/Packet.d' is newer than target `Obj/Packet.o'.
...
    Must remake target `Obj/Packet.o'.
Invoking recipe from Makefile:54 to update target `Obj/Packet.o'.
echo Building file: src/Packet.cpp
g++ -DLINUX -I"./src" -I"./header" -I"../Common++/header" -I"/root/rdmaip_root/rdmaip_sim/PcapPlusPlus/3rdParty/EndianPortable/include" -I"../3rdParty/hash-library" -I/usr/include/netinet -fPIC -O2 -g -Wall  -c -fmessage-length=0 -MMD -MP -MF "Obj/Packet.d" -MT "Obj/Packet.o" -o "Obj/Packet.o" "src/Packet.cpp"
    Successfully remade target file `Obj/Packet.o'.
```

It may be related to [this comment](http://make.mad-scientist.net/papers/advanced-auto-dependency-generation/#post-59:~:text=In%20some%20cases%20it%E2%80%99s%20been%20reported,spurious%20rebuilds%20of%20the%20object%20file.) regarding spurious rebuilds, follows by an example for how to solve this (POSTCOMPILE).

seladb:
thanks @amirgon ! I looked at the comment and managed to implement the fix they suggest using `POSTCOMPILE`.

However, there is a backward compatibility issue here. This solution requires `mv` and `touch` that are not natively available on Windows. Fortunately those commands are available in `MSYS` but in order to use them they should be defined in [`platform.mk.*`](https://github.com/seladb/PcapPlusPlus/tree/master/mk) as new variables `MV` and `TOUCH`. Now here is the problem: once I start using these variables in the `Makefile`, the build will fail for anyone who didn't run `configure-<OS>`. That means that if you have a working local copy of PcapPlusPlus and you pull the latest code (without running `configure-*`) the build will start failing until you run `configure-*` again.
I tried to make the `POSTCOMPILE` fix backward compatible but it's very difficult because it involves changing the file extension from `.d` to `.Td` (where `POSTCOMPILE` change it back to `.d`).

I can suggest a few options:
- Leave it as it is, meaning without `POSTCOMPILE`
- Split the fix into 2 phases: 
  1. Add these new variables `MV` and `TOUCH` in `platform.mk.*` now but NOT implement the `POSTCOMPILE` solution yet
  2. Add the `POSTCOMPILE` fix in a couple of months, hoping that most of the users already have a local copy of PcapPlusPlus that has these variables

amirgon:
> once I start using these variables in the `Makefile`, the build will fail for anyone who didn't run `configure-<OS>`.

Do you have some release schedule or strategy?  
You can require users to run `configure-<OS>` after each major release, for example, assuming they are a few months apart.
The spurious rebuilds problem may be a little annoying but it's not critical so you can add the `POSTCOMPILE` later on your next major release and ask users to run `configure-<OS>` then.  

Correctly handling of  depfiles, on the other hand (as fixed in this PR), is important because it can cause wrong builds.

seladb:
I'm targeting the next release during the next couple of months. I've added `mv` and `touch` commands to this PR and will add `POSTCOMPILE` after the next release.

PR #613, opened 2021-03-26
Initialize bpf_program struct to zero

nickamon:
Init the struct on the stack to zero to avoid segmentation fault or invalid pointer errors when trying to free the struct.  This occurs on my system when an invalid BPF filter is specified.
Error compiling filter. Error message is: syntax error
*** Error in `./Dist/examples/LiveStreamSplitter': free(): invalid pointer: 0x00007fac65049400 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x81329)[0x7fac64d02329]
        else if (method == SPLIT_BY_ROUND_ROBIN)
/lib64/libpcap.so.1(pcap_freecode+0x18)[0x7fac65a9f558]
./Dist/examples/LiveStreamSplitter[0x42cc21]
./Dist/examples/LiveStreamSplitter[0x4065bf]
/lib64/libc.so.6(__libc_start_main+0xf5)[0x7fac64ca3555]
./Dist/examples/LiveStreamSplitter[0x415dac]


Pcap++/src/PcapDevice.cpp
```diff
@@ -20,7 +20,7 @@ bool IPcapDevice::setFilter(std::string filterAsString)
 		return false;
 	}
 
-	struct bpf_program prog;
+	struct bpf_program prog = {};
```
seladb, line 23: This type of initialization is a C++11 feature. While I plan to move PcapPlusPlus to C++11 soon, as of now I wouldn't want to add C++11 specific code.

nickamon: OK, fair enough.  What do you suggest as an alternative, I'm not well-versed in C++?  Maybe explicitly setting the prog.bf_len = 0 and prog.bf_insns = NULL?

seladb: I'm not sure this struct needs to be initialized. I think libpcap initializes it internally: https://www.tcpdump.org/pcap.html

nickamon: It needs to be initialized because if the compilation of the filter fails, your code attempts to free the bpf_program structure, but if you review the code for [pcap_compile](https://github.com/the-tcpdump-group/libpcap/blob/bb02779a37fdf815da58e8828eb4ce99a4f86580/gencode.c#L724), it initializes the bpf_program structure only at the conclusion of the function (when the filter compilation succeeds).  In a filter-compilation-failure case, it is not initialized, resulting in a wild pointer that pcap_freecode attempts to free.  A simple test that can demonstrates this is to set the prog.bf_len to an arbitrary value and then check its value after pcap_compile fails (by specifying an invalid filter), you'll note that the value is unchanged.

seladb: So maybe we should just remove this call to `pcap_freecode`?

nickamon: lol, fair enough.

seladb: @nickamon would you like to make this change so we can merge the PR?

nickamon: Hmm, kinda of messed up my branch, let me know if you want me to close this pull request and reopen a new one w/ just the fix.

Issue #598, opened 2021-03-05
Sending packet exceeds MTU due to Ethernet layer


The7Tycoon:
I have a PCAP containing UDP packets that have been fragmented due to being well over the MTU. The first fragments fully use the MTU, meaning the IP layer is 1500 bytes long (IP header + payload). Then, when they were sent, an traditional 14 bytes Ethernet layer was added, so the packet is 1514 bytes long in total.

However, now that I want to send them back over the network, I'm having trouble with PcapLiveDevice::sendPacket(). It takes the whole raw packet, including the 14 bytes link layer, and determines that the packet is bigger than the 1500 bytes MTU, when I believe the Ethernet header should not be taken into consideration. 

I tried sending only the data starting at the IP layer (so exactly 1500 bytes), no error is shown but I can't see any packets on wireshark. The only temporary solution was to increase the MTU to 1514 so that the checks in sendPacket() passes.

Am I doing something wrong ? Or should sendPacket() only count the data length at the IP layer ?

All of this is tried on Linux (CentOS 8.2), with a default MTU of 1500.

seladb:
hi @The7Tycoon I'm not sure I fully understand your scenario: the first fragment is 1500 bytes long but doesn't contain the Ethernet header? How did you capture it then? Which interface are you using? 
And regarding sending these packets: how were these 14 bytes of the Ethernet header added? Did you add them?
Basically, you can't send packets that exceed the MTU but I'm not sure which interfaces are used for capturing and sending packets so it's hard to pinpoint the issue

SamuelBailey:
# Problem
I have come across this exact issue. I captured UDP packets from a PcapLiveDevice, and then attempted to forward the packet using PcapLiveDevice::sendPacket(). This worked fine for small packets, however the size check for the MTU (1500 bytes) meant that the device refused to send for larger packets, because it was recording a size of 1514 bytes (including the ethernet header). Looking this up, the MTU should only include the IP layer size, not the size of the Data Link Layer header, so the check should ether:
1. Compare length of the **packet** (**including data link layer**) against (**MTU** + **Data Link layer**)

or

2.  Find the size of the **IP layer** contained within the packet and compare this against **MTU**.

I have my own public fork of PcapPlusPlus in which I patched this for use in my own project. I hadn't opened a pull request because when I cloned the repository and just ran the tests (before making any changes), I couldn't get all of them to pass. 

[...]

PR #614, opened 2021-03-26
Fixed MTU check

SamuelBailey:
Altered MTU check to use the size of the data link payload, rather than the length of the entire packet. A `checkMtu` flag was added to the `sendPacket` and `sendPackets` methods for `PcapLiveDevice`. This determines whether checking the MTU should be performed since there is some overhead when a packet is required to be parsed. The `checkMtu` value defaults to `true` when parsing doesn't need to be performed and `false` when parsing is required to perform this check, in order to maintain performance.


Pcap++/header/PcapLiveDevice.h
```diff
@@ -406,17 +406,39 @@ namespace pcpp
...
+		 * @param[in] packetPayloadLength The length of the IP layer of the packet
+		 * @return True if the packetPayloadLength is less than or equal to the device MTU
+		 */
+		bool doMtuCheck(int packetPayloadLength);
+
 		/**
 		 * Send a RawPacket to the network
 		 * @param[in] rawPacket A reference to the raw packet to send. This method treats the raw packet as read-only, it doesn't change anything
 		 * in it
+		 * @param[in] checkMtu Whether the length of the packet's payload should be checked against the MTU. If enabled this comes with a small performance penalty.
```
seladb, line 420: I'd also mention that the default value is `false`

Pcap++/header/PcapLiveDevice.h
```diff
@@ -425,59 +447,63 @@ namespace pcpp
...
-		 * - Packet length is larger than device MTU
 		 * - Packet could not be sent due to some error in libpcap/WinPcap/Npcap
+		 * This method will not check the Mtu of the packet
 		 */
 		bool sendPacket(const uint8_t* packetData, int packetDataLength);
 
 		/**
 		 * Send a parsed Packet to the network
 		 * @param[in] packet A pointer to the packet to send. This method treats the packet as read-only, it doesn't change anything in it
+		 * @param[in] checkMtu Whether the length of the packet's payload should be checked against the MTU. If enabled this comes with a small performance penalty.
```
seladb, line 458: ditto - I'd also mention that the default value is `false`

Pcap++/header/PcapLiveDevice.h
```diff
@@ -425,59 +447,63 @@ namespace pcpp
...
 		 */
-		bool sendPacket(Packet* packet);
+		bool sendPacket(Packet* packet, bool checkMtu = true);
 
 		/**
 		 * Send an array of RawPacket objects to the network
 		 * @param[in] rawPacketsArr The array of RawPacket objects to send. This method treats all packets as read-only, it doesn't change anything
 		 * in them
 		 * @param[in] arrLength The length of the array
+		 * @param[in] checkMtu Whether to check the size of the packet payload against MTU size. Incurs a parsing overhead.
```
seladb, line 472: ditto

Pcap++/header/PcapLiveDevice.h
```diff
@@ -425,59 +447,63 @@ namespace pcpp
 		 * @return True if packet was sent successfully. False will be returned in the following cases (relevant log error is printed in any case):
 		 * - Device is not opened
 		 * - Packet length is 0
-		 * - Packet length is larger than device MTU
 		 * - Packet could not be sent due to some error in libpcap/WinPcap/Npcap
+		 * This method will not check the Mtu of the packet
```
seladb, line 451: `MTU` in capital letters? 🤔

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -594,44 +611,49 @@ bool PcapLiveDevice::sendPacket(const uint8_t* packetData, int packetDataLength)
...
 }
 
-bool PcapLiveDevice::sendPacket(Packet* packet)
+bool PcapLiveDevice::sendPacket(Packet* packet, bool checkMtu)
 {
 	RawPacket* rawPacket = packet->getRawPacket();
-	return sendPacket(*rawPacket);
+	if (checkMtu)
+	{
+		int packetPayloadLength = (int)packet->getFirstLayer()->getLayerPayloadSize();
```
seladb, line 619: this line assumes the first layer is a data link layer, but it doesn't have to be like that: https://github.com/seladb/PcapPlusPlus/blob/master/Packet%2B%2B/src/Packet.cpp#L713-L744

I'd suggest checking here if: `packet->getFirstLayer()->getOsiModelLayer() == pcpp::OsiModelDataLinkLayer`

Pcap++/header/PcapLiveDevice.h
```diff
@@ -406,17 +406,39 @@ namespace pcpp
...
+		 * @param[in] packetDataLength The length of the buffer (this is the entire packet, including link layer)
+		 * @param[in] packetPayloadLength The length of the payload for the data link layer. This includes all data apart from the header for the
+		 * data link layer.
+		 * @return True if the packet was sent successfully. False will be returned in the following cases (relevant log error is printed in any case):
+		 * - Device is not opened
+		 * - Packet data length is 0
+		 * - Packet payload length is larger than device MTU
+		 * - Packet could not be sent due to some error in libpcap/WinPcap/Npcap
+		 */
+		bool sendPacket(const uint8_t* packetData, int packetDataLength, int packetPayloadLength);
```
seladb, line 441: This method seems a bit confusing to me. I'm not sure users will really understand the difference between `packetDataLength` and `packetPayloadLength` (although it's explained very well in the doxygen documentation).
Also, how should users "guess" the `packetPayloadLength` without parsing the packet?

SamuelBailey: Given that this method takes a byte array rather than a `pcpp::Packet` or `pcpp::RawPacket` I thought it would be useful to be able to run the check for the MTU without incurring the penalty of packet parsing if you happened to have this information already. For example, in my own use case, I was using PcapPlusPlus to send packets produced by a network simulator. Given that the simulator had its own `packet` representation, it was possible to find this info without additional parsing. It is probably quite a niche use case where users would already have access to this information though.

If you would prefer to remove this method, considering it may be confusing, I can do that; or I could add a bit more to the documentation explaining when it should be used?

Also, for the method:
```c++
bool sendPacket(const uint8_t* packetData, int packetDataLength)
```
it made sense in my head to have this method not parse the MTU, and to have
```c++
bool sendPacket(const uint8_t* packetData, int packetDataLength, int packetPayloadLength)
```
for parsing the MTU, in order to keep the implementations which used byte arrays simple. In hindsight, in order to have more intuitive function signatures, it is probably worth including the default parameters:
```c++
bool checkMtu = false, pcpp::LinkLayerType linkType = pcpp::pcpp::LINKTYPE_ETHERNET
```
which were mentioned earlier in the thread #598, to give the option of parsing the packets and checking MTU. I'll add that back in unless there's any reason not to.

Pcap++/header/PcapLiveDevice.h
SamuelBailey: Good point. Will add this.

Pcap++/header/PcapLiveDevice.h
SamuelBailey: will do

Pcap++/header/PcapLiveDevice.h
SamuelBailey: will do

Pcap++/header/PcapLiveDevice.h
SamuelBailey: Thanks, will fix.

Pcap++/src/PcapLiveDevice.cpp
SamuelBailey: Thanks. Will add this.

I can add a check for if the first layer is `pcpp::OsiModelNetworkLayer` as well. In that instance, the MTU should just be the length of the packet.

Pcap++/header/PcapLiveDevice.h
```diff
@@ -425,59 +447,63 @@ namespace pcpp
 		 * @return True if packet was sent successfully. False will be returned in the following cases (relevant log error is printed in any case):
 		 * - Device is not opened
 		 * - Packet length is 0
-		 * - Packet length is larger than device MTU
 		 * - Packet could not be sent due to some error in libpcap/WinPcap/Npcap
+		 * This method will not check the Mtu of the packet
 		 */
 		bool sendPacket(const uint8_t* packetData, int packetDataLength);
```
SamuelBailey, line 453: For this method it made sense in my head to have this method not parse the MTU, and to have
```c++
bool sendPacket(const uint8_t* packetData, int packetDataLength, int packetPayloadLength)
```
for parsing the MTU, in order to keep the implementations which used byte arrays simple. In hindsight, in order to have more intuitive function signatures, it is probably worth including the default parameters:
```c++
bool checkMtu = false, pcpp::LinkLayerType linkType = pcpp::pcpp::LINKTYPE_ETHERNET
```
which were mentioned earlier in the thread #598, to give the option of parsing the packets and checking MTU. I'll add that back in unless there's any reason not to.

Pcap++/src/PcapLiveDevice.cpp
seladb: yes, that makes sense

Pcap++/header/PcapLiveDevice.h
seladb: Yes that makes sense. Let's keep that method and add a little more documentation

Pcap++/header/PcapLiveDevice.h
seladb: Yes, let's add these arguments with the default value you suggest. If `checkMtu == false` we can simply send the byte array without any check

Issue #611, opened 2021-03-23
Bad result on IPV4Address comparison


manuka89:
Hello,

I think that IPV4Address operator< does not work correctly. Small example to demonstrate:

    #include <IPv4Layer.h>
    #include <Packet.h>
    #include <PcapFileDevice.h>
    #include <iostream>
    
    using namespace pcpp;
    int main(int argc, char* argv[])
    {
        const IPv4Address smallerAddress { "212.0.0.1" };
        const IPv4Address largerAddress { "224.0.0.0" };

        std::cout << "SmallerAddress: " << smallerAddress.toString()
                  << " LargerAddress: " << largerAddress.toString()
                  << " Comparison (smallerAddress < largerAddress): " << (smallerAddress < largerAddress) << std::endl;

        return 0;
    }

The result is:

SmallerAddress: 212.0.0.1 LargerAddress: 224.0.0.0 Comparison (smallerAddress < largerAddress): 0

The operator< of IPv4Address uses the toInt() method:

	uint32_t IPv4Address::toInt() const
	{
		uint32_t addr;
		memcpy(&addr, m_Bytes, sizeof(m_Bytes));
		return addr;
	}

The byte array should be reversed in my opinion in a temporary variable before converting it to uint32_t.

seladb:
Yes, I believe you're right, thanks for reporting this issue!

Would you consider opening a PR with the fix?

manuka89:
Thanks for your response,. Yes I will do that.

seladb:
Thanks you so much @mmizi for providing the fix! 
Your PR #621 has been merge to `master` so I'll close this issue

PR #621, opened 2021-04-02
Bug #611 : Less operator does not work correctly due to toInt method (byte order…

github-euronext:
Less operator does not work correctly due to toInt method (byte order reversed)

### DIFF
--- a/Common++/header/IpAddress.h
+++ b/Common++/header/IpAddress.h
@@ -4,7 +4,7 @@
 #include <stdint.h>
 #include <string.h>
 #include <string>
-
+#include <algorithm>
 
 /// @file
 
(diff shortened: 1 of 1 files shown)


seladb:
@github-euronext your observation is correct! As you pointed out `toInt()` returns the wrong result.
However there are about 50 places in the code that use this method, and many of them assume that `toInt()` returns the reversed result. What we need to do is go over them one by one and change it... That's why some of the tests are failing

mmizi:
I think your second option can be acceptable. For me toInt() can return as well in a reversed byte order so that we do not break any code. There is no logical issue in returning a reversed order.

So, yes in my point of view we could correct only the operator<. So we go for the second option?

seladb:
> So we go for the second option?

Sounds good to me! In the future we might tackle the issue with `toInt()`

mmizi:
As requested, I repushed it with the correction only for the less operator. I've also changed the username and mail :) 

It looks like that all tests passed

Issue #609, opened 2021-03-22
PcapSplitter does not correctly splits UDP feed by server-port


manuka89:
When I invoke the PcapSplitter on an UDP feed and split by server-port I have some missing ports. After having checked the code I could see that, in case of UDP, always the default case is used:

	uint32_t getValue(pcpp::Packet& packet, PacketType packetType, uint16_t srcPort, uint16_t dstPort)
	{
		switch (packetType)
		{
		case SYN:
			return dstPort;
		case SYN_ACK:
			return srcPort;
		// other TCP packet or UDP packet
		default:
			return std::min<uint16_t>(srcPort, dstPort);
		}
	}

In my feed, the dstPort should be the server port but it is sometimes larger than the srcPort so that the port is not detected correctly.

Why it is not possible to check the source and destination IP whether it is in a multicast address range to decide whether it is the server port or not?

seladb:
This is a very good idea! I haven't thought of it!

Would you consider implementing it and open a PR?

manuka89:
Thanks, yes I will do it. Although, there is a limitation, it will only work for multicast addresses.

For the rest, I can use the the std:min algo.

seladb:
thank you so much @mmizi for working on this, I really appreciate the great time and efforts you put into this, BIG THANK YOU!
I encourage you to contribute more, it's because of people like this project can grow and evolve.

I'll close this issue now, please reopen if needed.

mmizi:
Thank you for your help and time during the PR process! It was a really good experience!

PR #622, opened 2021-04-02
Feature #609 : UDP Multicast

github-euronext:
Define algorithm for multicasting in case of port splitting, add client-port splitting feature


Common++/src/IpAddress.cpp
```diff
@@ -29,6 +31,10 @@ namespace pcpp
 		return std::string();
 	}
 
+	bool IPv4Address::isMulticast() const
```
seladb, line 34: I'd recommend adding this method to `IPAddress` as well

Examples/PcapSplitter/IPPortSplitters.h
```diff
@@ -233,6 +233,32 @@ class IPPortSplitter : public ValueBasedSplitter
...
+	 * An auxiliary method to indicate whether an IPv4/IPv6 source address is multicast or not
+	 */
+	bool isSrcIPMulticast(pcpp::Packet& packet)
+	{
+		if (packet.isPacketOfType(pcpp::IPv4))
+			return packet.getLayerOfType<pcpp::IPv4Layer>()->getSrcIPv4Address().isMulticast();
+		else if (packet.isPacketOfType(pcpp::IPv6))
+			return packet.getLayerOfType<pcpp::IPv6Layer>()->getSrcIPv6Address().isMulticast();
+		else
+			return false;
```
seladb, line 246: if we add `isMulticast()` to `IPAddress` we can simplify this method and `isDstIPMulticast()`

seladb (commented):
We cannot merge this PR before we fix and merge #621 , right?

Common++/src/IpAddress.cpp
```diff
@@ -70,6 +74,10 @@ namespace pcpp
 		return std::string();
 	}
 
+	bool IPv6Address::isMulticast() const
+	{
+		return !operator<(MulticastRangeLowerBound);
```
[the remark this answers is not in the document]
mmizi: Yes, we cannot merge this PR before #621 due to the operator< issue.

For multicast address ranges of IPv6 addresses I checked on [this page](https://www.ciscopress.com/articles/article.asp?p=2803866&seqNum=5#:~:text=Well%2Dknown%20multicast%20addresses%20are,255.255.)

If my understanding is correct an address larger than ff00 is a multicast address.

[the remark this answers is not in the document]
seladb: > If my understanding is correct an address larger than ff00 is a multicast address.

Yes, you're right. `ff00::/8` basically means that any IPv6 address that is equal or larger than `ff00:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0` would be considered as a multicast address.

seladb:
BTW, if you want your username to be shown in the project's [contributors list](https://github.com/seladb/PcapPlusPlus/graphs/contributors) I recommend pushing code with your GitHub username @mmizi

mmizi:
I've just seen that the same logic is used for IP splitting. Should we add this feature as well to IP splitting?

seladb:
> I've just seen that the same logic is used for IP splitting. Should we add this feature as well to IP splitting?
> 
> So as with ports we can consider that the server is who is sending the multicast feed whilst the client is the receptor of the feed (i.e. the multicast address itself)

Yes, I think that'd be a great idea!

Common++/header/IpAddress.h
```diff
@@ -81,7 +87,16 @@ namespace pcpp
...
+		bool operator<(const IPv4Address& rhs) const
+		{
+			uint32_t intVal = toInt();
+			std::reverse((uint8_t*)(&intVal), (uint8_t*)(&intVal) + sizeof(intVal));
+
+			uint32_t rhsIntVal = rhs.toInt();
+			std::reverse((uint8_t*)(&rhsIntVal), (uint8_t*)(&rhsIntVal) + sizeof(rhsIntVal));
+
+			return intVal < rhsIntVal;
+		}
```
seladb, line 99: Can you please update your branch from `PcapPlusPlus/dev`? This change is already merged...

mmizi: Done

mmizi:
By adding the support of multicast there are some test cases failing

mmizi:
By checking when split on client-ip this file is missing:

many-protocols-client-ip-ff02--1-2.pcap. Which is a multicast address. By putting some logs during execution:

Calling is Multicast with this: fe80::4dc7:f593:1f7b:dc11 Return: 1
Calling is Multicast with this: fe80::4dc7:f593:1f7b:dc11 Return: 1
Calling is Multicast with this: fe80::4dc7:f593:1f7b:dc11 Return: 1

For me the result is correct. I push the correction of the python test script to wait only for 4 files.

mmizi:
It looks like that I was mistaken.  The isMulticast returns 1 for these addresses but they are not multicast ones. It means that the less operator for IPv6 are also inversed. Am I correct?

seladb:
> to:
> 
>     * Save byteArray in a temporary variable for both
> 
>     * Reverse them
> 
>     * Then
>       return memcmp(toBytes(), rhs.toBytes(), sizeof(m_Bytes)) < 0;

Yes, I think that what we should do. Thanks for catching this!
While you're fixing the bug, can you also add tests for it? There are already tests here: https://github.com/seladb/PcapPlusPlus/blob/d2bba7d03c56c35da53981b55005bfa47b81c464/Tests/Pcap%2B%2BTest/Tests/IpMacTests.cpp#L104
but apparently they're not sufficient... Can you please extend them?

mmizi:
After having analysed the issue with the less operator for IPv6Address and having checked the unit tests, I found that there is no issue finally with the less operator for IPv6Address. 

The issue came from the bad definition of the multicast lower bound for IPv6. After this it works correctly. I also updated the expectation as now "many-protocols-server-ip-ff02--c.pcap" are considered client side and not server side.

Examples/PcapSplitter/IPPortSplitters.h
```diff
@@ -447,3 +497,76 @@ class ServerPortSplitter : public IPPortSplitter
...
+ * Splits a pcap file by server client (most of the time is similar to protocol). This means that all flows with a certain
+ * client port will be written to the same file. The client port for each flow is determined as follows: 1) if it's a TCP
+ * flow and we have the SYN packet - the client port is the source port of the SYN packet 2) if it's a TCP flow and we only
+ * have the SYN/ACK packet - the client port is the dest port of the SYN/ACK packet 3) if it's a partial TCP flow and
+ * we don't have the SYN or SYN/ACK packets, the server port will be determined by the port: the higher port is considered
+ * the client side 4) if it's a UDP multicast flow - if the sourceIP is a multicast address, the source port is considered
+ * as a client port, otherwise if the destIP is a multicast address, the dest port is considered as a client port 5) if
+ * it's a UDP flow - the client port will be determined by the port: the higher port is considered as client port
+ */
+class ClientPortSplitter : public IPPortSplitter
```
seladb, line 511: I'm not sure how useful this would be for users, that's why I didn't implement this in the first place. Server port is probably more useful. What do you think?

I'll leave it to you to decide because I don't have a hard opinion about this. However, if you decide to include it, would it be possible to add tests for it? Basically, you can add a test to [`test_pcapsplitter.py`](https://github.com/seladb/PcapPlusPlus/blob/master/Tests/ExamplesTest/tests/test_pcapsplitter.py)

mmizi: As an example for us it make sense as we want to analyse multicast feed by destination port which is in this case is the client port. 

I added a test for client port splitting.

mmizi:
There is a build which fails randomly. When I check the result it does not seem to be related to the code. Can you confirm?

seladb:
> There is a build which fails randomly. When I check the result it does not seem to be related to the code. Can you confirm?

Yes, it fails randomly, I'll re-run the build

PR #650, opened 2021-05-17
Linux: remove hardcoded netinet include folder

clementperon:
Hi @seladb,

Is there any reason to include this folder?
inet is provided by the libC no?  

When I tried both native and cross compilation under Linux, I didn't need it and it's polluting the cross compilation


clementperon:
I didn't touch to the one in mk/PcapPlusPlus.mk.macosx
`PCAPPP_INCLUDES += -I$(MACOS_SDK_HOME)/usr/include/netinet`

I think it's also not required, but as I can't build on MacOS...

PR #652, opened 2021-05-18
introduce CMake build system

clementperon:
Hi,

I have updated the cmake branch, and compilation of libs examples and tests are okay on my linux.

Unfortunately I don't know much about github-ci, so I din't backport travis jobs.


clementperon:
Hi @seladb,

I have some questions regarding the Cmake:
Why do you include the .h when declaring library/executable source?
`add_executable(PcapSplitter 
ConnectionSplitters.h
IPPortSplitters.h
main.cpp
SimpleSplitters.h
Splitters.h)`
instead of
`add_executable(PcapSplitter 
main.cpp)`

Also there are multiple cmake_minimum_required declared, I don't think it's good idea IMO

seladb:
@clementperon thanks for your work, it looks great!
The best way to verify that all platforms work is to add CMake to the CI. Would you consider doing that?

clementperon:
Hi @seladb,

Ok I will try, I have fix buid for PF_RING and DPDK on Linux.

clementperon:
@seladb seems good now!

seladb:
@clementperon thanks for working on this!
In the CI it looks like the Windows builds don't use CMake yet, is this still WIP?

clementperon:
@seladb only Linux is working at the moment with a "recent" CMake 3.0 so no Ubuntu 16.04 nor CentOS7 support. 

I have rework Finddpdk and it should work automatically with Pkg-config or if the package is installed in common folder (/usr/ or /usr/local) else we need to provide install directory with an env Variable DPDK_DIR.

CMake should also generate a proper Pkg-config and PCapPlusPlus.mk file for Linux.

I didn't touch Windows and don't understand why Windows is failing at the moment...

I will try to make it work on MacOS using a VM

Will see later for Android/Windows but merging Linux is a first step no ?

seladb:
Regarding Windows - there is some kind of an issue with `zstd` - @egecetin also saw it in [his PR](https://github.com/seladb/PcapPlusPlus/pull/775). I'll check and keep you posted.

Ideally we should replace the build system for all platforms. Assuming Linux and MacOS work, do you think it'll be a lot of efforts to add Windows?

clementperon:
@seladb I have no experience to support windows compilation but as it compile with Make if shouldn't be hard to support it with cMake.

I think we could also support centos if we bump to centos8.

seladb:
> @seladb I have no experience to support windows compilation but as it compile with Make if shouldn't be hard to support it with cMake.
> 
> I think we could also support centos if we bump to centos8.

I agree it should be easier to support Windows once Linux, Mac and Android work

clementperon:
@seladb mingw seems fine now !

.github/workflows/build_and_test.yml
```diff
@@ -537,5 +537,6 @@ jobs:
         working-directory: ./ToyVpn-PcapPlusPlus
         run: |
           sed -i.bak "s|abiFilters.*$|abiFilters '${{ matrix.target }}'|g" app/build.gradle
+          sed -i.bak "s|gradle-6.5-bin|gradle-7.3-bin'|g" gradle/wrappers/gradle-wrapper.properties
```
seladb, line 540: This is probably not needed anymore...

clementperon:
@seladb so Linux, MacOSX and MinGW are OK.

Todo:
- Android support.
- Shared libs
- generate a PcapPlusPlusConfig.cmake

seladb:
> Todo:
> 
>     * Android support.
> 
>     * Shared libs
> 
>     * generate a PcapPlusPlusConfig.cmake

We should also make sure Visual Studio works...

clementperon:
@seladb Android and FreeBSD are ok !

For windows, moving to C++11 Threads will help a lot !

seladb:
Woohoo 🎉 
Regarding moving to c++11 - it's a bigger change because it's a breaking change. We need to discuss when to do it.
Do you think we can decouple this cmake from c++11?

clementperon:
@seladb, i will try to link with pthread-w32. 

Do you want to keep both makefile and CMake in // in the CI?

seladb:
I think that if we move to cmake we can remove the current build system. Maintaining both will be harder.
BTW, much like moving to c++11, this is also a breaking change. Maybe we should release a version with the current changes, and then release another version with cmake only. WDYT?

clementperon:
Yes sounds good.

clementperon:
So introducing Cmake will drop Ubuntu 16.04 ? I think it's not that bad regarding that Ubuntu 22.04 is planned 3 months.

seladb:
> So introducing Cmake will drop Ubuntu 16.04 ? I think it's not that bad regarding that Ubuntu 22.04 is planned 3 months.

Yes, I think it makes sense. Ununtu 16.04 is pretty old...

seladb:
@clementperon as a general note I wanted to thank you for working on all these enhancements, I really appreciate your work and it's a great contribution to this project. BIG THANK YOU!! 🙏 🙏

clementperon:
@seladb if I understood correctly the project name is PcapPlusPlus and export 3 libs => Pcap++ Packet++ and Common++.

But it export only one PkgConfig file and all headers are installed in the same pcapplusplus folder.

Does some project only link with Packet++ or Common++ and without Pcap++ ?

Is it not simpler to just export one libPcap++.a or libPcapPlusPlus.a not sure about the naming?

clementperon:
@seladb any idea why the test_dnsresolver fails with MacOSX?

seladb:
> @seladb any idea why the test_dnsresolver fails with MacOSX?

I'm not sure why it fails... it gives this error:
```text
[ERROR: /Users/runner/work/PcapPlusPlus/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp: doOpen:276] (cannot open BPF device) /dev/bpf0: Permission denied
```

But I'm not really sure how it works in the current flow and doesn't work in your branch

clementperon:
@seladb Windows build is Ok, it will be better to move to C++ Threads and clean all the mess with pthreads-win32.

There is some missing things:
 - fix test on MacOS/FreeBSD
 - readd Asan/Msan test
 
Also do you want to keep pkg-config and PcapPlusPlus.mk files?

seladb:
> @seladb Windows build is Ok, it will be better to move to C++ Threads and clean all the mess with pthreads-win32.

Yes, I think moving to C++11 (and get rid of `pthreads-win32`) is one of the next things we need to invest in

> 
> There is some missing things:
> 
>     * fix test on MacOS/FreeBSD
> 
>     * readd Asan/Msan test
> 
> 
> Also do you want to keep pkg-config and PcapPlusPlus.mk files?

Once we have CMake support I don't think we need to keep the existing build system. But we need to make sure CMake supports all the options that are currently supported...

clementperon:
> >   
> > Also do you want to keep pkg-config and PcapPlusPlus.mk files?
> 
> Once we have CMake support I don't think we need to keep the existing build system. But we need to make sure CMake supports all the options that are currently supported...

You mean support ZSTD, LibPcap, WinPcap, NPCAP, etc.. ?

seladb:
> You mean support ZSTD, LibPcap, WinPcap, NPCAP, etc.. ?

Yes... do you see any other way that will make it simple for people to use this library?

clementperon:
@seladb I would say pkg-config is a nice to have as this allow user to have a Makefile without Using CMake in their project. But we can drop PcapPlusPlus.mk which is not standard...

seladb:
> @seladb I would say pkg-config is a nice to have as this allow user to have a Makefile without Using CMake in their project. But we can drop PcapPlusPlus.mk which is not standard...

Yes, I agree 👍

clementperon:
Hi @seladb,

So this means that now for Makefile only pkg-config works I will keep the same name PcapPlusPlus.pc just for retrocompatibility
But I think that Windows doesn't support pkg-config or maybe mingw, so this will break the windows build system and force user to move to CMake.

I will also rename the libs to libpcap++, libcommon++ and libpacket++ because uppercase are not standard in lib name.

Normally will all that the MR is almost ready, maybe can merge it in a "cmake" branch just for people to test it / contribute to it ? What do you think ?

ceggers-arri:
> @seladb if I understood correctly the project name is PcapPlusPlus and export 3 libs => Pcap++ Packet++ and Common++.
> 
Is installing a library named "libcommon++" into the main library folder (e.g. /usr/lib/) a good idea? What if other projects also want to claim the name "common" for them...

clementperon:
@ceggers-arri Agree we can install them in /usr/lib/pcapplusplus/(pcap++|common++|packet++)

@seladb what do you think?

seladb:
> @ceggers-arri Agree we can install them in ${LIBDIR}/pcapplusplus/(pcap++|common++|packet++) like for the include
> 
> @seladb what do you think?

Yes, I agree. We can put the lib files in `${LIBDIR}/pcapplusplus/`. This is how DPDK installation works, but in their case they create a folder with the target / arch name like `x86_64-linux-gnu`

tan-wei:
Wow, seems great. Waiting for the PR is merged.

3rdParty/MemPlumber/MemPlumber/CMakeLists.txt
```diff
@@ -1,24 +1,36 @@
...
 endif()
 
-target_include_directories(memplumber PUBLIC
-	${CMAKE_CURRENT_SOURCE_DIR}
-)
+target_include_directories(memplumber PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
+
+# execinfo for backtraces. This is only required on FreeBSD/Alpine, as both Mac
+# OS X and Linux include backtrace functions in libSystem / libc.
+find_package(Backtrace)
```
seladb, line 19: It looks like there is a built-in cmake package `FindBacktrace` that we can use:
https://cmake.org/cmake/help/latest/module/FindBacktrace.html

seladb, line 19: It looks like the is a `FindBacktrace` package we can use:
https://cmake.org/cmake/help/latest/module/FindBacktrace.html

Is running `find_package(Backtrace)`  similar to running `include(FindBacktrace)` like here: https://github.com/Wohlstand/TheXTech/blob/9b6244cda3d6d30b31b3b651b0699c3c745e147a/CMakeLists.txt#L9 ?

seladb (commented):
Some initial review comments, mostly for Linux configuration. I'll continue to review later

3rdParty/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+  FetchContent_Declare(
+    pcap
+    # ~~~
+    # URL https://www.tcpdump.org/release/libpcap-1.10.1.tar.gz
+    # URL ${CMAKE_CURRENT_SOURCE_DIR}/libpcap-1.10.1.tar.gz
+    # URL_HASH MD5=28e17495004036567c2cc884b51eba45
+    # ~~~
+    GIT_REPOSITORY https://github.com/clementperon/libpcap.git
+    GIT_TAG ce30044e01e99fb2139bf05c0d51be745959f1de
+)
```
seladb, line 30: I don't think we should install or build libpcap, package managers are meant for that...

Examples/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
+cmake_minimum_required(VERSION 3.12)
```
seladb, line 1: We should probably remove the `Makefile` from all examples, no?

clementperon: I keep Makefile because they are good example in case you want to use Pcapp++ in project that use make.

But I'm not against removing them

3rdParty/CMakeLists.txt
clementperon: Yes agree, this is something that I found in lot of cmake project but I'm more and more dubious about it. The lib is then internal to the project.

If we keep it external so we should also remove libzstd no ?

3rdParty/LightPcapNg/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+    FetchContent_Declare(
+      zstd
+      # URL
+      # https://github.com/facebook/zstd/releases/download/v1.5.2/zstd-1.5.2.tar.gz
+      URL ${CMAKE_CURRENT_SOURCE_DIR}/zstd-1.5.2.tar.gz
+      URL_HASH
+        SHA256=7c42d56fac126929a6a85dbc73ff1db2411d04f104fae9bdea51305663a83fd0
+      SOURCE_SUBDIR build/cmake)
+
+    set(ZSTD_BUILD_STATIC ON)
```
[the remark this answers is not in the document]
clementperon: Ok

3rdParty/CMakeLists.txt
seladb: yes, I commented about it here: https://github.com/seladb/PcapPlusPlus/pull/652#discussion_r928098805

Examples/CMakeLists.txt
seladb: I'm not sure these `Makefile`s will even work with the new CMake build system 🤔

seladb:
@clementperon I just saw there are conflicts between your branch and `dev`, I think we should resolve them before moving on...

CMakeLists.txt
```diff
@@ -0,0 +1,203 @@
+if(WIN32)
+  # We need 3.12 or later, so that we can set policy CMP0074; see below.
+  cmake_minimum_required(VERSION 3.12)
+else()
+  cmake_minimum_required(VERSION 3.6)
+endif()
```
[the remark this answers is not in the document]
clementperon: done

CMakeLists.txt
```diff
@@ -0,0 +1,203 @@
...
+
+# We want find_path() and find_library() to honor {packagename}_ROOT
+if(POLICY CMP0074)
+  cmake_policy(SET CMP0074 NEW)
+endif()
```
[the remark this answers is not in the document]
clementperon: I don't think it's a good idea else CMake warns when the policy is not set and uses OLD behavior.

CMakeLists.txt
```diff
@@ -0,0 +1,203 @@
...
+# Enable the option if it's available
+option(PCAPPP_HAS_PCAP_SET_DIRECTION "Enable Pcap Direction catpure"
+       ${HAVE_PCAP_DIRECTION})
+
+# Check in case user force it but it's not available
+if(PCAPPP_HAS_PCAP_SET_DIRECTION)
+  if(NOT HAVE_PCAP_DIRECTION)
+    message(FATAL_ERROR "PCAP library doesn't have Direction support!")
+  endif()
+  # add_definitions(-DHAS_SET_DIRECTION_ENABLED)
```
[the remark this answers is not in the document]
clementperon: I think it was making the CI fails, will try to enable it again.

3rdParty/MemPlumber/MemPlumber/CMakeLists.txt
clementperon: find_package(Backtrace) will use the FindBacktrace provide by CMake there is no FindBacktrace.cmake in our code

CMakeLists.txt
```diff
@@ -0,0 +1,199 @@
...
+include(VersionFromGit)
+
+# Get the git version
+version_from_git(LOG ON TIMESTAMP "%Y%m%d%H%M%S")
+
+project(
+  PcapPlusPlus
+  DESCRIPTION
+    "PcapPlusPlus is a multiplatform C++ library for capturing, parsing and crafting of network packets."
+  VERSION ${VERSION})
```
seladb, line 30: Does `${VERSION}` come from the `VersionFromGit` module?
The latest version is maintained in [`PcapPlusPlusVersion.h`](https://github.com/seladb/PcapPlusPlus/blob/master/Common%2B%2B/header/PcapPlusPlusVersion.h), however in order to make things simpler we can just hard-code it here, and maybe change `PcapPlusPlusVersion.h` to take the version from it?

Common++/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+  install(
+    TARGETS CommonPP
+    EXPORT PcapPlusPlusTargets
+    ARCHIVE DESTINATION ${PCAPP_INSTALL_LIBDIR}
+    LIBRARY DESTINATION ${PCAPP_INSTALL_LIBDIR}
+    PUBLIC_HEADER DESTINATION ${PCAPP_INSTALL_INCLUDEDIR})
+endif()
+
+set_property(TARGET CommonPP PROPERTY OUTPUT_NAME "common++")
+set_property(TARGET CommonPP PROPERTY VERSION ${VERSION_MAJOR}.${VERSION_MINOR})
```
seladb, line 48: Where do `${VERSION_MAJOR}` and `${VERSION_MINOR}` come from? PcapPlusPlus uses date-based versioning scheme so "major" and "minor" have less significance, it's better to use one variable for version.

Same in [`CMakeLists.txt`](https://github.com/clementperon/PcapPlusPlus/blob/cmake-ng/Packet%2B%2B/CMakeLists.txt) and [`CMakeLists.txt`](https://github.com/clementperon/PcapPlusPlus/blob/cmake-ng/Pcap%2B%2B/CMakeLists.txt)

Packet++/CMakeLists.txt
```diff
@@ -0,0 +1,128 @@
...
+  src/TelnetLayer.cpp
+  src/TextBasedProtocol.cpp
+  src/TLVData.cpp
+  src/UdpLayer.cpp
+  src/VlanLayer.cpp
+  src/VxlanLayer.cpp
+  # Force hash-library pcapng to be link fully static
+  $<TARGET_OBJECTS:hash-library>)
+
+set(public_headers
```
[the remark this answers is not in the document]
clementperon: this has been done is a another commit i have squashed them

Packet++/CMakeLists.txt
```diff
@@ -0,0 +1,128 @@
+add_library(
```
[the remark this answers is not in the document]
clementperon: this has been done is a another commit i have squashed them

Packet++/CMakeLists.txt
```diff
@@ -0,0 +1,128 @@
...
+  PacketPP
```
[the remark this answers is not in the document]
clementperon: if i remenver correctly i try to called it Packet++ the first time but it was not working with old version of CMake.

CMakeLists.txt
clementperon: Getting in from git allow you to know if the environement is clean or dirty.
And if you are on top of a version you get how many commit and the exact sha1 in the compiled executable.
Moreover the only human actions is now only to "git tag" which avoid human errors.

Common++/CMakeLists.txt
clementperon: Version is set based on git tag.

In case you in a dev branch or out of tree

version will be 22.05+g02b51784.20220731130355
And lib compile will get libPcap.so.22.05

clementperon: for official release
```
clement@Clement-Blade14:~/work/PcapPlusPlus/Dist$ git tag v23.01
clement@Clement-Blade14:~/work/PcapPlusPlus/Dist$ cmake ..
-- [MunkeiVersionFromGit] Version: 23.01
     Git tag:     [v23.01]
     Git branch:  [cmake-ng]
     Git hash:    []
     Decorated:   [v23.01]
     Identifiers: []
     Metadata:    []
     SemVer:      [23.01]
```

Examples/Tutorials/Tutorial-DpdkL2Fwd/Makefile
```diff
@@ -1,4 +1,8 @@
-include /usr/local/etc/PcapPlusPlus.mk
+PKGCONG := pkg-config
+
+PCAPPP_BUILD_FLAGS := "-fPIC"
+PCAPPP_INCLUDES := $(shell $(PKGCONF) --cflags PcapPlusPlus)
+PCAPPP_LIBS := $(shell $(PKGCONF) --libs PcapPlusPlus)
```
seladb, line 5: Why do we need to keep the makefiles for the tutorials?

Common++/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+add_library(
+  CommonPP
```
seladb, line 2: Why not calling it `Common++`?

Examples/ArpSpoofing/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+add_executable(ArpSpoofing main.cpp)
```
seladb, line 1: I think we should remove the Makefile from all the examples

Packet++/CMakeLists.txt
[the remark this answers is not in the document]
seladb: We anyway support CMake version 3.12 or higher, so I think it should work there

mk/setup_dpdk.py
```diff
@@ -1,1182 +0,0 @@
-#! /usr/bin/env python
```
seladb, line 1: We need to keep this file somehow in the new build system, it's the script to configure DPDK after installing

Tests/Fuzzers/Makefile
```diff
@@ -1,21 +0,0 @@
-include ../../mk/platform.mk
```
seladb, line 1: We removed the Makefile but didn't add `CMakeLists.txt` instead...

.github/workflows/build_and_test.yml
```diff
@@ -54,38 +73,28 @@ jobs:
...
+      - name: Build Tutorials w/ Make
         run: |
+          # Pkg-config on CentOS only look in /usr/share/pkgconfig and /usr/lib64/pkgconfig
+          export PKG_CONFIG_PATH=/usr/local/lib64/pkgconfig
+
           cd Examples/Tutorials/Tutorial-HelloWorld
           make -f Makefile.non_windows all
           ./Tutorial-HelloWorld
 
-  dpdk-legacy:
```
seladb, line 66: Why are we removing support for DPDK 18.11?

cmake/valgrind.cmake
```diff
@@ -0,0 +1,16 @@
+set(CTEST_SOURCE_DIRECTORY .)
```
seladb, line 1: What are we using valgrind for?

.cirrus.yml
```diff
@@ -21,72 +33,28 @@ task:
...
-  name: Regression Tests Linux (Undefined Behavior Sanitizer)
-  container:
-    image: seladb/ubuntu1804:latest
-  configure_script:
-    - ./configure-fuzzing.sh --sanitizer undefined
-    - ldconfig
-  build_script:
-    - make fuzzers
-  test_script:
-    - cd Tests/Fuzzers/RegressionTests && ./run_tests.sh
```
seladb, line 75: Why are we removing sanitizer tests?

.cirrus.yml
```diff
@@ -1,15 +1,27 @@
...
+    - pkg install -y libpcap bash gsed git gmake tcpreplay cmake
+  clone_script: |
+    if [ -z "$CIRRUS_PR" ]; then
+      git clone --recursive --branch=$CIRRUS_BRANCH https://x-access-token:${CIRRUS_REPO_CLONE_TOKEN}@github.com/${CIRRUS_REPO_FULL_NAME}.git $CIRRUS_WORKING_DIR
+      git reset --hard $CIRRUS_CHANGE_IN_REPO
+    else
+      git clone --recursive https://x-access-token:${CIRRUS_REPO_CLONE_TOKEN}@github.com/${CIRRUS_REPO_FULL_NAME}.git $CIRRUS_WORKING_DIR
+      git fetch origin pull/$CIRRUS_PR/head:pull/$CIRRUS_PR
+      git reset --hard $CIRRUS_CHANGE_IN_REPO
+    fi
```
seladb, line 20: I'm not sure I understand this change

.cirrus.yml
```diff
@@ -21,72 +33,28 @@ task:
...
     - GITHUB_TOKEN: ENCRYPTED[ab88249da2005c70f25f5910b08e907139f73bc0efb31a19b33f832e9e3bffe3cfc8e81932f21d7734c7aceca2a77188]
+  clone_script: |
+    if [ -z "$CIRRUS_PR" ]; then
+      git clone --recursive --branch=$CIRRUS_BRANCH https://x-access-token:${CIRRUS_REPO_CLONE_TOKEN}@github.com/${CIRRUS_REPO_FULL_NAME}.git $CIRRUS_WORKING_DIR
+      git reset --hard $CIRRUS_CHANGE_IN_REPO
+    else
+      git clone --recursive https://x-access-token:${CIRRUS_REPO_CLONE_TOKEN}@github.com/${CIRRUS_REPO_FULL_NAME}.git $CIRRUS_WORKING_DIR
+      git fetch origin pull/$CIRRUS_PR/head:pull/$CIRRUS_PR
+      git reset --hard $CIRRUS_CHANGE_IN_REPO
+    fi
```
seladb, line 51: ditto

.github/workflows/build_and_test.yml
```diff
@@ -17,28 +20,44 @@ jobs:
...
+            configure: cmake -DPCAPPP_BUILD_EXAMPLES=ON -DPCAPPP_BUILD_TESTS=ON -DLIGHT_PCAPNG_ZSTD=OFF -S . -B $BUILD_DIR
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: "0"
+
+      - name: Mark repository as safe (Checkout actions is not run with same UID)
```
seladb, line 51: I'm not sure I understand

Examples/Tutorials/Tutorial-DpdkL2Fwd/Makefile
clementperon: We can remove them, it's just an example how to build with make and the CI validate that it works fine.

.cirrus.yml
clementperon: If we get the tag from git we need to clone the repo with all the reference to get the proper tags

.github/workflows/build_and_test.yml
clementperon: Introduced with new version of git in Ubuntu 22.04
https://github.com/actions/checkout/issues/760

The docker is executed with a different user ID as the actions/checkout so the repository safe trig an error

Examples/ArpSpoofing/CMakeLists.txt
clementperon: removed

Common++/CMakeLists.txt
clementperon: It was not supported with previous version of cmake, but now that we require a recent version i have changed them

Tests/Fuzzers/Makefile
clementperon: Yes you're correct I will check how i can properly integrate Fuzzer and Asan to cmake.

clementperon:
I will keep the Branch as a reference but will close this MR in favor of #944

Issue #638, opened 2021-04-27
Changing / Crafting for IPv6 packets


jafar75:
Hi there, 

I was reading [tutorial ](https://pcapplusplus.github.io/docs/tutorials/packet-crafting)about changing and crafting packets. there is not any 'set' function for [IPv6Layer ](https://pcapplusplus.github.io/api-docs/classpcpp_1_1_i_pv6_layer.html)similar to IPv4Layer. 
How can I change some info like src/dst IP for a layer created by IPv6Layer?

regards

seladb:
Good catch! we should definitely add `set` method for `IPv6Layer`. Would you consider adding that and open a PR?

jafar75:
Sure, I'll try to provide a PR

seladb:
Thank you @jafar75 for your contribution to this project, much appreciated!

PR #654, opened 2021-05-22
Add setter for ip address (src/dst) for IPv6 layer

jafar75:
Adding 2 methods for setting ip address (src/dst)

### DIFF
--- a/Packet++/header/IPv6Layer.h
+++ b/Packet++/header/IPv6Layer.h
@@ -127,6 +127,19 @@ namespace pcpp
 		IPv6Address getSrcIPv6Address() const { return getIPv6Header()->ipSrc; }
 
 		/**
+		 * Set the source IP address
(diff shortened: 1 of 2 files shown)


jafar75:
#638

seladb (commented):
Do you think you can add tests to https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Packet%2B%2BTest/Tests/IPv6Tests.cpp ?

jafar75:
> Do you think you can add tests to https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Packet%2B%2BTest/Tests/IPv6Tests.cpp ?

@seladb

seladb:
@jafar75 I just noticed that the PR will be merged to `master`, can you please change it so it'll be merged to `dev`?

PR #655, opened 2021-05-25
add virtual IPLayer destructor, fixes #653

### LINKED ISSUE #653: Warnings about non-virtual destructor
In `IPLayer.h` and `IPv4Layer.h` I get the following warnings:

```
../include/pcapplusplus/IPLayer.h:22:8: warning: ‘class pcpp::IPLayer’ has virtual functions and accessible non-virtual destructor [-Wnon-virtual-dtor]
   22 |  class IPLayer
      |        ^~~~~~~
In file included from ../src/pcap_test.cpp:8:
../include/pcapplusplus/IPv4Layer.h:417:8: warning: base class ‘class pcpp::IPLayer’ has accessible non-virtual destructor [-Wnon-virtual-dtor]
  417 |  class IPv4Layer : public Layer, public IPLayer
      |        ^~~~~~~~~
```

These can be fixed by adding a virtual destructor in the class definition, i.e.:

```c++
virtual ~IPLayer() = default;
```

slhck:


seladb (commented):
- Please notice the build errors in CI: ``. I think this is because of the `default` keyword` 🤔 :
```
./header/IPLayer.h:42:24: warning: defaulted function definitions are a C++11 extension [-Wc++11-extensions]
                virtual ~IPLayer() = default;
                                     ^
In file included from src/EthLayer.cpp:4:
./header/IPv4Layer.h:417:8: error: exception specification of overriding function is more lax than base version
        class IPv4Layer : public Layer, public IPLayer
              ^
./header/IPLayer.h:42:11: note: overridden virtual function is here
                virtual ~IPLayer() = default;
                        ^
In file included from src/EthLayer.cpp:5:
./header/IPv6Layer.h:98:3: error: exception specification of overriding function is more lax than base version
                ~IPv6Layer();
                ^
./header/IPLayer.h:42:11: note: overridden virtual function is here
                virtual ~IPLayer() = default;
```
- Can you please change the PR so it will be merged to `dev` instead of a `master`?

slhck:
> * Please notice the build errors in CI: ``. I think this is because of the `default` keyword` 🤔 :


Hmm, I had tested this with my local checkout and there I'm using C++11 … I am not really an expert, and so I don't know what the change should be to accommodate older standards. Any idea?

> * Can you please change the PR so it will be merged to `dev` instead of a `master`?

Will do in case there is an easy solution for the above issue. Otherwise I suggest we just leave it as-is because it's too much of an effort for "just" a warning.

seladb:
> Hmm, I had tested this with my local checkout and there I'm using C++11 … I am not really an expert, and so I don't know what the change should be to accommodate older standards. Any idea?

I think you can try:
```cpp
virtual ~IPLayer() {}
```
And see if that works.

seladb:
> Changed like you suggested and created PR against dev.

Build is still failing, probably because you didn't add an empty implementation `{}`, you should add the empty curly brackets:
```cpp
virtual ~IPLayer() {}
```

slhck:
Thank you!

Issue #234, opened 2019-09-05
RawPacketsVector concurrent


KevinSmild:
Hi. I would like to thank you for this nice library, it's much more simpler and C++ish than npcap or libpcap.

I would like to ask a question. Is there any way of making the Async capture with a vector thread concurrent without changing the library code?

seladb:
thanks for the kind words :)

Could you please clarify your question? what do you mean by async capture with a vector thread?

KevinSmild:
Sorry. 
What I meant is if there is a way of making "Asynchronous packet capture using a packet list (vector)" thread safe.

KevinSmild:
I want to capture packets and process them at the same time (without stopping the capture). It would be nicer if I can process a bunch of packets at the same time instead of processing one by one. Hence, I would like to use the "Asynchronous packet capture using a packet list (vector)" capture mode of your library. 

That being said, I have (at least) two threads trying to use the same pcpp::RawPacketsVector, so I was wondering if there is a way of making it thread safe. Or do you have any other suggestion?

seladb:
Thanks for clarifying your question. If I understand your ask, you're currently using the API:

```cpp
PcapLiveDevice::startCapture(RawPacketVector& capturedPacketsVector)
```

And you're looking for a way to start using the packets in the `RawPacketVector` before stopping the capture.

Unfortunately `RawPacketVector` (which is based on `std::vector`) is not thread safe so this may not work. But it's should be pretty easy to make it thread safe by adding a locking mechanism to that class. Please let me know what you think

KevinSmild:
Thanks for your reply.
I'll work on that.

seladb:
Thanks. Please create a PR once you're done and I'll review and merge it

seladb:
hi @KevinSmild please let me know if you have the time to work on this

seladb:
@gx740 what do you think about this?

jafar75:
@seladb If it helps, i want to work on this issue and create PR if done.

seladb:
Sure @jafar75 you're more than welcome! Please go for it, thank you!

seladb:
To be honest, I don't think this is something worth investigating. I'll close this issue

muthusaravanan3186:
can i work on this issue? I need rawpointer vector to be thread safe

tigercosmos:
@muthusaravanan3186  I think it's easy to implement your own wrapper to achieve the goal.

seladb:
[...]

PR #662, opened 2021-06-06
thread-safe live capture

jafar75:
modified:   Pcap++/src/PcapLiveDevice.cpp

#234


jafar75:
I'll try to add some tests soon

seladb:
hi @jafar75 , I don't think this change resolves https://github.com/seladb/PcapPlusPlus/issues/234. The lock should be placed in the `RawPacketVector` class itself

PR #682, opened 2021-07-06
Added musl base flag. Alpine doesn't support backtrace

adubovikov:


seladb:
Thanks @adubovikov for your contribution!
I think a more standard way to support musl-based platforms would be to create a new `mk/platform.mk.musl` that adds `-DMUSL` to `GLOBAL_FLAGS`:

```makefile
# musl specific build flags
GLOBAL_FLAGS := -DMUSL
```

And then in `configure-linux.sh` just append the content of this file, same as we do with macOS M1:
https://github.com/seladb/PcapPlusPlus/blob/master/configure-mac_os_x.sh#L143-L148

seladb:
@adubovikov did you have a chance to look at my previous comment?

adubovikov:
@seladb yes, sorry, I was out these days and will do as you suggested.

adubovikov:
now should look better

.gitignore
```diff
@@ -38,6 +38,8 @@ mk/PcapPlusPlus.mk
 mk/platform.mk
 mk/PcapPlusPlus.mk-e
 3rdParty/LightPcapNg/zstd.mk
+3rdParty/MemPlumber/musl.mk
```
seladb, line 41: I don't think this is needed, is it?

3rdParty/MemPlumber/Makefile
```diff
@@ -8,7 +8,7 @@ include Makefile.defs
 
 %.o: %.cpp
 	@echo Building MemPlumber file: $<
-	@$(CC) $(GLOBAL_FLAGS) -Wall -O2 -g -c -o "$@" "$<"
+	@$(CC) $(GLOBAL_FLAGS) $(DEFS) -Wall -O2 -g -c -o "$@" "$<"
```
seladb, line 11: do we need to add `$(DEFS)` here?

.gitignore
adubovikov: sorry, yes, let me remove that

3rdParty/MemPlumber/Makefile
adubovikov: fixed.

seladb (approved):

PR #684, opened 2021-07-06
get the full timestamp resolution from a standard pcap file

ian-p-cooke:
pcap_open_offline_with_tstamp_precision() and pcap_fopen_offline_with_tstamp_precision() became available in libpcap release 1.5.1.

libpcap will read the magic number from the pcap file header and scale the timestamp up, down, or pass it through depending on how you opened it.  The existing method of opening a pcap file truncates the nanos from a pcap file even if the magic number indicates the scale is nanos.  If you open with PCAP_TSTAMP_PRECISION_NANO you can always use a timespec for setting RawPacket data.

this is pretty basic and I didn't find any tests that exercise this feature.  I have limited time but I will make any changes requested when possible; I don't want to maintain a fork for this feature.

### DIFF
--- a/Pcap++/src/PcapFileDevice.cpp
+++ b/Pcap++/src/PcapFileDevice.cpp
@@ -132,7 +132,11 @@ bool PcapFileReaderDevice::open()
 	}
 
 	char errbuf[PCAP_ERRBUF_SIZE];
+#if defined(PCAP_TSTAMP_PRECISION_NANO)
 	m_PcapDescriptor = pcap_open_offline_with_tstamp_precision(m_FileName.c_str(), PCAP_TSTAMP_PRECISION_NANO, errbuf);
+#else
+    m_PcapDescriptor = pcap_open_offline(m_FileName.c_str(), errbuf);
+#endif
 	if (m_PcapDescriptor == NULL)
 	{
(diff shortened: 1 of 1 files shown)


ian-p-cooke:
i see the symbols missing for Windows. I should be able to reproduce that.  Will post back with more info later.

seladb:
We should also think about how to support platforms that have older versions of libpcap

seladb:
@ian-p-cooke any update on this?
As I mentioned in my previous comment - We should also think about how to support platforms that have older versions of libpcap

ian-p-cooke:
I think we can address this by testing for PCAP_AVAILABLE_1_5 at compile time.  Both npcap and libpcap have defines for each version while winpcap does not.  I'll try that out.

ian-p-cooke:
> I think we can address this by testing for PCAP_AVAILABLE_1_5 at compile time.
>
the version of libpcap on my local host didn't have that define but I we can test for PCAP_TSTAMP_PRECISION_NANO instead.

ian-p-cooke:
I think the FreeBSD test failures are unrelated to my changes.

seladb:
> I think the FreeBSD test failures are unrelated to my changes.

Yes, there is an issue with FreeBSD build, I'll try to fix it

Issue #733, opened 2021-10-13
File-handle leak in lightpcapng


bbenne10:
While developing a program based on PcapPlusPlus's PcapSplitter example, we noticed that sufficiently large inputs caused some errors from LightPcapNG regarding the output file not being opened for writing. Upon investigating, we discovered a file handle leak in `light_pcapng_open_append`. In particular, [this line](https://github.com/seladb/PcapPlusPlus/blob/master/3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng_ext.c#L294) leaks the file handle that's allocated as part of `light_pcapng_open_read`. 

I've tested a small change that simply calls `light_close(pcapng->file)` before the following call to `light_open` and that caused the errors to disappear entirely for our particular problematic input. I'm unsure if this is the perfect solution for everyone, but I'm happy to put together a small PR for this if desired.

Additionally, there's a mistaken assumption that `light_open` always succeeds on line 297 in the same file. A simple call to the `DCHECK_NULLP` macro will allow a `return NULL` in that case (exactly as like 294 does). Happy to open a PR for this (as part of the same one that addresses the handle leak or separate) if that's also desired.

I know that this code doesn't technically belong to PcapPlusPlus (since it is 3rdParty) but the history suggests that contributors have made changes to this code in the past.

PR #734, opened 2021-10-13
Fix file handle leak in light_pcapng_open_append

bbenne10:
I mentioned in #733 that I'd wait for approval before opening the PR. Upon reading other tickets, I noticed a request for the reporter to open a PR, so that's what I am doing. Please feel free to merge just the first commit if you don't like the second (or vice versa).


bbenne10:
I am quite confused as to why the tests are failing on ubuntu 20.04 with dpdk 20.11. I don't think I've changed anything that should explicitly affect that particular test. If anyone has any knowledge, please chime in.

PR #757, opened 2021-11-18
Use CXXFLAGS when compiling hash library

catenacyber:
So as to fix oss-fuzz build

Without using these flags, std library for c++11 gets used when
compiling md5.cpp which is incompatible with the stdlib
used in the rest of libPacket++.a

Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28915


seladb:
@catenacyber thank you so much for the fix, much appreciated!

catenacyber:
You are welcome.

PR #766, opened 2021-12-22
make PcapLiveDevice constructor public

weyrick:
This PR makes `PcapLiveDevice` public for special use cases such as capturing from the same interface more than once (with different bpf filters). It does not change any default functionality: PcapLiveDeviceList still creates one PcapLiveDevice as before so is completely backwards compatible, however it _allows_ users to instantiate their own PcapLiveDevice if they wish.

This came out of a discussion on the mailing list: https://groups.google.com/g/pcapplusplus-support/c/cqyiiZaMCyk

[pktvisor](https://github.com/ns1labs/pktvisor) is using this change successfully in a fork, and I would like it to be merged upstream so that we can sync to the latest version.

### DIFF
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -109,8 +109,6 @@ namespace pcpp
 		bool m_CaptureCallbackMode;
 		LinkLayerType m_LinkType;
 
-		// c'tor is not public, there should be only one for every interface (created by PcapLiveDeviceList)
-		PcapLiveDevice(pcap_if_t* pInterface, bool calculateMTU, bool calculateMacAddress, bool calculateDefaultGateway);
 		// copy c'tor is not public
 		PcapLiveDevice( const PcapLiveDevice& other );
 		PcapLiveDevice& operator=(const PcapLiveDevice& other);
@@ -234,6 +232,11 @@ namespace pcpp
 			}
 		};
 
+		/**
+		 * Create new PcapLiveDevice. Typically handled by PcapLiveDeviceList but may be manually created for special use cases such
+		 * as capturing from the same interface more than once on platforms that support it
+		 */
+		PcapLiveDevice(pcap_if_t* pInterface, bool calculateMTU, bool calculateMacAddress, bool calculateDefaultGateway);
 
 		/**
 		 * A destructor for this class


Pcap++/header/PcapLiveDevice.h
```diff
@@ -234,6 +232,11 @@ namespace pcpp
 			}
 		};
 
+		/**
+		 * Create new PcapLiveDevice. Typically handled by PcapLiveDeviceList but may be manually created for special use cases such
+		 * as capturing from the same interface more than once on platforms that support it
+		 */
+		PcapLiveDevice(pcap_if_t* pInterface, bool calculateMTU, bool calculateMacAddress, bool calculateDefaultGateway);
```
seladb, line 239: @weyrick the reason this c'tor is private is because it requires `pcap_if_t` which is an internal libpcap structure. I looked in pktvisor and it seems it uses [libpcap's `pcap_findalldevs` API](https://github.com/ns1labs/pktvisor/blob/d421cd630b7fc377fbea6a9097cfc8044983fe9d/src/inputs/pcap/PcapInputStream.cpp#L164) to create `PcapLiveDevice` instances. I'm curious what is the use case to use libpcap directly and not use `PcapLiveDeviceList` 🤔

weyrick: @seladb yes the reason is just as it states in the comment there, and as we discussed in the group message I linked: there is a use case to capture multiple times on the same interface, especially when a different bpf filter is specified, so that the kernel can filter traffic efficiently. PcapPlusPlus makes the assumption that a device will only ever be opened and captured once, and this is an architectural problem for this use case.

seladb: @weyrick did you conclude that libpcap supports multiple concurrent captures? In the group message you mentioned that `pcap_t` might not be thread safe 🤔

weyrick: @seladb Right, the documentation isn't super clear but it's my belief that while `pcap_t` itself shouldn't be shared across threads, you can create `pcap_t` on the same interface in different threads just fine. This is the way I'm using it in pktvisor without a problem currently.

seladb:
@weyrick I may have a few ideas that will hide the underlying libpcap structs and calls, I'd love to hear your thoughts:
1. Clone a specific device - instead of making the `PcapLiveDevice` c'tor public, maybe we can add a `clone()` method to it, which will call `pcap_findalldevs()`, find the device name and then will create and return a  `PcapLiveDevice`. Here is how it can be used:
    ```cpp
    pcpp::PcapLiveDevice* myCloneDevice = myDevice->clone();
    ```
2. Clone all devices - we can add a `clone()` method to `PcapLiveDeviceList` which will create a non-singleton copy of it. Can be used like this:  
   ```cpp
   pcpp::PcapLiveDeviceList* myClonedDevList = pcpp::PcapLiveDeviceList::getInstance().clone();
   ```

weyrick:
> @weyrick I may have a few ideas that will hide the underlying libpcap structs and calls, I'd love to hear your thoughts:
> 
> 1. Clone a specific device - instead of making the `PcapLiveDevice` c'tor public, maybe we can add a `clone()` method to it, which will call `pcap_findalldevs()`, find the device name and then will create and return a  `PcapLiveDevice`. Here is how it can be used:
>    ```c++
>    pcpp::PcapLiveDevice* myCloneDevice = myDevice->clone();
>    ```

@seladb Yes this would satisfy our use case!

seladb:
@weyrick I'm closing this PR as @egecetin is implementing the `clone()` method in PR #775 (thank you so much @egecetin for working on this! 🙏 )

PR #769, opened 2021-12-30
External driver path for dpdk

egecetin:
Small contribution for dpdk. If the shared objects of external drivers can't found during the initialization DPDK returns 0 available devices. By using -d eal flag we can provide the target shared object or its path. I added the path as a last argument so it can be more backward compatible

### DIFF
--- a/Pcap++/header/DpdkDeviceList.h
+++ b/Pcap++/header/DpdkDeviceList.h
@@ -127,7 +127,7 @@ namespace pcpp
 		 * returned false it's impossible to use DPDK with PcapPlusPlus. You can get some more details about mbufs and pools in 
 		 * DpdkDevice.h file description or in DPDK web site
 		 */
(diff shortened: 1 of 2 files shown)


seladb:
@egecetin thanks for your contribution!

Please see my comment to this issue: https://github.com/seladb/PcapPlusPlus/issues/770#issuecomment-1008222525 , maybe instead of adding more parameters we can just add `argc` and `argv` arguments to [`DpdkDeviceList::initDpdk()`](https://github.com/seladb/PcapPlusPlus/blob/85e34dc165a7954d7aa28d4ccf56b49349b5cec5/Pcap%2B%2B/src/DpdkDeviceList.cpp#L69) so users can pass any additional arguments?

egecetin:
Updated my PR according to your comment. Argc and argv will be much more useful.

Issue #773, opened 2022-01-11
Alpine/Musl compilation not working


dehmarc:
Hello,

I used the latest version 21.11 as I saw that it now has an additional flag for musl.

**Steps to reproduce:**
Clone v21.11
`run configure-linux.sh --musl`

The following errors occur during build:
`3rdParty/MemPlumber/MemPlumber/memplumber.cpp:361:10: fatal error: execinfo.h: No such file or directory`
` 361 | #include <execinfo.h>`

**Analysis**
I checked the relevant files and it seems that MemPlumber [only looks at mk/platform.mk](https://github.com/seladb/PcapPlusPlus/blob/ddce58723a4ef7ab19bf1507e6323990b4d2d15e/3rdParty/MemPlumber/Makefile.defs#L5).
The problem is that the musl flag is only written to the [mk/PcapPlusPlus.m](https://github.com/seladb/PcapPlusPlus/blob/ddce58723a4ef7ab19bf1507e6323990b4d2d15e/configure-linux.sh#L343)

**Solution**
exemplary PR to fix the issue will be created separately. 
PR for reference #774

seladb:
Thanks for providing the fix @dehmarc !
I just merged your PR

PR #774, opened 2022-01-11
Fix missing musl flag in platform.mk

dehmarc:
Related issue has been created #773

### DIFF
--- a/configure-linux.sh
+++ b/configure-linux.sh
@@ -341,6 +341,7 @@ fi
 
 if [ -n "$MUSL" ]; then
    cat mk/platform.mk.musl >> $PCAPPLUSPLUS_MK
+   cat mk/platform.mk.musl >> $PLATFORM_MK
 fi
 
 # non-default libpcap include dir


seladb (approved):

clementperon:
@dehmarc could you add a Musl compilation in the CI or at least give me the instruction you do to build it with musl?

seladb:
@dehmarc  @clementperon we should probably also add this flag to the documentation:
https://pcapplusplus.github.io/docs/install/build-source/linux
https://github.com/PcapPlusPlus/pcapplusplus.github.io/blob/master/docs/install/linux.md

I thought I did but apparently I forgot...

dehmarc:
Do you prefer a native build in an alpine container (e.g. alpine:3.13) or cross crompile on ubuntu machine/image? 
Not sure what strategy you follow..

On a native alpine(=musl) build I required the following dependencies to be installed
`apk add git bash build-base libpcap-dev linux-headers`

other than that its just running
`configure-linux.sh --musl`
`make -j$(nproc)`

dehmarc:
> @dehmarc @clementperon we should probably also add this flag to the documentation: https://pcapplusplus.github.io/docs/install/build-source/linux https://github.com/PcapPlusPlus/pcapplusplus.github.io/blob/master/docs/install/linux.md
> 
> I thought I did but apparently I forgot...

yes definately! I guess I just searched the whole repo to check if you provide flags for a musl-based build

clementperon:
@dehmarc MUSL flag is just because there is no backtrace support on musl libc right?

what about installing libexecinfo on alpine? 

This can be easily detected by cmake with CHECK_INCLUDE_FILES()

dehmarc:
Then libexecinfo would need to be linked in case of musl. But it looks like libexecinfo provides the missing symbols (just checked the symbols, I have not tested it though)

seladb:
Backtrace is only needed for tests, so I wouldn't require users to install libexecinfo on musl just for that...
The flag is probably the easiest solution

PR #775, opened 2022-01-12
PcapLiveDevice/PcapLiveDeviceLists clone

egecetin:
PR for #766 

Still have work to do. I think test cases should be updated too. Any help/comment is welcomed.

### DIFF
--- a/Pcap++/header/PcapLiveDeviceList.h
+++ b/Pcap++/header/PcapLiveDeviceList.h
@@ -34,8 +34,6 @@ namespace pcpp
 		// private copy c'tor
 		PcapLiveDeviceList( const PcapLiveDeviceList& other );
 		PcapLiveDeviceList& operator=(const PcapLiveDeviceList& other);
-		// private d'tor
-		~PcapLiveDeviceList();
 
 		void init();
 
@@ -105,9 +103,18 @@ namespace pcpp
 		const std::vector<IPv4Address>& getDnsServers() const { return m_DnsServers; }
 
 		/**
+		 * Copies the current live devicelist
+		 * @return PcapLiveDeviceList* 
+		 */
+		PcapLiveDeviceList* clone();
+
+		/**
 		 * Reset the live device list and DNS server list, meaning clear and refetch them
 		 */
 		void reset();
+
+		// d'tor
+		~PcapLiveDeviceList();
 	};
 
 } // namespace pcpp
--- a/Pcap++/header/PcapLiveDeviceList.h
+++ b/Pcap++/header/PcapLiveDeviceList.h
@@ -103,7 +103,7 @@ namespace pcpp
 		const std::vector<IPv4Address>& getDnsServers() const { return m_DnsServers; }
 
 		/**
-		 * Copies the current live device list
+		 * Copies the current live device list 
 		 * @return A pointer to the cloned device list
 		 */
 		PcapLiveDeviceList* clone();
--- a/Pcap++/header/PcapLiveDeviceList.h
+++ b/Pcap++/header/PcapLiveDeviceList.h
@@ -103,8 +103,8 @@ namespace pcpp
 		const std::vector<IPv4Address>& getDnsServers() const { return m_DnsServers; }
 
 		/**
-		 * Copies the current live devicelist
-		 * @return PcapLiveDeviceList* 
+		 * Copies the current live device list
+		 * @return A pointer to the cloned device list
 		 */
 		PcapLiveDeviceList* clone();
 
--- a/Pcap++/header/PcapLiveDeviceList.h
+++ b/Pcap++/header/PcapLiveDeviceList.h
@@ -103,7 +103,7 @@ namespace pcpp
 		const std::vector<IPv4Address>& getDnsServers() const { return m_DnsServers; }
 
 		/**
-		 * Copies the current live device list 
+		 * Copies the current live device list
 		 * @return A pointer to the cloned device list
 		 */
 		PcapLiveDeviceList* clone();
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -105,66 +105,6 @@ PcapLiveDevice::PcapLiveDevice(pcap_if_t* pInterface, bool calculateMTU, bool ca
 		LOG_DEBUG("   Default Gateway: " << m_DefaultGateway);
 	}
 
-	//copy interface info
-	ifaceInfo = new pcap_if_t;
-	if(pInterface->name)
-	{
-		ifaceInfo->name = new char[strlen(pInterface->name) + 1];
-		strcpy(ifaceInfo->name, pInterface->name);
-	}
-	else
-		ifaceInfo->name = NULL;
-	if(pInterface->description)
-	{
-		ifaceInfo->description = new char[strlen(pInterface->description) + 1];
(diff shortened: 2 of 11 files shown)


seladb (commented):
Can you also add tests for these methods?

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -105,6 +105,66 @@ PcapLiveDevice::PcapLiveDevice(pcap_if_t* pInterface, bool calculateMTU, bool ca
...
+			memcpy(ifaceInfo->addresses->dstaddr, pInterface->addresses->dstaddr, sizeof(sockaddr));
+		}
+		else
+			ifaceInfo->addresses->dstaddr = NULL;
+	}
+	else
+		ifaceInfo->addresses = NULL;
+
+	ifaceInfo->next = NULL;
+	ifaceInfo->flags = pInterface->flags;
```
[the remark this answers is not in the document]
egecetin: Yeah it will works but will it really "clone"? For example during runtime, for some reason one of the addresses changed. Then if we call the clone() it will construct with new values. Same question valid for PcapLiveDeviceList too. Currently at the PR PcapDeviceList constructs the new values. I was planning to change that to the copy too but which one will be better? 

For tests I will try to add asap but it can take some time. Additionally I should prepare a windows environment for manual testing. I don't know it really works for reading multiple instances of same device. If it works maybe I can add an example code too.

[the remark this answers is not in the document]
seladb: I agree it might not be a perfect "clone", but in most cases devices don't change and even if they do, the clone should get the new params. If `clone()` can't find a device with the same name as the current device, it can return `NULL`.

Regarding tests - as mentioned in [this thread](https://groups.google.com/g/pcapplusplus-support/c/cqyiiZaMCyk) parallel packet capture from the same device might not be possible on all platforms, but "cloning" the device should be possible regardless capture is possible or not. It'd be great if you can test it on Windows, but even if you can't - for this PR it's enough that we can test the clone rather than capturing packets

egecetin:
@seladb i think there is a problem with windows release configuration in github actions. It can't find libzstd. Is there any change recently?

seladb:
@egecetin I'll check why the `zstd` build fails and keep you posted

seladb:
@egecetin I've disabled the zstd CI build for now so the Windows build should pass. Can you update your branch and try again?

egecetin:
Windows is ok, there is a just problem with ARM but it's caused by the server. I can't reach the requested URL from the browser too. I'll check later. 

` Could not resolve com.android.tools.build:gradle:4.1.2.
Required by:
         project :
      > Could not resolve com.android.tools.build:gradle:4.1.2.
         > Could not get resource 'https://jcenter.bintray.com/com/android/tools/build/gradle/4.1.2/gradle-4.1.2.pom'.
            > Could not GET 'https://jcenter.bintray.com/com/android/tools/build/gradle/4.1.2/gradle-4.1.2.pom'. Received status code 502 from server: Bad Gateway`

seladb:
> @seladb Windows is ok, there is a just problem with Android but it's caused by the server. I can't reach the requested URL from the browser too. I'll check later.
> 
> ` Could not resolve com.android.tools.build:gradle:4.1.2. Required by: project : > Could not resolve com.android.tools.build:gradle:4.1.2. > Could not get resource 'https://jcenter.bintray.com/com/android/tools/build/gradle/4.1.2/gradle-4.1.2.pom'. > Could not GET 'https://jcenter.bintray.com/com/android/tools/build/gradle/4.1.2/gradle-4.1.2.pom'. Received status code 502 from server: Bad Gateway`

Yes, I noticed it too, probably a temp issue with jcenter. Let's give it a couple of hours and try again?

egecetin:
I think it can be merged for now. 

As in your comment if parallel packet capture from the same device also possible on other platforms i will open another PR to support this.

egecetin:
@seladb Additionaly, I think you should check `Test Zstd MacOS` too :). It's waiting in queue forever.

Pcap++/header/PcapLiveDeviceList.h
```diff
@@ -104,10 +102,19 @@ namespace pcpp
 		 */
 		const std::vector<IPv4Address>& getDnsServers() const { return m_DnsServers; }
 
+		/**
+		 * Copies the current live devicelist
```
seladb, line 106: nit: `device list` instead of `devicelist`

Pcap++/header/PcapLiveDeviceList.h
```diff
@@ -104,10 +102,19 @@ namespace pcpp
 		 */
 		const std::vector<IPv4Address>& getDnsServers() const { return m_DnsServers; }
 
+		/**
+		 * Copies the current live devicelist
+		 * @return PcapLiveDeviceList*
```
seladb, line 107: I'd add more details:
```
@return A pointer to the cloned device list
```

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -370,6 +370,35 @@ void PcapLiveDevice::close()
...
+{
+	PcapLiveDevice *retval = NULL;
+
+	pcap_if_t *interfaceList;
+	char errbuf[PCAP_ERRBUF_SIZE];
+	int err = pcap_findalldevs(&interfaceList, errbuf);
+	if (err < 0)
+	{
+		LOG_ERROR("Error searching for devices: " << errbuf);
+	}
```
seladb, line 383: we should also add `return NULL` here

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -179,6 +179,20 @@ PTF_TEST_CASE(TestPcapLiveDeviceList)
...
+
+	std::vector<pcpp::PcapLiveDevice*> clonedDevListVector = clonedDevList->getPcapLiveDevicesList();
+	PTF_ASSERT_EQUAL(clonedDevListVector.size(), devList.size());	
+
+	std::vector<pcpp::PcapLiveDevice*>::iterator iterCloned = clonedDevListVector.begin();
+	for(std::vector<pcpp::PcapLiveDevice*>::iterator iter = devList.begin(); iter != devList.end(); iter++, iterCloned++)
+	{
+		PTF_ASSERT_EQUAL((*iter)->getName().length(), (*iterCloned)->getName().length());
+		PTF_ASSERT_BUF_COMPARE((*iter)->getName().c_str(), (*iterCloned)->getName().c_str(), (*iter)->getName().length());
+	}
```
seladb, line 193: Since `getName()` returns `std::string` we can make it more simple:
```cpp
PTF_ASSERT_EQUAL((*iter)->getName(), (*iterCloned)->getName());
```

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -248,6 +262,49 @@ PTF_TEST_CASE(TestPcapLiveDevice)
 
 
 
+PTF_TEST_CASE(TestPcapLiveDeviceClone)
```
seladb, line 265: Since this test requires a network connection it won't run in CI. Did you test it on your local machine?

egecetin: Yes, I checked with Ubuntu 20.04 WSL

PR #777, opened 2022-01-15
Bump to centos8

clementperon:
It will be easier to support CMake on CentOS8

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -14,7 +14,7 @@ jobs:
            python: python3.8
          - image: ubuntu1604
            python: python3.8
-         - image: centos7
+         - image: centos8
            python: python3.8
            test-flags: -s
          - image: fedora34


seladb:
@clementperon in order for the CI to work we need to create a docker image for CentOS 8 which is similar to this: https://github.com/seladb/PcapPlusPlus-DockerImages/blob/master/Dockerfile-centos7

Can you please create that image and open a PR in https://github.com/seladb/PcapPlusPlus-DockerImages?

seladb:
@clementperon please open PRs from `dev` instead of `master`

clementperon:
@seladb Ok will do, but IMO, the CI is the truth so if CI pass this mean branch is stable :)

seladb:
Yes I understand what you mean. `dev` is always a little behind `master` and is used as an extra precaution to make sure things work before merging to `master`. I try to sync `dev` to `master` very often

PR #778, opened 2022-01-16
Fix GCC warning

clementperon:
Some warning popup when compiling, let's try to fix them


Tests/Pcap++Test/Tests/FilterTests.cpp
```diff
@@ -757,7 +757,7 @@ PTF_TEST_CASE(TestPcapFilters_LinkLayer)
 			pcpp::BPFStringFilter bpfStringFilter("host " + ip4layer->getDstIPAddress().toString()); // checking against real filter, not the "" filter
 			if (bpfStringFilter.matchPacketWithFilter(*iter) && pcpp::IPcapDevice::matchPacketWithFilter(bpfStringFilter, *iter))
 			{
-				if((*iter)->getLinkLayerType(), pcpp::LINKTYPE_DLT_RAW1)
+				if((*iter)->getLinkLayerType() == pcpp::LINKTYPE_DLT_RAW1)
```
clementperon, line 760: @seladb I wonder if this is the correct fix as it trig the CI

clementperon: Debugging this 62 packets instead of 64

2 packets are marked with LINKTYPE_NULL

Open with wireshark and yes packet 3 and 4 are sent on Null/loopback interface

seladb (approved):

PR #781, opened 2022-01-18
CI: add visual studio 2022 and MacOS 11

clementperon:


clementperon:
@seladb seems that the executable built required MSVCR100.dll which is C++, which is now deprecated.

seladb:
> @seladb seems that the executable built required MSVCR100.dll which is now deprecated.
> 
> I don't see how to change microsoft project to update to a more recent target.

I'm not sure what you mean. Does the build fail because `MSVCR100.dll` is missing?

clementperon:
@seladb Yes, I'm not a Windows expert but MSVCR100.dll seems an old visual C++ libs not shipped in newer win11 platform no ?

seladb:
> @seladb Yes, I'm not a Windows expert but MSVCR100.dll seems an old visual C++ libs not shipped in newer win11 platform no ?
> 
> EDIT: to be more precise msvcr100.dll from VS C++ Redistributable 2010 which is no more supported.

PcapPlusPlus itself doesn't require `MSVCR100.dll`, but I think WinPcap does. Maybe try to build with Npcap instead?

clementperon:
@seladb, I will try with a new version of NPcap, is it possible to ask for an OEM license of NPCAP for the CI, they seems to offer it for open source project.

clementperon:
@seladb open the exe with dependency walker and indeed this MSVR100.dll is required by PTHREADVC2.dll ...

seladb:
> @seladb, I will try with a new version of NPcap, is it possible to ask for an OEM license of NPCAP for the CI, they seems to offer it for open source project.

I already have an OEM license for Npcap which is used in the CI, however the secrets are not available in forks which is why you can see them. If you want access to this license I can reach out to the Npcap team and ask if I can share it with you. Let's continue to discuss it over email: pcapplusplus@gmail.com

https://github.com/seladb/PcapPlusPlus/blob/master/.github/workflows/build_and_test.yml#L255-L256

seladb:
> @seladb open the exe with dependency walker and indeed this MSVR100.dll is required by PTHREADVC2.dll ...
> 
> The proper way can be to switch to C++11 Threads. The other way can be to switch to Pthreads4W lib Another solution could be to install MSCVR100 on this machine but not a verry good solution I think

you're right, now I remember that I once did this exercise myself... 
I do have plans to move PcapPlusPlus to C++11, and one of the reasons is to get rid of `pthread-win32` (there are other benefits of course 😃 ).
Do you think now is a good time to start this project or maybe release another version with the existing changes and then move?

clementperon:
@seladb I think moving to C++ 11 threads is not a be big deal, also this will make it easier to support CMake on windows.

Could you add the license as an encrypted value env in the CI ?

clementperon:
I will close this one until C++11 threads is properly supported.

PR #782, opened 2022-01-18
only define inet_pton() for old mingw + add mingw64 64bits support

clementperon:
To reduce the CMake MR.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -232,8 +232,15 @@ jobs:
       run: |
         cd Dist && ALL_LIBRARIES_WITHOUT_SYMLINKS=($(find . -name "*.a" -type f)) &&  for LIBRARY in ${ALL_LIBRARIES_WITHOUT_SYMLINKS[@]}; do echo "Checking for arm64 in $LIBRARY"; CHECK_ARCH=$(lipo $LIBRARY -verify_arch arm64 2>&1); RETVAL=$?; if [ $RETVAL -eq 1 ]; then echo "Couldn't find arm64 slice in $LIBRARY"; exit 1; else echo "Found arm64 slice in $LIBRARY"; fi; done
 
-  mingw:
+  mingw-w64:
     runs-on: windows-latest
+    strategy:
+      matrix:
+        include:
+         - env: i686
+           sys: mingw32
+         - env: x86_64
+           sys: mingw64
 
     steps:
     - name: Checkout code
@@ -242,8 +249,8 @@ jobs:
     - name: Setup MSYS2
       uses: msys2/setup-msys2@v2
       with:
-        msystem: MINGW32
-        install: mingw-w64-i686-gcc mingw-w64-i686-make
+        msystem: ${{matrix.sys}}
+        install: mingw-w64-${{matrix.env}}-gcc mingw-w64-${{matrix.env}}-make
 
     - name: Setup Python
       uses: actions/setup-python@v2
@@ -261,7 +268,7 @@ jobs:
     - name: Configure PcapPlusPlus
       working-directory: ${{env.GITHUB_WORKSPACE}}
       shell: msys2 {0}
-      run: ./configure-windows-mingw.bat mingw-w64 -m /mingw32 -w ${{ env.PCAP_SDK_DIR }} -s .
+      run: ./configure-windows-mingw.bat mingw-w64 -m  /${{ matrix.sys }} -w ${{ env.PCAP_SDK_DIR }} -s .
 
     - name: Build PcapPlusPlus
       working-directory: ${{env.GITHUB_WORKSPACE}}
--- a/Common++/header/IpUtils.h
+++ b/Common++/header/IpUtils.h
@@ -21,7 +21,11 @@
 
 /// @file
 
-#if defined(WIN32) && !defined(_MSC_VER)
+// Both Visual C++ Compiler and MinGW-w64 define inet_ntop() and inet_pton()
+// Add compatibility functions for old MinGW (aka MinGW32)
+// We use "__MINGW64_MAJOR_VERSION" and not __MINGW64__ to detect MinGW-w64 compiler
+// because the second one is not defined for MinGW-w64 in 32bits mode
+#if defined(WIN32) && !defined(_MSC_VER) && !defined(__MINGW64_MAJOR_VERSION)
 /**
  * Convert a network format address to presentation format.
  * @param[in] af Address family, can be either AF_INET (IPv4) or AF_INET6 (IPv6)
--- a/Common++/src/IpUtils.cpp
+++ b/Common++/src/IpUtils.cpp
@@ -62,7 +62,8 @@ namespace pcpp
 	} // namespace internal
 } // namespace pcpp
 
-#if defined(WIN32) && !defined(_MSC_VER)
+// Only MinGW32 doesn't have these functions (not MinGW-w64 nor Visual C++)
+#if defined(WIN32) && !defined(_MSC_VER) && !defined(__MINGW64_MAJOR_VERSION)
 /* const char *
  * inet_ntop4(src, dst, size)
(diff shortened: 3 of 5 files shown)


clementperon:
I got this issue when compiling with minGW64. This header was already set in winsock2.h and differs from this one.

I don't see a reason to keep these..

seladb:
The reason we have this implementation of `inet_pton` in `IpUtills` is to support [MinGW](https://osdn.net/projects/mingw/). 

MinGW is the original project and MinGW-w64 is actually a fork of it. You can read about it [here](https://stackoverflow.com/questions/25582110/what-is-the-difference-between-mingw-mingw-w64-and-mingw-builds).

Unfortunately MinGW doesn't have an implementation of  `inet_pton` and `inet_ntop` so the build fails:

```shell
C:\seladb\temp\pr782\PcapPlusPlus>mingw32-make
==== Building target: Common++ ====
Building file: src/Logger.cpp
Building file: src/IpUtils.cpp
src/IpUtils.cpp: In function 'void pcpp::internal::sockaddr2string(sockaddr*, char*)':
src/IpUtils.cpp:45:5: error: 'inet_ntop' was not declared in this scope
     inet_ntop(AF_INET, &(((sockaddr_in*)sa)->sin_addr), resultString, INET_ADDRSTRLEN);
     ^~~~~~~~~
src/IpUtils.cpp:45:5: note: suggested alternative: 'inet_ntoa'
     inet_ntop(AF_INET, &(((sockaddr_in*)sa)->sin_addr), resultString, INET_ADDRSTRLEN);
     ^~~~~~~~~
     inet_ntoa
src/IpUtils.cpp:50:5: error: 'inet_ntop' was not declared in this scope
     inet_ntop(AF_INET6, &(((sockaddr_in6*)sa)->sin6_addr), resultString, INET6_ADDRSTRLEN);
     ^~~~~~~~~
src/IpUtils.cpp:50:5: note: suggested alternative: 'inet_ntoa'
     inet_ntop(AF_INET6, &(((sockaddr_in6*)sa)->sin6_addr), resultString, INET6_ADDRSTRLEN);
     ^~~~~~~~~
     inet_ntoa
Makefile:51: recipe for target 'Obj/IpUtils.o' failed
mingw32-make[1]: *** [Obj/IpUtils.o] Error 1
Makefile:90: recipe for target 'libs' failed
mingw32-make: *** [libs] Error 2
```

clementperon:
Ok I will change the define to detect mingw and not mongw-w64. if mingw support still a requirement

.github/workflows/build_and_test.yml
```diff
@@ -232,8 +232,15 @@ jobs:
...
-  mingw:
+  mingw-w64:
     runs-on: windows-latest
+    strategy:
+      matrix:
+        include:
+         - env: i686
+           sys: mingw32
+         - env: x86_64
+           sys: mingw64
```
seladb, line 243: I'm not sure this is a valid configuration because WinPcap only supports 32-bit in MinGW (as mentioned [here](https://pcapplusplus.github.io/docs/install/build-source/mingw#additional-prerequisites))

mk/PcapPlusPlus.mk.mingw-w64
```diff
@@ -4,7 +4,7 @@
 PCAPPP_INCLUDES += -I$(PCAP_SDK_HOME)/Include
 
 # libs dir
-PCAPPP_LIBS_DIR += -L$(PCAP_SDK_HOME)/lib -L$(MINGW_HOME)/lib
+PCAPPP_LIBS_DIR += -L$(PCAP_SDK_HOME)/lib -L$(PCAP_SDK_HOME)/lib/x64 -L$(MINGW_HOME)/lib
```
seladb, line 7: it might be a little bit confusing but MinGW-w64 doesn't mean "MinGW in 64-bit". It's actually a different fork of MinGW and supports both 32-bit and 64-bit. But because WinPcap doesn't support 64-bit, only 32-bit is actually supported

Common++/src/IpUtils.cpp
```diff
@@ -62,7 +62,8 @@ namespace pcpp
 	} // namespace internal
 } // namespace pcpp
 
-#if defined(WIN32) && !defined(_MSC_VER)
+// Only MinGW doesn't support WinSock2 (not MinGW-w64)
+#if defined(__MINGW32__) && !defined(__MINGW64__)
```
seladb, line 66: ditto

Common++/header/IpUtils.h
```diff
@@ -21,7 +21,8 @@
 
 /// @file
 
-#if defined(WIN32) && !defined(_MSC_VER)
+// Only MinGW doesn't support WinSock2 (not MinGW-w64)
+#if defined(__MINGW32__) && !defined(__MINGW64__)
```
seladb, line 25: ditto

.github/workflows/build_and_test.yml
clementperon: In this build we are linking with NPcap not WinPcap.

Winpcap seems no more maintained regarding this 
https://npcap.com/vs-winpcap.html

Common++/header/IpUtils.h
clementperon: https://sourceforge.net/p/predef/wiki/Compilers/

Yes

#if defined(__MINGW32__) && !defined(__MINGW64_VERSION_MAJOR) ?

mk/PcapPlusPlus.mk.mingw-w64
clementperon: This will make the linker found the correct library in mingw-w64 64bits mode, but I agree that only Npcap will be supported in mingw-w64_x86_64

Common++/header/IpUtils.h
clementperon: this should match only mingw-w32 but not mingw-w64_x86_64 or mingw-w64_i686.

What do you think ?

.github/workflows/build_and_test.yml
seladb: oh, you're right!

Common++/header/IpUtils.h
seladb: What about:
```c++
#if defined(WIN32) && !defined(_MSC_VER) && !defined(__MINGW64__)
```

mk/PcapPlusPlus.mk.mingw-w64
seladb: I'm not sure how it could work... how does the linker know which library to choose from?

Common++/header/IpUtils.h
clementperon: This will not work for MinGW-w64 32 Bit

clementperon: #if defined(WIN32) && !defined(_MSC_VER) && !defined(__MINGW64_MAJOR_VERSION)

mk/PcapPlusPlus.mk.mingw-w64
clementperon: Yes linker first try to link with a 32bits library and return a warning because arch doesn't match and they try to link with the 64 one

Common++/header/IpUtils.h
seladb: yes you're right! your idea is better

mk/PcapPlusPlus.mk.mingw-w64
clementperon: D:/a/_temp/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: skipping incompatible C:/Npcap-sdk/lib\wpcap.lib when searching for -lwpcap

seladb: yes, I just saw it in the CI. Looks good to me!

seladb (approved):

PR #788, opened 2022-01-23
NTP protocol support

egecetin:


3rdParty/EndianPortable/include/EndianPortable.h
```diff
@@ -67,6 +67,10 @@
 #elif defined(WIN32) || defined(WINx64)
 #include <winsock2.h>
  #if(BYTE_ORDER == LITTLE_ENDIAN)
+ #if defined(PCAPPP_MINGW_ENV)
```
clementperon, line 70: IMO PCAPPP_MINGW_ENV is a bit ambigous.

```
#ifndef htonll
  #define htonll(x) ((1==htonl(1)) ? (x) : (((uint64_t)htonl((x) & 0xFFFFFFFFUL)) << 32) | htonl((uint32_t)((x) >> 32)))
#endif // htonll
#ifndef ntohll
  #define ntohll(x) ((1==ntohl(1)) ? (x) : (((uint64_t)ntohl((x) & 0xFFFFFFFFUL)) << 32) | ntohl((uint32_t)((x) >> 32)))
#endif // ntohll
```

And also this should be defined out of BYTE_ORDER == LITTLE_ENDIAN no ?

egecetin: Updated, thanks

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,456 @@
...
+         * @return std::string ISO8601 formatted string
+         */
+        static std::string convertToIsoFormat(const double timestamp);
+
+        /**
+         * A static method to convert timestamp value to ISO8601 date time format
+         * @param[in] timestampInNTPformat Value in NTP timestamp format
+         * @return std::string ISO8601 formatted string
+         */
+        static std::string convertToIsoFormat(const uint64_t timestampInNTPformat);
```
seladb, line 425: I'm thinking that since these methods are not unique to the NTP protocol, maybe we can move them to [`GeneralUtils.h`](https://github.com/seladb/PcapPlusPlus/blob/master/Common%2B%2B/header/GeneralUtils.h) and then provide methods like `getTransmitTimestampAsString()`, `getReceiveTimestampAsString()`, etc. WDYT?

Tests/Packet++Test/Tests/NtpTests.cpp
```diff
@@ -0,0 +1,158 @@
...
+#else
+    PTF_ASSERT_EQUAL(pcpp::NtpLayer::convertToIsoFormat(0.0), std::string("1970-01-01T00:00:00.000000000Z"));
+    PTF_ASSERT_EQUAL(pcpp::NtpLayer::convertToIsoFormat(uint64_t(0)), std::string("1900-01-01T00:00:00.000000000Z"));
+#endif
+
+
+    PTF_ASSERT_EQUAL(pcpp::NtpLayer::convertToIsoFormat(1642879410.0), "2022-01-22T19:23:30.000000000Z");
+    PTF_ASSERT_EQUAL(pcpp::NtpLayer::convertToIsoFormat(pcpp::NtpLayer::convertToTimestampFormat(1642879410.0)), "2022-01-22T19:23:30.000000000Z");
+
+} // NtpMethodsTests
```
seladb, line 31: If we move these methods to `GeneralUtils.h`  we can probably create a new test file `Packet++Test/Tests/GeneralUtilsTests.cpp` and move this test case there

Packet++/src/NtpLayer.cpp
```diff
@@ -0,0 +1,516 @@
...
+        return false;
+    }
+
+    std::string NtpLayer::toString() const
+    {
+        std::stringstream ss;
+
+        ss << "NTP Layer, ";
+        ss << "Version: " << (int)getVersion() << ", Mode: " << (int)getMode() << ", ";
+        ss << "Leap Indicator: " << (int)getLeapIndicator() << ", Stratum: " << (int)getStratum();
```
seladb, line 512: maybe to keep this string short we can do something similar to Wireshark:
`NTP Layer v4, mode: client`

Tests/Packet++Test/Tests/NtpTests.cpp
```diff
@@ -0,0 +1,158 @@
...
+    PTF_ASSERT_EQUAL(ntpLayer->getReceiveTimestamp(), be64toh(0xd944575531b4e978));
+    PTF_ASSERT_EQUAL(ntpLayer->getTransmitTimestamp(), be64toh(0xd94f51f42d26e2f4));
+
+    // Test Ipv6
+    READ_FILE_AND_CREATE_PACKET(2, "PacketExamples/ntpv4Ipv6_withAuth.dat");
+
+    ntpPacket = pcpp::Packet(&rawPacket2);
+    ntpLayer = ntpPacket.getLayerOfType<pcpp::NtpLayer>();
+
+    PTF_ASSERT_NOT_NULL(ntpLayer);
```
seladb, line 96: we don't have a test for `ntpLayer->toString()`

Packet++/src/NtpLayer.cpp
```diff
@@ -0,0 +1,516 @@
...
+    std::string NtpLayer::convertToIsoFormat(const uint64_t timestampInNTPformat)
+    {
+        return convertToIsoFormat(convertFromTimestampFormat(timestampInNTPformat));
+    }
+
+    bool NtpLayer::isDataValid(const uint8_t *data, size_t dataSize)
+    {
+        if (data && dataSize >= sizeof(ntp_header))
+            return true;
+        return false;
```
seladb, line 503: a shorter way could be:
```cpp
return data && dataSize >= sizeof(ntp_header);
```

3rdParty/EndianPortable/include/EndianPortable.h
```diff
@@ -207,4 +207,11 @@
 #define htole32(x) le32toh(x)
 #endif
 
+#ifndef htonll 
+#define htonll(x) ((1==htonl(1)) ? (x) : (((uint64_t)htonl((x) & 0xFFFFFFFFUL)) << 32) | htonl((uint32_t)((x) >> 32))) 
+#endif // htonll 
+#ifndef ntohll 
+#define ntohll(x) ((1==ntohl(1)) ? (x) : (((uint64_t)ntohl((x) & 0xFFFFFFFFUL)) << 32) | ntohl((uint32_t)((x) >> 32))) 
+#endif // ntohll
```
seladb, line 215: I didn't see that we use it anywhere in `NtpLayer`, why add it in this PR?

Tests/Packet++Test/Tests/NtpTests.cpp
```diff
@@ -0,0 +1,158 @@
...
+    PTF_ASSERT_EQUAL(ntpLayer->getRootDispersion(), 0);
+    PTF_ASSERT_EQUAL(ntpLayer->getReferenceIdentifier(), 0);
+    PTF_ASSERT_EQUAL(ntpLayer->getReferenceTimestamp(), 0);
+    PTF_ASSERT_EQUAL(ntpLayer->getOriginateTimestamp(), 0);
+    PTF_ASSERT_EQUAL(ntpLayer->getReceiveTimestamp(), be64toh(0xd94f4f1100000000));
+    PTF_ASSERT_EQUAL(ntpLayer->getTransmitTimestamp(), be64toh(0xd94f4f1100000000));
+
+} // NtpParsingV3Tests
+
+PTF_TEST_CASE(NtpParsingV4Tests)
```
seladb, line 62: If you want to find NTP packet that have more diverse information you can see here: https://weberblog.net/wp-content/uploads/2018/08/NTP-IPv6-with-Auth.zip

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,604 @@
...
+      .                                                               .
+      |                                                               |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+      |                          Key Identifier                       |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+      |                                                               |
+      |                   dgst (128 for v4, 64 for v3)                |
+      |                                                               |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+     * @endverbatim
```
seladb, line 81: This is nice!
However, what I meant in https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r791472635 is that it'd be good to document each field in each struct, like we do here:
https://github.com/seladb/PcapPlusPlus/blob/master/Packet++/header/ArpLayer.h#L24-L41

That way Doxygen will create an documentation per field like this:
https://pcapplusplus.github.io/api-docs/structpcpp_1_1arphdr.html

Packet++/src/NtpLayer.cpp
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+    void NtpLayer::setPrecision(int8_t val)
+    {
+        ((ntp_header *)m_Data)->precision = val;
+    }
+
+    double NtpLayer::getPrecisionInSecs() const
+    {
+        return pow(2, getPrecision());
```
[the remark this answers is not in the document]
seladb: You didn't change anything in the code, can you please elaborate one what this calculation does?

Packet++/src/NtpLayer.cpp
seladb: Right now it returns an empty string:

```cpp
std::string NtpLayer::toString() const
{
    return std::string();
    //std::stringstream ss;

    //ss << "NTP Layer v" << (int)getVersion() << ", Mode: " << getModeString();

    //return ss.str();
}
```

I guess we should fix it?

seladb:
@egecetin as a general note I wanted to thank you for adding NTP support, it's a great contribution to this project! 🙏 🙏

egecetin:
@seladb During the latest commit somehow `dhcpv6creation` and test with related pcap write failed for windows. I will temporarily disable them until the update `NtpLayer` comments/reviews. Please do not merge until fix the issue.

seladb:
> @seladb During the latest commit somehow `dhcpv6creation` and test with related pcap write failed (actually hangs during test run, possibly segv etc.) for windows. I will temporarily disable them until the update `NtpLayer` comments/reviews. Please do not merge until fix the issue.

Sure @egecetin , sounds good 👍

seladb:
@egecetin I see you're still working on some changes. Let me know when the PR is ready for re-review and I'll take another look

egecetin:
@seladb I think I'm done for now. Can you check again for anything i missed?

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,649 @@
...
+#include <stdlib.h>
+
+/// @file
+
+/// 2^16 as a double
+#define NTP_FRIC 65536.
+/// 2^32 as a double
+#define NTP_FRAC 4294967296.
+/// Epoch offset between Unix time and NTP
+#define EPOCH_OFFSET 2208988800ULL
```
seladb, line 21: We can move these consts to `NtpLayet.cpp`, they're not used in the header file

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,649 @@
...
+
+#include "Logger.h"
+#include "IPv4Layer.h"
+#include "Layer.h"
+
+#include "GeneralUtils.h"
+#include "SystemUtils.h"
+
+#include <math.h>
+#include <stdlib.h>
```
seladb, line 12: I think we can move all of these includes to `NtpLayer.cpp` as they're not used in the header file

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,649 @@
+#ifndef PACKETPP_NTP_LAYER
+#define PACKETPP_NTP_LAYER
+
+#include "Logger.h"
+#include "IPv4Layer.h"
```
seladb, line 5: Is this header used somewhere?

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,649 @@
...
+		 */
+        static bool isNTPPort(uint16_t port) { return port == 123; }
+
+        // overridden methods
+
+        void parseNextLayer() {}
+
+        size_t getHeaderLen() const { return m_DataLen; }
+
+        void computeCalculateFields() {}
```
seladb, line 640: Let's add a documentation for these methods also

egecetin:
@seladb Let's continue from here to follow up easily. 

Updated regarding your reviews. Except this,

> We don't have tests for all the `set...InSecs` methods (`setReferenceTimestampInSecs`, `setReceiveTimestampInSecs`, `setRootDelayInSecs`, `setRootDispersionInSecs`). Maybe the easiest way would be to add to `NtpCraftingTests` crafting of a NTPv3 packet which uses these methods 🤔

_Originally posted by @seladb in https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r795163539_

And my comment,

> 
> @seladb set....InSecs methods have some fundamental problem. Since they are using floating point, same as get...InSecs functions their result may or may not equal to provided value so the tests can fail. Because of this i lower the precision of this to 4 digits too. Because it can change sometimes from platform to platform too.
> 
> Because of the same problem NTP timestamp format interprets integer and fraction parts separately. I can replace these double function calls with structs,
> 
> struct X
> {
> uint32_t integer,
> float fraction
> }
> but this means they will equal to plain set....() methods because it is actually NTP Timestamp format.
> 
> Do you have any suggestions for to test these methods?

egecetin:
@seladb And thanks for your reviews🙏

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,649 @@
...
+	 */
+    class NtpLayer : public Layer
+    {
+    private:
+        ntp_header *getNtpHeader() const { return (ntp_header *)m_Data; }
+
+    public:
+        /**
+		 * A constructor that creates the layer from an existing packet raw data
+		 * @param[in] data A pointer to the raw data (will be casted to @ref ntphdr)
```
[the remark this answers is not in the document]
seladb: Now that `ntp_header` is private we need to remove this part: `(will be casted to @ref ntp_header)` because:
1. The user doesn't know what is `ntp_header` (this struct isn't exposed to him)
2. Doxygen gives an error because this struct is private

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,615 @@
...
+        enum Mode
+        {
+            Reserved = 0,
+            SymActive,
+            SymPassive,
+            Client,
+            Server,
+            Broadcast,
+            Control,
+            PrivateUse
```
seladb, line 178: These enum values are public and not documented, hence Doxygen can't document them:
image

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,615 @@
...
+{
+    /**
+	 * @class NtpLayer
+	 * Represents a NTP (Network Time Protocol) layer
+     * 
+     * @brief The NTP packet consists of an integral number of 32-bit (4 octet) words in network byte order. 
+     * The packet format consists of three components: the header itself, one or more optional extension fields (for v4), 
+     * and an optional message authentication code (MAC). Currently the extension fields are not supported. The NTP header is:
+     * 
+     * @verbatim
```
seladb, line 27: nit: this asterisk (`*`) is redundant:
image

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,615 @@
...
+      .                                                               .
+      |                                                               |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+      |                          Key Identifier                       |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+      |                                                               |
+      |                   dgst (128 for v4, 64 for v3)                |
+      |                                                               |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+     * @endverbatim
```
seladb, line 73: nit: this asterisk (`*`) is also redundant:
image

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,615 @@
...
+        /**
+         * Set the value of poll interval
+         * @param[in] val Poll interval in log2 seconds
+         */
+        void setPollInterval(int8_t val);
+
+        /**
+         * @return The value of poll interval in seconds
+         */
+        double getPollIntervalInSecs() const;
```
seladb, line 332: We don't have tests for this method

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,649 @@
...
+         * Get the value of reference timestamp
+         * @return Value in seconds from Unix Epoch (1 Jan 1970)
+         */
+        double getReferenceTimestampInSecs() const;
+
+        /**
+         * Set the value of reference timestamp
+         * @param[in] val Value in seconds from Unix Epoch (1 Jan 1970)
+         */
+        void setReferenceTimestampInSecs(double val);
```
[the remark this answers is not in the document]
seladb: > @seladb Let's continue from here to follow up easily. Searching from reviews a bit confusing
> 
> Updated regarding your reviews. Except this,
> 
> > We don't have tests for all the `set...InSecs` methods (`setReferenceTimestampInSecs`, `setReceiveTimestampInSecs`, `setRootDelayInSecs`, `setRootDispersionInSecs`). Maybe the easiest way would be to add to `NtpCraftingTests` crafting of a NTPv3 packet which uses these methods 🤔
> 
> _Originally posted by @seladb in [#788 (comment)](https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r795163539)_
> 
> And my comment,
> 
> > @seladb set....InSecs methods have some fundamental problem. Since they are using floating point, same as get...InSecs functions their result may or may not equal to provided value so the tests can fail. Because of this i lower the precision of this to 4 digits too. Because it can change sometimes from platform to platform too.
> > Because of the same problem NTP timestamp format interprets integer and fraction parts separately. I can replace these double function calls with structs,
> > struct X
> > {
> > uint32_t integer,
> > float fraction
> > }
> > but this means they will equal to plain set....() methods because it is actually NTP Timestamp format.
> > Do you have any suggestions for to test these methods?

If I understand correctly the value may change from platform to platform because the accuracy of `double` varies between platforms, is that correct?
In that case, can we use an input value that is low enough and will be supported on all platforms?

seladb (commented):
@egecetin I marked all resolved comments as "resolved". So all the ones that are still open are left

seladb:
For your convenience, here are the comments currently open:
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r795161351
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r796291181
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r795162934
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r796297936
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r795163919
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r795164057
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r796292108
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r796292770
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r796293014
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r796294877

egecetin:
I was trying to understand why you marking `EndianPortable.h` over and over. Then i noticed some of my comments are waiting at pending 🙄.

3rdParty/EndianPortable/include/EndianPortable.h
seladb: ok got it, thanks!

Packet++/header/NtpLayer.h
seladb: In that case, I think we can move `#include "IPv4Layer.h"` to `NtpLayer.cpp` 🤔

Packet++/header/NtpLayer.h
```diff
@@ -0,0 +1,649 @@
...
+        Broadcast,
+        Control,
+        PrivateUse
+    };
+
+    /**
+     * 32-bit code identifying the particular server or reference clock. 
+     * The interpretation depends on the value in the stratum field.
+     */
+    enum NTPClockSource
```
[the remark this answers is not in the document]
seladb: The argument of the `setReferenceIdentifier()` method is `uint32_t`. If we think it should be both IPv4 or `ClockSource`, why not create 2 overloads of this method - one that takes an `IPv4Address` as an argument and the other that takes `ClockSource` as an argument?

Packet++/header/NtpLayer.h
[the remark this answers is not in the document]
seladb: We still need to fix this (doxygen give an error)

seladb:
@egecetin thank you for all of your efforts!!

Here are the comments that remain open:
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r801323417
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r801315931
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r795163919
- https://github.com/seladb/PcapPlusPlus/pull/788#discussion_r801314527

Once we fix them I think we can merge the PR

Packet++/header/NtpLayer.h
egecetin: I already moved them. Probably you checked from outdated version

seladb: I still see:

```cpp
#include "GeneralUtils.h"
#include "SystemUtils.h"
```

In the header file, but I think they're only used in `NtpLayer.cpp` 🤔

egecetin: Understand now. I thought you comment for math.h

Packet++/header/NtpLayer.h
egecetin: Replaced with `#include "IpAddress.h"` because of `void setReferenceIdentifier(IPv4Address val);` since we only need `IPv4Address`

seladb (approved):

seladb:
thank you so much for all of your hard work @egecetin ! I really appreciate your efforts and I'm sorry for the time it took to review and merge this PR

PR #790, opened 2022-01-24
cirrus-ci: add FreeBSD 13 and bump FreeBSD 12 to 12.3

clementperon:


.cirrus.yml
```diff
@@ -3,7 +3,8 @@ task:
   freebsd_instance:
     matrix:
       image_family: freebsd-11-4
-      image_family: freebsd-12-2
+      image_family: freebsd-12-3
+      image_family: freebsd-13-0
```
seladb, line 7: I'm wondering if we should be running 3 versions of FreeBSD in our CI, isn't it too much?

clementperon: If you agree we can drop 11-4 which is no more maintained

seladb: Sure, let's do that!

clementperon: done

PR #791, opened 2022-01-24
Only use compiler defines

clementperon:
@seladb I think it's proper to use `_WIN32` and maybe we should also switch `__linux` and `__APPLE__`.

This defines are set by the compiler so we don't need to add it to Makefile or Cmake build system.
Also if someone want to include a PcapPlusPlus header. He doesn't need to define special macro.

What do you think ? 

From:
https://web.archive.org/web/20190409055410/http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system


mk/vs/Common++.vcxproj.template
```diff
@@ -147,7 +147,7 @@ xcopy "$(PcapPlusPlusHome)\Common++\header\*" "$(PcapPlusPlusHome)\Dist\header"
       <PrecompiledHeader>NotUsing</PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>WINx64;MSVC;_CRT_SECURE_NO_WARNINGS;GIT_BRANCH="$(GitBranch)";GIT_COMMIT="$(GitCommit)"</PreprocessorDefinitions>
```
seladb, line 150: I'm don't remember adding these `WINx64` and `MSVC` preprocessor definitions, I think VS adds them by default

mk/vs/Common++.vcxproj.template
```diff
@@ -121,7 +121,7 @@ xcopy "$(PcapPlusPlusHome)\Common++\header\*" "$(PcapPlusPlusHome)\Dist\header"
       <Optimization>Full</Optimization>
       <FunctionLevelLinking>true</FunctionLevelLinking>
       <IntrinsicFunctions>true</IntrinsicFunctions>
-      <PreprocessorDefinitions>WIN32;NDEBUG;_LIB;_CRT_SECURE_NO_WARNINGS;GIT_BRANCH="$(GitBranch)";GIT_COMMIT="$(GitCommit)";%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
seladb, line 124: Same for `WIN32`

mk/vs/Common++.vcxproj.template
clementperon: Maybe it's the default when creating a MS project. But this can be an issue if someone doesn't use Visual Code and include the Pcap++ header in his project the compiler will not define "WINx64" or "WIN32".

So I think it's better to not rely on it, What do you think?

seladb: Yes, I agree we don't need to rely on it, but I'm not sure we should remove them from all VS projects 🤔

clementperon: Yes but In case they are still some #ifdef WIN32 instead on _WIN32 we will not catch them on the CI.

seladb (commented):
- If we don't need `MAC_OS_X` anymore, maybe we should remove it from `platform.mk.macosx`?
- Same for `FREEBSD` and `platform.mk.freebsd`
- Same for `LINUX` and `platform.mk.linux`
- We probably can't do it for `WIN32` because the Makefiles use it to define a few things

clementperon:
Yes, but MAC_OS_X is still used 

`ifdef MAC_OS_X
INCLUDES += -I$(MACOS_SDK_HOME)/usr/include/malloc
endif`

Anyway this will be drop with Cmake

seladb:
> MAC_OS_X

You're right, I missed that one...

seladb (approved):

PR #793, opened 2022-01-26
Fix broken teardown in `TestPcapLiveDeviceClone`

seladb:
Since `DeviceTeardown` runs after `delete liveDev` it runs on invalid memory. This causes a failure which I observed with VS 2019.
The fix is to add a flag to `DeviceTeardown` to allow deleting the device.

### DIFF
--- a/Tests/Pcap++Test/Common/TestUtils.h
+++ b/Tests/Pcap++Test/Common/TestUtils.h
@@ -12,16 +12,21 @@ private:
 
 	pcpp::IDevice* m_Device;
 	bool m_CancelTeardown;
+	bool m_DeleteDevice;
 
 public:
 
-	DeviceTeardown(pcpp::IDevice* device) : m_Device(device), m_CancelTeardown(false) {}
+	DeviceTeardown(pcpp::IDevice* device, bool deleteDevice = false) : m_Device(device), m_CancelTeardown(false), m_DeleteDevice(deleteDevice) {}
 
 	~DeviceTeardown()
 	{
 		if (!m_CancelTeardown && m_Device != NULL && m_Device->isOpened())
 		{
 			m_Device->close();
+		}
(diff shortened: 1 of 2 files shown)


seladb:
@egecetin @clementperon please let me know what you think

clementperon:
@seladb I don't know a lot about the code but IMO the delete should be done by the same function/class doing the clone().

What's the issue with the acutal implem?

egecetin:
@seladb Is it work for not cloned devices if the flag set to true? If I'm not missing a part it will be free a value from `PcapLiveDeviceList` then if i call the same device from device list with for example with `getPcapLiveDeviceByName` it will return the deleted memory. 

If I am right maybe we can add a flag to `PcapLiveDeviceList` like `isCloned` and we can delete the memory when `if (isCloned && m_DeleteDevice)` at deconstructor

seladb:
@clementperon @egecetin  `DeviceTeardown` is a helper class for the tests. It's there to ensure that a device is properly closed at the end of the test, whether the test succeeds or fails. I added it because in the past I had devices that were opened during a test and if the test failed they would stay open, hence impacting other tests and making them fail too.
Until the recent change made by @egecetin (https://github.com/seladb/PcapPlusPlus/pull/775) a `PcapLiveDevice` was created only once and stored in a static list in `PcapLiveDeviceList`. Now, a device can be cloned - meaning it's possible to create and delete `PcapLiveDevice` instances throughout the program's lifetime. That's why I needed to add to `DeviceTeardown` an option to delete the cloned `PcapLiveDevice` that was created during the test.
Again - it's just for the purpose of the test. In a real program, the developer who creates the `PcapLiveDevice` object needs to delete it eventually.
Please let me know if it makes sense.

egecetin:
@seladb If `DeviceTeardown` is for only tests I think this PR is sufficient.

PR #794, opened 2022-01-27
ci: install latest npcap-oem

clementperon:
This USERNAME/Password is set for mingw but not for visual, also bump npcap to latest version

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -346,7 +346,8 @@ jobs:
 
     - name: Install NPcap
       env:
-        NPCAP_USERNAME: ${{ secrets.SuperSecret }}
+        NPCAP_USERNAME: ${{ secrets.NPCAP_USERNAME }}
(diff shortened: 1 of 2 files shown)


clementperon:
@seladb looks like   ${{ secrets.NPCAP_USERNAME }} and ${{ secrets.NPCAP_PASSWORD }} are not defined correct ?

seladb:
> @seladb looks like ${{ secrets.NPCAP_USERNAME }} and ${{ secrets.NPCAP_PASSWORD }} are not defined correct ?

What do you mean by not defined correct?

clementperon:
It seems that these env variables are not defined. Jobs still download the non OEM version in the CI pipeline.

seladb:
Yes, these variables aren't exposed to forks because they contain Npcap username and password for the OEM package that I'm not supposed to share. That's why forks are using the non-OEM version and only the main repo uses the OEM version as you can see here:
https://github.com/seladb/PcapPlusPlus/runs/4975418568?check_suite_focus=true#step:5:20

PR #797, opened 2022-02-01
Support debian/ubuntu installation of DPDK

szokovacs:
use the provided pkg-config if available


mk/PcapPlusPlus.mk.dpdk
```diff
@@ -1,7 +1,15 @@
 ### DPDK ###
 
+PKGCONF = pkg-config
+
 # includes
+TMPI = $(shell $(PKGCONF) --cflags libdpdk)
```
seladb, line 6: 1. If `libdpdk` isn't installed `pkg-config` gives an error:
    ```shell
    Package libdpdk was not found in the pkg-config search path.
    Perhaps you should add the directory containing `libdpdk.pc'
    to the PKG_CONFIG_PATH environment variable
    No package 'libdpdk' found
    ```
    This error will be prompt during the Makefile run and might be confusing to users. The only way I found to suppress it is adding `2> /dev/null`: `TMPI = $(shell $(PKGCONF) --cflags libdpdk 2> /dev/null)`

2.  Using `--cflags` gives more than just includes. Maybe we should consider using `--cflags-only-I` for the includes and then do:
    ```makefile
    TMPF = $(shell $(PKGCONF) --cflags-only-other libdpdk 2> /dev/null)
    PCAPPP_BUILD_FLAGS += $(TMPF)
    ```

mk/PcapPlusPlus.mk.dpdk
```diff
@@ -10,4 +18,9 @@ PCAPPP_LIBS_DIR += -L$(RTE_SDK)/build/lib -L/lib64
 PCAPPP_BUILD_FLAGS += -msse -msse2 -msse3 -Wall
 
 # libs
+TMPL = $(shell $(PKGCONF) --libs libdpdk)
```
seladb, line 21: Same here:
1. We can add `2> /dev/null` to suppress errors
2. We can separate into 2 parts:
    ```
    TMPLDIR = $(shell $(PKGCONF) --libs-only-L libdpdk)
    TMPL = $(shell $(PKGCONF) --libs-only-other libdpdk)
    ```

Pcap++/Makefile
```diff
@@ -46,7 +46,9 @@ ifdef PF_RING_HOME
 INCLUDES += -I$(PF_RING_HOME)/userland/lib -I$(PF_RING_HOME)/kernel
 endif
 ifdef USE_DPDK
+PKGCONF = pkg-config
 INCLUDES += -I"$(RTE_SDK)/build/include"
+INCLUDES += $(shell $(PKGCONF) --cflags libdpdk)
```
seladb, line 51: Why is this needed here if we already do it in `PcapPlusPlus.mk.dpdk`?

szokovacs: I did not investigate too deeply, but on is INCLUDES, the other is PCAPPP_INCLUDES

configure-linux.sh
```diff
@@ -274,13 +274,21 @@ fi
...
 function get_dpdk_version() {
    if [ -f $DPDK_HOME/pkg/dpdk.spec ]; then
       echo $(grep "Version" $DPDK_HOME/pkg/dpdk.spec | cut -d' ' -f2 | cut -d'.' -f 1,2);
-   else
+   elif [ -f $DPDK_HOME/VERSION ]; then
       echo $(cat $DPDK_HOME/VERSION | cut -d'.' -f 1,2);
+   elif [ -f $DPDK_HOME/rte_build_config.h ]; then
+      YEAR=$(grep RTE_VER_YEAR $DPDK_HOME/rte_build_config.h|cut -d ' ' -f3);
+      MONTH=$(grep RTE_VER_MONTH $DPDK_HOME/rte_build_config.h|cut -d ' ' -f3);
+      MINOR=$(grep RTE_VER_MINOR $DPDK_HOME/rte_build_config.h|cut -d ' ' -f3);
```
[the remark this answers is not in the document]
szokovacs: I meant it as a last fallback, also I think I confused $DPDK_HOME/build with $DPDK_HOME/, I will fix it.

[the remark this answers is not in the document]
szokovacs: Well, fixing it causes problems with my ubuntu-released 19.11.1, /usr/bin/ld: cannot find -lrte_pmd_vmxnet3_uio

Common++/Makefile
```diff
@@ -40,7 +40,7 @@ endif
 			
 Obj/%.o: src/%.cpp Obj/%.d
 	@echo Building file: $<
-	@$(CXX) $(DEPS) $(INCLUDES) $(PCAPPP_BUILD_FLAGS) -c -fmessage-length=0 -MMD -MP -MF "$(@:Obj/%.o=Obj/%.d)" -MT "$@" -o "$@" "$<"
+	@$(CXX) $(DEPS) $(INCLUDES) $(PCAPPP_INCLUDES) $(PCAPPP_BUILD_FLAGS) -c -fmessage-length=0 -MMD -MP -MF "$(@:Obj/%.o=Obj/%.d)" -MT "$@" -o "$@" "$<"
```
seladb, line 43: This change is not needed, `Common++` - `$(INCLUDES)` already has all the includes needed to compile this library

Packet++/Makefile
```diff
@@ -40,7 +40,7 @@ endif
 
 Obj/%.o: src/%.cpp Obj/%.d
 	@echo Building file: $<
-	@$(CXX) $(DEPS) $(INCLUDES) $(PCAPPP_BUILD_FLAGS) -c -fmessage-length=0 -MMD -MP -MF "$(@:Obj/%.o=Obj/%.d)" -MT "$@" -o "$@" "$<"
+	@$(CXX) $(DEPS) $(INCLUDES) $(PCAPPP_INCLUDES) $(PCAPPP_BUILD_FLAGS) -c -fmessage-length=0 -MMD -MP -MF "$(@:Obj/%.o=Obj/%.d)" -MT "$@" -o "$@" "$<"
```
seladb, line 43: This change is not needed, `Packet++` - `$(INCLUDES)` already has all the includes needed to compile this library

Pcap++/Makefile
```diff
@@ -46,7 +46,12 @@ ifdef PF_RING_HOME
 INCLUDES += -I$(PF_RING_HOME)/userland/lib -I$(PF_RING_HOME)/kernel
 endif
 ifdef USE_DPDK
+PKGCONF = pkg-config
+TMPI = $(shell $(PKGCONF) --cflags-only-I libdpdk 2>/dev/null)
 INCLUDES += -I"$(RTE_SDK)/build/include"
+ifneq ($(TMPI),)
+INCLUDES += $(TMPI)
+endif
```
seladb, line 54: Now I remember - `$(INCLUDES)` should have all the header files needed to compile `Pcap++`. So this change is correct. Just one thing we can improve:
```Makefile
ifdef USE_DPDK
PKGCONF = pkg-config
TMPI = $(shell $(PKGCONF) --cflags-only-I libdpdk 2>/dev/null)
ifneq ($(TMPI),)
INCLUDES += $(TMPI)
else
INCLUDES += -I"$(RTE_SDK)/build/include"
endif
endif
```

mk/PcapPlusPlus.mk.dpdk
```diff
@@ -1,13 +1,30 @@
...
 PCAPPP_LIBS_DIR += -L$(RTE_SDK)/build/lib -L/lib64
 
 #flags
 PCAPPP_BUILD_FLAGS += -msse -msse2 -msse3 -Wall
+ifneq ($(TMPFLAG),)
+PCAPPP_BUILD_FLAGS += $(TMPFLAG)
+endif
 
 # libs
+TMPL = $(shell $(PKGCONF) --libs libdpdk 2>/dev/null)
```
seladb, line 25: here we should use the `--libs-only-l` flag

mk/PcapPlusPlus.mk.dpdk
```diff
@@ -1,13 +1,30 @@
...
+TMPFLAG = $(shell $(PKGCONF) --cflags-only-other libdpdk 2>/dev/null)
+
+ifeq ($(TMPI),)
 PCAPPP_INCLUDES += -I$(RTE_SDK)/build/include
+else
+PCAPPP_INCLUDES += $(TMPI)
+endif
 
 # libs dir
 PCAPPP_LIBS_DIR += -L$(RTE_SDK)/build/lib -L/lib64
```
seladb, line 16: Here we can do a similar thing with the `--libs-only-L` flag:
```Makefile
TMPLD = $(shell $(PKGCONF) --libs-only-L libdpdk 2>/dev/null)
ifneq ($(TMPLD),)
PCAPPP_LIBS_DIR+= $(TMPLD)
else
PCAPPP_LIBS_DIR += -L$(RTE_SDK)/build/lib -L/lib64
endif
```

configure-linux.sh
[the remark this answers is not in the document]
seladb: ok, we can keep it like this since it's the 3rd fallback option

Pcap++/Makefile
seladb: You're right, we should use `INCLUDES` and not `PCAPPP_INCLUDES`, please see my comment: https://github.com/seladb/PcapPlusPlus/pull/797#discussion_r799228388

Common++/Makefile
szokovacs: I agree, it should not be, but --cflags-only-other is for me "-include rte_config.h -march=corei7" and we included it in $(PCAPPP_BUILD_FLAGS) in commit b9f00a9c85b312e9fe84a8b4093edc3985dd425d

seladb: I'm not sure I understand... 🤔

szokovacs: $ pkg-config --cflags-only-other libdpdk
-include rte_config.h -march=corei7

in mk/PcapPlusPlus.mk.dpdk :
TMPFLAG = $(shell $(PKGCONF) --cflags-only-other libdpdk 2>/dev/null)
ifneq ($(TMPFLAG),)
PCAPPP_BUILD_FLAGS += $(TMPFLAG)
endif

seladb: I'm not sure how this is related to the `Makefile` in `Common++`...
Basically `Common++` and `Packet++` libraries don't use any DPDK object, so they shouldn't need any additional DPDK include. Only the `Pcap++` library needs DPDK header files, which is why the change you made is necessary

mk/PcapPlusPlus.mk.dpdk
```diff
@@ -1,13 +1,35 @@
...
+TMPFLAG = $(shell $(PKGCONF) --cflags-only-other libdpdk 2>/dev/null)
+
+ifeq ($(TMPI),)
 PCAPPP_INCLUDES += -I$(RTE_SDK)/build/include
+else
+PCAPPP_INCLUDES += $(TMPI)
+endif
 
 # libs dir
+TMPLD = $(shell $(PKGCONF) --libs-l libdpdk 2>/dev/null)
```
seladb, line 16: I'm not sure the `--libs-l` flag exists, isn't it `--libs-only-l`?

szokovacs: Sorry about this, the 2>/dev/null hid the typo :(

Common++/Makefile
szokovacs: Common++/Makefile includes mk/PcapPlusPlus.mk which in turn has mk/PcapPlusPlus.mk.dpdk included, so PCAPPP_BUILD_FLAGS ends up including the DPDK stuff. $(INCLUDES) $(PCAPPP_BUILD_FLAGS) seems asymmetric to me, maybe there should be a BUILD_FLAGS too?
Or should the PcapPlusPlus.mk.dpdk restricted to Pcap++ ?

seladb: `PCAPPP_INCLUDES`, `PCAPPP_LIBS`, `PCAPPP_LIBS_DIR` and  `PCAPPP_BUILD_FLAGS` are mostly for external users who want to use PcapPlusPlus.
PcapPlusPlus own libraries (`Common++`, `Packet++`, `Pcap++`) have their own includes so they don't need to user `PCAPPP_*`

Common++/Makefile
```diff
@@ -40,7 +40,7 @@ endif
 			
 Obj/%.o: src/%.cpp Obj/%.d
 	@echo Building file: $<
-	@$(CXX) $(DEPS) $(INCLUDES) $(PCAPPP_BUILD_FLAGS) -c -fmessage-length=0 -MMD -MP -MF "$(@:Obj/%.o=Obj/%.d)" -MT "$@" -o "$@" "$<"
+	@$(CXX) $(DEPS) $(INCLUDES) $(BUILD_FLAGS) -c -fmessage-length=0 -MMD -MP -MF "$(@:Obj/%.o=Obj/%.d)" -MT "$@" -o "$@" "$<"
```
seladb, line 43: Let's just revert the change and keep it as it was?

Packet++/Makefile
```diff
@@ -40,7 +40,7 @@ endif
 
 Obj/%.o: src/%.cpp Obj/%.d
 	@echo Building file: $<
-	@$(CXX) $(DEPS) $(INCLUDES) $(PCAPPP_BUILD_FLAGS) -c -fmessage-length=0 -MMD -MP -MF "$(@:Obj/%.o=Obj/%.d)" -MT "$@" -o "$@" "$<"
+	@$(CXX) $(DEPS) $(INCLUDES) $(BUILD_FLAGS) -c -fmessage-length=0 -MMD -MP -MF "$(@:Obj/%.o=Obj/%.d)" -MT "$@" -o "$@" "$<"
```
seladb, line 43: Let's just revert the change and keep it as it was?

mk/PcapPlusPlus.mk.common
```diff
@@ -24,3 +24,5 @@ CXXFLAGS := -O2 -g -Wall
 endif
 
 PCAPPP_BUILD_FLAGS += $(CXXFLAGS)
+
+BUILD_FLAGS := $(PCAPPP_BUILD_FLAGS)
```
seladb, line 28: Let's just revert the change and keep it as it was?

szokovacs: Which version do you mean? Keep PCAPPP_BUILD_FLAGS as it was and extend PCAPPP_INCLUDES with the output of
"pkg-config --cflags" ?

seladb: We can remove `BUILD_FLAGS := $(PCAPPP_BUILD_FLAGS)` from this file and use `$(PCAPPP_BUILD_FLAGS)` in `Common++/Makefile` and `Packet++/Makefile` (which is the same as we currently have in `dev` and `master` branches).

I suggest we focus this PR on making DPDK build work with `pkg-config`. We can leave additional improvements to a separate PR. Please let me know what you think

configure-linux.sh
```diff
@@ -300,6 +308,7 @@ if (( $COMPILE_WITH_DPDK > 0 )) ; then
 
    # if DPDK ver < 20.11 concat additional definitions to PcapPlusPlus.mk
    CUR_DPDK_VERSION=$(get_dpdk_version)
+   echo $CUR_DPDK_VERSION
```
seladb, line 311: Is this a leftover from your debugging sessions? I'm not sure we need this print 🤔

szokovacs: Yes, sorry.

mk/PcapPlusPlus.mk.common
szokovacs: I agree, pushed a new version

seladb (approved):

seladb:
Thanks for your contribution @szokovacs ! I really appreciate your hard work and I'm sorry for the time it took to review and merge this PR

szokovacs:
> Thanks for your contribution @szokovacs ! I really appreciate your hard work and I'm sorry for the time it took to review and merge this PR

No problem at all, I agree it's important to iron out a details :)

PR #799, opened 2022-02-01
Renamed `LOG_DEBUG` to `PCPP_LOG_DEBUG` and `LOG_ERROR` to `PCPP_LOG_ERROR`

szokovacs:
because LOG_DEBUG is the debug priority macro in the widely used syslog.h


clementperon:
I think a better name would be PCAPP_LOG_DBG and PCAPP_LOG_ERR no? 

To avoid long macro name and be sure it doesn't conflict with another macro

szokovacs:
> I think a better name would be PCAPP_LOG_DBG and PCAPP_LOG_ERR no?

Sure, I'll make the change. Do you prefer PCAPP_ or PCAPPP_ ?

clementperon:
Ho yes PCAPPP*,  Just a recommendation maybe wait for @seladb opinion

seladb:
> > I think a better name would be PCAPP_LOG_DBG and PCAPP_LOG_ERR no?
> 
> Sure, I'll make the change. Do you prefer PCAPP_ or PCAPPP_ ?

We can change them to `PCPP_LOG_ERROR` and `PCPP_LOG_DEBUG`, please let me know what you think

seladb (approved):

PR #800, opened 2022-02-01
Drop WinPcap support

clementperon:
The WinPcap website is down and the project is quite old and has been superseeded by NPcap on Windows.

Comparaison between both: 
https://npcap.com/vs-winpcap.html

Let's remove it no ?

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -323,12 +323,10 @@ jobs:
             os: windows-2019
             platform: x86
             config: Release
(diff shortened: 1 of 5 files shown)


seladb:
@clementperon there are few reasons to continue supporting WinPcap:
- Many people and orgs still use WinPcap although it's deprecated
- Npcap has a restrictive license: https://github.com/nmap/npcap#license which requires orgs to pay if it's deployed on more than 5 systems. WinPcap, as far as I know, is free regardless of the number of deployments

PR #806, opened 2022-02-10
Fix missing pragma pop

clementperon:
@seladb @egecetin


egecetin:
@clementperon Yes definitely it should be there. Thanks

clementperon:
@seladb catch by the compiler not by me ;), IMO we should enable the Werror.

PR #807, opened 2022-02-11
pre-commit ci

egecetin:
At #806 trailing whitespaces removed manually. Maybe we can enforce similar changes by using pre-commit ci bot. @seladb @clementperon what do you think?

You can find my test run [here](https://github.com/egecetin/PcapPlusPlus/pull/1)


clementperon:
I think it's a good idea to have proper formatting check.

I would even push further with a clang-format/clang-tidy coding style.

seladb:
I agree it'd be great to have pre-commit hooks, but I'm not sure we want a bot to make those changes. Maybe we should just add the checks as part of the CI, which will make a PR fail until those checks are met 🤔

clementperon:
@seladb as the commit are squashed (I'm not fan of it),  we don't really care about each commit to be proper no?
This will also make it easier for user to contribute as the CI will fix their commit without them to install required tools.

seladb:
> @seladb as the commit are squashed (I'm not fan of it), we don't really care about each commit to be proper no? This will also make it easier for user to contribute as the CI will fix their commit without them to install required tools.

Yes, because PRs are squashed we don't care about each commit - only the entire PR matters. So when contributors are working on PRs they don't have to address those issues while a PR is still WIP, but they do need to address them in order to get the PR ready for review (otherwise the CI would fail). If we let a bot do it people won't care about formatting issues which I believe is not a good habit (and is what happening now since we don't have pre-commit hooks).

BTW, people don't have to install these pre-commit tools - they just need to see the CI errors and fix them which is usually not a big deal...

egecetin:
I think most of people don't want to install required tools. So manually fixing will be a good option for them. Additionally, even bot can fail sometimes (When clang-format is active, even the `pre-commit run` from my command line failed 2 or 3 times then resolved the all issues, I don't know maybe because of high number of changes) so probably not using the bot will be more accurate. Anyway, bot is only an option, both of solutions (with or without bot) is good for me.

As a side note for clang-format case (if we want to add it too), which style will you prefer? I'm the Microsoft style person 🙂

clementperon:
I'm the kernel style person :), but microsoft seems nice, very readable.

@seladb ?

seladb:
Regarding clang-format style:
- Is there an easy-to-use comparison of all styles? I googled it but couldn't find any
- What is the most popular style?
- Which style is the closest to what we have today, meaning which will require least changes?

Regarding the existing PR:
- I suggest we remove the bot for now, we can always add it later if we need to
- Maybe we also want to run the pre-commit hooks as part of the CI?
- This PR only contains the pre-commit hook, what about the code changes enforced by the pre-commit hook? do we leave it to a separate PR?
- Looking at [your PR](https://github.com/egecetin/PcapPlusPlus/pull/1) it seems the required changes are huge (358 files)! That's the main reason I didn't add pre-commit hooks before. Is there a way to do it gradually so we can control the changes?

clementperon:
You can try the different style here first line of the configuration. 

https://zed0.co.uk/clang-format-configurator/

seladb:
> You can try the different style here first line of the configuration.
> 
> https://zed0.co.uk/clang-format-configurator/

thank you @clementperon ! At first glance it looks like Microsoft + `UseTab = Always` is the closest to what we have now, so the least changes will be required. Please let me know what you think

egecetin:
> This PR only contains the pre-commit hook, what about the code changes enforced by the pre-commit hook? do we leave it to a separate PR?
> Looking at https://github.com/egecetin/PcapPlusPlus/pull/1 it seems the required changes are huge (358 files)! That's the main reason I didn't add pre-commit hooks before. Is there a way to do it gradually so we can control the changes?

For this reason I think we should separate the changes from this PR and probably you should commit them. I checked all changes with a commit at a different branch then removed. The changes are not raised an error on CI tests but we should check manually on different systems because of tests with network connection not supported at CI.

> I suggest we remove the bot for now, we can always add it later if we need to

Agree, only I will remove the part of ci section from yaml.

> Maybe we also want to run the pre-commit hooks as part of the CI?

I think we definitely should.

egecetin:
@seladb I'm not sure but most of changes are coming from mixed-line-endings. Maybe we can divide into 2 part (or 3 part) to changes to commit.

1. Black + pre-commit-hooks (except mixed-line-ending)
2. Mixed-line-ending
3. Clang-format

clementperon:
> > You can try the different style here first line of the configuration.
> > https://zed0.co.uk/clang-format-configurator/
> 
> thank you @clementperon ! At first glance it looks like Microsoft + `UseTab = Always` is the closest to what we have now, so the least changes will be required. Please let me know what you think

Yes let's add a .clang-format with this coding style. Also this would need a small paragraph in the documentation.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
+exclude: '.*\.(pcap|pcapng|dat|txt)'
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0  # Use the ref you want to point at
+    hooks:
+    -   id: trailing-whitespace
+    -   id: check-case-conflict
+    -   id: end-of-file-fixer
+    -   id: mixed-line-ending
+        args: ['--fix=lf']
```
seladb, line 10: why do we want `lf` and not the default? 🤔

.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+    -   id: black
+ci:
+    autofix_commit_msg: |
+        [pre-commit.ci] auto fixes from pre-commit.com hooks
+    autofix_prs: true
+    autoupdate_branch: 'dev'
+    autoupdate_commit_msg: '[pre-commit.ci] pre-commit autoupdate'
+    autoupdate_schedule: weekly
+    skip: []
+    submodules: false
```
seladb, line 23: As we discussed, let's remove the `ci` section for now

.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0  # Use the ref you want to point at
+    hooks:
+    -   id: trailing-whitespace
+    -   id: check-case-conflict
+    -   id: end-of-file-fixer
+    -   id: mixed-line-ending
+        args: ['--fix=lf']
+-   repo: https://github.com/psf/black
+    rev: 21.12b0
```
seladb, line 12: latest version is [`22.1.0`](https://github.com/psf/black/releases/tag/22.1.0)

.pre-commit-config.yaml
clementperon: Is the most frequent line ending per file?
 
Maybe if this is the case, this will not guarantee a cohenrency in all the repo.

I think we can agree that we want LF as EOL no?

seladb:
> @seladb I'm not sure but most of changes are coming from mixed-line-endings. Maybe we can divide into 2 part (or 3 part) to changes to commit.
> 
>     1. Black + pre-commit-hooks (except mixed-line-ending)
> 
>     2. Mixed-line-ending
> 
>     3. Clang-format

From a quick check I did we can probably make the changes in this order:
1. Black (which only affect Python files)
2. `end-of-file-fixer` - affects 173 but easy to review
3. `trailing-whitespace` - affect 177 files
4. The rest of the checks - total number of files affected is 517, need to understand the changes better
5. `clang-format`. After we do that we can add it to the pre-commit config: https://github.com/pocc/pre-commit-hooks

.pre-commit-config.yaml
egecetin: I agree on @clementperon, default will not guarantee consistency between files and most of developers work on Linux based systems so I think LF will be more suitable for an open source repo. What do you think?

.github/workflows/build_and_test.yml
```diff
@@ -2,6 +2,13 @@ name: Build and test
 on: [push, pull_request]
 
 jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+    - uses: pre-commit/action@v2.0.3
```
seladb, line 10: Let's not include pre-commit checks in the CI yet, otherwise it will keep failing until we resolve all issues. Let me open PRs to fix all issues, and then we can include this in the CI

.pre-commit-config.yaml
seladb: ok sure, I agree 👍

seladb:
@egecetin let's just fix this https://github.com/seladb/PcapPlusPlus/pull/807#discussion_r808738718 and it's good to merge

.github/workflows/build_and_test.yml
egecetin: Since we are planning to add I only comment out these lines, so we can add easily later

.clang-format
```diff
@@ -0,0 +1,136 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Microsoft
```
clementperon, line 3: Why not just Microsoft + useTabs ?

egecetin: It can be, I dump the config from command line then changed the useTabs, so all of configuration is there. If we want to make it simpler I can change it

clementperon: If you made this dump with clang-format-13 there are some params that would be not available on clang-format-10 for example. 

I think just Microsoft + useTabs Is the easiest to read and if we need a specific change we can explicit it no ?

seladb (approved):

PR #808, opened 2022-02-12
Fix tests w/ Python 3.10

clementperon:
We need to bump pytest to fix issue with python 3.10.

But let's also bump the rest of packages and remove the unused


Tests/ExamplesTest/requirements.txt
```diff
@@ -1,17 +1,8 @@
...
+iniconfig==1.1.1
+packaging==21.3
+py==1.11.0
+pyparsing==3.0.7
+pytest==6.2.5
 scapy==2.4.3
-six==1.15.0
-toml==0.10.1
-wcwidth==0.2.5
-wrapt==1.12.1
```
seladb, line 17: I think most of the "extra" packages are needed for `black` and `pylint` which we might want to add as [pre-commit hooks](https://github.com/seladb/PcapPlusPlus/pull/807) and check them in the CI as well

clementperon: Maybe but the black or pylint commit should introduce them properly.

seladb: ok, I agree 👍

PR #809, opened 2022-02-12
Telnet protocol support

egecetin:


seladb:
@egecetin this PR is still in draft, I assume it's not yet ready to review?

egecetin:
@seladb Yes it has a bit work to do. I need to check some command options of telnet. But if you have time you can review these parts.

egecetin:
@seladb could you take look?

seladb:
Thank you @egecetin ! Please give me a few more days to review. I already started but it'll take me some time

Packet++/header/TelnetLayer.h
```diff
@@ -0,0 +1,343 @@
...
+
+        struct telnet_field_data
+        {
+            // Header
+            struct telnet_header *hdr;
+            // Size of the header including data payload
+            uint16_t hdrSize;
+            // Offset of the current header
+            uint16_t currentOffset;
+        };
```
seladb, line 45: 1. Generally speaking we usually try to keep as less layer metadata as possible for several reasons:
   1. Avoid taking a lot of memory per packet
   2. In case the current layer of  the layers below it change, there is less data and pointers we need to update 
2. Keeping pointers might be an issue because if the `Packet` object is copied or if one of the other layers changes, these pointers become invalid and can even point to invalid memory. Instead, it's usually better to keep just the offset

You can see examples from other layers, but we usually keep just the number of "options" or "commands" and parse the packet again each time `getOption()` or `getOptionData()` is called

Packet++/src/TelnetLayer.cpp
```diff
@@ -0,0 +1,315 @@
...
+{
+
+    void TelnetLayer::parseTelnetFields()
+    {
+        // Since only contains data return immediately
+        if ((m_Data[0] != 255) || (m_Data[0] == 255 && m_Data[1] == 255))
+        {
+            isData = true;
+            return;
+        }
```
seladb, line 21: I've seen cases where the same packet contains both commands and data, here is one example:
[telnet-commands-and-data.zip](https://github.com/seladb/PcapPlusPlus/files/8152387/telnet-commands-and-data.zip)

Packet++/header/TelnetLayer.h
```diff
@@ -0,0 +1,343 @@
...
+
+        /// Return the number of detected Telnet Commands
+        uint16_t getNumberOfCommands();
+
+        /**
+         * Get the command of the given index
+         * @param[in] index Index to requested value
+         * @return Command of the given index, TelnetCommandInternalError if the requested index is larger than number of commands
+         */
+        TelnetCommands getCommand(size_t index);
```
seladb, line 258: I know that telnet data isn't similar to TLV options (Type-Length-Value), but maybe we can borrow some ideas from that interface. You can look at [`TLVData.h`](https://github.com/seladb/PcapPlusPlus/blob/master/Packet++/header/TLVData.h) and a few layers that implement it such as [`TcpLayer`](https://github.com/seladb/PcapPlusPlus/blob/master/Packet++/header/TcpLayer.h), [`IPv4Layer`](https://github.com/seladb/PcapPlusPlus/blob/master/Packet++/header/IPv4Layer.h), and more - and think if we want a somehow similar interface, for example:
- Create a struct/class for every command
- Build an interface that includes:
  - `getFirstCommand()`
  - `getNextCommand(prevCommand)`
  - `Command getCommand(commandType)`
  - `getCommandCount()`

Packet++/header/TelnetLayer.h
```diff
@@ -0,0 +1,343 @@
...
+        struct telnet_header
+        {
+            // "Interpret as Command" escape character (FF)
+            uint8_t interpretation,
+                // Command
+                command,
+                // Option of the command
+                subcommand,
+                // Data (Variable length)
+                data[];
```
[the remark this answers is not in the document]
egecetin: Yes. I try to get rid of it by,

https://github.com/egecetin/PcapPlusPlus/blob/5848c07a0510000d5ad04c25b091d9362f5926b5/Packet%2B%2B/src/TelnetLayer.cpp#L113

During the parsing I only placed same struct for every command then when the value returned I adjusted the position.  But I'm not sure this is the best practice. I will check TLV maybe I can find a solution

seladb:
hi @egecetin ! Do you know when you'll have time to continue the work on this?

egecetin:
Probably in next two weeks

egecetin:
@seladb for keep you updated. I'm a bit busy for now. Probably i will need more time to update this PR.

seladb:
> @seladb for keep you updated. I'm a bit busy for now. Probably i will need more time to update this PR.

Sure @egecetin thanks for the update!

Packet++/header/TelnetLayer.h
```diff
@@ -0,0 +1,337 @@
...
+         * A constructor that creates the layer from an existing packet raw data
+         * @param[in] data A pointer to the raw data
+         * @param[in] dataLen Size of the data in bytes
+         * @param[in] prevLayer A pointer to the previous layer
+         * @param[in] packet A pointer to the Packet instance where layer will be stored in
+         */
+        TelnetLayer(uint8_t *data, size_t dataLen, Layer *prevLayer, Packet *packet) : Layer(data, dataLen, prevLayer, packet)
+        {
+            m_Protocol = Telnet;
+            lastPositionOffset = UINT64_MAX;
```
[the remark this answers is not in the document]
egecetin: @seladb I checked `size_t` a bit more and it seems to vary from platform to platform. `UINT16_MAX` seems like a better choice or maybe we can use `SIZE_MAX`. Which one do you think is better?

Packet++/header/TelnetLayer.h
```diff
@@ -0,0 +1,337 @@
+#ifndef PACKETPP_TELNET_LAYER
```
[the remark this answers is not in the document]
egecetin: Looks like my vs code misses or overwrites `.clang-format` 🤔

Packet++/header/TelnetLayer.h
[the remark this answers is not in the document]
seladb: You're right, `SIZE_MAX` is a better choice!

Packet++/src/TelnetLayer.cpp
```diff
@@ -0,0 +1,483 @@
...
+		lastPositionOffset = pos - m_Data;
+		return static_cast<TelnetLayer::TelnetCommand>(pos[1]);
+	}
+	lastPositionOffset = UINT16_MAX;
+	return TelnetCommandEndOfPacket;
+}
+
+TelnetLayer::TelnetOption TelnetLayer::getOption()
+{
+	if (lastPositionOffset <= m_DataLen)
```
seladb, line 219: this should be `<` instead of `<=`, no? 🤔

egecetin: Definitely

seladb (approved):

Issue #758, opened 2021-11-23
ipv4::parseNextLayer doesn't support ipv6 protocol ( protocol field value equal to 41 ).


alexdrobinsky:
Hello dear developers,

I stumbled upon problem with parsing of pcap with ipv6 encapsulated into ipv4 , looks like switch case missing following lines:

    case PACKETPP_IPPROTO_IPV6:
        m_NextLayer = new IPv6Layer(payload, payloadLen, this, m_Packet);
        break;

I verified with test pcap file that fix works properly. Could you add it to IPV4Layer.cpp ?
Best regards,
Alexander

seladb:
You're right @alexdrobinsky , that's indeed missing. Would you consider opening a PR with the fix?
Also - can you provide a sample pcap file we can use in the [tests](https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Packet++Test/Tests/IPv4Tests.cpp)?

alexdrobinsky:
Sorry for huge delay in response, below attached pcap with relevant scenario

ср, 24 нояб. 2021 г. в 07:48, seladb ***@***.***>:

seladb:
thanks @alexdrobinsky, however I cannot see the pcap file. Can you attach it again to the issue?

alexdrobinsky:
Attached for sure, hopefully some dumb antivirus doesn't prevent it

пт, 3 дек. 2021 г. в 09:03, seladb ***@***.***>:

seladb:
@alexdrobinsky I still can't see this file, maybe it's because you're replying from an email? 🤔 
Can you reply from GitHub directly and drag/attach the file to the issue?

alexdrobinsky:
[FTP_ipv6_onlyControl.zip](https://github.com/seladb/PcapPlusPlus/files/7728004/FTP_ipv6_onlyControl.zip)
Looks like this e-mail ( even drag and drop ) doesn't support pcap files. Hopefully, storing it in zip file should solve the issue.

seladb:
thanks @alexdrobinsky , I can see the file now. I'll work on the fix and keep you posted

seladb:
Fixed in #810 , thank you @egecetin ! 🙏

PR #810, opened 2022-02-13
Bugfix #758 - support IPv4 encapsulated IPv6

egecetin:
Fix for #758 

Since issue is open for approx 3 months let's clean it up


clementperon (approved):

Packet++/src/IPv4Layer.cpp
```diff
@@ -321,6 +321,9 @@ void IPv4Layer::parseNextLayer()
 			? static_cast<Layer*>(new ESPLayer(payload, payloadLen, this, m_Packet))
 			: static_cast<Layer*>(new PayloadLayer(payload, payloadLen, this, m_Packet));
 		break;
+	case PACKETPP_IPPROTO_IPV6:
+		m_NextLayer = new IPv6Layer(payload, payloadLen, this, m_Packet);
```
seladb, line 325: I think we should check if IPv6 data is valid:
```cpp
m_NextLayer = IPv6Layer::isDataValid(payload, payloadLen)
	? static_cast<Layer*>(new IPv6Layer(payload, payloadLen, this, m_Packet))
	: static_cast<Layer*>(new PayloadLayer(payload, payloadLen, this, m_Packet));
```

Tests/Packet++Test/PacketExamples/IPv4-encapsulated-IPv6.dat
```diff
@@ -0,0 +1 @@
+0100010000001a432000010008004500006e2e0100001029432d8b12192151834383600440e80032063f2001063809020001020102fffee275962002518343830000000000005183438300150402e537a964626bf4b35018816099d30000323030204e4f4f5020636f6d6d616e64207375636365737366756c2e0d0a
```
seladb, line 1: Can you also include a `.pcap` file that contains this packet? It's easier to open it in Wireshark if needed...
Please note that [`.gitignore` disallows adding `.pcap` files](https://github.com/seladb/PcapPlusPlus/blob/master/.gitignore#L20-L24) so you need to use the `-f` flag in order to add it:

```shell
git add -f <YOUR_FILE>.pcap
```

seladb (commented):
Since we squash PRs, which makes every PR a single commit, it'd be good to describe the change in the PR header, maybe something like: `Bugfix #758 - support IPv6 over IPv4`

seladb (approved):

PR #813, opened 2022-02-18
Pcap++: fix segfault in case pcap_activate() fail

clementperon:
pcap is set to NULL and then used in pcap_setdirection() / pcap_datalink()
this generate a segfault.

Return NULL instead of continuing

### DIFF
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -275,7 +275,7 @@ pcap_t* PcapLiveDevice::doOpen(const DeviceConfiguration& config)
 	{
 		PCPP_LOG_ERROR(pcap_geterr(pcap));
 		pcap_close(pcap);
-		pcap = NULL;
+		return NULL;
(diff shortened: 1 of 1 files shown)


Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -275,7 +275,7 @@ pcap_t* PcapLiveDevice::doOpen(const DeviceConfiguration& config)
 	{
 		PCPP_LOG_ERROR(pcap_geterr(pcap));
 		pcap_close(pcap);
-		pcap = NULL;
+		return NULL;
```
seladb, line 278: @clementperon I guess this is the only fix that is related to the segfault, is that correct? 
I suggest we keep only that in this PR and move the rest of the changes (if they're needed) to a separate PR. Please let me know what you think

clementperon: done

seladb (approved):

PR #823, opened 2022-03-01
Inline the small functions in Eth and Arp layers

szokovacs:
Introduce small util functions to replace source and destination addresses

### DIFF
--- a/Packet++/header/ArpLayer.h
+++ b/Packet++/header/ArpLayer.h
@@ -127,7 +127,7 @@ namespace pcpp
 		 * - @ref arphdr#hardwareSize = 6
 		 * - @ref arphdr#protocolType = ETHERTYPE_IP (assume IPv4 over ARP)
 		 * - @ref arphdr#protocolSize = 4 (assume IPv4 over ARP)
-		 * - if it's an ARP requst: @ref arphdr#targetMacAddr = MacAddress("00:00:00:00:00:00")
+		 * - if it's an ARP request: @ref arphdr#targetMacAddr = MacAddress("00:00:00:00:00:00")
 		 */
 		void computeCalculateFields();
 
--- a/Packet++/header/ArpLayer.h
+++ b/Packet++/header/ArpLayer.h
@@ -109,12 +109,6 @@ namespace pcpp
 		 */
 		inline IPv4Address getTargetIpAddr() const { return getArpHeader()->targetIpAddr; }
 
-		/**
-		 * Transform the packet to a response, suitable for sending back 
-		 * @param[in] senderMacAddr The sender MAC address (will be put in arphdr#senderMacAddr)
-		 */
-		void createResponse(const MacAddress& senderMacAddr);
-
 		// implement abstract methods
 
 		/**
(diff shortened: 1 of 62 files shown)


Packet++/header/ArpLayer.h
```diff
@@ -83,32 +83,37 @@ namespace pcpp
...
 		 * @return An IPv4Address containing the target protocol address (TPA)
 		 */
-		IPv4Address getTargetIpAddr() const { return getArpHeader()->targetIpAddr; }
+		inline IPv4Address getTargetIpAddr() const { return getArpHeader()->targetIpAddr; }
 
+		/**
+		 * Transform the packet to a response, suitable for sending back 
+		 * @param[in] senderMacAddr The sender MAC address (will be put in arphdr#senderMacAddr)
+		 */
+		void createResponse(const MacAddress& senderMacAddr);
```
seladb, line 116: I wouldn't put this method in `ArpLayer` because it isn't related to packet parsing. This is a pretty simple method to write to I'm not even sure we need to add it to PcapPlusPlus... can you elaborate on why you need it?

Packet++/header/EthLayer.h
```diff
@@ -103,31 +103,36 @@ namespace pcpp
...
 		 * Set destination MAC address
 		 * @param destMac Destination MAC to set
 		 */
-		void setDestMac(const MacAddress& destMac) { destMac.copyTo(getEthHeader()->dstMac); }
+		inline void setDestMac(const MacAddress& destMac) { destMac.copyTo(getEthHeader()->dstMac); }
+
+		/**
+		 * Transform the packet to a response, suitable for sending back 
+		 */
+		void createResponse();
```
seladb, line 135: Same - this method probably doesn't belong to `EthLayer` (see my previous comment)

Packet++/header/ArpLayer.h
szokovacs: I needed to create a basic ARP response on the DPDK interface so the packets can be routed there. 
It does not have to be part of pcpp, of course, but I think it is a common enough problem that it could be useful. I put some hours of research in it, maybe other don't have to :)
I agree it is not strictly related to parsing the of the packet, but does it have to be? It is possible to use the Packet interface to create/modify the outgoing packets too, the EthLayer even have the set* functions, and they are useful, so I would propose to extend the ArpLayer with set* too.
If you think creating an ARP response doesn't belong to the ArpLayer, should we create an utility class for these kind of things?

Is there any other forum to discuss these proposals besides creating a PR? The googlegroup seems dead.

seladb: @szokovacs I'm sorry for not responding on PcapPlusPlus Google Group. It's not dead, I just missed your message...
I read your comment and went over the code again and I'm a little bit more convinced now 😃 
The main concern I still have is:
In most of the layers (maybe in all of them) we don't have methods that are directly related to the mechanism of the protocol, such as creating responses or requests. The reason is that in many protocols a single packet doesn't have enough context to understand the state of the protocol, and we need to look at the whole stream to understand which message should come next. Good examples of this are [TCP reassembly](https://github.com/seladb/PcapPlusPlus/blob/master/Packet%2B%2B/header/TcpReassembly.h) and [IP de/fragmentation](https://github.com/seladb/PcapPlusPlus/blob/master/Packet%2B%2B/header/IPReassembly.h) which "live" outside of the IP and TCP layers.  This is of course not true in simple protocols such as ARP where there is virtually no state and creating a response from a request needs just the current (request) packet.
Still, we need to figure out the right place for these kinds of methods. For example: what if the current ARP packet is already a response? What should `createResponse()` do in that case that will be clear enough to the user? 🤔

seladb (approved):

PR #829, opened 2022-03-05
Typos CI

egecetin:
Looks like codespell is missing too many errors. Maybe [typos](https://github.com/crate-ci/typos) can solve these


Packet++/header/SSLCommon.h
```diff
@@ -388,7 +388,7 @@ namespace pcpp
 	/**
 	 * SSL/TLS symmetric encryption algorithms
 	 */
-	enum SSLSymetricEncryptionAlgorithm
+	enum SSLSymmetricEncryptionAlgorithm
```
seladb, line 391: enums are more difficult to clone and dperecate. I suggest we just leave it with the typo. What do you think?

Packet++/header/SSLHandshake.h
```diff
@@ -656,7 +656,7 @@ class SSLServerHelloMessage : public SSLHandshakeMessage
 	/**
 	 * @return The size (in bytes) of all extensions data in this message. Extracted from the "extensions length" field
 	 */
-	uint16_t getExtensionsLenth() const;
+	uint16_t getExtensionsLength() const;
```
seladb, line 659: ditto

Packet++/header/DhcpLayer.h
```diff
@@ -688,7 +688,7 @@ namespace pcpp
 		 * @return DHCP message type as extracted from ::DHCPOPT_DHCP_MESSAGE_TYPE option. If this option doesn't exist the value of
 		 * ::DHCP_UNKNOWN_MSG_TYPE is returned
 		 */
-		DhcpMessageType getMesageType() const;
+		DhcpMessageType getMessageType() const;
```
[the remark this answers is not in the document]
egecetin: I agree, I will update exception list and functions to deprecated 👍

Packet++/header/SSLCommon.h
egecetin: We can duplicate values but probably not worth it 🤔. I think adding exception is easier.

seladb (approved):

Issue #737, opened 2021-10-14
`aarch64` support


seladb:
Continuing the discussion that started here: https://github.com/seladb/PcapPlusPlus/issues/732

@inSilo let's discuss here how we want to proceed

inSilo:
> Continuing the discussion that started here: #732
> 
> @inSilo let's discuss here how we want to proceed

Sorry, I have been too busy these few days, I haven't had time to come and see the reply. Today I saw "change the variable type that holds getopt_long result from char to int", which is pretty good, and the rest is to cross-compile to aarch64 platform. If you don't respect the integration of libpcap into the project and compile it is also possible. Cross-compilation is currently configured in platform.mk.linux.aarch64 and configure-linux.sh. The cross-target tool chain of aarch64 is configured in platform.mk.linux.aarch64, and then a "--" is configured in configure-linux.sh. The "aarch64" option allows users to specify the cross-compilation target as the aarch64 platform. The author's idea is great, I hope the author can integrate it according to the idea of ​​my last post.

sebastiandavalle:
any update regarding this? 
I need PcapPlusPlus compiled for aarch64 (need to run it on ubuntu aarch64 - Raspberry pi) . I've been trying to cross-compile it but without success

sjdavalle:
any update regarding this?
I need PcapPlusPlus compiled for aarch64 (need to run it on ubuntu aarch64 - Raspberry pi) . I've been trying to cross-compile it but without success

PR #833, opened 2022-03-11
added support for aarch64

sjdavalle:
The aim of this PR is to add support for aarch64 cross compilation.

pre-requisites:
sudo apt-get update && sudo apt-get install -y \
    wget \
    gcc \
    make \
    flex \
    bison \
    gcc-aarch64-linux-gnu \
    g++-aarch64-linux-gnu \
    binutils-aarch64-linux-gnu \
    libpcap-dev

usage:
./configure-linux.sh --default --aarch64
make libs

### DIFF
--- a/configure-linux.sh
+++ b/configure-linux.sh
@@ -251,17 +251,17 @@ fi
 PLATFORM_MK="mk/platform.mk"
 PCAPPLUSPLUS_MK="mk/PcapPlusPlus.mk"
 
-# copy the basic Linux platform.mk
-cp -f mk/platform.mk.linux $PLATFORM_MK
-
 # copy the common (all platforms) PcapPlusPlus.mk
 cp -f mk/PcapPlusPlus.mk.common $PCAPPLUSPLUS_MK
 
 # add the Linux definitions to PcapPlusPlus.mk
-cat mk/PcapPlusPlus.mk.linux >> $PCAPPLUSPLUS_MK
 if [ -n "$BUILD_FOR_ARM64" ]; then
+   cp -f mk/platform.mk.linux.aarch64 $PLATFORM_MK
    cat mk/PcapPlusPlus.mk.linux.aarch64 >> $PCAPPLUSPLUS_MK
-   cat mk/platform.mk.linux.aarch64 >> $PLATFORM_MK
+else
+   # copy the basic Linux platform.mk
+   cp -f mk/platform.mk.linux $PLATFORM_MK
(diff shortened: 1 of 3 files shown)


clementperon:
@sjdavalle the libpcap-dev should be the one cross compiled in aarch64 no ?

seladb:
@sjdavalle I see the only change between other Linux build and `aarch64` is the gcc/g++ compiler. In that case, there is no need to change anything! You can just do:
```shell
seladb@seladb:/PcapPlusPlus# export CXX=/usr/bin/aarch64-linux-gnu-g++
seladb@seladb:/PcapPlusPlus# export CC=/usr/bin/aarch64-linux-gnu-gcc
seladb@seladb:/PcapPlusPlus# ./configure-linux.sh --default
seladb@seladb:/PcapPlusPlus# make
```

I tried it locally and it works!
Of course, as @clementperon mentioned, in order to have a running app you also need `libpcap-dev` compiled for `aarch64`.

So I'm not sure these changes are needed. What might be needed is more documentation! I'd add it to here:
https://pcapplusplus.github.io/docs/next/install/linux
https://github.com/PcapPlusPlus/pcapplusplus.github.io/blob/master/docs/install/linux.mdx

sjdavalle:
Yes, I agree, it works... it was my first approach but I thought that it might be useful to have just a flag instead of setting those env variable manually.

In regards to `libpacap-dev`... yes it is installed in an aarch64 environment where the application is being built.

seladb:
> Yes, I agree, it works... it was my first approach but I thought that it might be useful to have just a flag instead of setting those env variable manually.

I guess there is some gray area when to create a new flag vs when it's not needed... My opinion is that if we're only changing the compiler there's no need to create a new flag. However if we need to change more than probably a flag and new `platform.mk.*` / `PcapPlusPlus.mk.*` are needed

seladb:
But I still think we need documentation on how to compile PcapPlusPlus using different gcc compilers 😃

seladb:
@sjdavalle I added this to the documentation, please let me know what you think:
https://pcapplusplus.github.io/docs/next/install/linux#cross-compilation-or-build-for-different-architectures

sjdavalle:
that looks great! Thanks!
If you need I can share a small Dockerfile and some bash script to cross-compile inside a container.

seladb:
Sure, I will add this to the documentation

seladb:
@sjdavalle can we close the PR and the corresponding issue #737 ?

sjdavalle:
cross-compilation steps and information added to the documentation by @seladb

PR #838, opened 2022-03-15
Fix internet checksum calculation

guebe:
Bugfix of internet checksum for odd length packets.
Odd length packets are possible when sending/receiving Tcp payload.

This fixes:
- invalid read of one byte past the end of the buffer. Verified with
  valgrind.
- wrong checksum on big endian systems. Verified with qemu-ppc.

Initially this fix was done for an older version of libpcapplusplus we are using.
The fix was used in a production system for quite some time now.
Without the fix we had some spurious segfaults on windows debug builds.
The backtrace and valgrind helped us narrowing down the problem to this function.

### DIFF
--- a/Tests/Packet++Test/Tests/TcpTests.cpp
+++ b/Tests/Packet++Test/Tests/TcpTests.cpp
@@ -287,3 +287,20 @@ PTF_TEST_CASE(TcpPacketCreation2)
 	PTF_ASSERT_TRUE(tcpSnackOption.isNotNull());
 	PTF_ASSERT_TRUE(tcpSnackOption.setValue(htobe32(1000)));
 } // TcpPacketCreation2
+
+PTF_TEST_CASE(TcpChecksumInvalidRead)
+{
+	uint8_t *m = new uint8_t[3];
+	m[0] = 0x01;
+	m[1] = 0x12;
+	m[2] = 0xF3;
+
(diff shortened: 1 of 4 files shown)


Tests/Packet++Test/Tests/TcpTests.cpp
```diff
@@ -287,3 +287,20 @@ PTF_TEST_CASE(TcpPacketCreation2)
...
+
+	pcpp::ScalarBuffer<uint16_t> vec[1];
+	vec[0].buffer = (uint16_t*)m;
+	vec[0].len = 3;
+
+	uint16_t c = pcpp::compute_checksum(vec, 1);
+	PTF_ASSERT_EQUAL(c, 0xbedU, size);
+
+	delete [] m;
+} // TcpChecksumInvalidRead
```
guebe, line 306: this testcase fails if you run it without the fix:
valgrind reports an invalid read of one byte past m.
when run under qemu-ppc a wrong checksum is calculated (assert in line 303 breaks).

seladb (approved):

seladb:
Thank you @guebe ! I really appreciate the time you spent on finding the bug and contributing back the fix. BIG THANK YOU! 🙏 🙏

guebe:
@seladb thx for very fast merge!
We are in the process of updating to latest pcapplusplus version.
In this process we may have more PRs coming.

PR #866, opened 2022-04-10
PacketUtils: cleaner computeChecksum code

clementperon:
Hi @guebe,

You recently submit a patch to fix compute checksum on bigendian and avoid an out of bound reading.

This caught my attention as this part is a bit hard to read. 

Here is a clean rework, could you review this and confirm if it works on BigEndian arch?

Thanks

### DIFF
--- a/Packet++/src/PacketUtils.cpp
+++ b/Packet++/src/PacketUtils.cpp
@@ -1,4 +1,5 @@
 #include <string.h>
+
 #include "PacketUtils.h"
 #include "IPv4Layer.h"
 #include "IPv6Layer.h"
@@ -32,7 +33,8 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
 			uint8_t *vecBytes = (uint8_t *)vec[i].buffer;
 			uint8_t lastByte = vecBytes[vec[i].len - 1];
 			PCPP_LOG_DEBUG("1 byte left, adding value: 0x" << std::uppercase << std::hex << lastByte);
-			// swap this in case we are on bigEndian arch
+			// We have read the latest byte manually but this byte should be interpreted
+			// as a 0xFF on LE and a 0xFF00 on BE to have a proper checksum computation
 			localSum += le16toh(lastByte);
 			PCPP_LOG_DEBUG("Local sum = " << localSum << ", 0x" << std::uppercase << std::hex << localSum);
 		}
@@ -42,7 +44,6 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
 		{
 			localSum = (localSum & 0xffff) + (localSum >> 16);
 		}
-		localSum = be16toh(localSum);
 		PCPP_LOG_DEBUG("Local sum = " << localSum << ", 0x" << std::uppercase << std::hex << localSum);
 		sum += localSum;
 	}
@@ -51,14 +52,16 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
 	{
 		sum = (sum & 0xffff) + (sum >> 16);
 	}
-
 	PCPP_LOG_DEBUG("Sum before invert = " << sum << ", 0x" << std::uppercase << std::hex << sum);
 
-	sum = ~sum;
+	// To obtain the checksum we take the ones' complement of this result
+	uint16_t result = sum;
+	result = ~result;
 
-	PCPP_LOG_DEBUG("Calculated checksum = " << sum << ", 0x" << std::uppercase << std::hex << sum);
+	PCPP_LOG_DEBUG("Calculated checksum = " << sum << ", 0x" << std::uppercase << std::hex << result);
 
-	return ((uint16_t) sum);
+	// We return the result in Network byte order
+	return htons(result);
 }
 
 static const uint32_t FNV_PRIME = 16777619u;
--- a/Packet++/src/PacketUtils.cpp
+++ b/Packet++/src/PacketUtils.cpp
@@ -17,25 +17,27 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
 	for (size_t i = 0; i<vecSize; i++)
 	{
 		uint32_t localSum = 0;
-		size_t buffLen = vec[i].len;
-		while (buffLen > 1)
+
+		// vec len is in bytes
+		for (size_t j = 0; j < vec[i].len / 2; j++)
 		{
-			PCPP_LOG_DEBUG("Value to add = 0x" << std::uppercase << std::hex << *(vec[i].buffer));
-			localSum += *(vec[i].buffer);
-			++(vec[i].buffer);
-			buffLen -= 2;
+			PCPP_LOG_DEBUG("Value to add = 0x" << std::uppercase << std::hex << vec[i].buffer[j]);
+			localSum += vec[i].buffer[j];
 		}
 		PCPP_LOG_DEBUG("Local sum = " << localSum << ", 0x" << std::uppercase << std::hex << localSum);
 
-		if (buffLen == 1)
-		{
-			uint16_t lastByte = 0;
-			*((uint8_t*)(&lastByte)) = *((uint8_t*)(vec[i].buffer));
+		// check if there is one byte left
+		if (vec[i].len % 2) {
+			// access to the last byte using an uint8_t pointer
+			uint8_t *vecBytes = (uint8_t *)vec[i].buffer;
+			uint8_t lastByte = vecBytes[vec[i].len - 1];
 			PCPP_LOG_DEBUG("1 byte left, adding value: 0x" << std::uppercase << std::hex << lastByte);
-			localSum += lastByte;
+			// swap this in case we are on bigEndian arch
+			localSum += le16toh(lastByte);
 			PCPP_LOG_DEBUG("Local sum = " << localSum << ", 0x" << std::uppercase << std::hex << localSum);
 		}
 
+		// carry count is added to the sum
 		while (localSum>>16)
 		{
 			localSum = (localSum & 0xffff) + (localSum >> 16);
--- a/Packet++/src/PacketUtils.cpp
+++ b/Packet++/src/PacketUtils.cpp
@@ -61,8 +61,8 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
 
 	PCPP_LOG_DEBUG("Calculated checksum = " << sum << ", 0x" << std::uppercase << std::hex << result);
 
-	// We return the result in Network byte order
-	return htons(result);
+	// We return the result in BigEndian byte order
+	return htobe16(result);
 }
 
 static const uint32_t FNV_PRIME = 16777619u;
--- a/Packet++/src/PacketUtils.cpp
+++ b/Packet++/src/PacketUtils.cpp
@@ -33,9 +33,10 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
 			uint8_t *vecBytes = (uint8_t *)vec[i].buffer;
 			uint8_t lastByte = vecBytes[vec[i].len - 1];
 			PCPP_LOG_DEBUG("1 byte left, adding value: 0x" << std::uppercase << std::hex << lastByte);
-			// We have read the latest byte manually but this byte should be interpreted
+			// We have read the latest byte manually but this byte should be properly interpreted
 			// as a 0xFF on LE and a 0xFF00 on BE to have a proper checksum computation
-			localSum += le16toh(lastByte);
+			localSum += be16toh(lastByte << 8);
+
 			PCPP_LOG_DEBUG("Local sum = " << localSum << ", 0x" << std::uppercase << std::hex << localSum);
 		}
 
--- a/Tests/Packet++Test/Tests/TcpTests.cpp
+++ b/Tests/Packet++Test/Tests/TcpTests.cpp
@@ -304,3 +304,30 @@ PTF_TEST_CASE(TcpChecksumInvalidRead)
 
 	delete [] m;
 } // TcpChecksumInvalidRead
+
+PTF_TEST_CASE(TcpChecksumMultiBuffer)
+{
+	// Taken from https://en.wikipedia.org/wiki/IPv4_header_checksum#Calculating_the_IPv4_header_checksum
+	uint16_t m[4] = {0x4500, 0x0073, 0x0000, 0x4000};
+	uint16_t n[3] = {0x4011, 0xc0a8, 0x0001};
+	uint16_t o[2] = {0xc0a8, 0x00c7};
+	uint16_t checksum_expected = 0xb861;
+
+	pcpp::ScalarBuffer<uint16_t> vec[4];
+	vec[0].buffer = m;
+	vec[0].len = 8;
+	vec[1].buffer = n;
+	vec[1].len = 6;
+	vec[2].buffer = o;
+	vec[2].len = 4;
+	vec[3].buffer = &checksum_expected;
+	vec[3].len = 2;
+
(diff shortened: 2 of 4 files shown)


Packet++/src/PacketUtils.cpp
```diff
@@ -17,30 +21,32 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
...
+		// check if there is one byte left
+		if (vec[i].len % 2) {
+			// access to the last byte using an uint8_t pointer
+			uint8_t *vecBytes = (uint8_t *)vec[i].buffer;
+			uint8_t lastByte = vecBytes[vec[i].len - 1];
 			PCPP_LOG_DEBUG("1 byte left, adding value: 0x" << std::uppercase << std::hex << lastByte);
-			localSum += lastByte;
+			// We have read the latest byte manually but this byte should be interpreted
+			// as a 0xFF on LE and a 0xFF00 on BE to have a proper checksum computation
+			localSum += le16toh(lastByte);
```
guebe, line 41: on le this is a nop
on be this swaps.
this should work.

in other words we want to have the same "format" here at line 41 in local sum, as we had in line 29.

Tests/Packet++Test/Tests/TcpTests.cpp
```diff
@@ -304,3 +308,30 @@ PTF_TEST_CASE(TcpChecksumInvalidRead)
...
+	vec[3].len = 2;
+
+	uint16_t c = pcpp::computeChecksum(vec, 3);
+	// computeChecksum return in network byte order
+	PTF_ASSERT_EQUAL(c, htons(checksum_expected));
+
+	// Adding the checksum should be equal to 0x0
+	c = pcpp::computeChecksum(vec, 4);
+	PTF_ASSERT_EQUAL(c, 0);
+} // TcpChecksumInvalidRead
```
guebe, line 337: nice test.

verfied on big endian:
`root@ATL00058:/mnt/c/devel# gcc PacketUtils.cpp
root@ATL00058:/mnt/c/devel# file a.out
a.out: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=421c75fb1259673a2f88a05c9a75bc844fed5b28, for GNU/Linux 3.2.0, not stripped
root@ATL00058:/mnt/c/devel# ./a.out
61b8
0000
root@ATL00058:/mnt/c/devel# powerpc-linux-gnu-gcc -static PacketUtils.cpp
root@ATL00058:/mnt/c/devel# file a.out
a.out: ELF 32-bit MSB executable, PowerPC or cisco 4500, version 1 (SYSV), statically linked, BuildID[sha1]=843d1a9cf062cf62d025d34a4899a07af97333fc, for GNU/Linux 3.2.0, not stripped
root@ATL00058:/mnt/c/devel# ./a.out
b861
0000`

guebe (approved):

Packet++/src/PacketUtils.cpp
guebe: would this be easier to read?:

            uint16_t lastSum = 0;
            memcpy(&lastSum, &lastByte, 1);
            localSum += lastSum;

Packet++/src/PacketUtils.cpp
```diff
@@ -49,14 +55,16 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
...
+	// To obtain the checksum we take the ones' complement of this result
+	uint16_t result = sum;
+	result = ~result;
 
-	PCPP_LOG_DEBUG("Calculated checksum = " << sum << ", 0x" << std::uppercase << std::hex << sum);
+	PCPP_LOG_DEBUG("Calculated checksum = " << sum << ", 0x" << std::uppercase << std::hex << result);
 
-	return ((uint16_t) sum);
+	// We return the result in Network byte order
+	return htons(result);
```
guebe, line 67: isn't there more change needed in the calling code if you change the result of this function?
I mean different output now on LE (the arch where this lib is most used on)

clementperon: the return was already returned in Network byte order:
I have removed the following line
`localSum = be16toh(localSum);`

Packet++/src/PacketUtils.cpp
```diff
@@ -1,4 +1,8 @@
+#ifndef _WIN32
+#include <arpa/inet.h>
```
guebe, line 2: perhaps use the macros from EndianPortable.h instead?
Its already included a few lines below.

Tests/Packet++Test/Tests/TcpTests.cpp
```diff
@@ -304,3 +308,30 @@ PTF_TEST_CASE(TcpChecksumInvalidRead)
```
guebe, line 308: the test above should fail on little endian after your change - it will need another htons.

Packet++/src/PacketUtils.cpp
clementperon: true

Tests/Packet++Test/Tests/TcpTests.cpp
clementperon: No the CI passed correctly

Packet++/src/PacketUtils.cpp
guebe, line 67: not obvious in the first place - but this is a nicer way than doing it on temp sums (line 43).

Tests/Packet++Test/Tests/TcpTests.cpp
guebe: yeah - I saw in the meantime - I also redid the test locally - then I found out that the format did not change, the code was just rearranged ...

Packet++/src/PacketUtils.cpp
```diff
@@ -17,30 +18,35 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
...
+		// check if there is one byte left
+		if (vec[i].len % 2) {
+			// access to the last byte using an uint8_t pointer
+			uint8_t *vecBytes = (uint8_t *)vec[i].buffer;
+			uint8_t lastByte = vecBytes[vec[i].len - 1];
 			PCPP_LOG_DEBUG("1 byte left, adding value: 0x" << std::uppercase << std::hex << lastByte);
-			localSum += lastByte;
+			// We have read the latest byte manually but this byte should be interpreted
+			// as a 0xFF on LE and a 0xFF00 on BE to have a proper checksum computation
+			uint16_t lastSum;
```
guebe, line 38:
```suggestion
			uint16_t lastSum = 0;

```

guebe: without this the second byte of lastSum has undefined content

guebe (changes requested):
see comment

Packet++/src/PacketUtils.cpp
```diff
@@ -35,7 +35,10 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
 			PCPP_LOG_DEBUG("1 byte left, adding value: 0x" << std::uppercase << std::hex << lastByte);
 			// We have read the latest byte manually but this byte should be interpreted
 			// as a 0xFF on LE and a 0xFF00 on BE to have a proper checksum computation
-			localSum += le16toh(lastByte);
+			uint16_t lastSum;
+			memcpy(&lastSum, &lastByte, 1);
```
clementperon, line 39: @seladb proposition from @guebe, IMO the le16toh() is more explicit.

Which one do you prefer?

Packet++/src/PacketUtils.cpp
```diff
@@ -35,7 +35,10 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
 			PCPP_LOG_DEBUG("1 byte left, adding value: 0x" << std::uppercase << std::hex << lastByte);
 			// We have read the latest byte manually but this byte should be interpreted
 			// as a 0xFF on LE and a 0xFF00 on BE to have a proper checksum computation
-			localSum += le16toh(lastByte);
+			uint16_t lastSum = 0;
+			memcpy(&lastSum, &lastByte, 1);
```
clementperon, line 39: @seladb proposition from @guebe, IMO the le16toh() is more explicit.

Which one do you prefer?

guebe: for me both would be fine ...

Packet++/src/PacketUtils.cpp
clementperon: maybe i'm wrong but i think the previous way was wrong as then the carry count would be wrong.

guebe: I think it was also OK. because of https://datatracker.ietf.org/doc/html/rfc1071 Page 3, (B)  Byte Order Independence

guebe: btw. why do we do the carry twice? once on the local sum and once on sum. Should be enough to do it only on sum.

please have a look at the original version of tcpdump. I very much like the simplicity of it:
https://github.com/the-tcpdump-group/tcpdump/blob/a91d20cc8491f2afcc04e51474244710481244b9/print-ip.c#L276

also the implementation and comment of the next function is notable:
https://github.com/the-tcpdump-group/tcpdump/blob/a91d20cc8491f2afcc04e51474244710481244b9/print-ip.c#L312

Packet++/src/PacketUtils.cpp
```diff
@@ -17,30 +18,35 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
...
+			localSum += lastSum;
+
 			PCPP_LOG_DEBUG("Local sum = " << localSum << ", 0x" << std::uppercase << std::hex << localSum);
 		}
 
+		// carry count is added to the sum
 		while (localSum>>16)
 		{
 			localSum = (localSum & 0xffff) + (localSum >> 16);
 		}
```
guebe, line 49: btw. why do we do the carry twice? once on the local sum and once on sum. Should be enough to do it only on sum.

please have a look at the original version of tcpdump. I very much like the simplicity of it:
https://github.com/the-tcpdump-group/tcpdump/blob/a91d20cc8491f2afcc04e51474244710481244b9/print-ip.c#L276

also the implementation and comment of the next function is notable:
https://github.com/the-tcpdump-group/tcpdump/blob/a91d20cc8491f2afcc04e51474244710481244b9/print-ip.c#L312

guebe: I think I was wrong. Not adding the carrys of local sum will give us problems for very long messages.

3rdParty/EndianPortable/include/EndianPortable.h
```diff
@@ -207,6 +207,10 @@
 #define htole32(x) le32toh(x)
 #endif
 
+#ifndef _WIN32
+#include <arpa/inet.h> // for htonl()
+#endif
```
seladb, line 212: I'm not sure why this change is needed, where do we use `htonl` in `computeChecksum()`?

Packet++/src/PacketUtils.cpp
seladb: yes, I think it'd be cleaner to use `be16toh()`. We can assume `be` because network order is always BE

Packet++/src/PacketUtils.cpp
seladb: Yes, we need to add the carry 😃

Packet++/src/PacketUtils.cpp
```diff
@@ -49,14 +55,16 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
 	{
 		sum = (sum & 0xffff) + (sum >> 16);
 	}
-
 	PCPP_LOG_DEBUG("Sum before invert = " << sum << ", 0x" << std::uppercase << std::hex << sum);
 
-	sum = ~sum;
+	// To obtain the checksum we take the ones' complement of this result
+	uint16_t result = sum;
+	result = ~result;
```
seladb, line 62: I'm not sure why the extra `result` variable is needed. What's wrong with the previous code?

Packet++/src/PacketUtils.cpp
clementperon: done

Packet++/src/PacketUtils.cpp
clementperon: to have a proper print std::hex << result;

3rdParty/EndianPortable/include/EndianPortable.h
clementperon: We use htons() i Added the comment because we use htonl() in htonll()

seladb: I'm not sure I understand... where in `computeChecksum()` do we use `htonl` or htonll`? In other words - what will fail if we remove this change?

Packet++/src/PacketUtils.cpp
```diff
@@ -49,14 +53,16 @@ uint16_t computeChecksum(ScalarBuffer<uint16_t> vec[], size_t vecSize)
...
+	// To obtain the checksum we take the ones' complement of this result
+	uint16_t result = sum;
+	result = ~result;
 
-	PCPP_LOG_DEBUG("Calculated checksum = " << sum << ", 0x" << std::uppercase << std::hex << sum);
+	PCPP_LOG_DEBUG("Calculated checksum = " << sum << ", 0x" << std::uppercase << std::hex << result);
 
-	return ((uint16_t) sum);
+	// We return the result in Network byte order
+	return htons(result);
```
seladb, line 65: Since we have `3rdParty/EndianPortable/include/EndianPortable.h` We don't need to use `htons`, `htonl` or `htonll` directly (which are sometimes platform specific). Instead we should use `htobe16()`

Packet++/src/PacketUtils.cpp
seladb: oh ok, that makes sense 👍

3rdParty/EndianPortable/include/EndianPortable.h
clementperon: i have switched htons to htobe16

Packet++/src/PacketUtils.cpp
clementperon: done

3rdParty/EndianPortable/include/EndianPortable.h
```diff
@@ -61,25 +61,25 @@
...
   #define htole32(x) (x)
-  #define be32toh(x) ntohl(x)
+  #define be32toh(x) __builtin_bswap32(x)
   #define le32toh(x) (x)
 
-  #define htobe64(x) htonll(x)
+  #define htobe64(x) __builtin_bswap64(x)
   #define htole64(x) (x)
-  #define be64toh(x) ntohll(x)
+  #define be64toh(x) __builtin_bswap64(x)
```
seladb, line 82: Why is this change needed? Wouldn't it be simpler to keep the previous code? That way we don't need to add a special `#ifedf` for MSVC

clementperon: But htonll isn't defined and is it guarantee that network byte order is always big endian ?

clementperon: If you want to keep the previous format,  the Big Endian on windows will fail as it's using GCC built-in

seladb: `htonll` isn't defined in which platform? We currently don't use it so I'm not sure it matters...
Anyway I assume this is not needed for this PR, is that correct?
Maybe we can remove this change from this PR and think later if we want to add it or not. What do you think?

clementperon: We don't use but C preprocessor comply that the macro doesn't exist.

Yes not related to this MR, I will drop these changes

clementperon: @seladb done

PR #887, opened 2022-05-19
FTP Protocol Support

egecetin:
Since FTP-Data requires session tracking only FTP control commands (Req/Resp) added.


seladb:
Sure @egecetin I'll take a look soon. I just want to finish the C++11 work first...

Packet++/header/FtpLayer.h
```diff
@@ -0,0 +1,458 @@
...
+		/**
+		 * Empty c'tor
+		 */
+		FtpRequestLayer();
+
+		/**
+		 * Set the command of request message
+		 * @param[in] code Value to set command
+		 */
+		void setCommand(FtpCommand code);
```
seladb, line 223: Maybe this method should accept as parameters both the request code and the options/arguments?
```cpp
void setCommand(FtpCommand code, const std::string& arguments);
```

This will simplify the code in `LineBasedProtocol`

Tests/Packet++Test/main.cpp
```diff
@@ -250,8 +250,12 @@ int main(int argc, char* argv[])
 	PTF_RUN_TEST(NtpParsingV4Tests, "ntp");
 	PTF_RUN_TEST(NtpCraftingTests, "ntp");
 
-	PTF_RUN_TEST(TelnetCommandParsingTests, "telnet");
+  	PTF_RUN_TEST(TelnetCommandParsingTests, "telnet");
 	PTF_RUN_TEST(TelnetDataParsingTests, "telnet");
 
+	PTF_RUN_TEST(FtpParsingIpv4Tests, "ftp");
+	PTF_RUN_TEST(FtpParsingIpv6Tests, "ftp");
```
seladb, line 257: I'm not sure we need 2 separate tests for IPv4 and IPv6 - there is no difference in parsing the FTP layer itself. The test can include both IPv4  and IPv6 packets

Tests/Packet++Test/main.cpp
```diff
@@ -250,8 +250,12 @@ int main(int argc, char* argv[])
 	PTF_RUN_TEST(NtpParsingV4Tests, "ntp");
 	PTF_RUN_TEST(NtpCraftingTests, "ntp");
 
-	PTF_RUN_TEST(TelnetCommandParsingTests, "telnet");
+  	PTF_RUN_TEST(TelnetCommandParsingTests, "telnet");
 	PTF_RUN_TEST(TelnetDataParsingTests, "telnet");
 
+	PTF_RUN_TEST(FtpParsingIpv4Tests, "ftp");
+	PTF_RUN_TEST(FtpParsingIpv6Tests, "ftp");
+	PTF_RUN_TEST(FtpCraftingTests, "ftp");
```
seladb, line 258: Can we add another test for updating an existing FTP request/response?

Packet++/header/Layer.h
```diff
@@ -5,6 +5,7 @@
 #include <stdio.h>
 #include "ProtocolType.h"
 #include <string>
+#include <sstream> // VS removed the iostreams stuff from string in VS2019
```
seladb, line 8: Why is this change needed? We didn't change anything in this file in this PR... 🤔

Packet++/src/LineBasedProtocol.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+		// Find Hyphen "-" if exists
+		pos = (uint8_t *)memchr(m_Data, 0x2d, maxLen);
+		if (pos)
+			return pos - m_Data;
+
+		return m_DataLen;
+	}
+
+	void LineBasedProtocolMessage::changeDelimiter(bool toHyphen)
```
seladb, line 30: nit: we can call this method `setDelimiter(bool hyphen)`

Packet++/src/LineBasedProtocol.cpp
```diff
@@ -0,0 +1,126 @@
...
+	{
+		if (toHyphen)
+			memset(&m_Data[getOptionOffset()], 0x2d, 1);
+		else
+			memset(&m_Data[getOptionOffset()], 0x20, 1);
```
seladb, line 35: it'd be nicer to add macros at the top of the file for hyphen (`0x2d`) space (`0x20`)

Packet++/src/LineBasedProtocol.cpp
```diff
@@ -0,0 +1,126 @@
...
+		uint8_t *pos = (uint8_t *)memchr(m_Data, 0x20, maxLen);
+		if (pos)
+			return pos - m_Data;
+
+		// Find Hyphen "-" if exists
+		pos = (uint8_t *)memchr(m_Data, 0x2d, maxLen);
+		if (pos)
+			return pos - m_Data;
+
+		return m_DataLen;
```
seladb, line 27: if the command doesn't have arguments (i.e `FEAT\r\n`) this method will return the extra `\r\n`. I think we should omit that, no?

Packet++/header/LineBasedProtocol.h
```diff
@@ -0,0 +1,55 @@
...
+ * \namespace pcpp
+ * \brief The main namespace for the PcapPlusPlus lib
+ */
+namespace pcpp
+{
+
+	/**
+	 * Class for general line based protocol (FTP, SMTP) message
+	 */
+	class LineBasedProtocolMessage : public Layer
```
seladb, line 18: Is there a "formal" name for this format that is used in FTP and SMTP?
I'm not sure "line-based" is very clear...

Packet++/header/LineBasedProtocol.h
```diff
@@ -0,0 +1,55 @@
...
+		void setCommandField(std::string value);
+		void setOptionField(std::string value);
+
+		std::string getCommandField() const;
+		std::string getOptionField() const;
+
+	public:
+
+		/**
+		 * Checks if the current message is a multi-line reply. Multi-line messages are indicatied with a Hyphen (-) immediately after reply code.
```
seladb, line 38: There is a typo in the word `indicatied`

Packet++/header/FtpLayer.h
```diff
@@ -0,0 +1,458 @@
...
+
+		/**
+		 * A static method that checks whether the port is considered as FTP
+		 * @param[in] port The port number to be checked
+		 */
+		static bool isFtpPort(uint16_t port) { return port == 21; }
+
+		// overridden methods
+
+		/// Parses the next layer. FTP is the always last so does nothing for this layer
```
seladb, line 34: nit: no need in "Parses the next layer."

Packet++/header/FtpLayer.h
```diff
@@ -0,0 +1,458 @@
...
+		/**
+		 * @return Get the size of the layer
+		 */
+		size_t getHeaderLen() const { return m_DataLen; }
+
+		/// Does nothing for this layer
+		void computeCalculateFields() {}
+
+		/**
+		 * @return The OSI layer level of Telnet (Application Layer).
```
seladb, line 46: Typo: "Telnet" instead of "FTP"

seladb (commented):
@egecetin I did a first pass, please take a look at the comments and I'll do a second pass a bit later

Packet++/header/Layer.h
egecetin: I'm not sure why the changes affect the whole code but when I remove this line it gives error on the other layers. I will share the compile log for Windows. Maybe, you will have an idea

Packet++/header/LineBasedProtocol.h
egecetin: To be honest, I don't know too 😁. They are actually TextBased but differ slightly from HTTP. I will look for a better naming until your second pass.

Packet++/src/LineBasedProtocol.cpp
egecetin: I think it should return extra characters to indicate there is no option. But I added the fix to `LineBasedProtocolMessage::getCommandField()` to return the correct stripped string to the user.

Tests/Packet++Test/main.cpp
seladb: @egecetin did you add a test to update an existing layer (`FtpEditTest`)

egecetin: @seladb I added lines to `FtpCraftingTests`,

https://github.com/egecetin/PcapPlusPlus/blob/c841c2d7370eac399d708a2690e4d74355ec7a0e/Tests/Packet%2B%2BTest/Tests/FtpTests.cpp#L149-L176

Tests/Packet++Test/Tests/FtpTests.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+	pcpp::IPv4Layer ipv4Layer1(*ftpPacket1.getLayerOfType<pcpp::IPv4Layer>());
+	PTF_ASSERT_TRUE(craftedPacket1.addLayer(&ipv4Layer1));
+
+	pcpp::TcpLayer tcpLayer1(*ftpPacket1.getLayerOfType<pcpp::TcpLayer>());
+	PTF_ASSERT_TRUE(craftedPacket1.addLayer(&tcpLayer1));
+
+	pcpp::FtpRequestLayer ftpReqLayer1;
+	ftpReqLayer1.setCommand(pcpp::FtpRequestLayer::USER);
+	ftpReqLayer1.setCommandOption(std::string("csanders"));
```
seladb, line 119: I don't think `std::string()` is required here, we can just do:
```cpp
ftpReqLayer1.setCommandOption("csanders");
```

Tests/Packet++Test/Tests/FtpTests.cpp
```diff
@@ -0,0 +1,177 @@
...
+	pcpp::IPv4Layer ipv4Layer2(*ftpPacket2.getLayerOfType<pcpp::IPv4Layer>());
+	PTF_ASSERT_TRUE(craftedPacket2.addLayer(&ipv4Layer2));
+
+	pcpp::TcpLayer tcpLayer2(*ftpPacket2.getLayerOfType<pcpp::TcpLayer>());
+	PTF_ASSERT_TRUE(craftedPacket2.addLayer(&tcpLayer2));
+
+	pcpp::FtpResponseLayer ftpReqLayer2;
+	ftpReqLayer2.setStatusCode(pcpp::FtpResponseLayer::SYSTEM_STATUS);
+	ftpReqLayer2.setStatusOption(
+		std::string("Extensions supported:\r\n CLNT\r\n MDTM\r\n PASV\r\n REST STREAM\r\n SIZE\r\n211 End."));
```
seladb, line 142: ditto: `std::string()` is not required here

Tests/Packet++Test/Tests/FtpTests.cpp
```diff
@@ -0,0 +1,177 @@
...
+	READ_FILE_AND_CREATE_PACKET(3, "PacketExamples/ftpIpv4Req.dat");
+	pcpp::Packet ftpPacket3(&rawPacket3);
+	pcpp::FtpRequestLayer *ftpLayer3 = ftpPacket3.getLayerOfType<pcpp::FtpRequestLayer>();
+
+	PTF_ASSERT_NOT_NULL(ftpLayer3);
+	ftpLayer3->setCommand(pcpp::FtpRequestLayer::FEAT);
+	PTF_ASSERT_EQUAL(ftpLayer3->getCommand(), pcpp::FtpRequestLayer::FEAT);
+	PTF_ASSERT_EQUAL(ftpLayer3->getCommandOption(), "csanders");
+
+	ftpLayer3->setCommandOption(std::string("Test option"));
```
seladb, line 158: ditto: `std::string()` is not required here

Packet++/header/FtpLayer.h
```diff
@@ -0,0 +1,458 @@
...
+namespace pcpp
+{
+
+	/**
+	 * Class for general FTP message
+	 */
+	class FtpLayer : public LineBasedProtocolMessage
+	{
+	protected:
+		FtpLayer(uint8_t *data, size_t dataLen, Layer *prevLayer, Packet *packet) : LineBasedProtocolMessage(data, dataLen, prevLayer, packet) {};
```
seladb, line 21: I think this c'tor can set `m_Protocol = FTP;` instead of setting it in the c'tor of `FtpRequestLayer` and `FtpResponseLayer`

Packet++/header/FtpLayer.h
```diff
@@ -0,0 +1,458 @@
...
+		 * @param[in] dataLen Size of the data in bytes
+		 * @param[in] prevLayer A pointer to the previous layer
+		 * @param[in] packet A pointer to the Packet instance where layer will be stored in
+		 */
+		FtpRequestLayer(uint8_t *data, size_t dataLen, Layer *prevLayer, Packet *packet) : FtpLayer(data, dataLen, prevLayer, packet) { m_Protocol = FTP; };
+
+		/**
+		 * Empty c'tor
+		 */
+		FtpRequestLayer();
```
seladb, line 217: Maybe this c'tor shouldn't be empty and accept the command and (optional) argument as parameters?
There is no point in an FTP message that doesn't have a command so it makes sense to "force" the user to supply it in the c'tor. The argument is optional but we should also enable to pass it for easier creation of this layer

Packet++/header/FtpLayer.h
```diff
@@ -0,0 +1,458 @@
...
+		 * @param[in] dataLen Size of the data in bytes
+		 * @param[in] prevLayer A pointer to the previous layer
+		 * @param[in] packet A pointer to the Packet instance where layer will be stored in
+		 */
+		FtpResponseLayer(uint8_t *data, size_t dataLen, Layer *prevLayer, Packet *packet) : FtpLayer(data, dataLen, prevLayer, packet) { m_Protocol = FTP; };
+
+		/**
+		 * Empty c'tor
+		 */
+		FtpResponseLayer();
```
seladb, line 409: Same here: this c'tor shouldn't be empty and accept the status code and (optional) argument, like:
```cpp
pcpp::FtpResponseLayer ftpResLayer(pcpp::FtpResponseLayer::SYSTEM_STATUS, "bla bla");
```

Tests/Packet++Test/Tests/FtpTests.cpp
```diff
@@ -0,0 +1,177 @@
...
+	pcpp::EthLayer ethLayer2(*ftpPacket2.getLayerOfType<pcpp::EthLayer>());
+	PTF_ASSERT_TRUE(craftedPacket2.addLayer(&ethLayer2));
+
+	pcpp::IPv4Layer ipv4Layer2(*ftpPacket2.getLayerOfType<pcpp::IPv4Layer>());
+	PTF_ASSERT_TRUE(craftedPacket2.addLayer(&ipv4Layer2));
+
+	pcpp::TcpLayer tcpLayer2(*ftpPacket2.getLayerOfType<pcpp::TcpLayer>());
+	PTF_ASSERT_TRUE(craftedPacket2.addLayer(&tcpLayer2));
+
+	pcpp::FtpResponseLayer ftpReqLayer2;
```
seladb, line 139: this variable should be called `ftpResponeLayer`, no?

Packet++/src/FtpLayer.cpp
```diff
@@ -0,0 +1,396 @@
...
+
+	void FtpRequestLayer::setCommand(FtpCommand code)
+	{
+		setCommandField(getCommandAsString(code));
+	}
+
+	FtpRequestLayer::FtpCommand FtpRequestLayer::getCommand() const
+	{
+		size_t val = 0;
+		std::string field = getCommandField();
```
seladb, line 24: Maybe it's better to use `getCommandString()` here? 🤔

Packet++/src/FtpLayer.cpp
```diff
@@ -0,0 +1,396 @@
...
+
+		for (size_t idx = 0; idx < field.size(); ++idx)
+			val |= (field.c_str()[idx] << (idx * 8));
+
+		return static_cast<FtpCommand>(val);
```
seladb, line 29: there is a question of what would happen if the command in the packet is an invalid command, for example: `XXXX`. Usually it's a good practice to check the value before using `static_cast`, but here it would be very difficult. I'm not sure I have a good suggestion here...

Packet++/src/FtpLayer.cpp
```diff
@@ -0,0 +1,396 @@
...
+	void FtpResponseLayer::setStatusCode(FtpStatusCode code)
+	{
+		std::ostringstream oss;
+		oss << code;
+		setCommandField(oss.str());
+	}
+
+	FtpResponseLayer::FtpStatusCode FtpResponseLayer::getStatusCode() const
+	{
+		return static_cast<FtpStatusCode>(atoi(getCommandField().c_str()));
```
seladb, line 241: Same here: what should happen if the status code is not valid?

Packet++/header/FtpLayer.h
```diff
@@ -0,0 +1,458 @@
...
+		 * @return std::string Value of command argument
+		 */
+		std::string getCommandOption(bool removeEscapeCharacters = true) const;
+
+		/**
+		 * Convert the command info to readable string
+		 * @param[in] code Command code to convert
+		 * @return std::string Returns the command info as readable string
+		 */
+		static std::string getCommandInfoAsString(FtpCommand code);
```
seladb, line 255: We can rename to just `getCommandInfo()`, the user can see it returns a `std::string` 😄

Packet++/header/LineBasedProtocol.h
seladb: Actually maybe `SingleCommandTextProtocol`?

Tests/Packet++Test/Tests/FtpTests.cpp
```diff
@@ -0,0 +1,177 @@
...
+	ftpReqLayer2.setStatusCode(pcpp::FtpResponseLayer::SYSTEM_STATUS);
+	ftpReqLayer2.setStatusOption(
+		std::string("Extensions supported:\r\n CLNT\r\n MDTM\r\n PASV\r\n REST STREAM\r\n SIZE\r\n211 End."));
+	PTF_ASSERT_TRUE(craftedPacket2.addLayer(&ftpReqLayer2));
+
+	PTF_ASSERT_EQUAL(bufferLength2, craftedPacket2.getRawPacket()->getRawDataLen());
+	PTF_ASSERT_BUF_COMPARE(buffer2, craftedPacket2.getRawPacket()->getRawData(), bufferLength2);
+
+	// Modify existing packets
```
seladb, line 148: Usually when modifying packets, one of the best ways to test is by saving it to raw data and check that we get what we expect. What you can do is:
- Write custom code to save the packet raw data to a pcap file
- Open it in Wireshark to make sure you get the expected result
- Convert the pcap file to `.dat` and save it in `PacketExamples/` folder
- Remove the code that saves the raw data to file
- Load the `.dat` file in the test
- Use `PTF_ASSERT_BUF_COMPARE` to compare the packet data with the content of the `.dat` file, like [here](https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Packet%2B%2BTest/Tests/SipSdpTests.cpp#L624)

seladb, line 148: In most cases we separate "crafting" and "edit" to different test cases: `[ProtocolLayer]CreationTest`, `[ProtocolLayer]EditTest`. Maybe we should separate here too?

Packet++/header/FtpLayer.h
egecetin: @seladb I'm not sure (or understand) how will this change simplify the code in `LineBasedProtocol` (or `SingleCommandTextProtocol` 😉) because there are four different combinations theoretically (code empty but argument, code and argument, code but argument empty, both empty), Additionally current packet can have or not have an option, this increases the combinations too. So, all of them should be checked.

Packet++/src/FtpLayer.cpp
egecetin: @seladb Looks like for C+11 there is no problem (no Undefined Behaviour) if I understand correctly (see accepted answer [here](https://stackoverflow.com/questions/18195312/what-happens-if-you-static-cast-invalid-value-to-enum-class)) but for ensure safety maybe we can use an underlying type for enum (discussed [here](https://stackoverflow.com/questions/55392083/c-enum-class-cast-to-non-existing-entry)). Since the commands have at most four characters `int` is sufficient. What do you think?

For the user side, if anyone is parsing a non-standard command value, the user should convert it to integer and compare it with its custom values.

Packet++/header/Layer.h
egecetin: Looks like I found the solution. I should include `sstream` before the `Layer.h` at `SingleCommandTextProtocol.h`

Packet++/src/FtpLayer.cpp
seladb: Yes that's a good idea! we can convert the enum to `enum class FtpCommand : int`, and add an `Unknown` value as the default value. Then casting will be safe,

We can probably do the same for `FtpStatusCode`

Packet++/header/FtpLayer.h
```diff
@@ -0,0 +1,466 @@
...
+		 * @param[in] packet A pointer to the Packet instance where layer will be stored in
+		 */
+		FtpResponseLayer(uint8_t *data, size_t dataLen, Layer *prevLayer, Packet *packet) : FtpLayer(data, dataLen, prevLayer, packet) {};
+
+		/**
+		 * A constructor that creates layer with provided input values
+		 * @param[in] code Status code
+		 * @param[in] option Argument of the status code
+		 */
+		FtpResponseLayer(const FtpStatusCode &code, std::string option = "") : FtpLayer(std::to_string(code), option) {};
```
seladb, line 415: The `option` argument can also be `const &`, no? 🤔

Tests/Packet++Test/Tests/FtpTests.cpp
egecetin: I'll update it but I want to work on other comments first

seladb (approved):

PR #895, opened 2022-05-30
Add SOME/IP protocol

kolbex:
Add SOME/IP layer support
* Add SomeIPLayer class for handling SOME/IP packets
* Support for crafting and parsing packets
* Add tests
* Update README.md file

SOME/IP TP and SOME/IP SD support will follow in separate PRs.


Packet++/src/SomeIpLayer.cpp
```diff
@@ -0,0 +1,379 @@
...
+}
+
+bool SomeIpLayer::isSomeIpProtocol(uint8_t *data, uint32_t dataLen)
+{
+	/* Ideas taken from wireshark some ip dissector */
+	if (dataLen < SOMEIP_HDR_LEN)
+		return false;
+
+	uint32_t *p32 = (uint32_t *)data;
+	uint32_t lengthBE = *(p32 + 1); // length field in someip header
```
guebe, line 143: this is not allowed. undefined behaviour. unaligned access.

guebe: the problem might be strict aliasing here (automatically enabled in gcc in -O2, -O3 and -Os):
The same object shall not be references by a pointer to uint32_t and to struct someiphdr.
That would also explain why it does not happen in similar code in other layers which also violates alignment constraints.

guebe:
@seladb can you please restart build?, we tried to fix a regression, however it seams now an unrelated memory leak check fired.

guebe:
> Done!

thx. We cant reproduce the regressions locally. We tried both locally mingw-w64 and vs 2019 release. But all tests are passing locally using those compilers. Is there a way to get the build artifacts from the test system, so we can debug them locally? Do you have other suggestions?

Packet++/src/SomeIpLayer.cpp
```diff
@@ -0,0 +1,379 @@
...
+	hdr->length = htobe32(sizeof(someiphdr) - sizeof(hdr->serviceID) - sizeof(hdr->methodID) - sizeof(hdr->length) +
+						  payloadLength);
+}
+
+uint32_t SomeIpLayer::getHeaderLengthWithPayload() const
+{
+	someiphdr *hdr = (someiphdr *)m_Data;
+
+	/* Add outer header of someip to length (messageId and length field itself) */
+	return sizeof(uint32_t) * 2 + be32toh(hdr->length);
```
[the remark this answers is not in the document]
guebe: thats how all other layers do it. makes no sense to make this upper layer differently. -> closed

Tests/Packet++Test/PacketExamples/someip.dat
```diff
@@ -0,0 +1 @@
+027dfa001001027dfa0117408100000286dd600d18bf003a0640fd537cb8038300020000000000010117fd537cb80383000e0000000000000014727471fc0026ae1b000019755018102c7e2200006059410c0000001e0003000a0105000040001000000000000000000085000000000000400100
```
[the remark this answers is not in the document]
guebe: thank you very much! 

I looked into this in more detail.
Reason for different behavior seems we are relying on implementation defined tellg() function output in TestUtils.cpp function getFileLength(). Most likely the test system uses other runtime than my local one (in mingw-64 you can use msvc runtime or UCRT (a newer one)).

Similar issue and reason for wrong behavior was discussed here:
https://stackoverflow.com/questions/22984956/tellg-function-give-wrong-size-of-file
https://stackoverflow.com/questions/5840148/how-can-i-get-a-files-size-in-c

It seems to get defined behavior we have to use fstat() function.

Anyhow this bug does not relate to this ticket - so we will just fix the .dat file 0x0a at end of file for now.

Packet++/src/SomeIpLayer.cpp
```diff
@@ -0,0 +1,379 @@
...
+	// TODO: handle SomeIPSD/SomeIPTP packets
+}
+
+bool SomeIpLayer::isSomeIpPort(uint16_t port)
+{
+	return std::any_of(mSomeIpPorts.begin(), mSomeIpPorts.end(),
+					   [&](const uint16_t &someIpPort) { return someIpPort == port; });
+}
+
+void SomeIpLayer::addSomeIpPort(uint16_t port)
```
miketsukerman, line 131: Just trying to understand. Is it expected that one who uses the implementation should first set some range of ports used for SOME/IP communication?

Packet++/src/SomeIpLayer.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+	uint8_t *payload = m_Data + headerLen;
+	size_t payloadLen = m_DataLen - headerLen;
+
+	// TODO: handle SomeIPSD/SomeIPTP packets
+
+	if (isSomeIpProtocol(payload, payloadLen))
+	/* else parse as SomeIpLayer (i.e. respect chaining of SomeIpLayers in one Ethernet frame) */
+	{
+		m_NextLayer = new SomeIpLayer(payload, payloadLen, this, m_Packet);
```
miketsukerman, line 372: Just for my understanding. There are some comments related to the SOME/IP service discovery. But here we assume that next layer is also plain SOME/IP layer. I was thinking about making a separate layer for SOME/IP SD messages. Please excuse me if I misunderstood the logic.

miketsukerman:
> SOME/IP TP and SOME/IP SD support will follow in separate PRs.

Would be cool to see or to, at least, understand how do you plan to parse SD messages.

Packet++/src/SomeIpLayer.cpp
maruu: Hi Michael, thanks for having time to take a look at the code.
Our current implementation works like this: We get a SOME/IP-SD packet and read the endpoints from it. We add the ports of the endpoints to this list of ports, so further SOME/IP communication on these ports is detected correctly. We could also do this autonomously in the SOME/IP-SD layer, but our first implementation outsourced this action to the application. 
Since SOME/IP-SD traffic is usually sent on the port 30490, we are checking for that port specifically, so we do not miss the first SOME/IP-SD packets.

What do you think about it?

Packet++/src/SomeIpLayer.cpp
maruu: We currently removed the code which creates the SOME/IP-SD layer, because we first wanted to upstream the base protocol, and later extend it with the SD and TP implementation. We decided to not upstream commented out code, which is why the reference to SD and TP is missing here. Sorry for the confusion.
The question why we are parsing further payload as a plain SOME/IP layer is a very good one. AUTOSAR specifies that a UDP datagram can contain multiple SOME/IP layers behind each other. This code implements this logic for the plain SOME/IP protocol.

maruu:
> > SOME/IP TP and SOME/IP SD support will follow in separate PRs.
> 
> Would be cool to see or to, at least, understand how do you plan to parse SD messages.

Our preference would have been to base the PRs for adding SOME/IP-SD and SOME/IP-TP on origin/dev as soon as the plain SOME/IP PR is merged, but if it makes review easier for you, we can also already create the PRs and base them on the plain SOME/IP PR branch.
Please let me know what you would prefer, so we know whether we should create the PRs already. Thanks

Packet++/src/SomeIpLayer.cpp
miketsukerman: Hi @maruu,

Sorry for the late reply. I was thinking as well about how to detect SOME/IP traffic properly. You have an interesting way of doing it. I do not have anything against this approach at the moment. ;)

Packet++/src/SomeIpLayer.cpp
miketsukerman: Frankly, on practice never seen nested SOME/IP packets except SD messages (which is a bit of different stuff). Yes, you are right, I have forgotten about this 

https://www.autosar.org/fileadmin/user_upload/standards/foundation/1-0/AUTOSAR_PRS_SOMEIPProtocol.pdf
```
[PRS_SOMEIP_00140] d The header format allows transporting more than one
SOME/IP message in a single UDP packet. The SOME/IP implementation shall identify
the end of a SOME/IP message by means of the SOME/IP length field. Based on the
UDP length field, SOME/IP shall determine if there are additional SOME/IP messages
in the UDP packet. c(RS_SOMEIP_00010)
```

miketsukerman:
> > > SOME/IP TP and SOME/IP SD support will follow in separate PRs.
> > 
> > 
> > Would be cool to see or to, at least, understand how do you plan to parse SD messages.
> 
> Our preference would have been to base the PRs for adding SOME/IP-SD and SOME/IP-TP on origin/dev as soon as the plain SOME/IP PR is merged, but if it makes review easier for you, we can also already create the PRs and base them on the plain SOME/IP PR branch. Please let me know what you would prefer, so we know whether we should create the PRs already. Thanks

I am okay with the idea to finish current MR first. But, in theory, you can also split changes into several commits even in this MR. But, well, up to you.

Packet++/header/SomeIpLayer.h
```diff
@@ -0,0 +1,435 @@
...
+		return (m_PayloadPtr == nullptr) ? m_Data + sizeof(someiphdr) : m_PayloadPtr;
+	}
+
+	/**
+	 * @return The size in bytes of the payload
+	 */
+	size_t getPduPayloadSize() const
+	{
+		return getHeaderLengthWithPayload() - sizeof(someiphdr);
+	}
```
seladb, line 370: We already have `getLayerPayloadSize()` in `pcpp::Layer` which should give the same result if we consider the payload as a separate layer

Packet++/header/SomeIpLayer.h
```diff
@@ -0,0 +1,435 @@
...
+	 * @param[in] type Type of the message
+	 * @param[in] returnCode Return Code
+	 * @param[in] data A pointer to the raw data
+	 * @param[in] dataLen Size of the data in bytes
+	 * @param[in] dontCopy If set to true, the payload is not copied to the layer object. In this case the m_PayloadPtr
+	 * holds the reference to a data buffer. This option can be used to reduce the number of copies to generate packets.
+	 */
+	SomeIpLayer(uint16_t serviceID, uint16_t methodID, uint16_t clientID, uint16_t sessionID, uint8_t interfaceVersion,
+				MsgType type, uint8_t returnCode, const uint8_t *const data = nullptr, size_t dataLen = 0,
+				bool dontCopy = false);
```
seladb, line 129: I think `data`, `dataLen` and `dontCopy` are redundant. Assuming we consider the payload as a separate layer, a layer shouldn't create additional layers. The user should create the `SomeIpLayer` and the payload layer separately

Packet++/src/SomeIpLayer.cpp
```diff
@@ -0,0 +1,349 @@
...
+#include "PayloadLayer.h"
+#include <algorithm>
+#include <sstream>
+#include <string.h>
+
+namespace pcpp
+{
+std::unordered_set<uint16_t> SomeIpLayer::mSomeIpPorts{};
+
+void SomeIpLayer::optionallyCopyPayloadData(const uint8_t *const data, size_t dataLen, bool dontCopy)
```
seladb, line 15: Assuming `SomeIpLayer` c'tor doesn't include the payload, this method can probably be removed and the simple initiation of  `m_Data`, `m_DataLen` and `m_Protocol` can be done in the c'tor itself

seladb (commented):
Hi @kolbex ! I left comments on the layer implementation, but didn't yet go over the tests. I suggest we address the existing comments and I'll review the tests later

kolbex:
@seladb thx for the review, i currently will focus on NDP/ICMPv6 so there will be no progress here the next week

Packet++/src/SomeIpLayer.cpp
```diff
@@ -0,0 +1,349 @@
...
+	{
+		/* If there is a protocol coming after the normal SomeIp header, return just the header length so PcapPlusPlus
+		handles the oncoming protocols correctly */
+		return sizeof(someiphdr);
+	}
+	else
+	{
+		/* Else, return the length of the header + payload, since it is a normal SomeIp PDU */
+		return getHeaderLengthWithPayload();
+	}
```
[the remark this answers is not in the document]
kolbex: @seladb: we do it this way because there can be more SOME/IP (SOME/IP-TP) frames in one packet, if getHeaderLen() always returns sizeof(someiphdr) and PayloadLayer whould contain the rest, this whould not be possible because the PayloadLayer is always the last layer. Currently subProtocolFollows() returns always false but this will be extended with SOME/IP-TP/SD. What do you think about this solution, could we let it that way?

Packet++/header/SomeIpLayer.h
kolbex: i removed the dontCopy flag

Packet++/src/SomeIpLayer.cpp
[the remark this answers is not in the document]
seladb: The way PcapPlusPlus packet parser works is that each layer is responsible for parsing the next layer, via implementing the `parseNextLayer()` method.
In this PR SOME/IP-TP and SOME/IP-SD are still missing, hence we can do the following:
- Check if the next layer is SOME/IP-TP or SOME/IP-SD
  - If yes - create a `PayloadLayer` (since we don't know how to parse these layers yet)
  - If no - check `SomeIpLayer::isDataValid(...)`:
     - If yes - create a `SomeIpLayer`
     - If no - create a `PayloadLayer` 

With this design we can keep `getHeaderLen()` very simple and can always return `sizeof(someiphdr)`, and `SomeIpLayer::parseNextLayer()` does the more heavy lifting.

Packet++/header/ProtocolType.h
```diff
@@ -267,6 +267,11 @@ namespace pcpp
 	 */
 	const ProtocolType FTP = 0x10000000000;
 
+	/**
+	 * SOME/IP Base protocol
+	 */
+	const ProtocolType SomeIP = 0x20000000000;
```
seladb, line 273: Please merge your branch from `dev` - the value `0x20000000000` is already taken by `ICMPv6`, it should be `0x40000000000`

Packet++/header/SomeIpLayer.h
```diff
@@ -0,0 +1,341 @@
...
+	OsiModelLayer getOsiModelLayer() const { return OsiModelApplicationLayer; }
+
+private:
+	static const uint8_t SOMEIP_PROTOCOL_VERSION = 1;
+
+	/* Using unordered_set since insertion and search should be almost constant time */
+	static std::unordered_set<uint16_t> m_SomeIpPorts;
+
+	bool subProtocolFollows() const;
+	static void splitUint32Id(uint32_t uint32Id, uint16_t &uint16IdUpper, uint16_t &uint16IdLower);
```
seladb, line 338: I just noticed this is a helper function that is only used within `SomeIpLayer.cpp`, hence we don't need to keep it as a private class method, instead it can be a method inside `SomeIpLayer.cpp` (without a class)

Packet++/header/SomeIpLayer.h
seladb: Since SOME/IP-SD and SOME/IP-TP will be separate layers in the next PRs, I think we can also remove `data` and `dataLen`. If users want to create multiple layers they can create a `SomeIpLayer` and on top of that they can create other `SomeIpLayer`/`SomeIpSdLayer`/`SomeIpTpLayer`

Packet++/src/TcpLayer.cpp
```diff
@@ -381,6 +382,8 @@ void TcpLayer::parseNextLayer()
 		m_NextLayer = new FtpResponseLayer(payload, payloadLen, this, m_Packet);
 	else if (FtpLayer::isFtpPort(portDst) && FtpLayer::isDataValid(payload, payloadLen))
 		m_NextLayer = new FtpRequestLayer(payload, payloadLen, this, m_Packet);
+	else if (SomeIpLayer::isSomeIpPort(portSrc) || SomeIpLayer::isSomeIpPort(portDst) || SomeIpLayer::isDataValid(payload, payloadLen))
```
seladb, line 385: This if statement means that every payload which meets the `SomeIpLayer::isDataValid()` criteria will be considered as SOME/IP, regardless of SOME/IP ports. Is this the intended behavior?

I think a safer approach could be:
```cpp
else if ((SomeIpLayer::isSomeIpPort(portSrc) || SomeIpLayer::isSomeIpPort(portDst)) && SomeIpLayer::isDataValid(payload, payloadLen))
{
    ...
}
```

Packet++/src/UdpLayer.cpp
```diff
@@ -133,6 +134,8 @@ void UdpLayer::parseNextLayer()
 		m_NextLayer = new DhcpV6Layer(udpData, udpDataLen, this, m_Packet);
 	else if ((NtpLayer::isNTPPort(portSrc) || NtpLayer::isNTPPort(portDst)) && NtpLayer::isDataValid(udpData, udpDataLen))
 		m_NextLayer = new NtpLayer(udpData, udpDataLen, this, m_Packet);
+	else if (SomeIpLayer::isSomeIpPort(portSrc) || SomeIpLayer::isSomeIpPort(portDst) || SomeIpLayer::isDataValid(udpData, udpDataLen))
```
seladb, line 137: Ditto

Packet++/header/SomeIpLayer.h
```diff
@@ -0,0 +1,435 @@
...
+	 * @param[in] version version to set
+	 */
+	void setInterfaceVersion(uint8_t version);
+
+	/**
+	 * @brief Get the message type
+	 *
+	 * @return uint8_t
+	 */
+	uint8_t getMessageType() const;
```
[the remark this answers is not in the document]
kolbex: implemented this, for the getMessageType() i used the enum class and static cast ... i hope this is the right way

Packet++/src/SomeIpLayer.cpp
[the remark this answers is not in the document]
kolbex: i updated the Branch with adding the PayloadLayer for SD/TP frames

for the getHeaderLen() method i am not sure how you how you mean this because in a SOME/IP frame there is always the header + payload and when there are two SOME/IP frames in one packet than there is header + payload + header + payload (i added a test for such a frame and updated the pcap file). So in my opinion this method can not be reduced to only sizeof(someiphdr) because it must also contain the length of the payload when it is a SOME/IP frame. Can you explain it again please?

Packet++/header/SomeIpLayer.h
kolbex: i think the data and dataLen cannot be reduced because the SOME/IP Layer also contains a payload, see also https://github.com/seladb/PcapPlusPlus/pull/895#discussion_r968200981

Packet++/src/SomeIpLayer.cpp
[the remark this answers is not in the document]
seladb: Oh ok, I missed that SOME/IP frames can have a payload, I'm sorry about that. In that case, your code is correct

Packet++/header/SomeIpLayer.h
seladb: Yes, that is correct!

Packet++/header/SomeIpLayer.h
```diff
@@ -0,0 +1,361 @@
...
+ * @class SomeIpLayer
+ * Represents a SOME/IP protocol layer
+ */
+class SomeIpLayer : public Layer
+{
+public:
+	/**
+	 * SOME/IP message types
+	 */
+	enum class MsgType : int
```
seladb, line 26: nit: since message type is 1-byte I think it can be `enum class MsgType : uint8_t`

Packet++/src/SomeIpLayer.cpp
```diff
@@ -0,0 +1,324 @@
...
+	uint32_t length = be32toh(lengthBE);
+	if ((length < 8) || (length > dataLen - 8))
+		return false;
+
+	if (data[12] != SOMEIP_PROTOCOL_VERSION)
+		return false;
+
+	someiphdr *hdr = (someiphdr *)data;
+
+	switch ((MsgType)(hdr->msgType & ~(uint8_t)MsgType::TP_REQUEST))
```
seladb, line 122: here I think we should use `static_cast`:
```cpp
switch (static_cast<MsgType>(hdr->msgType & ~(uint8_t)MsgType::TP_REQUEST))
```

Packet++/header/SomeIpLayer.h
```diff
@@ -0,0 +1,361 @@
...
+	 * @return true If the packet contains a SomeIpSd layer, false ottherwise
+	 */
+	static bool isSomeIpSdPacket(uint16_t serviceId, uint16_t methodId);
+
+	/**
+	 * @brief Checks whether the packet contains a SomeIpTp layer by checking the message type for the TP flag
+	 * @param[in] msgType MsgType that has to be checked
+	 * @return true if packet contains a SomeIpTp layer, false if otherwise
+	 */
+	static bool isSomeIpTpPacket(uint8_t msgType);
```
seladb, line 164: I'm not sure these methods should be public. For now it's ok because we don't yet have SD and TP, but once we have them the type of the layer would be `SomeIpSdLayer` or `SomeIpTpLayer`, so those methods might be confusing. What do you think?

Tests/Packet++Test/Tests/SomeIpTests.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "IPv6Layer.h"
+#include "Packet.h"
+#include "SomeIpLayer.h"
+#include "SystemUtils.h"
+#include "UdpLayer.h"
+#include "VlanLayer.h"
+#include <array>
+#include <cstring>
+
+PTF_TEST_CASE(SomeIpParsingTest)
```
seladb, line 13: We don't use `SomeIpLayer::addSomeIpPort(...)` anywhere in the test, we probably should so PcapPlusPlus knows these are SOME/IP packets

Tests/Packet++Test/Tests/SomeIpTests.cpp
```diff
@@ -0,0 +1,115 @@
...
+	gettimeofday(&time, NULL);
+
+	READ_FILE_AND_CREATE_PACKET(1, "PacketExamples/someip.dat");
+	READ_FILE_AND_CREATE_PACKET(2, "PacketExamples/someip2.dat");
+
+	pcpp::Packet someIpPacket(&rawPacket1);
+	pcpp::Packet someIpPacket2(&rawPacket2);
+
+	// Test with one SOME/IP layer
+	pcpp::SomeIpLayer *layer = someIpPacket.getLayerOfType<pcpp::SomeIpLayer>();
```
seladb, line 25: I'd add a few more tests here:
```cpp
PTF_ASSERT_TRUE(someIpPacket.isPacketOfType(pcpp::SomeIP));
pcpp::SomeIpLayer *someIpLayer = someIpPacket.getLayerOfType<pcpp::SomeIpLayer>();
PTF_ASSERT_NOT_NULL(someIpLayer);
```

Tests/Packet++Test/Tests/SomeIpTests.cpp
```diff
@@ -0,0 +1,115 @@
...
+	PTF_ASSERT_EQUAL(layer->getProtocolVersion(), 1);
+	PTF_ASSERT_EQUAL(layer->getInterfaceVersion(), 0x5);
+	PTF_ASSERT_EQUAL(layer->getMessageTypeAsInt(), (uint8_t)pcpp::SomeIpLayer::MsgType::REQUEST);
+	PTF_ASSERT_EQUAL((int)layer->getMessageType(), (int)pcpp::SomeIpLayer::MsgType::REQUEST);
+	PTF_ASSERT_EQUAL(layer->getReturnCode(), 0);
+
+	PTF_ASSERT_NULL(layer->getNextLayer());
+
+	// Test with two SOME/IP layers
+	pcpp::SomeIpLayer *layer2_1 = someIpPacket2.getLayerOfType<pcpp::SomeIpLayer>();
```
seladb, line 41: ditto: we can add:
```cpp
PTF_ASSERT_NOT_NULL(layer2_1);
```

Tests/Packet++Test/Tests/SomeIpTests.cpp
```diff
@@ -0,0 +1,115 @@
...
+	PTF_ASSERT_EQUAL(layer2_1->getClientID(), 0x3);
+	PTF_ASSERT_EQUAL(layer2_1->getSessionID(), 0xa);
+	PTF_ASSERT_EQUAL(layer2_1->getProtocolVersion(), 1);
+	PTF_ASSERT_EQUAL(layer2_1->getInterfaceVersion(), 0x5);
+	PTF_ASSERT_EQUAL(layer2_1->getMessageTypeAsInt(), (uint8_t)pcpp::SomeIpLayer::MsgType::REQUEST);
+	PTF_ASSERT_EQUAL((int)layer2_1->getMessageType(), (int)pcpp::SomeIpLayer::MsgType::REQUEST);
+	PTF_ASSERT_EQUAL(layer2_1->getReturnCode(), 0);
+	PTF_ASSERT_NOT_NULL(layer2_1->getNextLayer());
+
+	pcpp::SomeIpLayer *layer2_2 =  someIpPacket2.getNextLayerOfType<pcpp::SomeIpLayer>(layer2_1);
```
seladb, line 55: ditto: we can add:
```cpp
PTF_ASSERT_NOT_NULL(layer2_2);
```

Tests/Packet++Test/Tests/SomeIpTests.cpp
seladb, line 13: We should also test the following class methods:
- `getPduPayload()`
- `getPduPayloadSize()`
- `toString()`
- `getHeaderLen()`

Packet++/src/SomeIpLayer.cpp
```diff
@@ -0,0 +1,324 @@
...
+}
+
+void SomeIpLayer::computeCalculateFields()
+{
+	getSomeIpHeader()->protocolVersion = 0x01;
+
+	if (!m_NextLayer)
+		return;
+
+	// TODO: handle SomeIPSD/SomeIPTP packets
```
seladb, line 90: I think `SomeIpLayer` doesn't need to do anything else, right?

In the next PR `SomIpSdLayer` and `SomeIpTpLayer` will have their own `computeCalculateFields()`

Packet++/header/SomeIpLayer.h
```diff
@@ -0,0 +1,361 @@
...
+	size_t getPduPayloadSize() const { return getHeaderLengthWithPayload() - sizeof(someiphdr); }
+
+	/**
+	 * Get the Length of the SOME/IP header inc payload
+	 * @return size_t
+	 */
+	size_t getHeaderLen() const;
+
+	/**
+	 * Sets missing fields and calculates length
```
seladb, line 333: Maybe we can make this comment a bit more clear? Currently it only sets the protocol version to `0x01`

Packet++/src/SomeIpLayer.cpp
kolbex: only sets the protocol version, i currently moved this to the constructor which i find a better solution. In the next PR for SD/TP it will add code for setting the payload of the length field.

Packet++/src/TcpLayer.cpp
kolbex: not sure about this, currently there is a problem with the `SomeIpLayer::addSomeIpPort(...)` method which produces a memory leak if it is called (uses a `std::unordered_set<uint16_t>`). We could add a method for deallocation but i think it is very annoying if this is not done automatically. Do you have a suggestion?

Packet++/src/SomeIpLayer.cpp
seladb: We could still keep setting the version in `computeCalculateFields()`, but I don't have a strong opinion. Let's keep it as it is in the c'tor

Packet++/src/TcpLayer.cpp
seladb: @kolbex I have a suggestion: we can go with an approach that combines Pcap++Test's [`testSetUp`](https://github.com/seladb/PcapPlusPlus/blob/4708cfbeb901551c013fbad0da13600233e254b7/Tests/Pcap%2B%2BTest/Common/TestUtils.cpp#L94) and [`DeviceTeardown`](https://github.com/seladb/PcapPlusPlus/blob/4708cfbeb901551c013fbad0da13600233e254b7/Tests/Pcap%2B%2BTest/Common/TestUtils.h#L9):

1. We can add `pcpp::SomeIpLayer::removeSomeIpPort(uint16_t port)` so we can remove a port from the static `unordered_set`
2. Then we can add a new `testSetUp()` method in [`Packet++Test/Utils/TestUtils.h`](https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Packet%2B%2BTest/Utils/TestUtils.h) and [`Packet++Test/Utils/TestUtils.cpp`](https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Packet%2B%2BTest/Utils/TestUtils.cpp) that will add and remove a dummy port, just to initialize the `unordered_set`:

   ```cpp
   void testSetUp()
   {
   	pcpp::SomeIpLayer::addSomeIpPort(1);
   	pcpp::SomeIpLayer::removeSomeIpPort(1);
   }
   ```

3. `testSetUp()` should run in `Packet++Test/main.cpp` just about [here](https://github.com/seladb/PcapPlusPlus/blob/4708cfbeb901551c013fbad0da13600233e254b7/Tests/Packet%2B%2BTest/main.cpp#L109) before the tests start running
4. Now we can add a simple class `SomeIpTeardown` in `SomeIpTests.cpp` that will add a port in its c'tor and remove it in its d'tor, like this:

   ```cpp
   class SomeIpTeardown
   {
   private:
   	uint16_t m_SomeIpPort;
   public:
   	SomeIpTeardown(uint16_t someIpPort)
   	{
   		m_SomeIpPort = someIpPort;
   		pcpp::SomeIpLayer::addSomeIpPort(someIpPort);
   	}
   
   	~SomeIpTeardown()
   	{
   		pcpp::SomeIpLayer::removeSomeIpPort(m_SomeIpPort);
   	}
   };
   ``` 
5. The last step is to create an instance of `SomeIpTeardown` in each test, like this:

   ```cpp
   PTF_TEST_CASE(SomeIpParsingTest)
   {
   	timeval time;
   	gettimeofday(&time, NULL);
   
   	SomeIpTeardown(1111);
   
   	...
   ```

   What this does is ensure that a port is added in the beginning of the test and remove when the test ends or fails (which is when the `SomeIpTeardown` d'tor is called


I know these are quite a lot of changes, but I think it's a pretty clean solution to the memleak problem. Please let me know what you think

kolbex: Looks nice, two Questions:

> What is the `testSetUp()` method for?

> I would also add a method `pcpp::SomeIpLayer::removeAllSomeIpPorts(uint16_t port)` which removes all entries of the unordered_set. In order when more ports are added that you do not have to remember each port. What do you think?

seladb: > Looks nice, two Questions:
> 
>     1. What is the `testSetUp()` method for?
> 

It's just to initialize the `unordered_set` in the first time, otherwise you'll see "leaked memory" for the object initialization (even if the `unordered_set` is empty)

>     2. I would also add a method `pcpp::SomeIpLayer::removeAllSomeIpPorts(uint16_t port)` which removes all entries of the unordered_set. In order when more ports are added that you do not have to remember each port. What do you think?

Do you mean  `pcpp::SomeIpLayer::removeAllSomeIpPorts()` without the `port` parameter? 🤔

kolbex: Yes, without the port:
```
void SomeIpLayer::removeAllSomeIpPorts()
{
	m_SomeIpPorts.clear();
}
```

seladb: Sounds good!

kolbex: I currently updated the PR, i did not use the `SomeIpTeardown` class because i wanted to test both the `removeAllSomeIpPorts` and `removeSomeIpPort` methods, please take a look, i hope we can merge this PR.

seladb: Sure, I think we can merge the PR as is. I suggest we still add the `SomeIpTeardown` class in a separate PR to make sure we remove all SOME/IP ports after the test. It can have a simpler implementation:

```cpp
class SomeIpTeardown
{
public:
	SomeIpTeardown() {}

	~SomeIpTeardown()
	{
		pcpp::SomeIpLayer::removeAllSomeIpPorts();
	}
};
```

seladb (approved):

Issue #893, opened 2022-05-29
DNS response NX domains are classified as DNS query because there are no answers


AlienX2001:
Hey, from me again,
During my work with PcapPlusPlus I noticed that dns response NX domains were treated as DNS queries as seen in the screenshot
![image](https://user-images.githubusercontent.com/64488123/170861312-b85963f6-d6d0-4d3b-9bf6-8614f41953a1.png)

The code for the following pcapplusplus is as follows
```C++
#include "PcapFileDevice.h"
#include "DnsLayer.h"
#include "IPv4Layer.h"
#include "Packet.h"
#include <iostream>
#include <fstream>
#include <stdlib.h>
#include <string.h>
#include <unordered_map>

int main(int argc, char* argv[])
{
    //if less arguments
    if(argc<3){std::cout<<"Usage: "<<argv[0]<<" PcapFile BatchSize"<<std::endl; return -1;}
    //checking if pcap file exists
    std::ifstream f;
    f.open(argv[1]);
    if(!f){std::cout<<"PcapFile doesnt Exist!!"<<std::endl; return -1;}
    f.close();
    // use the IFileReaderDevice interface to automatically identify file type (pcap/pcap-ng)
    // and create an interface instance that both readers implement
    pcpp::IFileReaderDevice* reader = pcpp::IFileReaderDevice::getReader(argv[1]);

    // verify that a reader interface was indeed created
    if (reader == NULL)
    {
        std::cerr << "Cannot determine reader for file type" << std::endl;
        return -1;
    }
    // open the reader for reading
    if (!reader->open())
    {
        std::cerr << "Cannot open "<<argv[1]<<" for reading" << std::endl;
        return -1;
    }

    pcpp::RawPacket rawPacket; //raw packet taken from pcap
    std::vector<std::string> srcIP; // dest IP of the amplification
    std::unordered_map<std::string,int> floodScore; // map of dest IP and count of packets to it
    unsigned long long batch = 0;
    while(reader->getNextPacket(rawPacket)){
        pcpp::Packet parsedPacket(&rawPacket); // making parsed packet from raw packet to extract information layer by layer
        if (parsedPacket.isPacketOfType(pcpp::IPv4)) // if the packet is of IPv4
        {
            pcpp::DnsLayer* dnsLayer = parsedPacket.getLayerOfType<pcpp::DnsLayer>();
            if(dnsLayer == NULL){continue;} // continue if the packet is not of dns
[...]

PR #897, opened 2022-05-31
Update DnsLayer.cpp

AlienX2001:
Fixing https://github.com/seladb/PcapPlusPlus/issues/893


seladb:
@AlienX2001 can you add a test to [`DnsTests.cpp`](https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Packet%2B%2BTest/Tests/DnsTests.cpp)?
- You can add a pcap file in [`PacketExamples`](https://github.com/seladb/PcapPlusPlus/tree/master/Tests/Packet%2B%2BTest/PacketExamples)
- Convert it to hex stream and put that file in the same folder
- Then you can write a quick test that verifies that NXDOMAIN is parsed correctly

seladb:
Thanks! But you didn't write any test that uses this file :)

AlienX2001:
I saw the code structure of the tests and i think i got the gist of it, but my pcap file has many packets and not just 1, and so does my .dat file, so how would I proceed with it?

seladb:
> I saw the code structure of the tests and i think i got the gist of it, but my pcap file has many packets and not just 1, and so does my .dat file, so how would I proceed with it?

Each `.dat` file should contain just 1 packet, and I think 1 packet should be enough for this test. You can follow these steps:
- Take your pcap file, extract 1 NXDOMAIN packet and save it to a separate file
- Replace the current pcap file (that has many packets) with this one (that has only 1 packet)
- Open this 1-packet pcap file in Wireshark
- Select the packet and right click -> copy -> ...as a Hex stream
- Replace the existing content of the `.dat` file with the content of the packet
- Write a test that uses this packet from the `.dat` file

Please let me know if you have any questions.

image

seladb:
BTW, instead of naming these files `test.pcap` and `test.dat` you can give them more meaningful names like `dns_nxdomain.pcap` and `dns_nxdomain.dat`

seladb:
@AlienX2001 can you please rename `test.pcap` and `test.dat` to more meaningful names like `dns_nxdomain.pcap` and `dns_nxdomain.dat`? I think this would be the last comment and we can merge the PR after that

AlienX2001:
Done

AlienX2001:
And sorry for my inactivity on the PR, I had a deadline to catch up for my university

seladb:
@AlienX2001 the test itself is not there anymore 😕 
I remember you added a test in [`DnsTests.cpp`](https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Packet%2B%2BTest/Tests/DnsTests.cpp) but I don't see it anymore...

AlienX2001:
Ohk, I dont really understand why is it breaking, The READ_FILE_AND_CREATE_PACKET macro read the dat and creates a packet right?, naming convetion is "rawpacket<the number passed>", I simply copy pasted that code from above test cases, but somehow its breaking in that line.

egecetin:
Your test case should start with,

```
timeval time;
gettimeofday(&time, NULL);
```

AlienX2001:
Ohk, It passed all the tests, but could you explain, why that piece of code was needed? Thanks

egecetin:
@AlienX2001 Probably @seladb can explain better but,  `READ_FILE_AND_CREATE_PACKET` macro uses `pcpp::RawPacket` constructor which you can see below,

https://github.com/seladb/PcapPlusPlus/blob/b6fd30cd9e0fdd88b9f4d36e30a0bd7a7cc0b91a/Tests/Packet%2B%2BTest/Utils/TestUtils.h#L23-L25

and this constructor requires an input `timeval` called as time, you can find constructor here,

https://github.com/seladb/PcapPlusPlus/blob/b6fd30cd9e0fdd88b9f4d36e30a0bd7a7cc0b91a/Packet%2B%2B/header/RawPacket.h#L265-L276

therefore, compilation will fail if you do not define a `timeval time` value.

AlienX2001:
I see, thanks @egecetin .
I also dont seem to understand why this 1 pre commit test is failing

PR #899, opened 2022-06-01
Update protocols at README

egecetin:
People can browse more easily with supported protocols grouped and sorted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,41 +173,41 @@ PcapPlusPlus currently supports parsing, editing and creation of packets of the 
 1. Ethernet II
 2. IEEE 802.3 Ethernet
 3. Null/Loopback
-4. PPPoE
-5. SLL (Linux cooked capture)
-6. VLAN
-7. VXLAN
+4. Packet trailer (a.k.a footer or padding)
+5. PPPoE
+6. SLL (Linux cooked capture)
+7. VLAN
+8. VXLAN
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -157,36 +169,42 @@ PcapPlusPlus is currently supported on __Windows__, __Linux__, __MacOS__, __Andr
...
+12. IPv6
+13. MPLS
+14. Raw IP (IPv4 & IPv6)
+### Transport Layer (L4)
+15. GTP (v1)
+16. IPSec AH & ESP - parsing only (no editing capabilities)
+17. TCP
+18. UDP
+### Session Layer (L5)
+19. GRE
```
seladb, line 194: GRE is actually L3. PPP that comes after GRE can be considered L5

README.md
```diff
@@ -157,36 +169,42 @@ PcapPlusPlus is currently supported on __Windows__, __Linux__, __MacOS__, __Andr
...
+20. SDP
+21. SIP
 22. Radius
```
seladb, line 197: I think RADIUS belongs to L7, we should probably fix that in the code as well

seladb:
@egecetin We should probably make a similar PR in the PcapPlusPlus web-site:
https://github.com/PcapPlusPlus/pcapplusplus.github.io/blob/master/docs/features.mdx#supported-network-protocols

PR #902, opened 2022-06-13
Add ICMPv6 and NDP protocol

kolbex:
Add ICMPv6 and NDP layer support
- Add layer classes for handling ICMPv6 and NDP packets
- Supported packet types: Echo Request, Echo Reply, Neighbor Advertisement, Neighbor Solicitation
- Supports crafting and parsing
- Add tests
- Update README.md file


Packet++/src/IcmpV6Layer.cpp
```diff
@@ -0,0 +1,138 @@
...
+	case ICMPv6_NEIGHBOR_ADVERTISEMENT:
+		messageTypeAsString = "Neighbor Advertisement";
+		break;
+	default:
+		messageTypeAsString = "Unknown";
+		break;
+	}
+
+	std::ostringstream typeStream;
+	typeStream << (int)getMessageType();
```
maruu, line 82: You are getting the message type twice in this function:

`ICMPv6MessageTypes type = getMessageType();`

`typeStream << (int)getMessageType();`

I would propose to reduce it to one call.

Packet++/src/IcmpV6Layer.cpp
```diff
@@ -0,0 +1,138 @@
...
+namespace pcpp
+{
+IcmpV6Layer::IcmpV6Layer(ICMPv6MessageTypes type, uint8_t code) : Layer()
+{
+	m_DataLen = sizeof(icmpv6hdr);
+	m_Data = new uint8_t[m_DataLen];
+	memset(m_Data, 0, m_DataLen);
+	m_Protocol = ICMPv6;
+
+    icmpv6hdr *hdr = getIcmpv6Header();
```
maruu, line 21: There seems to be a minor tab/space mismatch here

Packet++/header/NdpLayer.h
```diff
@@ -0,0 +1,329 @@
...
+	 */
+	size_t getHeaderLen() const
+	{
+		return hasLinkLayerAddress() ? sizeof(ndpneighborsolicitationhdr) + sizeof(ndpoptionlinklayer)
+									 : sizeof(ndpneighborsolicitationhdr);
+	}
+
+	/**
+	 * This class is currently only used for parsing - so no implementation of this method
+	 */
```
maruu, line 315: I think this comment is not up to date anymore.. I guess in the case of this layer nothing needs to be done here, even for the crafting usecase?

Packet++/src/NdpLayer.cpp
```diff
@@ -0,0 +1,160 @@
...
+	pHdr->router = byRouter;
+	pHdr->solicited = unicastResponse;
+	pHdr->override = override;
+
+	memcpy(pHdr->targetIP, targetIP.toBytes(), 16);
+}
+
+std::string NDPNeighborAdvertisementLayer::toString() const
+{
+	return "NDP Neighbor Advertisement Layer, TargetIP: " + getTargetIP().toString();
```
maruu, line 87: Printing the target mac (if available) could also be interesting here, I guess?

maruu (changes requested):

kolbex:
Thanks @maruu for the review, i currently pushed the changes, @seladb this PR is ready for review

kolbex:
@clementperon could you do the review? @seladb seems to be unavailable at the moment

seladb:
> @clementperon could you do the review? @seladb seems to be unavailable at the moment

Hi @kolbex , I'm sorry for the late reply, I was out for a few weeks. I'll try to review this PR over the weekend. Thank you so much for this contribution!

seladb (commented):
@kolbex thank you so much for working on this!

I did a first pass on the PR and have some comments on the general design:
- `IcmpV6Layer` contains only the basic ICMPv6 header. While I understand this decision, it might be a bit problematic for non-NDP ICMPv6 messages, for example - echo/reply messages. The parsing of these packets will be partial and a portion of the packet will not be parsed at all. what I suggest is include the entire ICMPv6 data in the layer
- NDP is implemented as a separate layer from ICMPv6, but eventually, NDP is a type of ICMPv6 message, so I think it should inherit from `IcmpV6Layer` and we should find a way for `IcmpV6Layer` to become a "generic" ICMPv6 message, and the NDP layers to include more specific data for these type of messages
- NDP based message contain options - there is already an option parsing/crafting infrastructure in PcapPlusPlus in [`TLVData.h`](https://github.com/seladb/PcapPlusPlus/blob/master/Packet%2B%2B/header/TLVData.h). I think we should use it both for parsing and for crafting options instead of re-implementing it
- Would you consider implementing more ICMPv6 messages? Especially echo/reply which are pretty basic

seladb:
@kolbex what should we do with this PR?
Please see my comment on the general design: https://github.com/seladb/PcapPlusPlus/pull/902#pullrequestreview-1019350061

kolbex:
@seladb thanks for the review and sorry for the late reply
i also find it a good idea to move the NDP Layer together with the ICMPv6 Layer into one layer, and i was also thinking about a solution with tlv, i will let you next week if i can provide a solution with tlv
Additionally i added ICMPv6_ECHO_REQUEST and ICMPv6_ECHO_REPLY messages, but further messages are not planed from my side to implement. I will push the changes next week.

seladb:
> @seladb thanks for the review and sorry for the late reply i also find it a good idea to move the NDP Layer together with the ICMPv6 Layer into one layer, and i was also thinking about a solution with tlv, i will let you next week if i can provide a solution with tlv Additionally i added ICMPv6_ECHO_REQUEST and ICMPv6_ECHO_REPLY messages, but further messages are not planed from my side to implement. I will push the changes next week.

Sounds good @kolbex , thank you! 🙏

seladb:
Hi @kolbex ! if you're interested we have Slack workspace for PcapPlusPlus where we chat outside of GitHub. If you're interested in joining please send me your email and I'll send you an invite (you can either paste your email here or send an email to pcapplusplus@gmail.com)

kolbex:
Hi @seladb, regarding tlv: in the NDP option types the length is stored in units of 8 octets, how can this be handled in pcpp?

seladb:
> Hi @seladb, regarding tlv: in the NDP option types the length is stored in units of 8 octets, how can this be handled in pcpp?

@kolbex from the examples I saw and the (little) documentation I read it looks like ICMPv6 options are in the format of Type (1 byte), Length (1 byte), Value (variable length). In that case you can look at [`IPv4Layer`](https://github.com/seladb/PcapPlusPlus/blob/master/Packet%2B%2B/header/IPv4Layer.h) as a good example: 
- It implements [`IPv4Option`](https://github.com/seladb/PcapPlusPlus/blob/f6d57cf25d4f63cac75aa4bd6c707448e609121f/Packet%2B%2B/header/IPv4Layer.h#L212) to represent the option
- [`IPv4OptionBuilder`](https://github.com/seladb/PcapPlusPlus/blob/f6d57cf25d4f63cac75aa4bd6c707448e609121f/Packet%2B%2B/header/IPv4Layer.h#L352) to encapsulate option creation
- The `IPv4Layer` holds [`TLVRecordReader<IPv4Option>`](https://github.com/seladb/PcapPlusPlus/blob/f6d57cf25d4f63cac75aa4bd6c707448e609121f/Packet%2B%2B/header/IPv4Layer.h#L636) object to help iterating over options 
- The `IPv4Layer` exposes methods to read, create, and delete IPv4 options

kolbex:
Hi @seladb,
i updated the PR according to your review:
- TLV support for NDP messages
- ICMPv6 and NDP is combined to one layer
- ICMPv6Layer as a abstract layer base for icmpv6 messages
- Added messages: echo request and echo reply
can you please do another review? Please let me know what do you think

Packet++/src/IPv6Layer.cpp
```diff
@@ -258,6 +260,21 @@ void IPv6Layer::parseNextLayer()
...
+	{
+		ProtocolType icmpv6Ver = IcmpV6Layer::getIcmpv6Version(payload, payloadLen);
+		if(icmpv6Ver == ICMPv6EchoRequest)
+			m_NextLayer = new ICMPv6EchoRequestLayer(payload, payloadLen, this, m_Packet);
+		else if(icmpv6Ver == ICMPv6EchoReply)
+			m_NextLayer = new ICMPv6EchoReplyLayer(payload, payloadLen, this, m_Packet);
+		else if(icmpv6Ver == NDPNeighborSolicitation)
+			m_NextLayer = new NDPNeighborSolicitationLayer(payload, payloadLen, this, m_Packet);
+		else if(icmpv6Ver == NDPNeighborAdvertisement)
+			m_NextLayer = new NDPNeighborAdvertisementLayer(payload, payloadLen, this, m_Packet);
```
seladb, line 273: I think we can move this logic to `IcmpV6Layer` and have a static method `IcmpV6Layer::parseIcmpV6Message(...)` instead of `IcmpV6Layer::getIcmpv6Version()` that takes `payload`, `payloadLen`, `this` and `m_Packet` and creates the right layer.

We do something similar in BGP layer: https://github.com/seladb/PcapPlusPlus/blob/1a7e27c066d68d2c36f8c1f4c306fa15da1e6e7b/Packet%2B%2B/src/BgpLayer.cpp#L31

Packet++/src/IcmpV6Layer.cpp
```diff
@@ -0,0 +1,199 @@
...
+	case ICMPv6_NEIGHBOR_ADVERTISEMENT:
+		return NDPNeighborAdvertisement;
+	}
+
+	return UnknownProtocol;
+}
+
+ICMPv6MessageType IcmpV6Layer::getMessageType() const
+{
+	return (ICMPv6MessageType)getIcmpv6Header()->type;
```
seladb, line 42: If we convert `ICMPv6MessageType` to `enum class` we can safely do a static cast:
```cpp
return static_cast<ICMPv6MessageType>(getIcmpv6Header()->type);
```

Packet++/src/IcmpV6Layer.cpp
```diff
@@ -0,0 +1,199 @@
...
+	size_t payloadLen = m_DataLen - getHeaderLen();
+
+	switch (getMessageType())
+	{
+	case ICMPv6_NEIGHBOR_SOLICITATION:
+	case ICMPv6_NEIGHBOR_ADVERTISEMENT:
+		break;
+	default:
+		if (m_DataLen > headerLen)
+			m_NextLayer = new PayloadLayer(m_Data + headerLen, m_DataLen - headerLen, this, m_Packet);
```
seladb, line 68: Isn't ICMPv6 always the last layer? If that's the case let's always consider all of the data as part of the ICMPv6 message even if we don't know how to parse it

Packet++/src/IcmpV6Layer.cpp
```diff
@@ -0,0 +1,199 @@
...
+{
+	switch (getMessageType())
+	{
+	case ICMPv6_ECHO_REQUEST:
+	case ICMPv6_ECHO_REPLY:
+	case ICMPv6_NEIGHBOR_SOLICITATION:
+	case ICMPv6_NEIGHBOR_ADVERTISEMENT:
+		return m_DataLen;
+	default:
+		return sizeof(icmpv6hdr);
```
seladb, line 83: Same thing here: we can always return `m_DataLen` even if we don't know how to parse the data that comes after `icmpv6hdr`

Packet++/header/IcmpV6Layer.h
```diff
@@ -0,0 +1,300 @@
...
+	 */
+	void computeCalculateFields();
+
+	OsiModelLayer getOsiModelLayer() const { return OsiModelNetworkLayer; }
+
+  protected:
+	IcmpV6Layer() = default;
+
+	IcmpV6Layer(uint8_t *data, size_t dataLen, Layer *prevLayer, Packet *packet)
+		: Layer(data, dataLen, prevLayer, packet) {}
```
seladb, line 207: I think we can make this method public for parsing ICMPv6 messages we don't yet know how to parse

Packet++/header/IcmpV6Layer.h
```diff
@@ -0,0 +1,300 @@
...
+
+	/**
+	 * Calculate ICMPv6 checksum field
+	 */
+	void computeCalculateFields();
+
+	OsiModelLayer getOsiModelLayer() const { return OsiModelNetworkLayer; }
+
+  protected:
+	IcmpV6Layer() = default;
```
seladb, line 204: We can make this c'tor public and add parameters to it:
```cpp
IcmpV6Layer(ICMPv6MessageType msgType, uint8_t code);
```

Packet++/header/IcmpV6Layer.h
```diff
@@ -0,0 +1,300 @@
...
+	size_t dataLength;
+	/** most echo requests/replies contain some payload data. This is a pointer to this data */
+	uint8_t *data;
+} icmpv6_echo_request;
+
+/**
+ * @typedef icmpv6_echo_reply
+ * ICMPv6 echo reply message structure, same as icmpv6_echo_request
+ */
+typedef icmpv6_echo_request icmpv6_echo_reply;
```
seladb, line 145: I'm not sure we need this struct, now that we have dedicated layers for the echo/reply messages. The layer classes can return the header, the data length and a pointer to the data

Packet++/header/IcmpV6Layer.h
```diff
@@ -0,0 +1,300 @@
...
+	 */
+
+	ICMPv6EchoRequestLayer(uint16_t id, uint16_t sequence, const uint8_t *data, size_t dataLen);
+
+	virtual ~ICMPv6EchoRequestLayer() {}
+
+	/**
+	 * @return ICMP echo request data.
+	 */
+	icmpv6_echo_request *getEchoRequestData();
```
seladb, line 249: This should probably return `icmpv6_echo_hdr*`

Packet++/header/IcmpV6Layer.h
```diff
@@ -0,0 +1,300 @@
...
+
+	std::string toString() const;
+
+  private:
+	icmpv6_echo_request m_EchoData;
```
seladb, line 254: This member is probably redundent

Packet++/header/IcmpV6Layer.h
```diff
@@ -0,0 +1,300 @@
...
+	 * @param[in] dataLen The length of the echo reply payload
+	 */
+	ICMPv6EchoReplyLayer(uint16_t id, uint16_t sequence, const uint8_t *data, size_t dataLen);
+
+	virtual ~ICMPv6EchoReplyLayer()	{}
+
+	/**
+	 * @return ICMPv6 echo reply data.
+	 */
+	icmpv6_echo_reply *getEchoReplyData();
```
seladb, line 291: This method should probably return `icmpv6_echo_hdr*`

Packet++/header/IcmpV6Layer.h
```diff
@@ -0,0 +1,300 @@
...
+
+	std::string toString() const;
+
+  private:
+	icmpv6_echo_reply m_EchoData;
```
seladb, line 296: This member is probably redundant

Packet++/header/IcmpV6Layer.h
```diff
@@ -0,0 +1,300 @@
...
+
+  private:
+	icmpv6_echo_request m_EchoData;
+};
+
+/**
+ * @class ICMPv6EchoReplyLayer
+ * Represents an ICMPv6 echo reply protocol layer
+ */
+class ICMPv6EchoReplyLayer : public IcmpV6Layer
```
seladb, line 261: If we decide to remove `ICMPv6EchoReply` and `ICMPv6EchoRequest` then both `ICMPv6EchoReplyLayer` and `ICMPv6EchoRequestLayer` will be pretty similar. The only difference will be in the c'tor that sets a different ICMPv6 message type when building a new layer...

Tests/Packet++Test/PacketExamples/IcmpV6_EchoReply.dat
```diff
@@ -0,0 +1 @@
+00155da5c4c500155d70fc2e86dd6000000000403a80fe80000000000000dd05dae074bc7341fe8000000000000002155dfffea5c4c58100794c00180014bdcecb6200000000f3a1090000000000101112131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f3031323334353637
```
seladb, line 1: Can we add the pcap file that contains the echo request and reply packets?

seladb (commented):
@kolbex I reviewed the `IcmpV6Layer` and added some comments. Let's fix them first and then I'll review `NdpLayer.h`, that way the PR review will be a little bit easier. Let me know it that's ok with you

Packet++/header/IcmpV6Layer.h
```diff
@@ -0,0 +1,300 @@
...
+ */
+typedef icmpv6_echo_request icmpv6_echo_reply;
+
+/**
+ * @class IcmpV6Layer
+ * Abstract base class for ICMPv6 protocol layers. Cannot be instantiated and contains common logic for derived classes.
+ */
+class IcmpV6Layer : public Layer
+{
+  public:
```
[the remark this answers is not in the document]
kolbex: sure i can change it, i have done it that way because the automatic code formatting did it that way (.clang-format -> Microsoft style)... i am not sure if parameters from a style sheet can be overridden, but maybe it would make sense to adapt this parameter also in the .clang-format file

Packet++/header/IcmpV6Layer.h
kolbex: @seladb not sure about making this public, because then also the header space has to be allocated, which is currently done in the classes which inherits from this class. The default constructor is necessary for the specific layer classes.

kolbex:
@seladb can you please do the next review

Packet++/header/IcmpV6Layer.h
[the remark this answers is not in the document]
seladb: Yeah, we should fix the formatting at some point and add `.clang-format`....

Packet++/header/IcmpV6Layer.h
seladb: That's a good point!
The c'tor should probably also take a byte buffer and a length:
```cpp
IcmpV6Layer(ICMPv6MessageType msgType, uint8_t code, const uint8_t *data, size_t dataLen);
```

Then it can create `m_Data` in the size of `dataLen` and copy `data` into `m_Data`

Tests/Packet++Test/Tests/IcmpV6Tests.cpp
```diff
@@ -0,0 +1,287 @@
...
+	pcpp::Packet neighSoliPacket(&rawPacket3);
+	pcpp::Packet neighAdvPacket(&rawPacket4);
+	pcpp::Packet neighAdvPacketNoOpt(&rawPacket5);
+
+	// Echo request
+	PTF_ASSERT_TRUE(echoRequestPacket.isPacketOfType(pcpp::ICMPv6));
+	pcpp::ICMPv6EchoLayer *echoRequestLayer = echoRequestPacket.getLayerOfType<pcpp::ICMPv6EchoLayer>();
+	PTF_ASSERT_NOT_NULL(echoRequestLayer);
+	PTF_ASSERT_TRUE(echoRequestLayer->isMessageOfType(pcpp::ICMPv6MessageType::ICMPv6_ECHO_REQUEST));
+	PTF_ASSERT_EQUAL((int)echoRequestLayer->getMessageType(), 128);
```
seladb, line 36: `getMessageType()` returns an enum, why casing to an `int`? 🤔 
Same in lines 51, 65, 84, 93

Tests/Packet++Test/Tests/IcmpV6Tests.cpp
```diff
@@ -0,0 +1,287 @@
...
+	PTF_ASSERT_EQUAL(neighSoliLayer->getLinkLayerAddress(), pcpp::MacAddress("00:54:af:e9:4d:80"));
+	PTF_ASSERT_EQUAL(neighSoliLayer->getNdpOptionCount(), 1);
+
+	pcpp::NdpOption sourceLinkLayerOption = neighSoliLayer->getNdpOption(pcpp::NDPNeighborOptionTypes::NDP_OPTION_SOURCE_LINK_LAYER);
+	PTF_ASSERT_TRUE(sourceLinkLayerOption.isNotNull());
+	PTF_ASSERT_EQUAL(sourceLinkLayerOption.getNdpOptionType(), pcpp::NDPNeighborOptionTypes::NDP_OPTION_SOURCE_LINK_LAYER);
+	PTF_ASSERT_EQUAL(sourceLinkLayerOption.getDataSize(), 6);
+	PTF_ASSERT_EQUAL(sourceLinkLayerOption.getTotalSize(), 8);
+	PTF_ASSERT_EQUAL(pcpp::MacAddress(sourceLinkLayerOption.getValue()), pcpp::MacAddress("00:54:af:e9:4d:80"));
+	pcpp::NdpOption sourceLinkLayerOption2 = neighSoliLayer->getNdpOption(pcpp::NDPNeighborOptionTypes::NDP_OPTION_TARGET_LINK_LAYER);
```
seladb, line 79: nit: we can call this variable `targetLinkLayerOption`, no?

Tests/Packet++Test/Tests/IcmpV6Tests.cpp
```diff
@@ -0,0 +1,287 @@
...
+#include "IPv6Layer.h"
+#include "IcmpV6Layer.h"
+#include "Logger.h"
+#include "MacAddress.h"
+#include "NdpLayer.h"
+#include "Packet.h"
+#include "SystemUtils.h"
+#include <sstream>
+
+PTF_TEST_CASE(IcmpV6ParsingTest)
```
seladb, line 14: Can we find another ICMPv6 packet that is not a request, reply, neighbor solicitation or neighbor advertisement and try to parse it too?

Tests/Packet++Test/Tests/IcmpV6Tests.cpp
```diff
@@ -0,0 +1,287 @@
...
+	pcpp::Packet echoReplyPacket(100);
+	PTF_ASSERT_TRUE(echoReplyPacket.addLayer(&ethLayer2));
+	PTF_ASSERT_TRUE(echoReplyPacket.addLayer(&ipLayer2));
+	PTF_ASSERT_TRUE(echoReplyPacket.addLayer(&echoRepLayer));
+	echoReplyPacket.computeCalculateFields();
+	PTF_ASSERT_EQUAL(echoReplyPacket.getRawPacket()->getRawDataLen(), bufferLength2);
+	PTF_ASSERT_BUF_COMPARE(echoReplyPacket.getRawPacket()->getRawData()+54, buffer2+54, bufferLength2-54);
+
+	// Neighbor solicitation with source link-layer option
+	pcpp::IPv6Layer *pIpv6SoliLayer = new pcpp::IPv6Layer(pcpp::IPv6Address("fd53:7cb8:383:4::67"), pcpp::IPv6Address("fd53:7cb8:383:2::1:117"));
```
seladb, line 173: nit: we usually don't start a pointer variable with a `p`, I'd rename `pIpv6SoliLayer` to `ipv6SoliLayer`

Tests/Packet++Test/Tests/IcmpV6Tests.cpp
```diff
@@ -0,0 +1,287 @@
...
+	pcpp::NDPNeighborSolicitationLayer *pNdpSoliLayer = new pcpp::NDPNeighborSolicitationLayer(0, pcpp::IPv6Address("fd53:7cb8:383:2::1:117"), pcpp::MacAddress("00:54:af:e9:4d:80"));
```
seladb, line 174: same here: rename `pNdpSoliLayer` to `ndpSoliLayer`

Tests/Packet++Test/Tests/IcmpV6Tests.cpp
kolbex: currently i don't know how to handle the enum class together with PTF_ASSERT_EQUAL

PTF_ASSERT_EQUAL(echoRequestLayer->getMessageType(), pcpp::ICMPv6MessageType::ICMPv6_ECHO_REQUEST);
results in the error: error C2679: binary '<<': no operator found which takes a right-hand operand of type 'pcpp::ICMPv6MessageType'

same result when typing 'enum' as third argument

when i use 'enum class' as third argument, it compiles and runs correctly, but i'm a little bit confused about this... __VA_ARGS__##_PTF_PRINT_TYPE(actual) will be replaced by the precompiler to enum class_PTF_PRINT_TYPE(actual) where i currently dont know why this compiles?

what will be the correct solution?

Tests/Packet++Test/Tests/IcmpV6Tests.cpp
seladb: @kolbex you resolved a few comments but didn't push any commits, maybe you forgot to `git push`? 🤔

Tests/Packet++Test/Tests/IcmpV6Tests.cpp
seladb: you're right, I'm sorry. I don't think `PTF_ASSERT_EQUAL` works with `enum class`. I'll fix that, and for now let's keep the existing cast to `int`

Packet++/header/NdpLayer.h
```diff
@@ -0,0 +1,371 @@
...
+ * \namespace pcpp
+ * \brief The main namespace for the PcapPlusPlus lib
+ */
+namespace pcpp
+{
+
+/**
+ * An enum representing the available option types for Neighbor Discovery in IPv6 (see RFC 4861)
+ */
+enum NDPNeighborOptionTypes
```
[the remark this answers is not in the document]
kolbex: please check if the implementation is correct which handling the NDP_OPTION_UNKNOWN

Tests/Packet++Test/Tests/IcmpV6Tests.cpp
kolbex: sorry for the late push

Tests/Packet++Test/Tests/IcmpV6Tests.cpp
kolbex: currently i don't have time to do this, generating new/other ipv6 frames can sometimes be very time consuming, maybe someone else could do this or maybe we can bring this PR to dev branch and adding new frames later on?

seladb: Don't worry about it, I'll open another PR right after we merge this one and add this

seladb (approved):

seladb:
Thank you so much @kolbex for working on this, I really appreciate your contribution! 🙏 ❤️ 
ICMPv6 was missing in PcapPlusPlus, and it's a great addition to this project!

PR #910, opened 2022-07-15
Add SOME/IP TP protocol

kolbex:
Add SOME/IP TP layer support
- Add SomeIpTpLayer class for handling SOME/IP TP packets
- Support for crafting and parsing packets
- Add tests
- Update README.md file

This PR is the second of three PR for SOME/IP implementation. The third PR for SOME/IP SD will follow.


kolbex:
@seladb can you please do a review on this PR (TP related code)

seladb:
Sure @kolbex but maybe we should complete the work on the first PR before reviewing this one?

https://github.com/seladb/PcapPlusPlus/pull/895

We still have open comments there. Once we merge that one, it'll be easier to review this PR

seladb:
@kolbex after we merged the [first SOME/IP PR](https://github.com/seladb/PcapPlusPlus/pull/895) there are a lot of conflicts with `dev`. Can you resolve them?

kolbex:
@seladb currently updated this branch, please review it, i used the first version of `SomeIpTeardown` the second seemed to be optimized away with the compiler

Packet++/src/SomeIpLayer.cpp
```diff
@@ -269,14 +270,25 @@ bool SomeIpLayer::subProtocolFollows() const
...
 
+void SomeIpLayer::computeCalculateFields()
+{
+	if (!m_NextLayer)
+		return;
+
+	if ((m_NextLayer->getProtocol() == SomeIPTP)) // TODO: handling of SomeIpSd
+	{
+		setPayloadLength(m_NextLayer->getHeaderLen());
+	}
```
seladb, line 289: Before making any other comments, I wanted to discuss the design decision to make SOME/IP TP a layer on top of SOME/IP, so basically every TP packet will look like this:

```text
--------------------
| Eth layer        |
--------------------
| IPv4 layer       |
--------------------
| UDP layer        |
--------------------
| SOME/IP layer    |
--------------------
| SOME/IP TP layer |
--------------------
```

An alternative could be that `SomeIpTpLayer` inherits from `SomeIpLayer` and then the same packet will look like this: 

```text
--------------------
| Eth layer        |
--------------------
| IPv4 layer       |
--------------------
| UDP layer        |
--------------------
| SOME/IP TP layer |
--------------------
```

I'm not an expert in SOME/IP, so I'll need your guidance here: which of the designs makes more sense in terms of usage and user experience?

kolbex: it should be ok with both designs, maybe the handling is a little bit easier with one layer but i have not really a preference. What do you think which one we should use?

seladb: I agree, I think if we have the TP layer inherit from Some/IP layer it will be easier to manage. So basically the second option of the 2 I proposed. Will you make the relevant changes?

kolbex: Yes, currently working on it.

kolbex: currently updated, can you please take a look on this PR?

seladb: Sure, thanks @kolbex ! I'll review it soon

Packet++/header/ProtocolType.h
```diff
@@ -287,6 +287,11 @@ namespace pcpp
 	 */
 	const ProtocolType SomeIP = 0x100000000000;
 
+	/**
+	 * SOME/IP Transport Protocol
+	 */
+	const ProtocolType SomeIPTP = 0x200000000000;
```
seladb, line 293: Now that `SomeIpTpLayer` inherits from `SomeIpLayer` we probably don't need another const here. To get the TP layer one can do: 
```cpp
packet.getLayerOfType<pcpp::SomeIpTpLayer>();
```

If it not `nullptr` it means the packet has a TP layer

Packet++/header/SomeIpTpLayer.h
```diff
@@ -0,0 +1,134 @@
+#ifndef PACKETPP_SOMEIPTP_LAYER
```
seladb, line 1: We usually try to keep all classes of a layer in the same file. I'd move `SomeIpTpLayer` to `SomeIpLayer.h`, and the implementation to `SomeIpLayer.cpp`

Common++/header/Logger.h
```diff
@@ -69,6 +69,7 @@ namespace pcpp
 		PacketLogModuleNdpLayer, ///< NdpLayer module (Packet++)
 		PacketLogModuleFtpLayer, ///< FtpLayer module (Packet++)
 		PacketLogModuleSomeIpLayer, ///< SomeIpLayer module (Packet++)
+		PacketLogModuleSomeIpTpLayer, ///< SomeIpTpLayer module (Packet++)
```
seladb, line 72: If we move `SomeIpTpLayer` to `SomeIpLayer.h` we probably don't need this

Tests/Packet++Test/Tests/SomeIpTpTests.cpp
```diff
@@ -0,0 +1,171 @@
+#include "../TestDefinition.h"
```
seladb, line 1: Let's move these test cases to [`SomeIpTests.cpp`](https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Packet%2B%2BTest/Tests/SomeIpTests.cpp)

mk/vs/Packet++.vcxproj.filters
```diff
@@ -144,6 +144,9 @@
     <ClInclude Include="..\..\Packet++\header\SomeIpLayer.h">
       <Filter>Header Files</Filter>
     </ClInclude>
+    <ClInclude Include="..\..\Packet++\header\SomeIpTpLayer.h">
```
seladb, line 147: If we move `SomeIpTpLayer` to `SomeIpLayer.h`/`SomeIpLayer.cpp`  the changes in VS files aren't necessary (both this file as well as the other VS files)

seladb (commented):
@kolbex I have a few comments on the file structure etc. I suggest we fix them first, making the PR contain fewer changes, hence easier to review

kolbex:
@seladb i updated this PR, its ready for the next review

Packet++/src/SomeIpLayer.cpp
```diff
@@ -326,10 +293,90 @@ bool SomeIpLayer::isSomeIpSdPacket(uint16_t serviceId, uint16_t methodId)
...
+bool SomeIpTpLayer::isDataValid(uint8_t *data, uint32_t dataLen)
+{
+	someiphdr *hdr = (someiptphdr *)data;
+	return SomeIpTpLayer::isSomeIpTpPacket(hdr->msgType);
+}
+
+uint8_t SomeIpTpLayer::setTpFlag(uint8_t messageType)
+{
+	return messageType | (uint8_t)SomeIpLayer::MsgType::TP_REQUEST;
+}
```
seladb, line 344: Should this method be public?
I guess we need to use it in 2 places:
- `SomeIpTpLayer::computeCalculateFields()`
- SomeIpTpLayer::SomeIpTpLayer(...)`

Maybe we can make it private?

Packet++/src/SomeIpLayer.cpp
```diff
@@ -286,20 +251,20 @@ void SomeIpLayer::parseNextLayer()
...
 	size_t payloadLen = m_DataLen - headerLen;
 
-	/* Check if the header is a valid SomeIpSd header */
-	if (isSomeIpSdPacket(getServiceID(), getMethodID()))
-	{
-		m_NextLayer = new PayloadLayer(payload, payloadLen, this, m_Packet);
-	}
-	/* Check if the payload is a valid SomeIp TP layer */
-	else if (isSomeIpTpPacket(getMessageTypeAsInt()))
+	if (SomeIpLayer::isDataValidSd(payload, payloadLen))
```
seladb, line 254: if we implement the `parseSomeIpLayer(...)`  I spoke about in one of the previous comments, we can probably use it here too

Packet++/header/SomeIpLayer.h
```diff
@@ -303,58 +303,174 @@ class SomeIpLayer : public Layer
...
-	 * Identifies next layer and tries to create it
+	 * Identifies the following next layers: SomeIpLayer, SomeIpTpLayer. Otherwise sets PayloadLayer
 	 */
 	void parseNextLayer();
 
 	/**
 	 * @return The string representation of the SOME/IP layer
 	 */
-	std::string toString() const;
+	virtual std::string toString() const;
```
seladb, line 335: We don't override `toString()` in `SomeIpTpLayer` but we probably should, so `SomeIpLayer::toString()` only handles SOME/IP packets, and `SomeIpTpLayer::toString()` handles TP packets

README.md
```diff
@@ -222,7 +222,7 @@ PcapPlusPlus currently supports parsing, editing and creation of packets of the
 33. HTTP headers (request & response)
 34. NTP (v3, v4)
 35. Radius
-36. SOME/IP
+36. SOME/IP (SOME/IP, SOME/IP-TP)
```
seladb, line 225: I'm not sure if we should provide this extra info. Soon we'll merge the SD PR and then we'll cover all SOME/IP packets, right?

Tests/Packet++Test/Tests/SomeIpTests.cpp
```diff
@@ -148,3 +149,161 @@ PTF_TEST_CASE(SomeIpCreationTest)
 	delete [] buffer1;
 	delete [] buffer2;
 }
+
+class SomeIpTeardown
```
seladb, line 153: It's great that you added it!

I think we should also use it in `SomeIpParsingTest` (and move `SomeIpTeardown` to the top of the file)

Tests/Packet++Test/Tests/SomeIpTests.cpp
```diff
@@ -148,3 +149,161 @@ PTF_TEST_CASE(SomeIpCreationTest)
...
+	pcpp::Packet someIpTpPacket2(&rawPacket2);
+
+	// Test SomeIpTpLayer::setTpFlag()
+	PTF_ASSERT_EQUAL((uint8_t)pcpp::SomeIpLayer::MsgType::TP_REQUEST_NO_RETURN, pcpp::SomeIpTpLayer::setTpFlag((uint8_t)pcpp::SomeIpLayer::MsgType::REQUEST_NO_RETURN));
+
+	// Test SOME/IP-TP start packet
+	PTF_ASSERT_TRUE(someIpTpPacket1.isPacketOfType(pcpp::SomeIP));
+	pcpp::SomeIpTpLayer *someIpTpLayer1 = someIpTpPacket1.getLayerOfType<pcpp::SomeIpTpLayer>();
+	PTF_ASSERT_NOT_NULL(someIpTpLayer1);
+	PTF_ASSERT_EQUAL(someIpTpLayer1->getHeaderLen(), 1412);
```
seladb, line 190: Something in the math confuses me here:
The length field is 1404 + 2 bytes service ID + 2 bytes method ID = 1408.

How come the header len is 1412? 🤔 
I see that we calculate header len as:


I see that in Wireshark there are 4 extra bytes in this packet that are not included in this layer:

image

I think this is because we might miss something in `getHeaderLen()` calculation?

https://github.com/kolbex/PcapPlusPlus/blob/c97b2f6dfaf6209a7c5c7f43684f34de74fcf3cb/Packet%2B%2B/header/SomeIpLayer.h#L320

Maybe it should be:

```cpp
virtual size_t getHeaderLen() const { return sizeof(uint16_t) * 2 + getLengthField(); }
```

Packet++/src/SomeIpLayer.cpp
```diff
@@ -326,10 +293,90 @@ bool SomeIpLayer::isSomeIpSdPacket(uint16_t serviceId, uint16_t methodId)
...
 {
 	/* Check if TP flag is set in msgType */
 	return (msgType & (uint8_t)SomeIpLayer::MsgType::TP_REQUEST) != 0;
 }
 
+bool SomeIpTpLayer::isDataValid(uint8_t *data, uint32_t dataLen)
+{
+	someiphdr *hdr = (someiptphdr *)data;
+	return SomeIpTpLayer::isSomeIpTpPacket(hdr->msgType);
+}
```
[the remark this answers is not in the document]
kolbex: I thought about the inheritance for SD from the SOME/IP class and yes it should be also possible. This changes from above should also be implemented.

Tests/Packet++Test/Tests/SomeIpTests.cpp
seladb: That's a good point!
Maybe we can do something like this:
- Modify `SomeIpTeardown()` c'tor to not take any argument (and then we don't need `m_SomeIpPort`)
- Modify `~SomeIpTeardown()` d'tor to use `removeAllSomeIpPorts()` - so it will clean up all the ports in the end of the test
- Add another test case for testing `addSomeIpPort()`, `removeSomeIpPort()`, `removeAllSomeIpPorts()` and `isSomeIpPort()`, it can do something like this:
 ```cpp
PTF_TEST_CASE(SomeIpPortTest)
{
	SomeIpTeardown tearDown;

	pcpp::SomeIpLayer::addSomeIpPort(1234);
	PTF_ASSERT_TRUE(pcpp::SomeIpLayer::isSomeIpPort(1234));

	pcpp::SomeIpLayer::removeSomeIpPort(1234);
	PTF_ASSERT_FALSE(pcpp::SomeIpLayer::isSomeIpPort(1234));

	pcpp::SomeIpLayer::addSomeIpPort(1235);
	pcpp::SomeIpLayer::addSomeIpPort(1236);
	PTF_ASSERT_TRUE(pcpp::SomeIpLayer::isSomeIpPort(1235));
	PTF_ASSERT_TRUE(pcpp::SomeIpLayer::isSomeIpPort(1236));

	pcpp::SomeIpLayer::removeAllSomeIpPorts();
	PTF_ASSERT_FALSE(pcpp::SomeIpLayer::isSomeIpPort(1235));
	PTF_ASSERT_FALSE(pcpp::SomeIpLayer::isSomeIpPort(1236));
}
```

kolbex: ok, for some reason `SomeIpTeardown tearDown();` did not work, but now it should be fine. Can we bring this to the dev branch?

seladb (approved):

PR #914, opened 2022-07-17
Fixes Heap-buffer-overflow READ in pcpp::cross_platform_memmem

sashashura:
Fixes Heap-buffer-overflow READ in pcpp::cross_platform_memmem:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43408

The `haystackLen` for `cross_platform_memmem` call in `SipRequestFirstLine::parseVersion` is calculated as `m_SipRequest->getDataLen() - m_UriOffset`. However when `m_DataLen` is 4 and `m_UriOffset` is 5 it results in -1 which is converted into max uint64. This leads to memory read out of data buffer bounds.


seladb:
@sashashura pre-commit fails because you have trailing whitespace...

seladb:
@sashashura pre-commit fails because you have trailing whitespace...

sashashura:
I'm not sure why it fails the android build. Let me know if I need to do something.

seladb:
> I'm not sure why it fails the android build. Let me know if I need to do something.

Lately the Android builds on macOS tend to fail, I'm not sure why... I reran the failed jobs

seladb (approved):

PR #916, opened 2022-07-17
Fixes Heap-buffer-overflow READ in pcpp::IPv6Layer::parseExtensions

sashashura:
Fixes Heap-buffer-overflow READ in pcpp::IPv6Layer::parseExtensions:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43358 and
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43290

When `dataLen` in `IPv6Layer::IPv6Layer` constructor call is smaller than `sizeof(ip6_hdr)` it leads to out of bounds memory read when `data` is casted to `ip6_hdr*` and its members are accessed.


sashashura:
...and this one?

Packet++/src/IPv4Layer.cpp
```diff
@@ -261,72 +261,79 @@ void IPv4Layer::parseNextLayer()
...
+			m_NextLayer = IcmpLayer::isDataValid(payload, payloadLen)
+				? static_cast<Layer*>(new IcmpLayer(payload, payloadLen, this, m_Packet))
+				: static_cast<Layer*>(new PayloadLayer(payload, payloadLen, this, m_Packet));
+			break;
+		case PACKETPP_IPPROTO_IPIP:
+			ipVersion = *payload >> 4;
+			if (ipVersion == 4)
+				m_NextLayer = new IPv4Layer(payload, payloadLen, this, m_Packet);
+			else if (ipVersion == 6)
+				m_NextLayer = new IPv6Layer(payload, payloadLen, this, m_Packet);
```
seladb, line 287: Instead of `try..catch` a better solution would be to check `IPv6Layer::isDataValid(...)` or `IPv4Layer::isDataValid(...)`, like this:
```cpp
ipVersion = *payload >> 4;
if (ipVersion == 4 && IPv4Layer::isDataValid(payload, payloadLen))
	m_NextLayer = new IPv4Layer(payload, payloadLen, this, m_Packet);
else if (ipVersion == 6 && IPv6Layer::isDataValid(payload, payloadLen))
	m_NextLayer = new IPv6Layer(payload, payloadLen, this, m_Packet);
else
```

Packet++/src/IPv6Layer.cpp
```diff
@@ -33,6 +33,12 @@ IPv6Layer::IPv6Layer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* pa
 	m_LastExtension = NULL;
 	m_ExtensionsLen = 0;
 
+	if (dataLen < sizeof(ip6_hdr))
```
seladb, line 36: This is not needed if before getting to the c'tor we call `IPv6Layer::isDataValid(...)`

seladb (changes requested):

Packet++/src/IPv4Layer.cpp
sashashura: Didn't notice the pattern.

seladb (approved):

Issue #856, opened 2022-03-30
Heap buffer overflow (read) in IPv6Layer::parseExtensions() at IPv6Layer.cpp:82:37


vivin:
There is a buffer overflow at the specified location:

```
==4179529==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x606000000178 at pc 0x000000560e53 bp 0x7ffce59dc690 sp 0x7ffce59dc688
READ of size 1 at 0x606000000178 thread T0
    #0 0x560e52 in pcpp::IPv6Layer::parseExtensions() /home/vivin/Projects/phd/subjects/pcapplusplus/PcapPlusPlus-master/Packet++/src/IPv6Layer.cpp:82:37
    #1 0x560861 in pcpp::IPv6Layer::IPv6Layer(unsigned char*, unsigned long, pcpp::Layer*, pcpp::Packet*) /home/vivin/Projects/phd/subjects/pcapplusplus/PcapPlusPlus-master/Packet++/src/IPv6Layer.cpp:36:2
    #2 0x663643 in pcpp::IPv4Layer::parseNextLayer() /home/vivin/Projects/phd/subjects/pcapplusplus/PcapPlusPlus-master/Packet++/src/IPv4Layer.cpp:285:22
    #3 0x54a40b in pcpp::Packet::setRawPacket(pcpp::RawPacket*, bool, unsigned long, pcpp::OsiModelLayer) /home/vivin/Projects/phd/subjects/pcapplusplus/PcapPlusPlus-master/Packet++/src/Packet.cpp:80:13
    #4 0x54d82b in pcpp::Packet::Packet(pcpp::RawPacket*, bool, unsigned long, pcpp::OsiModelLayer) /home/vivin/Projects/phd/subjects/pcapplusplus/PcapPlusPlus-master/Packet++/src/Packet.cpp:126:2
    #5 0x4f33c3 in main /home/vivin/Projects/phd/resources/archives/pcapplusplus/readpcap.cpp:30:20
    #6 0x7fc0983670b2 in __libc_start_main /build/glibc-eX1tMB/glibc-2.31/csu/../csu/libc-start.c:308:16
    #7 0x4441bd in _start (/home/vivin/Projects/phd/workspace/smartdsf/pcapplusplus/binaries/vanilla-asan/readpcap+0x4441bd)

0x606000000178 is located 2 bytes to the right of 54-byte region [0x606000000140,0x606000000176)
allocated by thread T0 here:
[...]

PR #925, opened 2022-07-29
Add getNextPacket

sashashura:
Use the idea from https://github.com/seladb/PcapPlusPlus/issues/856#issuecomment-1086590528 of parsing packets in a loop. It improves the fuzzer's speed (from 65575 exec/s to 521916 exec/s on my machine) and coverage.

### DIFF
--- a/Tests/Fuzzers/FuzzTarget.cpp
+++ b/Tests/Fuzzers/FuzzTarget.cpp
@@ -54,18 +54,21 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
 		return 1;
 	}
 
-	// parse the raw packet into a parsed packet
-	pcpp::Packet parsedPacket(&rawPacket);
-
-	// verify the packet is IPv4
-	if (parsedPacket.isPacketOfType(pcpp::IPv4))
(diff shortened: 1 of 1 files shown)


Tests/Fuzzers/FuzzTarget.cpp
```diff
@@ -54,18 +54,21 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
 		return 1;
 	}
 
-	// parse the raw packet into a parsed packet
-	pcpp::Packet parsedPacket(&rawPacket);
-
-	// verify the packet is IPv4
-	if (parsedPacket.isPacketOfType(pcpp::IPv4))
+	while (reader.getNextPacket(rawPacket))
```
seladb, line 57: If we do the while loop, we lose the first packet which is read in line 51 (`reader.getNextPacket(rawPacket)`)...

seladb (changes requested):

Tests/Fuzzers/FuzzTarget.cpp
sashashura: Good catch. Made do/while.

seladb (approved):

PR #926, opened 2022-07-29
Refactoring. Use cpp style casting instead c-style cast.

Minge3:
c-style cast is more difficult to understand what it's actually doing.
CPP cast makes it more clear what is happening.

Re-using a pointer variable "RawPacket* const rPacket" instead of casting the same thing again. That change should make the code more readable.


seladb:
@Minge3 thank you for opening this PR! It looks like many of the CI jobs fail, I'm not sure which change causes these failures, can you please take a look?

PR #929, opened 2022-07-30
3rdparty: dirent: update to latest version

clementperon:
New [CMakeLists.txt](https://github.com/tronkko/dirent/blob/master/CMakeLists.txt) of dirent allow to include dirent without building tests


seladb:
@clementperon we use `dirent.h` only in one example (`PcapSearch`) and we anyway only need the header file and not the whole project. Maybe it's a good opportunity to remove `3rdParty/dirent-for-Visual-Studio` and instead create a new folder `Examples/PcapSearch/3rdParty/dirent-for-Visual-Studio` and add just this header file. What do you think? 🤔

clementperon:
This will  need to update Vscode template I will close and do the clean after cmake

PR #931, opened 2022-08-01
Add memory sanitizer regressions check

sashashura:
Just checking some things...


sashashura:
nope

seladb:
@sashashura just FYI - you can keep the PR open as "draft" if you want to test things. You don't have to close and reopen it :)

seladb:
@sashashura is this PR ready for review?

sashashura:
Depends... I was able to make it work with the `gcr.io/oss-fuzz-base/base-clang` docker image. It is not ideal because it is not the same image is used in other regression test tasks.

seladb:
> Depends... I was able to make it work with the `gcr.io/oss-fuzz-base/base-clang` docker image. It is not ideal because it is not the same image that is used in other regression test tasks.

I wonder what is different between `gcr.io/oss-fuzz-base/base-clang` and the image we use for the other tests `seladb/ubuntu1804`. Maybe Ubuntu 18.04 is too old and we should use `seladb/ubuntu2204`? 🤔

sashashura:
I narrowed it down to https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-clang/checkout_build_install_llvm.sh. It must be about building custom libcxx as described in https://github.com/google/sanitizers/wiki/MemorySanitizerLibcxxHowTo

seladb:
> I narrowed it down to https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-clang/checkout_build_install_llvm.sh. It must be about building custom libcxx as described in https://github.com/google/sanitizers/wiki/MemorySanitizerLibcxxHowTo

Got it, that makes in a way... It's interesting that the other sanitizers (Address and Undefined Behavior) don't need a custom libcxx build.
What's also interesting is that it passes although libpcap is not MSan-ified.
If it works I guess we can use the `oss-fuzz` image, what do you think?

sashashura:
Yes, let's use the image.

.cirrus.yml
```diff
@@ -50,6 +50,20 @@ task:
...
+task:
+  name: Regression Tests Linux (Memory Behavior Sanitizer)
+  container:
+    image: gcr.io/oss-fuzz-base/base-clang
+  configure_script:
+    - apt-get update && apt-get install -y libpcap-dev git make
+    - ./configure-fuzzing.sh --sanitizer memory
+    - ldconfig
+  build_script:
+    - export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"
```
seladb, line 62: @sashashura why is this line needed?

sashashura: It wasn't working without it. I think it is reading `LIB_FUZZING_ENGINE` from env somewhere. Let me quickly check.

seladb: Makes sense, thanks for the explanation! 🙏

seladb (approved):

PR #935, opened 2022-08-08
Update clangformat

egecetin:
Not important for now but It seems that these options are better match to existing code (Especially for *Layer headers) . What do you think?


seladb (approved):

seladb:
We don't yet use `clang-format`. I'm wondering if we can add it without making tons of changes in the code 🤔

PR #934, opened 2022-08-05
Auto OUI Database Update

egecetin:
Automatically update OUI database and commit to GitHub if any changes detected.

This PR requires merging of #904

PR #937, opened 2022-08-10
Auto precommit Update

egecetin:
Automatically update precommit hooks every week and create PR if any change detected

See test PR [here](https://github.com/egecetin/PcapPlusPlus/pull/5)

### DIFF
--- a/.github/workflows/auto_update.yml
+++ b/.github/workflows/auto_update.yml
@@ -12,7 +12,8 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


egecetin:
@seladb If you merge this PR we can check the auto PR behavior less harmful (Instead of #934). What do you think?

seladb (approved):

seladb:
> @seladb If you merge this PR we can check the auto PR behavior less harmful (Instead of #934).

That's a good idea! Let's see if we need a weekly update or a less frequent one. We can start with one week and see how it goes

PR #947, opened 2022-08-20
Pytest rootpath option

clementperon:
The examples binaries outpath is actually hardcoded in the Makefile.

it's not the case with the cmake build system.

Introduce an option to pytest to give the example binaries path

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -51,7 +51,7 @@ jobs:
           cd Tests/ExamplesTest
           ${{ matrix.python }} -m pip install -U pip
           ${{ matrix.python }} -m pip install -r requirements.txt
-          ${{ matrix.python }} -m pytest --interface eth0
+          ${{ matrix.python }} -m pytest --interface eth0 --root-path=../../Dist/examples
 
       - name: Check installation
         run: |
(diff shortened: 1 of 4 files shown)


seladb (commented):
@clementperon instead of passing `root_path` as a fixture to every test, I propose a simpler approach which is based on this StackOverflow thread:
https://stackoverflow.com/a/50135020

Add `root_path` to `run_example`:
```python
def run_example(
    example_name, args, root_path, timeout=10, expected_return_code=0, requires_root=False
):
    command_to_run = (
        ["sudo"]
        if requires_root
        and (platform.system() == "Linux" or platform.system() == "Darwin")
        else []
    ) + [os.path.join(root_path, example_name)]
    ...
    ...
```

Then add this to the `ExampleTest` class:
```python
class ExampleTest:
    @pytest.fixture(autouse=True)
    def _root_path(self, request):
        self.root_path = request.config.getoption("--root-path")

    def run_example(
        self, args, timeout=10, expected_return_code=0, requires_root=False
    ):
        return run_example(
            example_name=self.__class__.__name__[4:],
            args=args,
            root_path=self.root_path,
            timeout=timeout,
            expected_return_code=expected_return_code,
            requires_root=requires_root,
        )
```

And of course add `root-path` to `pytest_addoption` in `conftest.py`:
```python
DEFAULT_EXAMPLE_DIR = os.path.abspath("../../Dist/examples/")

def pytest_addoption(parser):
    ...
    ...
    parser.addoption(
        "--root-path",
        action="store",
        default=DEFAULT_EXAMPLE_DIR,
        help="Examples root dir",
    )
```

clementperon:
@seladb yes much more better! I have updated the MR

clementperon:
maybe root-path should be examples-binaries-path no?

seladb (approved):

PR #950, opened 2022-08-23
Fix potential memory leak

BaderEddineOuaich:
The method `RawSocketDevice::receivePacket(...)` performs several error checks, some of which free the buffer before returning, whereas others do not (memory leak).
The buffer is now wrapped in a smart pointer, which will handle its deletion properly.

### DIFF
--- a/Pcap++/src/RawSocketDevice.cpp
+++ b/Pcap++/src/RawSocketDevice.cpp
@@ -15,6 +15,7 @@
 #include "SystemUtils.h"
 #include "Packet.h"
 #include "EthLayer.h"
+#include <memory>
 
 namespace pcpp
 {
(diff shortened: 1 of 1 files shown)


seladb:
@BaderEddineOuaich thank you for opening this PR!

CI fails on multiple platforms, can you please check?

BaderEddineOuaich:
Hello, I mistakenly assumed that [RawSocketDevice](https://github.com/seladb/PcapPlusPlus/blob/dev/Pcap%2B%2B/src/RawSocketDevice.cpp) was responsible of deleting the [buffer](https://github.com/seladb/PcapPlusPlus/blob/dev/Pcap%2B%2B/src/RawSocketDevice.cpp#L157), it turns out that the buffer will be [served](https://github.com/seladb/PcapPlusPlus/blob/dev/Pcap%2B%2B/src/RawSocketDevice.cpp#L201) to the `RawPacket` which will [take care of its deletion](https://github.com/seladb/PcapPlusPlus/blob/dev/Packet%2B%2B/src/RawPacket.cpp#L44).
Please ignore this PR.
Thanks for the great work and have a nice day!

PR #952, opened 2022-08-23
Add CodeQL SAST

sashashura:
CodeQL is the continuation LGTM. It integrates into GitHub Security tab and is more configurable. I have set it up to scan only for C++ issues.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -2,10 +2,10 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "master" ]
+    branches: [ "dev" ]
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [ "master" ]
+    branches: [ "dev" ]
   workflow_dispatch:
   schedule:
     - cron: '15 14 * * 1'
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -16,8 +16,6 @@ jobs:
     runs-on: ubuntu-20.04
     container: seladb/ubuntu2204
     permissions:
-      actions: read
-      contents: read
       security-events: write
 
     strategy:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -39,7 +39,7 @@ jobs:
         # If you wish to specify custom queries, you can do so here or in a config file.
         # By default, queries listed here will override any specified in a config file.
         # Prefix the list here with "+" to use these queries and those in the config file.
-        
+
         # Details on CodeQL's query packs refer to : https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
         # queries: security-extended,security-and-quality
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,51 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
+  workflow_dispatch:
+  schedule:
+    - cron: '15 14 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-20.04
+    container: seladb/ubuntu2204
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: ${{ matrix.language }}
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file.
+        # Prefix the list here with "+" to use these queries and those in the config file.
+        
+        # Details on CodeQL's query packs refer to : https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
+        # queries: security-extended,security-and-quality
+
+    - run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,51 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
```
seladb, line 8: PRs are usually merged to `dev` and not to `master`. Should we change it to `dev` and change `push` to both `master` and `dev`?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,51 @@
...
+  workflow_dispatch:
```
seladb, line 9: is this needed? 🤔

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,51 @@
...
+    - cron: '15 14 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-20.04
+    container: seladb/ubuntu2204
+    permissions:
+      actions: read
+      contents: read
```
seladb, line 20: According to the documentation, this is only needed for private repos...

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,51 @@
...
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
```
seladb, line 28: We can probably remove this comment?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,51 @@
...
+    container: seladb/ubuntu2204
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp' ]
```
seladb, line 26: is `matrix` needed here? there's only one language...

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,51 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
```
sashashura, line 5:
```suggestion
    branches: [ "dev" ]
```

.github/workflows/codeql.yml
sashashura, line 8:
```suggestion
    branches: [ "dev" ]
```

sashashura: I think `dev` is enough.

.github/workflows/codeql.yml
sashashura: It is a manual trigger. It allows maintainers to go to the Actions tab and to click the button "Start Workflow". Not mandatory, but quite useful.

.github/workflows/codeql.yml
sashashura: Makes it easier to add additional languages if one day you would like to scan python too for example.

.github/workflows/codeql.yml
sashashura: It were just some hints for the future, but if you prefer clean code, sure.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,51 @@
...
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
```
sashashura, line 28:
```suggestion
```

.github/workflows/codeql.yml
sashashura: Sorry, I don't follow. `Permissions:` at all or some specific `right`?

.github/workflows/codeql.yml
seladb: Ok sure, let's keep it 👍

.github/workflows/codeql.yml
seladb: According to the CodeQL Github Actions README file:

image

https://github.com/github/codeql-action#usage

.github/workflows/codeql.yml
seladb: Sure, we can keep it 👍

.github/workflows/codeql.yml
seladb: I don't think we'll need to examine the python code because it's mostly for tests or configuration, but we can keep it as is

seladb:
@sashashura the build is failing because CodeQL check fail: `Use of a broken or risky cryptographic algorithm` in ` Packet++/src/SSLHandshake.cpp`

These errors should be ignored because PcapPlusPlus uses these to parse SSL/TLS packets. However I'm not sure there is a way to suppress/ignore errors on specific lines as this issue says: https://github.com/github/codeql/issues/9298#issuecomment-1136938130

Should we add `paths-ignore` for the entire file? 🤔

.github/workflows/codeql.yml
sashashura: I see. I'll remove it.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,49 @@
...
+    - cron: '15 14 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-20.04
+    container: seladb/ubuntu2204
+    permissions:
+      actions: read
+      contents: read
```
sashashura, line 20:
```suggestion
```

sashashura:
Interesting behavior. From my experience all security issues found by a codeql.yml already merged are places in Security tab and are visible only to maintainers. It is easy to click a button and suppress specific warnings there if they are false.
While it makes sense to fail the check if changes from pull request introduce a potential security vulnerability, it is inconvenient to suppress them. I think it is better to not use it for pull requests and use it only as a scheduled check. I'll remove the trigger.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,47 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "dev" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "dev" ]
```
sashashura, line 8:
```suggestion
```

seladb (approved):
@sashashura the next step would be to add a badge to README.md, and then let it run for some time. If CodeQL is stable and works as expected we can probably remove LGTM

PR #965, opened 2022-09-16
add a method to get ipv6 similar to ipv4 for pcapLiveDevice

jafar75:
modified:   Pcap++/header/PcapLiveDevice.h
	modified:   Pcap++/src/PcapLiveDevice.cpp
	modified:   Tests/Pcap++Test/TestDefinition.h
	modified:   Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
	modified:   Tests/Pcap++Test/main.cpp

### DIFF
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -939,6 +939,27 @@ IPv4Address PcapLiveDevice::getIPv4Address() const
 	return IPv4Address::Zero;
 }
 
+IPv6Address PcapLiveDevice::getIPv6Address() const
+{
+	for (std::vector<pcap_addr_t>::const_iterator addrIter = m_Addresses.begin(); addrIter != m_Addresses.end();
+		 addrIter++)
+	{
+		char addrAsString[INET6_ADDRSTRLEN];
+		internal::sockaddr2string(addrIter->addr, addrAsString);
+		if (Logger::getInstance().isDebugEnabled(PcapLogModuleLiveDevice) && addrIter->addr != NULL)
+		{
+			PCPP_LOG_DEBUG("Searching address " << addrAsString);
+		}
+		in6_addr *currAddr = internal::sockaddr2in6_addr(addrIter->addr);
+		if (currAddr == NULL)
+		{
+			PCPP_LOG_DEBUG("Address is NULL");
+			continue;
+		}
+		return IPv6Address(addrAsString);
+	}
+	return IPv6Address::Zero;
+}
 
 IPv4Address PcapLiveDevice::getDefaultGateway() const
 {
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -944,10 +944,10 @@ IPv6Address PcapLiveDevice::getIPv6Address() const
 	for (std::vector<pcap_addr_t>::const_iterator addrIter = m_Addresses.begin(); addrIter != m_Addresses.end();
 		 addrIter++)
 	{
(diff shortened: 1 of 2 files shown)


jafar75:
#965

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -334,7 +334,30 @@ PTF_TEST_CASE(TestPcapLiveDeviceNoNetworking)
 
 } // TestPcapLiveDeviceNoNetworking
 
+PTF_TEST_CASE(TestPcapLiveDeviceNoNetworkingIPv6)
```
seladb, line 337: This test assumes the machine running it has a NIC with IPv6 address. This is not always the case, for example in the VMs running the CI: https://github.com/seladb/PcapPlusPlus/actions/runs/3068629818

There is no easy way around it because we need a live device to test this functionality.

One option I can think of is [creating a virtual interface](https://linuxconfig.org/configuring-virtual-network-interfaces-in-linux) but it's only available on Linux and won't work on other platforms.

Another option that requires much more work is creating a mock library for libpcap/WinPcap/Npcap where we can mock everything these libraries do and then have full control, but it's probably out-of-scope for this PR.

Unfortunately, as I see no other option, I suggest we omit the test for this functionality

jafar75: @seladb at this stage, I omitted the test for the feature. maybe later we can develop about mock library or any other solution. thank you

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -1,35 +1,35 @@
 #define LOG_MODULE PcapLogModuleLiveDevice
```
seladb, line 1: @jafar75 this file contains many changes, maybe you ran a code formatter?

I'd suggest we limit this PR to adding `getIPv6Address()` and remove the other formatting changes. 
We have a task in our backlog to run code formatter on the entire code, we'll do it at some point...

seladb: @jafar75 this file still has a lot of changes:

image

I'd expect to only see the implementation of `getIPv6Address()`...

jafar75:
Thanks, I will correct it in the next 2 days

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -939,6 +939,27 @@ IPv4Address PcapLiveDevice::getIPv4Address() const
...
+		{
+			PCPP_LOG_DEBUG("Searching address " << addrAsString);
+		}
+		in6_addr *currAddr = internal::sockaddr2in6_addr(addrIter->addr);
+		if (currAddr == NULL)
+		{
+			PCPP_LOG_DEBUG("Address is NULL");
+			continue;
+		}
+		return IPv6Address(addrAsString);
```
seladb, line 959: I think we can implement it in a more efficient way using `currAddr->s6_addr`:

```cpp
if (Logger::getInstance().isDebugEnabled(PcapLogModuleLiveDevice) && addrIter->addr != NULL)
{
	char addrAsString[INET6_ADDRSTRLEN];
	internal::sockaddr2string(addrIter->addr, addrAsString);
	PCPP_LOG_DEBUG("Searching address " << addrAsString);
}
in6_addr *currAddr = internal::sockaddr2in6_addr(addrIter->addr);
if (currAddr == NULL)
{
	PCPP_LOG_DEBUG("Address is NULL");
	continue;
}
return IPv6Address(currAddr->s6_addr);
```

seladb (approved):

PR #985, opened 2022-10-18
Fix variableScope

egecetin:
Fix cppcheck warning variableScope

### DIFF
--- a/Packet++/src/HttpLayer.cpp
+++ b/Packet++/src/HttpLayer.cpp
@@ -795,7 +795,7 @@ std::string HttpResponseFirstLine::getStatusCodeString() const
 	std::string result;
 	if (m_StatusCode != HttpResponseLayer::HttpStatusCodeUnknown)
 	{
-		int statusStringOffset = 13;
+		const int statusStringOffset = 13;
 		int statusStringEndOffset = m_FirstLineEndOffset - 2;
 		if ((*(m_HttpResponse->m_Data + statusStringEndOffset)) != '\r')
 			statusStringEndOffset++;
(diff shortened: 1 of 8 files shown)


Packet++/src/HttpLayer.cpp
```diff
@@ -793,9 +793,9 @@ int HttpResponseFirstLine::getStatusCodeAsInt() const
 std::string HttpResponseFirstLine::getStatusCodeString() const
 {
 	std::string result;
-	int statusStringOffset = 13;
 	if (m_StatusCode != HttpResponseLayer::HttpStatusCodeUnknown)
 	{
+		int statusStringOffset = 13;
```
seladb, line 798: I'm not sure I understand this change. Maybe we should turn this variable into a `const`?

Packet++/src/SipLayer.cpp
```diff
@@ -753,9 +753,9 @@ int SipResponseFirstLine::getStatusCodeAsInt() const
 std::string SipResponseFirstLine::getStatusCodeString() const
 {
 	std::string result;
-	int statusStringOffset = 12;
 	if (m_StatusCode != SipResponseLayer::SipStatusCodeUnknown)
 	{
+		int statusStringOffset = 12;
```
seladb, line 758: ditto

Packet++/src/HttpLayer.cpp
seladb: What I meant is that now we can move this back to line 796 and `cppcheck` should not give a warning?

egecetin: Oh sorry I understand now fixing

seladb (approved):

Issue #966, opened 2022-09-18
Is it possible to support LINKTYPE_NFLOG  for pcap file in protocols module


ligewei:
Now, more and more pcap files from Android debugging are basing LINKTYPE_NFLOG, it is very important to support it. 

<img width="566" alt="image" src="https://user-images.githubusercontent.com/11908819/190884996-96b15446-c43c-4c26-b608-7501ab12b768.png">

seladb:
@ligewei thank you for reaching out!

I agree that can be a nice addition to PcapPlusPlus. It's not only about adding `LINKTYPE_NFLOG` but also supporting the Linux Netfilter NFLOG protocol. 

If you or anyone else would consider picking it up that'd be amazing!

Here is a link to a sample pcap file from the Wireshark web-site: https://wiki.wireshark.org/uploads/__moin_import__/attachments/SampleCaptures/nflog.pcap

jafar75:
@seladb If it helps I can to work on this issue

jafar75:
I investigated the examples and I think for now, default linktype is LINKTYPE_ETHERNET to read pcap files.

if I could to distinguish the link type of input, I can manage it to read the desired protocol.
@seladb Did I understand the scope of the issue correctly (completely)?

regards

seladb:
@jafar75 thank you for working on this issue!

The default link type is `LINKTYPE_ETHERNET`, but `PcapFileReaderDevice` supports other link types too. Here is where it reads it from the pcap file:
https://github.com/seladb/PcapPlusPlus/blob/6a864322ac1cbed0188ae5735597d28696ad3f94/Pcap%2B%2B/src/PcapFileDevice.cpp#L265

`PcapNgFileReaderDevice` also supports different link types, but in the pcapng format each packet can have its own link type, so getting it is done per packet:
https://github.com/seladb/PcapPlusPlus/blob/6a864322ac1cbed0188ae5735597d28696ad3f94/Pcap%2B%2B/src/PcapFileDevice.cpp#L387

The link type is read from the pcap/pcapng file and put in the `RawPacket` object, then when parsing the packet [`Packet::createFirstLayer()`](https://github.com/seladb/PcapPlusPlus/blob/6a864322ac1cbed0188ae5735597d28696ad3f94/Packet%2B%2B/src/Packet.cpp#L706) is called and uses the link type to know what should be the first layer of the packet (which matches the link type).

So basically what we need to do here is:
[...]

PR #987, opened 2022-10-29
Implement nflog linktype protocol 966

jafar75:
#966


Tests/Packet++Test/Tests/NflogTests.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "MacAddress.h"
+#include "Packet.h"
+#include "NflogLayer.h"
+#include "IPv4Layer.h"
+#include "SystemUtils.h"
+
+
+PTF_TEST_CASE(NflogPacketParsingTest)
+{
+    timeval time;
```
egecetin, line 13: nit: convert to tab

Tests/Packet++Test/Tests/NflogTests.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+	READ_FILE_AND_CREATE_PACKET_LINKTYPE(1, "PacketExamples/nflogPacket.dat", pcpp::LINKTYPE_NFLOG);
+
+	pcpp::Packet nflogPacket(&rawPacket1);
+
+	PTF_ASSERT_TRUE(nflogPacket.isPacketOfType(pcpp::NFLOG));
+	PTF_ASSERT_EQUAL(nflogPacket.getFirstLayer()->getProtocol(), pcpp::NFLOG, enum);
+	pcpp::NflogLayer* nflogLayer = nflogPacket.getLayerOfType<pcpp::NflogLayer>();
+	PTF_ASSERT_NOT_NULL(nflogLayer->getNextLayer());
+    pcpp::nflog_packet_header* pck_hdr = nflogLayer->getPacketHeader();
```
egecetin, line 24: nit: convert to tab

Tests/Packet++Test/PacketExamples/nflogPacket.dat
```diff
@@ -0,0 +1 @@
+0200002a080001000000030005000a0000000000080005000000000208000b000000000008000e0000000000410009004500003d021040004011208f0a00020f0a000203a542003500294156c04e0100000100000000000003777777076578616d706c65036e657400000100012f0a31
```
egecetin, line 1: To ease review in case of bug/problem etc. all `.dat` files have `pcap` versions. Could you add `pcap` of this packet?

Packet++/src/NflogLayer.cpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+std::pair<uint8_t*, int> NflogLayer::getPayload()
+{
+	uint8_t* data = m_Data + sizeof(nflog_header);
+	nflog_tlv* current_tlv = (nflog_tlv*)data;
+
+	uint16_t offset = sizeof(nflog_header);
+
+	while (current_tlv->tlv_type != NFULA_PAYLOAD) {
```
egecetin, line 26: Looks like this is a standard TLV structure if I'm not mistaken. What about using parsers in `TLVData.h` for better support/maintainability?

jafar75:
@egecetin 
thanks a lot for review, I'll resolve the issues ASAP

seladb:
@jafar75 any progress on this PR?

Packet++/src/NflogLayer.cpp
jafar75: @egecetin thanks
I think using parsers in TLVData.h will be a good idea,
I will work on it

seladb:
@egecetin can you take a look at this PR?

egecetin:
@seladb Sure. There is only one comment regarding `TLVData.h` except this part it is ok for me. @jafar75 Do you have any idea when will you have time to work on this?

jafar75:
> @seladb Sure. There is only one comment regarding `TLVData.h` except this part it is ok for me. @jafar75 Do you have any idea when will you have time to work on this?

@egecetin sorry for delay. I will finish the work at last the next 6 days

Packet++/src/NflogLayer.cpp
jafar75: @egecetin I have a some difficulties about this
in the example pcap file that @seladb added to issue page, format of TLVs are as len/type/data. but in implementation of TLVData.h there is type/len/data

https://github.com/seladb/PcapPlusPlus/blob/3e764a88a8fef38f0007673f2be9dd4f3f9c1597/Packet%2B%2B/header/TLVData.h#L29

So, in assign() method, when filling m_Data, the type/len will become different with the format of NFLOG. and I cannot use getTLVRecord() properly. 

replacing order of type/len in struct TLVRawData will affect other use cases rather than Nflog (because of calling getType())
I appreciate your suggestion.

jafar75: https://wiki.wireshark.org/uploads/__moin_import__/attachments/SampleCaptures/nflog.pcap

jafar75: I think its better to define a struct inside my class as TLVRawData with correct order of variables.

egecetin: > @egecetin I have a some difficulties about this in the example pcap file that @seladb added to issue page, format of TLVs are as len/type/data. but in implementation of TLVData.h there is type/len/data
> 
> https://github.com/seladb/PcapPlusPlus/blob/3e764a88a8fef38f0007673f2be9dd4f3f9c1597/Packet%2B%2B/header/TLVData.h#L29
> 
> So, in assign() method, when filling m_Data, the type/len will become different with the format of NFLOG. and I cannot use getTLVRecord() properly.
> 
> replacing order of type/len in struct TLVRawData will affect other use cases rather than Nflog (because of calling getType())

Looks like I missed that difference you are right. I will review your changes as soon as possible.

egecetin:
@jafar75 I will review your changes as soon as possible.

egecetin (commented):
@jafar75 some minor last comments. @seladb can you also review if there is anything I missed?

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,229 @@
...
+		 * Get a pointer to the Nflog header.
+		 * @return A pointer to the nflog_header
+		 */
+		nflog_header* getNflogHeader() const { return (nflog_header*)m_Data; }
+
+		/**
+		 * Get address family of the packet. e.g. 2 for ipv4 and 10 for ipv6
+		 * @return an unsigned char of address famliy
+		*/
+		uint8_t getFamily();
```
seladb, line 184: don't we also want to have a similar method to get the version `getVersion()` and resource ID `getResourceId()`?

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,229 @@
...
+		 * @return pointer to nflog_packet_header
+		*/
+		nflog_packet_header* getPacketHeader();
+
+		/**
+		 * returns a pair of pointer to tlv data and the length of the tlv
+		 * @param[in] type type of tlv by using enum defined as NflogTlvType
+		 * @return pair of <uint8_t*, int>
+		*/
+		std::pair<uint8_t*, int> getTlvByType(uint32_t type) const;
```
seladb, line 197: I think this method should return `NflogTlv` instead of `std::pair<uint8_t*, int>`

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,289 @@
...
+		/**
+		 * Currently identifies the following next layers: IPv4Layer, IPv6Layer using address family
+		 * Otherwise sets PayloadLayer
+		 */
+		void parseNextLayer();
+
+		/**
+		 * @return Size of nflog_header
+		 */
+		size_t getHeaderLen() const { return sizeof(nflog_header); }
```
seladb, line 268: The header len should be `sizeof(nflog_header)` + the size of all TLV records besides the data (payload) of `NflogTlvType::NFULA_PAYLOAD`.

There should be some logic that goes over the TLV records and sums their length until it reaches `NflogTlvType::NFULA_PAYLOAD` and then adds only the first 4 bytes of it (type + length)

Packet++/src/NflogLayer.cpp
```diff
@@ -0,0 +1,118 @@
...
+	// NFULA_PACKET_HDR is the first tlv
+	uint8_t* data = m_Data + sizeof(nflog_header) + sizeof(nflog_tlv);
+	return (nflog_packet_header*)data;
+}
+
+void NflogLayer::parseNextLayer()
+{
+	if (m_DataLen <= sizeof(nflog_header))
+		return;
+	auto payloadInfo = getTlvByType(NflogTlvType::NFULA_PAYLOAD);
```
seladb, line 87: After fetching the `payloadInfo` we should check if it's null:
```cpp
auto payloadInfo = getTlvByType(NflogTlvType::NFULA_PAYLOAD);
if (payloadInfo.isNull())
{
	return;
}
```

Packet++/src/NflogLayer.cpp
```diff
@@ -0,0 +1,118 @@
...
+	uint8_t* payload = payloadInfo.getValue();
+	size_t payloadLen = payloadInfo.getTotalSize() + sizeof(nflog_tlv);
```
seladb, line 89: the payload length is `payloadInfo.getTotalSize() - sizeof(nflog_tlv)`, no? 🤔

Packet++/src/NflogLayer.cpp
```diff
@@ -0,0 +1,118 @@
...
+	switch (family)
+	{
+	case PCPP_BSD_AF_INET:
+		m_NextLayer = IPv4Layer::isDataValid(payload, payloadLen)
+			? static_cast<Layer*>(new IPv4Layer(payload, payloadLen, this, m_Packet))
+			: static_cast<Layer*>(new PayloadLayer(payload, payloadLen, this, m_Packet));
+		break;
+	case PCPP_BSD_AF_INET6_BSD:
+	case PCPP_BSD_AF_INET6_FREEBSD:
+	case PCPP_BSD_AF_INET6_DARWIN:
```
seladb, line 102: are you sure all of these values are possible? 

In Wireshark they seem to only support these values:
https://github.com/wireshark/wireshark/blob/b4ef671fba521da6e10bd69b325c0540d0dd1271/epan/dissectors/packet-nflog.c#L253

https://github.com/wireshark/wireshark/blob/f5a40ada4d5c10dc266897945dfc77aa2e278f7e/epan/dissectors/packet-netlink.h#L82

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,289 @@
+#ifndef PACKETPP_NFLOG_LAYER
```
seladb, line 1: Please also add this protocol to `README.md`

Packet++/src/NflogLayer.cpp
jafar75: I added ipv4 and ipv6 related constants in the NflogLayer.h

Packet++/src/NflogLayer.cpp
jafar75: correct, I made a mistake here

Packet++/header/NflogLayer.h
jafar75: I thought only sizeof(nflog_header) would be sufficient. I changed this

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,229 @@
...
+
+		/** local-in, local-out, post-routing, etc */
+		uint8_t netfilter_hook;
+
+		/** one byte padding */
+		uint8_t padding;
+	};
+#pragma pack(pop)
+
+	enum NflogTlvType
```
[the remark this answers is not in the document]
jafar75: sounds strange, I changed and added doc for this

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,289 @@
...
+ 		NFULA_HWTYPE				= 15,
+        /// skbuff's MAC-layer header
+ 		NFULA_HWHEADER				= 16,
+        /// length of skbuff's MAC-layer header
+ 		NFULA_HWLEN					= 17,
+	};
+
+	/**
+	 * @class NflogTlv
+	 * A wrapper class for NFLOG tlv fields. This class does not create or modify TLVs related to NFLOG, but rather
```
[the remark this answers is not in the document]
seladb: What I meant was, instead of `A wrapper class for NFLOG tlv fields.` we should have `A wrapper class for NFLOG TLV fields.`

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,272 @@
...
+		/** local-in, local-out, post-routing, etc */
+		uint8_t netfilterHook;
+
+		/** one byte padding */
+		uint8_t padding;
+	};
+
+    /**
+     * @enum NflogTlvType
+     * represents different types of TLV types that may be find in a packet with NFLOG layer.
```
seladb, line 69: nit: better wording: 
```cpp
/**
 * @enum NflogTlvType
 * Represents TLV types of NFLOG packets
 */
```

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,272 @@
...
+        /** sequence number of pakets on all NFLOG sockets */
+ 		NFULA_SEQ_GLOBAL			= 13,
+        /** GID owning socket on which packet was sent/received */
+ 		NFULA_GID					= 14,
+        /** ARPHRD_ type of skbuff's device */
+ 		NFULA_HWTYPE				= 15,
+        /** skbuff's MAC-layer header */
+ 		NFULA_HWHEADER				= 16,
+        /** length of skbuff's MAC-layer header */
+ 		NFULA_HWLEN					= 17,
```
seladb, line 108: Please replace whitespaces with tabs in the comments

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,272 @@
...
+         * The version field is 0 for the current version of the pseudo-header
+         * @return an unsigned char for version
+        */
+        uint8_t getVersion();
+
+        /**
+         * Get Resource Id in packet header
+         * On one netlink socket it's possible to listen to several nflog groups; the resource ID is the nflog group for the packet
+        */
+        uint16_t getResourceId();
```
seladb, line 225: Please replace whitespaces with tabs

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,289 @@
...
+	class NflogLayer : public Layer
+	{
+	public:
+		/**
+		 * A constructor that creates the layer from an existing packet raw data
+		 * @param[in] data A pointer to the raw data (will be casted to ether_header)
+		 * @param[in] dataLen Size of the data in bytes
+		 * @param[in] packet A pointer to the Packet instance where layer will be stored in
+		 */
+		NflogLayer(uint8_t* data, size_t dataLen, Packet* packet) : Layer(data, dataLen, NULL, packet) { m_Protocol = NFLOG; }
```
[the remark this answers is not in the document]
seladb: @jafar75 this was not fixed

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,272 @@
...
+			{
+				// to pass from some unknown paddings after tlv with type NFULA_PREFIX
+				while (*recordRawData == 0)
+					recordRawData += 1;
+				m_Data = (NflogTLVRawData*)recordRawData;
+			}
+		}
+
+		/**
+		 * @return True if the TLV record raw data is NULL, false otherwise
```
seladb, line 162: please replace the word `NULL` with the word `nullptr`

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,272 @@
...
+		 * nothing to do for now
+		*/
+		void computeCalculateFields() {};
+
+		std::string toString() const;
+
+		OsiModelLayer getOsiModelLayer() const { return OsiModelDataLinkLayer; }
+
+	private:
+        void initLayer();
```
seladb, line 263: Remove `initLayer()`

Packet++/src/NflogLayer.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+size_t NflogLayer::getHeaderLen() const
+{
+	size_t headerLen = 0;
+	NflogTLV currentTLV =  m_TlvReader.getFirstTLVRecord(
+		getTlvsBasePtr(),
+		m_DataLen - sizeof(nflog_header));
+
+	while (currentTLV.getType() != static_cast<uint16_t> (NflogTlvType::NFULA_PAYLOAD))
```
seladb, line 88: I'd also add here, just in case: `!currentTLV.isNull()`, like this:
```cpp
while (!currentTLV.isNull() && currentTLV.getType() != static_cast<uint16_t> (NflogTlvType::NFULA_PAYLOAD))
{
   ...
}
```

Packet++/src/NflogLayer.cpp
```diff
@@ -0,0 +1,107 @@
...
+		break;
+	default:
+		m_NextLayer = new PayloadLayer(payload, payloadLen, this, m_Packet);
+	}
+
+}
+
+size_t NflogLayer::getHeaderLen() const
+{
+	size_t headerLen = 0;
```
seladb, line 83: nit: you can do:
```cpp
size_t headerLen = sizeof(nflog_header);
```

And then in line 99 you can just return the `headerLen`:
```cpp
return headerLen;
```

README.md
```diff
@@ -211,6 +211,7 @@ PcapPlusPlus currently supports parsing, editing and creation of packets of the
 9. VLAN
 10. VXLAN
 11. Wake on LAN (WoL)
+12. NFLOG (Linux Netfilter NFLOG)
```
seladb, line 214: Since we only have parsing capabilities for now please add `parsing only (no editing capabilities)`:
```markdown
NFLOG (Linux Netfilter NFLOG) - parsing only (no editing capabilities)
```

Tests/Packet++Test/Tests/NflogTests.cpp
```diff
@@ -0,0 +1,37 @@
...
+	pcpp::Packet nflogPacket(&rawPacket1);
+
+	PTF_ASSERT_TRUE(nflogPacket.isPacketOfType(pcpp::NFLOG));
+	PTF_ASSERT_EQUAL(nflogPacket.getFirstLayer()->getProtocol(), pcpp::NFLOG, enum);
+	pcpp::NflogLayer* nflogLayer = nflogPacket.getLayerOfType<pcpp::NflogLayer>();
+	PTF_ASSERT_NOT_NULL(nflogLayer->getNextLayer());
+
+	PTF_ASSERT_EQUAL(nflogLayer->getFamily(), pcpp::IPv4);
+	PTF_ASSERT_EQUAL(nflogLayer->getVersion(), 0);
+	PTF_ASSERT_EQUAL(be16toh(htobe16(nflogLayer->getResourceId())), 42);
```
seladb, line 27: Since `getResourceId()` already use `htobe16()` the check here should be simpler:
```cpp
PTF_ASSERT_EQUAL(nflogLayer->getResourceId(), 42);
```

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,272 @@
...
+         * Get Resource Id in packet header
+         * On one netlink socket it's possible to listen to several nflog groups; the resource ID is the nflog group for the packet
+        */
+        uint16_t getResourceId();
+
+		/**
+		 * get packet header of the packet as the tlv value of 1
+		 * @return pointer to nflog_packet_header
+		*/
+		nflog_packet_header* getPacketHeader();
```
seladb, line 231: This method parses the first TLV, why is it important to provide a special method for it?

Tests/Packet++Test/Tests/NflogTests.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+	pcpp::nflog_packet_header* pck_hdr = nflogLayer->getPacketHeader();
+	PTF_ASSERT_EQUAL(pck_hdr->hardwareProtocol, 0);
+	PTF_ASSERT_EQUAL((int)pck_hdr->netfilterHook, 3);
+	PTF_ASSERT_EQUAL(nflogLayer->getNextLayer()->getProtocol(), pcpp::IPv4, enum);
+
+	pcpp::NflogTLV $payloadInfo = nflogLayer->getTlvByType(pcpp::NflogTlvType::NFULA_PAYLOAD);
```
seladb, line 34: why is this `$` sign needed? Can't we call the variable `payloadInfo`? 🤔

Tests/Packet++Test/Tests/NflogTests.cpp
```diff
@@ -0,0 +1,37 @@
...
+	PTF_ASSERT_EQUAL($payloadInfo.getTotalSize(), 65);
+	PTF_ASSERT_EQUAL($payloadInfo.getValue()[0], 'E');
```
seladb, line 36: I think it would be nice to parse all TLVs and make sure each has the correct type, length and value, similar to what we do here:
https://github.com/seladb/PcapPlusPlus/blob/1b6a8e2ff9f151bae6ae7f5f61e0eeda9eea34df/Tests/Packet%2B%2BTest/Tests/DhcpV6Tests.cpp#L25-L52

Tests/Packet++Test/Tests/NflogTests.cpp
```diff
@@ -0,0 +1,37 @@
...
+}
```
seladb, line 37: More tests we can add:
- Check that the size of `getHeaderLen()` is correct
- Get the next layer of `nflogPacket` and make sure it's an `IPv4Layer` (tests the `parseNextLayer()` method)
- If you have a packet example of NF_LOG followed by an IPv6 layer it'd be nice to add

Packet++/header/NflogLayer.h
jafar75: I think in the beginning of the development, I got it wrong instead of nflog_header

Tests/Packet++Test/Tests/NflogTests.cpp
jafar75: I was coding PHP in my job that day :))

Tests/Packet++Test/Tests/NflogTests.cpp
jafar75: developed a similar testing process for all tlvs

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,264 @@
+#ifndef PACKETPP_NFLOG_LAYER
+#define PACKETPP_NFLOG_LAYER
+
+#include "Layer.h"
+#include "TLVData.h"
+#include "GeneralUtils.h"
```
seladb, line 6: Do we use this include anywhere? If not, we should remove it.
If we only use it in `NflogLayer.cpp` it should move there

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,264 @@
...
+ * \namespace pcpp
+ * \brief The main namespace for the PcapPlusPlus lib
+ */
+namespace pcpp
+{
+
+	/** IPv4 protocol */
+	#define PCPP_WS_NFPROTO_IPV4		2
+	/** IPv6 protocol */
+	#define PCPP_WS_NFPROTO_IPV6		10
```
seladb, line 20: I think this can move the `NflogLayer.cpp`, no?

Tests/Packet++Test/Tests/NflogTests.cpp
```diff
@@ -0,0 +1,61 @@
...
+	for (int i = 0; i < 6; i++) {
+		pcpp::NflogTlv tlv = nflogLayer->getTlvByType(expectedTypes[i]);
+
+		PTF_ASSERT_EQUAL(tlv.getTotalSize(), optSizes[i]);
+		PTF_ASSERT_EQUAL(pcpp::byteArrayToHexString(tlv.getRecordBasePtr(), optSizes[i]), optDataAsHexString[i]);
+	}
+
+	/// sum of all TLVs before payload + size of nflog_header + size of (recordLength + recordType) variables of payload TLV
+	int headerLen = 45;
+	PTF_ASSERT_EQUAL(nflogLayer->getHeaderLen(), headerLen, enum);
```
seladb, line 60: We can simplify this:
```cpp
PTF_ASSERT_EQUAL(nflogLayer->getHeaderLen(), 45);
```

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,264 @@
...
+#pragma pack(push, 1)
+	struct nflog_tlv
+	{
+		/** tlv length */
+		uint16_t tlvLength;
+
+		/** tlv type */
+		uint16_t tlvType;
+	};
+#pragma pack(pop)
```
seladb, line 48: We don't really use this struct. The only place we use it is here: https://github.com/jafar75/PcapPlusPlus/blob/4e06ec9d58a27706c85af4e82aad0324eb418885/Packet%2B%2B/src/NflogLayer.cpp#L53

I think we can remove it from this file and simply do:

```cpp
size_t payloadLen = payloadInfo.getTotalSize() - sizeof(uint16_t) * 2;
```

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,264 @@
...
+	{
+		/** e.g. ipv4, unknown, etc */
+		uint16_t hardwareProtocol;
+
+		/** local-in, local-out, post-routing, etc */
+		uint8_t netfilterHook;
+
+		/** one byte padding */
+		uint8_t padding;
+	};
```
seladb, line 65: We don't use this struct anywhere so I think we can remove it

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,264 @@
...
+	/**
+	 * @struct nflog_header
+	 * Represents Nflog header
+	 */
+#pragma pack(push, 1)
+	struct nflog_header
+	{
+		uint8_t addressFamily;
+		uint8_t version;
+		uint16_t resourceId;
```
seladb, line 31: Please add documentation for the struct members

Packet++/header/NflogLayer.h
```diff
@@ -0,0 +1,264 @@
...
+		void parseNextLayer();
+
+		/**
+		 * @return Size of nflog_header
+		 */
+		size_t getHeaderLen() const;
+
+		/**
+		 * nothing to do for now
+		*/
```
seladb, line 249: Slightly better wording (which also match our convention):
```cpp
/**
 * Does nothing for this layer
 */
```

Packet++/src/Packet.cpp
```diff
@@ -771,6 +772,10 @@ Layer* Packet::createFirstLayer(LinkLayerType linkType)
 			? static_cast<Layer*>(new IPv6Layer((uint8_t*)rawData, rawDataLen, nullptr, this))
 			: static_cast<Layer*>(new PayloadLayer((uint8_t*)rawData, rawDataLen, nullptr, this));
 	}
+	else if (linkType == LINKTYPE_NFLOG)
+	{
+		return new NflogLayer((uint8_t*)rawData, rawDataLen, this);
```
seladb, line 777: I think we should add a static method:
```cpp
/**
 * A static method that validates the input data
 * @param[in] data The pointer to the beginning of a byte stream of an NFLOG packet
 * @param[in] dataLen The length of the byte stream
 * @return True if the data is valid and can represent an NFLOG packet
 */
static bool isDataValid(const uint8_t* data, size_t dataLen)
{
	return data && dataLen >= sizeof(nflog_header);
}
```

Then we can do here as we do in other places:
```cpp
else if (linkType == LINKTYPE_NFLOG)
{
	return NflogLayer::isDataValid(rawData, rawDataLen)
		? static_cast<Layer*>(new NflogLayer((uint8_t*)rawData, rawDataLen, this))
		: static_cast<Layer*>(new PayloadLayer((uint8_t*)rawData, rawDataLen, nullptr, this));
```

seladb (approved):

PR #996, opened 2022-11-17
Added extra matchSubnet definition

yishai1999:
Added another defintion for IPv4Address::matchSubnet where the parameter is a string representing a subnet with a mask prefix in the well known format 'X.X.X.X/Y'

Signed-off-by: Yishai Jaffe <yishai1999@gmail.com>

### DIFF
--- a/Common++/header/IpAddress.h
+++ b/Common++/header/IpAddress.h
@@ -112,6 +112,14 @@ namespace pcpp
 
 		/**
 		 * Checks whether the address matches a subnet.
+		 * For example: if subnet is 10.1.1.1/24 and address is 10.1.1.9 then the method will return true
+		 * Another example: if subnet is 10.1.1.1/16 and address is 11.1.1.9 then the method will return false
+		 * @param[in] subnet A string in X.X.X.X/Y format representing the masked subnet to compare with the address
+		 */
+		bool matchSubnet(const std::string& subnet) const;
+
+		/**
+		 * Checks whether the address matches a subnet.
 		 * For example: if subnet is 10.1.1.X, subnet mask is 255.255.255.0 and address is 10.1.1.9 then the method will return true
 		 * Another example: if subnet is 10.1.X.X, subnet mask is 255.0.0.0 and address is 11.1.1.9 then the method will return false
 		 * @param[in] subnet The subnet to be verified. Notice it's an IPv4Address type, so subnets with don't-cares (like 10.0.0.X) must have some number
--- a/Common++/src/IpAddress.cpp
+++ b/Common++/src/IpAddress.cpp
@@ -1,6 +1,11 @@
 #define LOG_MODULE CommonLogModuleIpUtils
 
+#include <algorithm>
+#include <cmath>
 #include <errno.h>
+#include <sstream>
+#include <stdexcept>
+#include <stdint.h>
 #include "Logger.h"
 #include "IpUtils.h"
 #include "IpAddress.h"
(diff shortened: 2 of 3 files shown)


Common++/header/IpAddress.h
```diff
@@ -110,6 +110,14 @@ namespace pcpp
 		 */
 		bool operator!=(const IPv4Address& rhs) const	{ return !(*this == rhs); }
 
+		/**
+		 * Checks whether the address matches a subnet.
+		 * For example: if subnet is 10.1.1.1/24 and address is 10.1.1.9 then the method will return true
+		 * Another example: if subnet is 10.1.1.1/16 and address is 11.1.1.9 then the method will return false
+		 * @param[in] subnet A string in X.X.X.X/Y format representing the masked subnet to compare with the address
+		 */
+		bool matchSubnet(const std::string& subnet) const;
```
clementperon, line 119: whould be nice to introduce a proper IPv4Network class to check the CIDR notation is valid.

But it's a bit out of scope for this MR

yishai1999: So can we merge?

Common++/src/IpAddress.cpp
```diff
@@ -43,6 +46,29 @@ namespace pcpp
...
 
 
+	bool IPv4Address::matchSubnet(const std::string& subnet) const
+	{
+		std::stringstream ss(subnet);
+		std::string subnetOnly, subnetPrefixStr;
+		std::getline(ss, subnetOnly, '/');
+		std::getline(ss, subnetPrefixStr);
+
+		uint32_t subnetPrefix = std::stoi(subnetPrefixStr);
```
seladb, line 56: `std::stoi` throws a [`std::invalid_argument`](https://en.cppreference.com/w/cpp/error/invalid_argument) exception if no conversion could be performed  (https://en.cppreference.com/w/cpp/string/basic_string/stol), maybe we should surround it with `try...catch` to prevent this exception which might be difficult for users to understand?

seladb, line 56: We can also check that `subnetPrefix` is lower or equal to 32

Tests/Pcap++Test/Tests/IpMacTests.cpp
```diff
@@ -40,12 +40,15 @@ PTF_TEST_CASE(TestIPAddress)
 	PTF_ASSERT_EQUAL(ip4AddrFromIpAddr, secondIPv4Address);
 
 	pcpp::IPv4Address ipv4Addr("10.0.0.4"), subnet1("10.0.0.0"), subnet2("10.10.0.0"), mask("255.255.255.0");
+	std::string maskedSubnet1("10.0.0.0/24"), maskedSubnet2("10.10.0.0/24");
```
seladb, line 43: I'd add a few negative tests when the subnet isn't valid, for example: `"aaaa"`, `"10.0.0.0"`, "`10.0.0.0/aa"`, `"10.0.0.0/33"`, `"999.999.1.1/24"`

Common++/header/IpAddress.h
seladb: @yishai1999 I'm not sure if you saw the remaining comments. Let's fix them and merge the PR?

yishai1999: Yup, I'm on it.

Common++/src/IpAddress.cpp
yishai1999: Done

yishai1999: Done

Tests/Pcap++Test/Tests/IpMacTests.cpp
yishai1999: Done

Common++/header/IpAddress.h
yishai1999: I fixed whatever you commented about but the mac and openBSD CI jobs don't pass. Any clue why?

Tests/Pcap++Test/Tests/IpMacTests.cpp
```diff
@@ -40,12 +40,21 @@ PTF_TEST_CASE(TestIPAddress)
...
 	PTF_ASSERT_TRUE(mask.isValid());
 	PTF_ASSERT_TRUE(ipv4Addr.matchSubnet(subnet1, mask));
 	PTF_ASSERT_FALSE(ipv4Addr.matchSubnet(subnet2, mask));
+	PTF_ASSERT_TRUE(ipv4Addr.matchSubnet(maskedSubnet1));
+	PTF_ASSERT_FALSE(ipv4Addr.matchSubnet(maskedSubnet2));
+	PTF_ASSERT_FALSE(ipv4Addr.matchSubnet(invalidMaskedSubnet1));
+	PTF_ASSERT_FALSE(ipv4Addr.matchSubnet(invalidMaskedSubnet2));
+	PTF_ASSERT_FALSE(ipv4Addr.matchSubnet(invalidMaskedSubnet3));
+	PTF_ASSERT_FALSE(ipv4Addr.matchSubnet(invalidMaskedSubnet4));
+	PTF_ASSERT_FALSE(ipv4Addr.matchSubnet(invalidMaskedSubnet5));
```
seladb, line 57: Please add `pcpp::Logger::getInstance().suppressLogs()` before the negative tests and `pcpp::Logger::getInstance().enableLogs()`, same as we do here for example: https://github.com/seladb/PcapPlusPlus/blob/b1c4e99c40071bf0639b773c17837f096d4f8cfa/Tests/Pcap%2B%2BTest/Tests/RawSocketTests.cpp#L27-L32

Otherwise the test output contains log errors:

image

https://github.com/seladb/PcapPlusPlus/actions/runs/3694904138/jobs/6271844098

Common++/src/IpAddress.cpp
```diff
@@ -43,6 +48,46 @@ namespace pcpp
...
+		std::getline(ss, subnetPrefixStr);
+
+		if (subnetPrefixStr.empty() || !std::all_of(subnetPrefixStr.begin(), subnetPrefixStr.end(), ::isdigit)) {
+			PCPP_LOG_ERROR("subnet prefix '" << subnetPrefixStr << "' must be an integer");
+			return false;
+		}
+
+		uint32_t subnetPrefix;
+		try {
+			subnetPrefix = std::stoi(subnetPrefixStr);
```
seladb, line 65: I noticed that if `subnetPrefixStr` is too large `std:stoi` throws `std::out_of_range` exception, here is an example:
https://www.online-cpp.com/JuthUqrVyI

We probably want to catch that too...

seladb (approved):

seladb:
Thank you @yishai1999 for this great contribution to the project, highly appreciated! 🙏 ❤️

PR #999, opened 2022-11-27
Fix noExplicitConstructor

egecetin:
Fix cppcheck warning noExplicitConstructor


clementperon (approved):

Common++/header/MacAddress.h
```diff
@@ -40,21 +40,21 @@ namespace pcpp
...
-		MacAddress(const char* addr) { init(addr); }
+		explicit MacAddress(const char* addr) { init(addr); }
 
 		/**
 		 *  A constructor that creates an instance of the class out of a std::string.
 		 *  If the string doesn't represent a valid MAC address, instance will be invalid, meaning isValid() will return false
 	 	 *	@param[in] addr A pointer to the string
 		 */
-		MacAddress(const std::string& addr) { init(addr.c_str()); }
+		explicit MacAddress(const std::string& addr) { init(addr.c_str()); }
```
seladb, line 57: Maybe this c'tor should not be `explicit`?

That way we can do something like this:

```cpp
void foo(const MacAddress& mac)
{
    std::cout << mac;
}

int main()
{
    foo("00:00:5e:00:53:af");
    return 0;
}
```

Pcap++/header/PcapFileDevice.h
```diff
@@ -117,7 +117,7 @@ namespace pcpp
 		 * isn't opened yet, so reading packets will fail. For opening the file call open()
 		 * @param[in] fileName The full path of the file to read
 		 */
-		PcapFileReaderDevice(const std::string& fileName) : IFileReaderDevice(fileName), m_PcapLinkLayerType(LINKTYPE_ETHERNET) {}
+		explicit PcapFileReaderDevice(const std::string& fileName) : IFileReaderDevice(fileName), m_PcapLinkLayerType(LINKTYPE_ETHERNET) {}
```
seladb, line 120: ditto for this c'tor all the classes in this file

Tests/Packet++Test/Tests/WakeOnLanTests.cpp
```diff
@@ -100,7 +100,7 @@ PTF_TEST_CASE(WakeOnLanEditTests)
 	pcpp::WakeOnLanLayer *wolLayer3 = wolPacketEdited2.getLayerOfType<pcpp::WakeOnLanLayer>();
 	PTF_ASSERT_NOT_NULL(wolLayer3);
 
-	wolLayer1->setTargetAddr("00:90:27:85:cf:01");
+	wolLayer1->setTargetAddr(pcpp::MacAddress("00:90:27:85:cf:01"));
```
seladb, line 103: If we keep the `MacAddress` c'tor non-explicit we can revert this change

Common++/header/MacAddress.h
egecetin: The same goes for constructors in `IpAddress.h` which I have excluded for now in #1003 to double check. Probably excluding both `MacAddress.h` and `IpAddress.h` might do the trick to make the functions easier to use.

Pcap++/header/PcapFileDevice.h
egecetin: But I'm not sure about these classes. Since `char*` currently is a valid input. For example https://github.com/seladb/PcapPlusPlus/blob/0b64a41ffcb4a1d7dbb7f1b97a3c84dc24792661/Tests/Pcap%2B%2BTest/Tests/IPFragmentationTests.cpp#L419

seladb (approved):

PR #1000, opened 2022-11-27
Fix out of bounds access at SnoopFileReaderDevice

egecetin:
Index 9 is out of bounds

### DIFF
--- a/Pcap++/src/PcapFileDevice.cpp
+++ b/Pcap++/src/PcapFileDevice.cpp
@@ -163,7 +163,7 @@ bool SnoopFileReaderDevice::open()
 		LINKTYPE_FDDI		/* FDDI */
 	};
 	uint32_t datalink_type = be32toh(snoop_file_header.datalink_type);
-	if (datalink_type > 9)
+	if (datalink_type > 8)
 	{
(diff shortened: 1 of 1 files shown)


Pcap++/src/PcapFileDevice.cpp
```diff
@@ -163,7 +163,7 @@ bool SnoopFileReaderDevice::open()
 		LINKTYPE_FDDI		/* FDDI */
 	};
 	uint32_t datalink_type = be32toh(snoop_file_header.datalink_type);
-	if (datalink_type > 9)
+	if (datalink_type > 8)
```
clementperon, line 166: Would be better to avoid this kind of magic number and prefer something like 
```#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))```

or the C++ way 
```
template <typename T, size_t N>
constexpr size_t array_size(T (&)[N]) { return N; }
```

egecetin: Good idea. Thanks

clementperon (approved):

seladb (approved):

PR #1001, opened 2022-11-27
Fix unsigned less than zero checks + redundant assignments

egecetin:


Examples/PfRingExample-FilterTraffic/main.cpp
```diff
@@ -326,7 +326,7 @@ int main(int argc, char* argv[])
 			case 'p':
 			{
 				srcPortToMatch = atoi(optarg);
-				if (srcPortToMatch <= 0)
+				if (atoi(optarg) <= 0)
 				{
 					EXIT_WITH_ERROR_AND_PRINT_USAGE("Source port to match isn't a valid TCP/UDP port");
 				}
```
clementperon, line 332: Not cool to call atoi() two times IMO maybe something like this ?
```suggestion
				int ret = atoi(optarg);
				if (ret <= 0)
				{
					EXIT_WITH_ERROR_AND_PRINT_USAGE("Source port to match isn't a valid TCP/UDP port");
				}
				srcPortToMatch = ret;
```

Pcap++/src/DpdkDevice.cpp
```diff
@@ -792,7 +792,7 @@ uint16_t DpdkDevice::receivePackets(MBufRawPacketVector& rawPacketsArr, uint16_t
 	//the following line trashes the log with many messages. Uncomment only if necessary
 	//PCPP_LOG_DEBUG("Captured %d packets", numOfPktsReceived);
 
-	if (unlikely(numOfPktsReceived <= 0))
+	if (unlikely(!numOfPktsReceived))
```
clementperon, line 795: is this check really needed ?

the for loop will do nothing and we will return numOfPktsReceived at the end

egecetin: To be honest I didn't check all the code when replacing but you are right. This line is unnecessary

Examples/DpdkExample-FilterTraffic/main.cpp
```diff
@@ -379,20 +379,22 @@ int main(int argc, char* argv[])
 			}
 			case 'p':
 			{
-				srcPortToMatch = atoi(optarg);
-				if (srcPortToMatch <= 0)
+				int ret = atoi(optarg);
+				if (ret <= 0)
```
clementperon, line 383: Also check the max valid port 

```suggestion
				if (ret <= 0 || ret > 65535)
```

Examples/DpdkExample-FilterTraffic/main.cpp
```diff
@@ -379,20 +379,22 @@ int main(int argc, char* argv[])
 			}
 			case 'p':
 			{
-				srcPortToMatch = atoi(optarg);
-				if (srcPortToMatch <= 0)
+				int ret = atoi(optarg);
+				if (ret <= 0 && ret > 65535)
```
clementperon, line 383: || instead of &&

Packet++/src/TelnetLayer.cpp
```diff
@@ -14,7 +14,7 @@ namespace pcpp
 bool TelnetLayer::isDataField(uint8_t *pos)
 {
 	// "FF FF" means data
-	return pos[0] != InterpretAsCommand || (pos[0] == InterpretAsCommand && pos[1] == InterpretAsCommand);
+	return pos[0] != InterpretAsCommand || pos[1] == InterpretAsCommand;
```
clementperon, line 17: or return !isCommandField(pos);

Packet++/src/TelnetLayer.cpp
```diff
@@ -471,7 +471,7 @@ std::string TelnetLayer::getTelnetOptionAsString(TelnetOption val)
 
 std::string TelnetLayer::toString() const
 {
-	if (m_Data[0] != InterpretAsCommand || (m_Data[0] == InterpretAsCommand && m_Data[1] == InterpretAsCommand))
+	if (m_Data[0] != InterpretAsCommand || m_Data[1] == InterpretAsCommand)
```
clementperon, line 474: maybe reuse isDataField() no ?

Packet++/src/TelnetLayer.cpp
egecetin: Good idea!

Pcap++/src/DpdkDevice.cpp
```diff
@@ -840,7 +835,7 @@ uint16_t DpdkDevice::receivePackets(MBufRawPacket** rawPacketsArr, uint16_t rawP
 	struct rte_mbuf* mBufArray[rawPacketArrLength];
 	uint16_t packetsReceived = rte_eth_rx_burst(m_Id, rxQueueId, mBufArray, rawPacketArrLength);
 
-	if (unlikely(packetsReceived <= 0))
+	if (unlikely(!packetsReceived))
```
clementperon, line 838: i think we can remove this check too

Pcap++/src/DpdkDevice.cpp
```diff
@@ -883,7 +878,7 @@ uint16_t DpdkDevice::receivePackets(Packet** packetsArr, uint16_t packetsArrLeng
 	struct rte_mbuf* mBufArray[packetsArrLength];
 	uint16_t packetsReceived = rte_eth_rx_burst(m_Id, rxQueueId, mBufArray, packetsArrLength);
 
-	if (unlikely(packetsReceived <= 0))
+	if (unlikely(!packetsReceived))
```
clementperon, line 881: this one too

Pcap++/src/KniDevice.cpp
```diff
@@ -489,7 +489,7 @@ uint16_t KniDevice::receivePackets(MBufRawPacketVector& rawPacketsArr)
 	//the following line trashes the log with many messages. Uncomment only if necessary
 	//PCPP_LOG_DEBUG("KNI Captured %d packets", numOfPktsReceived);
 
-	if (unlikely(numOfPktsReceived <= 0))
+	if (unlikely(!numOfPktsReceived))
```
clementperon, line 492: same

Pcap++/src/KniDevice.cpp
```diff
@@ -572,7 +567,7 @@ uint16_t KniDevice::receivePackets(Packet** packetsArr, uint16_t packetsArrLengt
 	struct rte_mbuf** mBufArray = CPP_VLA(struct rte_mbuf*, packetsArrLength);
 	uint16_t packetsReceived = rte_kni_rx_burst(m_Device, mBufArray, MAX_BURST_SIZE);
 
-	if (unlikely(packetsReceived <= 0))
+	if (unlikely(!packetsReceived))
```
clementperon, line 570: same

clementperon (approved):

Examples/HttpAnalyzer/main.cpp
```diff
@@ -582,10 +582,6 @@ int main(int argc, char* argv[])
 	{
 		EXIT_WITH_ERROR("Please input a number between 0 to 65535");
 	}
-	if (nPort < 0 || nPort > 65535)
```
seladb, line 585: I think there is a bug here, not in your fix but in the previous code: if `port` represents a number that is larger than 65535 or smaller than 0 then `nPort == 65535`. Also if `port` is not a number then `nPort == 0`. Maybe we want to fix this instead of removing these lines?

Pcap++/src/DpdkDevice.cpp
```diff
@@ -792,11 +792,6 @@ uint16_t DpdkDevice::receivePackets(MBufRawPacketVector& rawPacketsArr, uint16_t
 	//the following line trashes the log with many messages. Uncomment only if necessary
 	//PCPP_LOG_DEBUG("Captured %d packets", numOfPktsReceived);
 
-	if (unlikely(numOfPktsReceived <= 0))
-	{
-		return 0;
-	}
```
seladb, line 798: Maybe this is not significant, but if `numOfPktsReceived == 0` we can return here and don't need to run `clock_gettime` and a for loop with zero iterations.

The same comment applies to the rest of the similar changes in this PR

egecetin: clock_gettime calls are very cheap, but I agreed, every little bit counts.

Pcap++/src/DpdkDevice.cpp
```diff
@@ -841,9 +839,7 @@ uint16_t DpdkDevice::receivePackets(MBufRawPacket** rawPacketsArr, uint16_t rawP
 	uint16_t packetsReceived = rte_eth_rx_burst(m_Id, rxQueueId, mBufArray, rawPacketArrLength);
 
 	if (unlikely(packetsReceived <= 0))
-	{
```
seladb, line 844: It's a better practice to add curly brackets `{...}` even when the `if` clause includes one statement - it's more readable that way.

Same for the other changes in this file and in `Pcap++/src/KniDevice.cpp`

egecetin: Also while revoking i forgot to fix unsigned less than zero warning. Fixed now.

seladb (approved):

PR #1003, opened 2022-11-27
pre-commit cppcheck

egecetin:
Still work in progress. Warnings suppressed between L35-L37 should be fixed/double checked before merging


egecetin:
Added final minor fixes. I have suppressed warnings below for now because I'm a bit lazy to fix them, but I think they also should be fixed. (Postfix and cstyle cast will fix themselves when porting the code to C++11.)

```
missingOverride:*
postfixOperator:*
cstyleCast:*
```

cppcheckSuppressions.txt
```diff
@@ -0,0 +1,38 @@
...
+uninitdata:Pcap++/src/PcapLiveDevice.cpp:870
+uninitMemberVar:Packet++/header/IcmpLayer.h:399
+unreadVariable:Packet++/src/DnsResource.cpp:111
+unreadVariable:Packet++/src/DnsResource.cpp:137
+unusedVariable:Tests/Pcap++Test/Tests/LoggerTests.cpp:168
+unusedVariable:Tests/Pcap++Test/Tests/LoggerTests.cpp:196
+unusedVariable:Tests/Packet++Test/Tests/SomeIpTests.cpp:23
+unusedVariable:Tests/Packet++Test/Tests/SomeIpTests.cpp:46
+unusedVariable:Tests/Packet++Test/Tests/SomeIpTests.cpp:173
+useStlAlgorithm:Tests/Pcap++Test/Tests/LoggerTests.cpp:187
```
seladb, line 31: For the line-specific suppressions, maybe we should use `--inline-suppr` and put them inline?

.github/workflows/build_and_test.yml
```diff
@@ -9,6 +9,9 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.8.x"
+      - name: Install dependencies
+        run: |
+          sudo apt install cppcheck
```
seladb, line 14: This pre-commit hook requires an external installation of `cppcheck`. I tried to run it on a Windows machine without installing it and it give a generic error:
```shell
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

This might be a pretty bad user experience for new contributors...

Regardless, we don't have any documentation about our pre-commit hooks - probably the best place to put it is in [`CONTRIBUTING.md`](https://github.com/seladb/PcapPlusPlus/blob/master/CONTRIBUTING.md), and there we can mention the requirement of installing `cppcheck`. What do you think?

egecetin: I updated `CONTRIBUTING.md`. One option might be removing cppcheck from pre-commit and only checking at Github actions while encouraging people to use from command line, but it might be worse experience.

seladb:
@egecetin I just installed the latest version of `cppcheck` - 2.9 and it found a few additional issues (hence it fails):

```shell
Pcap++\header\DpdkDevice.h:287:4: style: Struct 'DpdkDeviceConfiguration' has a 
constructor with 1 argument that is not explicit. [noExplicitConstructor]       
   DpdkDeviceConfiguration(uint16_t receiveDescriptorsNumber = 128,
   ^
Pcap++\header\DpdkDevice.h:287:4: style: Struct 'DpdkDeviceConfiguration' has a 
constructor with 1 argument that is not explicit. [noExplicitConstructor]       
   DpdkDeviceConfiguration(uint16_t receiveDescriptorsNumber = 128,
   ^
Examples\HttpAnalyzer\HttpStatsCollector.h:436:19: performance: Ineffective call
 of function 'substr' because a prefix of the string is assigned to itself. Use 
resize() or pop_back() instead. [uselessCallsSubstr]
    contentType = contentType.substr(0, charsetPos);
Packet++\src\NtpLayer.cpp:519:25: style: Operator '|' with one operand equal to zero is redundant. [badBitmaskCheck]
        retval = retval | uint32_t(integerPartInt);
                        ^
Packet++\src\NtpLayer.cpp:537:25: style: Operator '|' with one operand equal to zero is redundant. [badBitmaskCheck]
        retval = retval | uint64_t(integerPartInt);
```

I think the build passes because the latest version in `apt-get` is 2.7: https://packages.ubuntu.com/jammy/devel/cppcheck

Maybe we should pin the `cppcheck` version? Otherwise the build may fail each time `cppcheck` has an update...
We should also mention which version to be used in `CONTRIBUTING.md`.

If we decide to use `brew` (MacOS, Linux) or `choco` (Windows) for installation we can pin the version, for example:
```shell
brew install cppcheck@2.9.3
```

```shell
choco install cppcheck --version=2.9
```

CONTRIBUTING.md
```diff
@@ -22,6 +22,11 @@ Every code contribution to this project is highly valued and appreciated. I enco
    - PcapPlusPlus compiles successfully on your machine, including all unit-tests and examples (just run `make` from PcapPlusPlus main directory)
    - Unit-tests pass succssfully on your machine (both `Tests/Packet++Test` and `Tests/Pcap++Test`)
    - All new APIs are well documented using Doxygen (please use @ for keywords)
+   - Make sure that pre-commit hooks are passing by using `pre-commit run --all-files`. For pre-commit hooks you need to install `cppcheck` on your system. Recommended and tested version is 2.7. You can install by using
+     - For Linux (apt) `sudo apt install cppcheck`
+     - For MacOS/Linux (brew) `brew install cppcheck@2.7`
```
egecetin, line 27: @seladb I fixed at 2.7 for now to keep simple github workflow yml. Can you check the brew command on a MacOS?

seladb: With `homebrew` it's not easy to install an old version of a package, `brew install cppcheck@2.7` doesn't work 😕  

Maybe instead of `brew` we can just add a link to the release page: https://github.com/danmar/cppcheck/releases/tag/2.7 ?

We can do something like this:
```markdown
For pre-commit hooks you need to install `cppcheck` on your system. Recommended and tested version is 2.7. You can install by using:
- For Linux (apt) `sudo apt install cppcheck=2.7-1`
- For Windows (using `choco`): `choco install cppcheck --version=2.7`
- For Windows (MSI install): https://github.com/danmar/cppcheck/releases/download/2.7/cppcheck-2.7-x64-Setup.msi
- Build from source: https://github.com/danmar/cppcheck/releases/tag/2.7
```

seladb (approved):
Fix the remaining comment and merge!

PR #1019, opened 2022-12-23
Improve the test framework

seladb:
This PR includes the following improvements to the test framework:
- Evaluate `actual` and `expected` only once at the beginning of the assertion macros instead of twice (once inside the `if` statement and then again when printing)
- Better printing of assertion failures
- Support `PTF_ASSERT_EQUAL` with enum classes
- Remove `object_no_str_PTF_PRINT_TYPE` which is not really needed (no point using `PTF_ASSERT_EQUAL` with objects)
- Increase the width from 30 to 35 chars to fix this: 
image

### DIFF
--- a/Tests/PcppTestFramework/PcppTestFramework.h
+++ b/Tests/PcppTestFramework/PcppTestFramework.h
@@ -6,20 +6,44 @@
 #include "memplumber.h"
 #include "PcppTestFrameworkCommon.h"
 
-#define _PTF_PRINT_TYPE(val) "[" << val << "]"
-#define object_no_str_PTF_PRINT_TYPE(val) "'" << #val << "'"
-#define hex_PTF_PRINT_TYPE(val) "0x" << std::hex << val
-#define enum_PTF_PRINT_TYPE(val) #val << "[" << val << "]"
-#define ptr_PTF_PRINT_TYPE(val) #val << "[" << val << "]"
+#define _PTF_PRINT_TYPE_ACTUAL(exp, val) val
+#define _PTF_PRINT_TYPE_EXPECTED(exp, val) val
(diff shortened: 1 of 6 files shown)


Tests/PcppTestFramework/PcppTestFramework.h
```diff
@@ -6,20 +6,44 @@
...
+#define enum_PTF_PRINT_TYPE_ACTUAL(exp, val) "enum[" << val << "]"
+#define enum_PTF_PRINT_TYPE_EXPECTED(exp, val) exp << "[" << val << "]"
+#define ptr_PTF_PRINT_TYPE_ACTUAL(exp, val) exp << "[ptr: " << val << "]"
+#define ptr_PTF_PRINT_TYPE_EXPECTED(exp, val) exp << "[ptr: " << val << "]"
+#define enumclass_PTF_PRINT_TYPE_ACTUAL(exp, val) "enum[" << +static_cast<std::underlying_type<decltype(val)>::type>(val) << "]"
+#define enumclass_PTF_PRINT_TYPE_EXPECTED(exp, val) exp << "[" << +static_cast<std::underlying_type<decltype(val)>::type>(val) << "]"
+
+#define PTF_PRINT_ASSERTION(severity, op) \
+	std::cout << std::left << std::setw(35) << __FUNCTION__ << ": " \
+    << severity \
```
egecetin, line 22: nit: alignment (not checked probably space to tab)

seladb: Good catch! Fixed in: https://github.com/seladb/PcapPlusPlus/pull/1019/commits/558a259db9679488400b52bee0e6322da55256c2

egecetin (approved):

PR #1020, opened 2022-12-24
NULL to nullptr for some files

hpirlo:
I started to port pcap++ to c++11. this is the first commit!

### DIFF
--- a/Examples/DpdkBridge/AppWorkerThread.h
+++ b/Examples/DpdkBridge/AppWorkerThread.h
@@ -68,7 +68,7 @@ public:
 		// free packet array (frees all mbufs as well)
 		for (int i = 0; i < MAX_RECEIVE_BURST; i++)
 		{
-			if (packetArr[i] != NULL)
+			if (packetArr[i] != nullptr)
 				delete packetArr[i];
 		}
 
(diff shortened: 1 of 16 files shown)


seladb:
@hpirlo thank you for opening this PR! I'm currently out and won't be able to review. Maybe @egecetin or @clementperon can review?

clementperon:
Pcap++ is already built with C++11 standards.
I agree it's proper to change null to nullptr but if you do it use clang-tidy and add the check in the CI.
https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-nullptr.html

seladb (approved):
I think it's ok to convert `NULL` to `nullptr` in phases. It's probably too early to add the `clang-tidy` check at this point until we finish converting everything

egecetin:
There are still other `NULL`s on the source code. For example, in `Common++`  

https://github.com/hpirlo/PcapPlusPlus/blob/7eff9470821052f114d7fcee2f0c4c5ebcadaba4/Common%2B%2B/src/GeneralUtils.cpp#L82.

Do you think all the `NULL`s should be converted before merging or can this be accepted as a part of it?

seladb:
@hpirlo a more extensive PR by @clementperon was already merged which converts all of `NULL` occurrences to `nullptr`: https://github.com/seladb/PcapPlusPlus/pull/1028

I think we can close this PR for now.

Thank you so much @hpirlo for contributing to this project 🙏 ❤️ and sorry for the slow response. If you're still interested in contributing we have a lot of open issues waiting for someone to work on them 😃

hpirlo:
@seladb Surely I will contribute to other open issues. Thanks for your support :heart: :pray:

PR #1021, opened 2022-12-25
NTP convert enum to strong typed + add KissODeath codes

egecetin:


seladb (approved):

Tests/Packet++Test/Tests/NtpTests.cpp
```diff
@@ -16,7 +16,6 @@
 
 PTF_TEST_CASE(NtpMethodsTests)
 {
-
```
seladb, line 19: nit: do you have an NTP packet example with Kiss-O-Death? If yes we can add it to the tests, otherwise we can merge the PR as is

egecetin: Unfortunately no, I don't have any.

seladb: ok, we can merge this PR as is 👍

PR #1025, opened 2023-01-04
Fix macos cirrus

clementperon:
Since January 1st 2023 cirrus CI macos have migrated to another instance see:

https://cirrus-ci.org/blog/2022/11/08/sunsetting-intel-macos-instances/

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -70,8 +70,8 @@ task:
 
 task:
   name: Test Zstd MacOS
-  osx_instance:
-    image: monterey-xcode
+  macos_instance:
+    image: ghcr.io/cirruslabs/macos-ventura-xcode:14.1
   configure_script:
     - brew install zstd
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -70,8 +70,8 @@ task:
 
 task:
   name: Test Zstd MacOS
-  osx_instance:
-    image: monterey-xcode
+  macos_instance:
+    image: ghcr.io/cirruslabs/macos-ventura-xcode:14.1
```
seladb, line 74: Since this fails we can probably keep it as it was. We anyway plan to remove the old build system soon

egecetin (approved):

.cirrus.yml
clementperon: The previous task also failed since 1st of January. Maybe we should remove it instead no ?

seladb (approved):

PR #1027, opened 2023-01-04
Codecov Integration for coverage reports

egecetin:
Check uploaded coverage report [here](https://app.codecov.io/gh/seladb/PcapPlusPlus/tree/coverage)
Also check flag (Every OS name/version used as flag) based coverage reports [here](https://app.codecov.io/gh/seladb/PcapPlusPlus/flags)

- Github Actions
   - [x] Linux
   - [x] ~~DPDK~~ (Since cant test DPDK on CI not affects coverage)
   - [x] ~~Pfring~~ (Since cant test PF_RING on CI not affects coverage)
   - [x] MacOS
   - [x] Mingw
   - [x] Visual Studio (19, 22)
   - [x] ~~Android~~ (ToyVPN fails and my knowledge about Android is so limited so disabled for now)
 - Cirrus
   - [x] ~~FreeBSD~~ (Coverage not works for 12.3 but works for 13 but still can't upload to codecov https://github.com/codecov/uploader/issues/849 )
   - [x] Ubuntu Zstd
   - [x] MacOS Zstd
   - [x] ~~Fuzzers~~ (Not affects coverage)
 - Appveyor
   - [x] Visual Studio (15, 17)


egecetin:
@seladb You can also enable automatic codecov bot comments for PRs. I do not remember exactly but just enabling integration from codecov website should be enough. Check docs step 3 [here](https://docs.codecov.com/docs). You can find a screenshot of an example comment [here](https://docs.codecov.com/docs/pull-request-comments)

.cirrus.yml
```diff
@@ -35,8 +35,14 @@ task:
   build_script:
     - cmake --build Dist -j 2
   test_pcapplusplus_script:
+    - python3 -m pip install gcovr
     - python3 -m pip install -r ci/run_tests/requirements.txt
     - python3 ci/run_tests/run_tests.py --interface eth0
+  coverage_report_script:
+    - gcovr -v -r . --gcov-ignore-parse-errors --exclude-throw-branches --filter Common --filter Pcap --filter Packet --xml -o coverage.xml
```
clementperon, line 42: I think there is a GitHub actions for this no ?
https://github.com/codecov/codecov-action

Examples/CMakeLists.txt
```diff
@@ -20,6 +20,11 @@ set(CMAKE_CXX_STANDARD 11)
 # popen()/pclose() are not C++ standards
 set(CMAKE_CXX_EXTENSIONS ON)
 
+if(PCAPPP_BUILD_COVERAGE AND NOT ANDROID)
+  add_compile_options(--coverage)
```
clementperon, line 24: You don't need to add it here no? Except if you build examples out of three

.cirrus.yml
egecetin: You mean for uploading part (L43-L45), right? If yes, it is but I don't know how to use github actions at cirrus. Let me check it.

clementperon: Sorry my comment was for GitHub ci only

Examples/CMakeLists.txt
egecetin: It raised link errors in some CI runs while working on it. I'll check again, might the behavior of the ci/compile has changed due to other PRs.

clementperon: Yes we are building tutorials out of three here, so as libpcap++ has been built with coverage and not linked.

We could either build libraries as shared libraries or test the building of tutorial out-of three in another jobs ?

egecetin: My bet is on shared libraries because I aimed to gather coverage of all ci workflows individually without changing them. I mean, what we test what we miss when a PR run. So, separating tutorials to another job might lead to misinterpretation and also increase the CI work.

README.md
```diff
@@ -7,6 +7,7 @@
 [![Cirrus CI - Base Branch Build Status](https://img.shields.io/cirrus/github/seladb/PcapPlusPlus?label=Cirrus%20CI&logo=cirrusci&style=flat)](https://cirrus-ci.com/github/seladb/PcapPlusPlus)
 [![AppVeyor](https://img.shields.io/appveyor/build/seladb/PcapPlusPlus/master?label=AppVeyor&logo=appveyor&logoColor=white&style=flat)](https://ci.appveyor.com/project/seladb/pcapplusplus/branch/master)
 [![GitHub Workflow Status](https://img.shields.io/github/actions/workflow/status/seladb/PcapPlusPlus/codeql.yml?branch=master&label=CodeQL&logo=github&style=flat)](https://github.com/seladb/PcapPlusPlus/actions?query=workflow%3A%22CodeQL%22)
+[![Codecov](https://img.shields.io/codecov/c/github/seladb/PcapPlusPlus?logo=codecov&logoColor=white)]()
```
seladb, line 10: For some reason the badge shows "unknown". Any idea why?

.github/workflows/build_and_test.yml
```diff
@@ -204,13 +243,26 @@ jobs:
...
+        run: |
+          python3 -m pip install gcovr
+          gcovr -v -r . --gcov-ignore-parse-errors --exclude-throw-branches --filter Common --filter Pcap --filter Packet --xml -o coverage.xml
+
+      - name: Upload Coverage Results
+        uses: codecov/codecov-action@v3
+        with:
+          files: ./coverage.xml
+          flags: ${{ matrix.os-version }},unittest
+          fail_ci_if_error: true
```
seladb, line 263: do we want the CI to fail on coverage upload errors? 🤔

.github/workflows/build_and_test.yml
```diff
@@ -117,6 +130,19 @@ jobs:
           cd Tests/Pcap++Test
           if [ -n "$(grep avx512 /proc/cpuinfo)" ]; then Bin/Pcap++Test -n; else echo AVX-512 SUPPORT NOT FOUND, CANNOT RUN Pcap++Test; fi
 
+      - name: Create Cobertura Report
```
seladb, line 133: The coverage report is a bit misleading for DPDK and PF_RING because we don't run these tests in CI, hence the coverage shows 0%. The reason is they require special hardware / kernel configuration which is not easy to achieve on CI machines. Maybe we can exclude these files from the report?

README.md
egecetin: It looks for master branch. Currently coverage is unknown for master branch. Once the PR merged it should be fixed

.github/workflows/build_and_test.yml
egecetin: Both options are ok for me. It is up to you

seladb: IMO it's should be set to `false` so if codecov is down our CI won't fail

.github/workflows/build_and_test.yml
```diff
@@ -277,6 +330,24 @@ jobs:
...
 
+      - name: Install Coverage Requirements
+        run: python3 -m pip install gcovr
+
+      - name: Process Coverage Files
+        shell: msys2 {0}
+        run: find . -name "*.gcno" -exec gcov -b -l -p -c {} +
+
+      - name: Create Coberture Report
+        run: gcovr -v -g -k -r . --gcov-ignore-parse-errors --exclude-throw-branches --filter Common --filter Pcap --filter Packet --xml -o coverage.xml
```
egecetin, line 341: I don't know why but I tried to use $GCOVR_FLAGS variable but it generated empty report.

egecetin, line 341: For mingw run, I tried to use $GCOVR_FLAGS but generated empty report.

egecetin:
Codecov lost the base commit for this PR so you should use direct [link](https://app.codecov.io/github/seladb/PcapPlusPlus/commit/a94f26d9f204ee0a217c530ccf5a832d392e268c/tree) to latest report.

A minor thing, since Windows coverage generated by `OpenCppCoverage` instead of `gcov` excluding `PcapRemoteDevice` is a bit tricky. It should be excluded from inside `run_tests_windows.py` so we can easily forget. For this reason I'm not sure it should be excluded but I can add if you want.

.cirrus.yml
```diff
@@ -46,29 +53,40 @@ task:
...
+    - ./codecov -v -f coverage.xml -F zstd -F macos-ventura -F unittest || true
 
 task:
   name: Regression Tests Linux (Address Sanitizer)
   container:
     image: gcr.io/oss-fuzz-base/base-clang
   configure_script:
-    - apt-get update && apt-get install -y cmake git libpcap-dev
-    - cmake -S . -B $BUILD_DIR -DPCAPPP_USE_SANITIZER=AddressSanitizer -DPCAPPP_BUILD_FUZZERS=ON -DPCAPPP_BUILD_EXAMPLES=OFF
+    - apt-get update && apt-get install -y cmake git libpcap-dev gcovr curl
```
seladb, line 76: Do we need to run `gcovr` in Address Sanitizer tests? 🤔

.cirrus.yml
```diff
@@ -90,14 +108,20 @@ task:
   container:
     image: gcr.io/oss-fuzz-base/base-clang
   configure_script:
-    - apt-get update && apt-get install -y cmake git libpcap-dev
-    - cmake -S . -B $BUILD_DIR -DPCAPPP_USE_SANITIZER=UndefinedBehaviorSanitizer -DPCAPPP_BUILD_FUZZERS=ON -DPCAPPP_BUILD_EXAMPLES=OFF
+    - apt-get update && apt-get install -y cmake git libpcap-dev gcovr curl
```
seladb, line 111: ditto

.github/workflows/build_and_test.yml
```diff
@@ -121,13 +135,26 @@ jobs:
...
+          verbose: true
+
   pfring:
     runs-on: ubuntu-20.04
     container: seladb/ubuntu2004-pfring
     strategy:
       matrix:
         include:
-          - configure: cmake -DPCAPPP_USE_PF_RING=ON -DPF_RING_ROOT="/PF_RING" -S . -B $BUILD_DIR
+          - configure: cmake -DPCAPPP_USE_PF_RING=ON -DPF_RING_ROOT="/PF_RING" -DPCAPPP_BUILD_COVERAGE=ON -S . -B $BUILD_DIR
```
seladb, line 157: ditto

.github/workflows/build_and_test.yml
```diff
@@ -175,13 +215,13 @@ jobs:
         # Ensure user have access to network devices instead of giving super-user right
         run: |
           sudo chmod a+rw /dev/bpf*
-          cmake -DLIGHT_PCAPNG_ZSTD=${{ matrix.config-zstd }} -DCMAKE_OSX_ARCHITECTURES=${{ matrix.arch }} -S . -B $BUILD_DIR
+          cmake -DLIGHT_PCAPNG_ZSTD=${{ matrix.config-zstd }} -DCMAKE_OSX_ARCHITECTURES=${{ matrix.arch }} -DBUILD_SHARED_LIBS=ON -DPCAPPP_BUILD_COVERAGE=ON -S . -B $BUILD_DIR
```
seladb, line 218: why is it needed to build share libs for coverage to work? 🤔

.github/workflows/build_and_test.yml
seladb: Maybe because this job runs in powershell on Windows, you can try: `$env:GCOVR_FLAGS`

.cirrus.yml
egecetin: It is an option to see where current regression tests touch. Not affect coverage much, just for visualization. It can be removed if you want.

seladb: I don't think these tests cover code that is not already covered in other tests, am I wrong?

.github/workflows/build_and_test.yml
egecetin: It is because of out of build of tutorials (@clementperon mentioned [here](https://github.com/seladb/PcapPlusPlus/pull/1027#discussion_r1087214317)) If it is static linked there is a need of coverage flags inside examples since Tutorials build out of tree

seladb: I read the thread but I'm not sure I understand. Can you please elaborate?

egecetin: @seladb Sorry for late response. I'm busy nowadays. It is because of linker step. While the `Pcap++` libraries are being built because of the coverage flags compiler and linker adds coverage related functions from `llvm` or `gcov` libraries. Since the tutorials stage is after the library builds the out of tree build of tutorials uses the already compiled `Pcap++` binaries. So, during the linker stage the `llvm`/`gcov` functions should be linked (because `Pcap++` binaries require them) but since there is no information about it in CMakeLists of tutorials about coverage flags, linker does not have any idea about where the functions are located, and it raises an error of undefined reference.

So basically, there is a two option, adding coverage flags to tutorials `CMakeLists` and using a boolean to add coverage linker flags or compiling shared libraries. Since shared libraries already linked there is no need to additional change in CMake

.cirrus.yml
egecetin: Not check all of them but yes, they are already covered. I'll remove

.github/workflows/build_and_test.yml
clementperon: Last option is to add a find_dependency like we do for DPDk and PF_RING

.github/workflows/build_and_test.yml
egecetin: Yes, it works (unfortunately partially). `gcovr` thinks they are directory 🙄. Check [this](https://github.com/seladb/PcapPlusPlus/actions/runs/4092523031/jobs/7057319238)

seladb: Powershell is a weird language... try this:

```powershell
gcovr -v -g -k -r . $env:GCOVR_FLAGS.split() -o coverage.xml
```

.github/workflows/build_and_test.yml
seladb: We're building the tutorials on all platforms, it only bothers `llvm/gcov`?

egecetin: Actually, it affects all platforms (except VS, it is a completely different story, OpenCppCoverage is non-intrusive but I don't have any deep knowledge how it works). Since compiler wraps all `Pcap++` function calls with coverage related functions internally during to build  (invisible to us), to generate coverage report on both build-time and runtime (`gcda`, `gcno` files) there is a dependency to coverage libraries for all systems _if the coverage build option is enabled for library build_, otherwise not. I mean, the problem is not related with tutorials itself. It is because how the library was built. If the coverage is enabled `libPcap++` etc. requires these invisible functions and if you want to use them you need to link `llvm/gcov` libraries too.

seladb: got it, I just noticed you also changed the linux job to also build shared libs.

IMHO we shouldn't change our CI to build shared libs just because of coverage. So I guess the only option is to add the coverage flags to the tutorials `CMakeLists`?

@clementperon please let us know what you think

egecetin: But on the other hand, using shared libraries because of coverage also means to checking of shared lib builds 😁

.github/workflows/build_and_test.yml
egecetin: It works, thanks

.github/workflows/build_and_test.yml
seladb: Yes, that's a good point 😄 
Maybe we should have another job to test shared libraries. We don't have to run it on all platforms, maybe pick just one Linux option (Ubuntu?). But that's probably a separate topic to discuss (and a separate PR).

@egecetin do you think you can fix the tutorials `CMakeLists` to work with the coverage flags?
Or maybe @clementperon has a simpler solution to make it work?

README.md
```diff
@@ -7,6 +7,7 @@
 [![Cirrus CI - Base Branch Build Status](https://img.shields.io/cirrus/github/seladb/PcapPlusPlus?label=Cirrus%20CI&logo=cirrusci&style=flat)](https://cirrus-ci.com/github/seladb/PcapPlusPlus)
 [![AppVeyor](https://img.shields.io/appveyor/build/seladb/PcapPlusPlus/master?label=AppVeyor&logo=appveyor&logoColor=white&style=flat)](https://ci.appveyor.com/project/seladb/pcapplusplus/branch/master)
 [![GitHub Workflow Status](https://img.shields.io/github/actions/workflow/status/seladb/PcapPlusPlus/codeql.yml?branch=master&label=CodeQL&logo=github&style=flat)](https://github.com/seladb/PcapPlusPlus/actions?query=workflow%3A%22CodeQL%22)
+[![Codecov](https://img.shields.io/codecov/c/github/seladb/PcapPlusPlus?logo=codecov&logoColor=white)]()
```
seladb, line 10: Don't we want to provide a link to Codecov?

cmake/modules/FindCoverage.cmake
```diff
@@ -0,0 +1,3 @@
+set(Coverage_FOUND TRUE)
```
egecetin, line 1: @clementperon I prepared a simple script. Do you have any advice?

.github/workflows/build_and_test.yml
egecetin: Done. I worked on find_dependency option

seladb:
@clementperon can you review the CMake fixes?

Otherwise - LGTM

seladb (approved):

CMakeLists.txt
```diff
@@ -210,6 +211,12 @@ if(MINGW)
   add_link_options(-static)
 endif()
 
+if(PCAPPP_BUILD_COVERAGE AND NOT ANDROID)
```
clementperon, line 214: Add a comment to explain why the NOT ANDROID

egecetin: Done with [453b322](https://github.com/seladb/PcapPlusPlus/pull/1027/commits/453b322a56b2318a06c21f2909bb923de8204d86)

PR #1028, opened 2023-01-04
Clang tidy: modernize use nullptr instead of NULL

clementperon:
This MR add the clang-tidy support + convert NULL to nullptr.

This superseed https://github.com/seladb/PcapPlusPlus/pull/1020

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,14 @@ if(PCAPPP_ENABLE_PCAP_SET_DIRECTION)
   add_definitions(-DHAS_SET_DIRECTION_ENABLED)
 endif()
 
+option(PCAPPP_ENABLE_CLANG-TIDY "Run Clang-Tidy static analysis during build" OFF)
+
+if (PCAPPP_ENABLE_CLANG-TIDY)
+  find_program(CLANG_TIDY_EXE NAMES "clang-tidy" REQUIRED)
+  set(CLANG_TIDY_COMMAND "${CLANG_TIDY_EXE}" "--fix" "--checks=modernize-use-nullptr")
+  set(CMAKE_CXX_CLANG_TIDY ${CLANG_TIDY_COMMAND})
+endif()
+
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 if(MSVC)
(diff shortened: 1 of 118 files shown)


seladb:
@clementperon the build on centos fails: https://github.com/seladb/PcapPlusPlus/actions/runs/3842342225/jobs/6543558857

And there are a few failures in CirrusCI as well.

In addition - it's hard to review such big PRs, even though the change is pretty straightforward...

clementperon:
@seladb agree that's why I have split into a separate commit and only enabled modernize nullptr. 

This is done automatically by clang-tidy

CMakeLists.txt
```diff
@@ -102,6 +102,14 @@ if(PCAPPP_ENABLE_PCAP_SET_DIRECTION)
   add_definitions(-DHAS_SET_DIRECTION_ENABLED)
 endif()
 
+option(PCAPPP_ENABLE_CLANG-TIDY "Run Clang-Tidy static analysis during build" OFF)
```
seladb, line 105: nit: I'd change it to `PCAPPP_ENABLE_CLANG_TIDY` (with `-` instead of `_`)

CMakeLists.txt
```diff
@@ -102,6 +102,14 @@ if(PCAPPP_ENABLE_PCAP_SET_DIRECTION)
...
+
+if (PCAPPP_ENABLE_CLANG-TIDY)
+  find_program(CLANG_TIDY_EXE NAMES "clang-tidy" REQUIRED)
+  set(CLANG_TIDY_COMMAND "${CLANG_TIDY_EXE}" "--fix" "--checks=modernize-use-nullptr")
```
seladb, line 109: - do we only want to check `nullptr`? Maybe enable all checks?
- Do we always want the `--fix` flag or should it be configurable?

clementperon: For the moment we only check for nullptr yes, I would like toenable the other checks progressively in another MR

seladb: ok sounds reasonable, what about the `--fix` flag?

seladb (approved):
Just fix this small issue: https://github.com/seladb/PcapPlusPlus/pull/1028#discussion_r1064490751

And merge the PR!

PR #1044, opened 2023-01-21
Move `setup_dpdk.py` to CMake directory

seladb:
- Move `setup_dpdk.py` to `cmake/` dir and copy it from there
- Fix old mentions of the `setup-dpdk.sh` to `setup_dpdk.py`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,8 +43,6 @@ mk/PcapPlusPlus.mk-e
 #auto generated files
 install.sh
 uninstall.sh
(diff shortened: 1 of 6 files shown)


clementperon (approved):

clementperon:
This is to simplify the move to cmake and only rm the MK folder right ?

seladb:
> This is to simplify the move to cmake and only rm the MK folder right ?

Yes, so that we can later remove the `mk/` folder altogether

PR #1046, opened 2023-01-21
Avoid using sudo

clementperon:
Instead of giving sudo write, let's try to allow the user access the network interfaces.


ci/legacy/.travis.yml
```diff
@@ -12,6 +12,7 @@ matrix:
 
 before_script:
   - ./configure-mac_os_x.sh
+  - sudo chmod a+rw /dev/bpf*
```
egecetin, line 15: Is there a need for travis? Maybe we should remove, no?

clementperon: @seladb what do you think ?

seladb: I didn't remember we still keep the TravisCI file. Sure, we can remove it

.cirrus.yml
```diff
@@ -7,6 +7,7 @@ task:
   install_script:
     - pkg install -y libpcap bash gsed gmake tcpreplay
   configure_script:
+    - chmod a+rw /dev/bpf*
```
seladb, line 10: @clementperon what's the purpose of allowing the user access to the network interfaces instead of using `sudo`? How is this better than using `sudo`? 🤔

clementperon: We explicitly control that only network should be accessible whereas with sudo you don't know what your python will do / touch.

clementperon: Also this will make it easier, in the future we only need to pass the interface name for running ctest, we don't need to tell to use sudo anymore

seladb: ok sure, but maybe put a comment above every `chmod` so we know we did it for?

I'll definitely forget after 2-3 months... 😃

ci/legacy/.travis.yml
clementperon: done

.cirrus.yml
clementperon: done

seladb (approved):

PR #1047, opened 2023-01-24
Delete the previous build system

seladb:


clementperon (approved):

.github/workflows/build_and_test.yml
```diff
@@ -1,419 +0,0 @@
-name: Build and test
```
egecetin, line 1: Wouldn't it be better to rename the [build_and_test_cmake.yml](https://github.com/seladb/PcapPlusPlus/blob/delete-previous-build-system/.github/workflows/build_and_test_cmake.yml) to build_and_test.yml?

seladb: I thought about it when working on the PR, but since we already have a file called `build_and_test.yml` - deleting it and then renaming this to `build_and_test.yml` will make git think that we're changing the content of `build_and_test.yml` which will make this PR very messy.

I'll rename it in a separate PR once this one is merged

PR #1049, opened 2023-01-25
Rename GitHub Actions workflow

seladb:
Remove "cmake" from the workflow name

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1,4 +1,4 @@
-name: Build and test w/ CMake
+name: Build and test
 on: [push, pull_request]
 
 env:


.github/workflows/build_and_test.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and test w/ CMake
+name: Build and test
 on: [push, pull_request]
```
clementperon, line 2: I think we should only choose one else it will create a double Actions when pushing in a PR that is already open no ?

Push this will trig the action on all push 
Whereas Pull_Request will only trig when the PR is opened

seladb: I'm probably missing something, but here is what I observed:
- When people fork the repo and create a PR the CI is triggered once because of `pull_request`
- When I push `dev` to `master` the CI is triggered only once because of `push`
- When I create a branch in the main repo and open a PR the CI gets triggered twice

I'm pretty sure we can modify this so it'll apply to all scenarios, but I need to research more. When I'll have time I'll try it on a dummy repo with a dummy CI (that runs faster 😄 ) and will open a PR to fix it

clementperon (approved):

PR #1057, opened 2023-01-29
Clang tidy fixes

clementperon:
Some fixes to use override and pass by const reference when we can


Packet++/src/HttpLayer.cpp
```diff
@@ -215,45 +216,47 @@ HttpRequestFirstLine::HttpRequestFirstLine(HttpRequestLayer* httpRequest) : m_Ht
 	}
 }
 
-HttpRequestFirstLine::HttpRequestFirstLine(HttpRequestLayer* httpRequest, HttpRequestLayer::HttpMethod method, HttpVersion version, std::string uri)
-try		// throw(HttpRequestFirstLineException)
+HttpRequestFirstLine::HttpRequestFirstLine(HttpRequestLayer* httpRequest, HttpRequestLayer::HttpMethod method, HttpVersion version, const std::string &uri)
 {
-	if (method == HttpRequestLayer::HttpMethodUnknown)
+	try		// throw(HttpRequestFirstLineException)
```
egecetin, line 221: not important but maybe this comment might be removed?

egecetin (approved):

seladb (commented):
@clementperon I think it's a great idea moving `dirent-for-Visual-Studio` under `PcapSearch`!

But when doing so, we don't need all the files - just `dirent-for-Visual-Studio/include/dirent.h`. We can remove the rest of the files, they are redundent.

clementperon:
> @clementperon I think it's a great idea moving `dirent-for-Visual-Studio` under `PcapSearch`!
> 
> But when doing so, we don't need all the files - just `dirent-for-Visual-Studio/include/dirent.h`. We can remove the rest of the files, they are redundent.

Drop them but I kept the LICENSE file

seladb (approved):

PR #1058, opened 2023-01-30
Enable Clang tidy in CI

clementperon:


.github/workflows/build_and_test.yml
```diff
@@ -49,17 +51,32 @@ jobs:
...
+            clang-tidy: ON
 
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
 
-      - name: Configure PcapPlusPlus
+      - if: ${{ matrix.clang-tidy }}
+        name: Configure PcapPlusPlus w/ Clang-Tidy
+        run: CXX=clang++ CC=clang cmake -DLIGHT_PCAPNG_ZSTD=${{ matrix.config-zstd }} -DPCAPPP_ENABLE_CLANG_TIDY=ON -S . -B $BUILD_DIR
```
seladb, line 65: All we want is to build PcapPlusPlus with `clang-tidy`, no?
In that case, maybe it'd be nicer to add another `job` that just builds PcapPlusPlus without running the tests etc. I think it's a simpler solution than having this `if`... what do you think?

egecetin (approved):

egecetin:
What is the `cppcheck` version inside alpine container?

clementperon:
> What is the `cppcheck` version inside alpine container?

2.9.1 it's apine package that's why they are new warnings. I will fix them :)
https://pkgs.alpinelinux.org/packages?name=cppcheck&branch=v3.17&repo=&arch=&maintainer=

.github/workflows/build_and_test.yml
```diff
@@ -1,28 +1,41 @@
...
-    branches: [ "dev" ]
+    branches: ["dev"]
 
 env:
   BUILD_DIR: Dist
 
 jobs:
   pre-commit:
     runs-on: ubuntu-latest
+    container: seladb/alpine317
```
seladb, line 14: Why is it needed to run `pre-commit` in a container? 🤔

.github/workflows/build_and_test.yml
```diff
@@ -1,28 +1,41 @@
...
+          ./ci/cmake-format-all.sh
+          git diff --exit-code
+
+      - name: Configure PcapPlusPlus for Static analysis
+        run: CXX=clang++ CC=clang cmake -DLIGHT_PCAPNG_ZSTD=ON -DPCAPPP_ENABLE_CLANG_TIDY=ON -S . -B $BUILD_DIR
+
+      - name: Build PcapPlusPlus and check any diff
+        run: |
+          cmake --build $BUILD_DIR -j
+          git diff --exit-code
```
seladb, line 38: I'm not sure this job should configure and build PcapPlusPlus, we have other jobs that do that. I think this job should be focused on running pre-commit hooks

.github/workflows/build_and_test.yml
clementperon: we need cmake, clang, clang-extra-tools

seladb: ah ok, that's because we enable `clang-tidy`

seladb (approved):

PR #1060, opened 2023-01-31
workflow: test building out of three for DPDK and PF_RING

clementperon:


.github/workflows/build_and_test.yml
```diff
@@ -121,6 +120,17 @@ jobs:
...
+      - name: Check installation
+        run: cmake --install $BUILD_DIR
+
+      - name: Build Tutorials
+        run: |
+          cmake -DPCAPPP_BUILD_TUTORIALS=ON -S Examples -B build_examples
+          cmake --build build_examples -j
+
+      - name: Test Tutorials
+        run: cd build_examples/tutorials_bin && ./Tutorial-HelloWorld
```
seladb, line 132: We probably don't need that in DPDK CI because `Tutorial-HelloWorld` doesn't use DPDK

.github/workflows/build_and_test.yml
```diff
@@ -149,6 +159,17 @@ jobs:
...
+      - name: Check installation
+        run: cmake --install $BUILD_DIR
+
+      - name: Build Tutorials
+        run: |
+          cmake -DPCAPPP_BUILD_TUTORIALS=ON -DPF_RING_ROOT="/PF_RING" -S Examples -B build_examples
+          cmake --build build_examples -j
+
+      - name: Test Tutorials
+        run: cd build_examples/tutorials_bin && ./Tutorial-HelloWorld
```
seladb, line 171: ditto for PF_RING

.github/workflows/build_and_test.yml
clementperon: yes but even if we don't use it, Pcap++ is built with and linking will fail if the lib is not found

That's what I fix in the commit "Fix compilation with PF_RING"

seladb (approved):

Issue #1012, opened 2022-12-12
Padding question


ogozman:
Hello, I seem to have issue with the location of packet trailer layer in payload. When I get UDPLayer, the actual data in payload that goes after header is offset by the number of bytes that is always equal to the length of packet trailer layer that is present in both datagram types (there are 2 types of PDUs and both have the same offset). The question is, is it a bug or is it a consistent behavior that is supposed to happen? I thought that padding should go after the data, not between payload and header. I am using IPv4 reassembly in this case. I have no issue with short PDU where no reassembly is needed. Thank you!

seladb:
Hi @ogozman ! I'm not sure I understand the issue. Can you provide 2 pcap files that reproduce the issue?
- One with the fragments before running IP reassembly
- Another with the output of IP reassembly and point to where the problem is?

ogozman:
Thank you for the fast response, @seladb. I will make a setup and upload the pcap file soon.

ogozman:
Hey, as discussed, here is a zip file with 2 pcap files, one is defragmented packet (ipv4reassembly) and the other one is fragmented packet and (if I understand correctly) packet reassembled by OS. I have actually a question in regards to my assumption, does PCAP++ also capture reassembled packets? Currently I am gathering IPv4 packets and reassembling myself, but the sender actually transmits UDP packets. Can I just capture those and avoid reassembly? 
If you compare OS reassembled package and my reassembled package, you can clearly see a bunch of 0s (14 to be exact) that are only in manually reassembled package.

seladb:
Thanks for sharing your pcap files @ogozman , I'll take a look and get back to you

seladb:
I think I found the issue. Packet no. 8 in your file, which is IP fragment no. 1 has a packet trailer of 12 bytes. The IP reassembly logic uses [`getLayerPayloadSize()`](https://github.com/seladb/PcapPlusPlus/blob/f6baff3c0eac1b66cc96861ed2be3e6a9c43f8b8/Packet%2B%2B/header/Layer.h#L112) to calculate the IP layer payload size which is actually the total size of the IP payload minus the IP header size:

https://github.com/seladb/PcapPlusPlus/blob/f6baff3c0eac1b66cc96861ed2be3e6a9c43f8b8/Packet%2B%2B/src/IPReassembly.cpp#L129

[...]

PR #1066, opened 2023-02-03
Bugfix #1012 - Fix IP defragmentation to ignore packet trailer in the first fragment

seladb:


Packet++/src/IPReassembly.cpp
```diff
@@ -342,7 +342,15 @@ Packet* IPReassembly::processPacket(Packet* fragment, ReassemblyStatus& status,
 			PCPP_LOG_DEBUG("[FragID=0x" << std::hex << fragWrapper->getFragmentId() << "] Got first fragment, allocating RawPacket");
 
 			// create the reassembled packet and copy the fragment data to it
-			fragData->data = new RawPacket(*(fragment->getRawPacket()));
+
+			// copy only data from the beginning of the fragment to the end of IP layer payload.
+			// Don't copy data beyond it such as packet trailer
+			auto fragmentRawPacket = fragment->getRawPacket();
+			auto rawDataLen = fragWrapper->getIPLayerPayload() - fragmentRawPacket ->getRawData() + fragWrapper->getIPLayerPayloadSize();
```
clementperon, line 349: Can't we create a getFragmentLength() ?

clementperon (approved):

Packet++/src/IPReassembly.cpp
seladb: The `fragment` variable here is actually an instance of type `Packet` which is the first IP fragment. Since it's a packet it doesn't make sense to add a method like `getFragmentLength()`

PR #1074, opened 2023-02-18
Package and push artifact

clementperon:


CMakeLists.txt
```diff
@@ -239,6 +240,12 @@ if(PCAPPP_INSTALL)
 
   if(APPLE)
     set(CPACK_SYSTEM_NAME "Darwin-${CMAKE_SYSTEM_PROCESSOR}")
+  elseif(LINUX)
+    # Easier for Packing but requires CMake 3.22
```
clementperon, line 244: @seladb what do you think about requirering a recent cmake version for packaging? 
I can make this limited to Package only.
So user can still build and install with old version of cmake

seladb: I'm not sure I understand the difference...

Yes, I'd like people to be able to build and install PcapPlusPlus with the oldest version of CMake we can allow

clementperon: This mean that I case we want to package we have to add -DPCAPP_ENABLE_PACKAGE=ON

clementperon: see: https://github.com/seladb/PcapPlusPlus/pull/1074/commits/5251c0c3c3ce91dea340db4ee2f8740973a9bf81

seladb: So can we allow an older CMake version for packaging and a newer version for building? 🤔

clementperon: older for building and installing, newer for building, installing and packaging

seladb: ok sounds good!

seladb:
I looked at the current artifacts, and here are a few comments:
1. FreeBSD package is missing because it only runs in GitHub Actions (FreeBSD is only available in CirrusCI
2. Android package is missing. It's a unique package because it doesn't contain the examples, but it does contain the libraries for the different android archs, as you can see here: https://github.com/seladb/PcapPlusPlus/releases/download/v22.11/pcapplusplus-22.11-android.tar.gz
3. A source package is missing (like [this](https://github.com/seladb/PcapPlusPlus/releases/download/v22.11/pcapplusplus-22.11-source-linux.tar.gz)). This package is essential because it includes scripts to build for DPDK and PF_RING: https://github.com/seladb/PcapPlusPlus-Deploy/tree/master/source-linux
4. macOS - what do the versions `20.6.0` and `21.6.0` mean? In previous releases we mention the macOS version and the GCC version, here is an example: https://github.com/seladb/PcapPlusPlus/releases/download/v22.11/pcapplusplus-22.11-macos-10.15.7-xcode-12.2-gcc-12.0.0.tar.gz
5. MSVC package:
    - I see 2 packages there. Are they VS 2019, VS 2022 or both?
    - Which configuration is being built? The MSVC package currently includes 4 configurations: `86/debug`, `x86/release`, `x64/debug`, `x64/release`. Here are the scripts that build them: https://github.com/seladb/PcapPlusPlus-Deploy/tree/master/vs
6. The MinGW build should be called `mingw-s64` instead of `GNU`, here is an example: https://github.com/seladb/PcapPlusPlus/releases/download/v22.11/pcapplusplus-22.11-windows-mingw-w64-gcc-8.1.0.zip
8. Packages names include `GNU` - maybe we want it to be `gcc`?
9. The packages should not include the standard `README.md` because it doesn't give a lot of value to users. Instead we create a `README.release` file that contains information such as the package content, how to use the libraries, release notes and more. Thus, each platform has a slightly different `README.release`. Here are the different versions: https://github.com/seladb/PcapPlusPlus-Deploy/tree/master/READMEs, and here is an example of how we build it for android (each platform has it in its deploy script): https://github.com/seladb/PcapPlusPlus-Deploy/blob/master/android/deploy-android.sh#L19
10. As we discussed, we usually include an `example-app` that has the "hello world" tutorial cpp + pcap file + makefile (which we should probably replace with a CMakeLists file)

clementperon:
> I looked at the current artifacts, and here are a few comments:
> 
> 1. FreeBSD package is missing because it only runs in GitHub Actions (FreeBSD is only available in CirrusCI

I will add FreeBSD running inside a VM to create this package is it ok?

> 2. Android package is missing. It's a unique package because it doesn't contain the examples, but it does contain the libraries for the different android archs, as you can see here: 
https://github.com/seladb/PcapPlusPlus/releases/download/v22.11/pcapplusplus-22.11-android.tar.gz
> 3. A source package is missing (like [this](https://github.com/seladb/PcapPlusPlus/releases/download/v22.11/pcapplusplus-22.11-source-linux.tar.gz)). This package is essential because it includes scripts to build for DPDK and PF_RING: https://github.com/seladb/PcapPlusPlus-Deploy/tree/master/source-linux

When creating a release, github automatically generate a source archive is it not enough ?

> 4. macOS - what do the versions `20.6.0` and `21.6.0` mean? In previous releases we mention the macOS version and the GCC version, here is an example: https://github.com/seladb/PcapPlusPlus/releases/download/v22.11/pcapplusplus-22.11-macos-10.15.7-xcode-12.2-gcc-12.0.0.tar.gz

I have fixed it

> 5. MSVC package:
>    
>    * I see 2 packages there. Are they VS 2019, VS 2022 or both?
>    * Which configuration is being built? The MSVC package currently includes 4 configurations: `86/debug`, `x86/release`, `x64/debug`, `x64/release`. Here are the scripts that build them: https://github.com/seladb/PcapPlusPlus-Deploy/tree/master/vs

There is lot of different configuration possible debug, non-debug, ZSTD, Npcap, Winpcap, etc..
I have moved to a separate pipeline, and we can configure which package we want.

> 6. The MinGW build should be called `mingw-s64` instead of `GNU`, here is an example: https://github.com/seladb/PcapPlusPlus/releases/download/v22.11/pcapplusplus-22.11-windows-mingw-w64-gcc-8.1.0.zip
> 7. Packages names include `GNU` - maybe we want it to be `gcc`?
Fixed

> 8. The packages should not include the standard `README.md` because it doesn't give a lot of value to users. Instead we create a `README.release` file that contains information such as the package content, how to use the libraries, release notes and more. Thus, each platform has a slightly different `README.release`. Here are the different versions: https://github.com/seladb/PcapPlusPlus-Deploy/tree/master/READMEs, and here is an example of how we build it for android (each platform has it in its deploy script): https://github.com/seladb/PcapPlusPlus-Deploy/blob/master/android/deploy-android.sh#L19
> 9. As we discussed, we usually include an `example-app` that has the "hello world" tutorial cpp + pcap file + makefile (which we should probably replace with a CMakeLists file)

CMakeLists.txt
seladb, line 1: One thing that is not clear to me is how to install a package and then use it in an app. For example, let's say I download this package: https://github.com/clementperon/PcapPlusPlus/releases/download/99.10.0/PcapPlusPlus-22.11+-ubuntu-22.04-gcc-11.2.0-x86_64.tar.gz

I see this file structure:
```text
|- lib
  |- cmake
    |- pcapplusplus
      |- modules
        |- ... a lot of modules...
      |- ... a bunch of cmake files...
  |- libCommon++.a
  |- libPacket++.a
  |- libPcap++.a
|- include
  |- pcapplusplus
    |- ... all .h files...
```

Few questions:
1. How do I install this package on my Ubuntu machine?
2. Let's say I want to build an app that uses PcapPlusPlus and I already have `main.cpp`. What's the minimal `CMakeLists.txt` I need to use in order to link my project with the PcapPlusPlus package installed on my machine?

.github/workflows/package.yml
```diff
@@ -0,0 +1,300 @@
+name: Package and release
+
+on:
+  push:
+    tags:
+      - '*'
```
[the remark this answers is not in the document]
clementperon: The workflow needs a tag which will be the release to know where to push the artifacts

CMakeLists.txt
clementperon: .tar.gz are not meant to be installed.
They should be extracted near the customer project and manually link/import.

For installation we should package a .deb / .rpm / .exe

seladb: in the previous deployment we provided scripts to install the downloaded package, which basically means copying the relevant files to `/usr/local/bin`, `/usr/local/lib`, `/usr/local/include`, etc. 
Here is an example: https://github.com/seladb/PcapPlusPlus-Deploy/blob/master/source-linux/install.sh

I think we should provide something similar in the new packaging otherwise users need to do it manually...

clementperon: It would be easier and cleaner to provide .deb / .rpm / .pkg regarding the platform.

CMakeLists.txt
```diff
@@ -285,3 +281,58 @@ if(PCAPPP_INSTALL)
...
+  set(CPACK_RESOURCE_FILE_README "${CMAKE_SOURCE_DIR}/README.md")
+
+  # used in CPackAdditionalInstall.cmake
+  set(CPACK_ADDITIONAL_INSTALL_RESOURCES "${CPACK_RESOURCE_FILE_LICENSE}" "${CPACK_RESOURCE_FILE_README}")
+  set(CPACK_ADDITIONAL_EXAMPLE_APP
+      "${CMAKE_SOURCE_DIR}/Examples/Tutorials/Tutorial-HelloWorld/1_packet.pcap"
+      "${CMAKE_SOURCE_DIR}/Examples/Tutorials/Tutorial-HelloWorld/CMakeLists.txt"
+      "${CMAKE_SOURCE_DIR}/Examples/Tutorials/Tutorial-HelloWorld/main.cpp"
+      "${CMAKE_SOURCE_DIR}/Examples/Tutorials/Tutorial-HelloWorld/Makefile.non_windows"
+      "${CMAKE_SOURCE_DIR}/Examples/Tutorials/Tutorial-HelloWorld/Makefile.windows"
```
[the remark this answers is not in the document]
clementperon: agree removed

CMakeLists.txt
```diff
@@ -285,3 +281,58 @@ if(PCAPPP_INSTALL)
...
+    set(CPACK_SYSTEM_NAME
+        "${DISTRO_ID}-${DISTRO_VERSION_ID}-${PCAPPP_TARGET_COMPILER}-${CMAKE_CXX_COMPILER_VERSION}-${CMAKE_SYSTEM_PROCESSOR}"
+    )
+  elseif(WIN32)
+    set(CPACK_GENERATOR "ZIP")
+    set(CPACK_SYSTEM_NAME "windows-${PCAPPP_TARGET_COMPILER}-${CMAKE_VS_PLATFORM_NAME}-${CMAKE_CXX_COMPILER_VERSION}")
+  endif()
+
+  include(CPack)
+endif()
```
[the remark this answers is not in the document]
clementperon: This is just reshipping the same source with a script that configure the build system to build DPDK.

I think enabling DPDK with CMake is not a big deal IMO

.github/workflows/package.yml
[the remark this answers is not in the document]
seladb: We anyway need the option to choose a tag for a release, no? The current [version numbers](https://github.com/clementperon/PcapPlusPlus/releases) are just for tests... 
Maybe this can help? 
https://github.com/community/community/discussions/45871

CMakeLists.txt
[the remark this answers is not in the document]
seladb: Yes that's a good point. We can probably remove this

CMakeLists.txt
seladb: I guess we can provide `.deb`/`.rpm`/`.pkg` also.

But let's say I download a specific package for Linux/macOS/FreeBSD from [Releases](https://github.com/clementperon/PcapPlusPlus/releases) - how do I use it?

Many people want to be able to "install" the package, meaning have everything copied to `/usr/local/...`.
If not - at least get an example  of `CMakeLists.txt` they can use to compile their app with PcapPlusPlus lib/header files

Ideally the `example-app` should contain a `CMakeLists.txt`

.github/workflows/package.yml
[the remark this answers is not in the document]
clementperon: When you push the tag this workflow is then triggered, create an associated release to the tag  and push the artifacts.

The release will have the tag name that you can edit after

CMakeLists.txt
clementperon: You should be able to configure your app with -DPCAPPLUSPLUS_ROOT=/home/downloads/my_folder/ 
I will add this in the readme example

.github/workflows/package.yml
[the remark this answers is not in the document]
seladb: thanks for clarifying, I think I got it now

.github/workflows/package.yml
```diff
@@ -0,0 +1,293 @@
...
+          artifacts: ${{ env.BUILD_DIR }}/*.tar.gz
+
+  freebsd:
+    runs-on: macos-12
+    strategy:
+      matrix:
+        include:
+          - freebsd-version: "12.3"
+          - freebsd-version: "13.0"
+          - freebsd-version: "13.1"
```
seladb, line 69: Do you think we need both 13.0 and 13.1? Shouldn't we choose just one of them?

clementperon: I don't know the release scheduled of FreeBSD, one artifact is ~9MB and this job is run only once per tag.

So up to you, we can keep only 13.1

CMakeLists.txt
seladb: Is this standard for other open-source libraries?
Meaning - download the package, un-tar it, and use it without installing it first?

clementperon: if you look for example a libzstd which is IMO a very famous library, you will get a zip for Win32 / Win64 which the static and dll library. Which is a good example of what we should achieve.

CMakeLists.txt
```diff
@@ -285,3 +281,58 @@ if(PCAPPP_INSTALL)
...
+      "PcapPlusPlus is a multiplatform C++ library for capturing, parsing and crafting of network packets. It is designed to be efficient, powerful and easy to use."
+  )
+  set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_SOURCE_DIR}/LICENSE")
+  set(CPACK_RESOURCE_FILE_README "${CMAKE_SOURCE_DIR}/README.md")
+
+  # used in CPackAdditionalInstall.cmake
+  set(CPACK_ADDITIONAL_INSTALL_RESOURCES "${CPACK_RESOURCE_FILE_LICENSE}" "${CPACK_RESOURCE_FILE_README}")
+  set(CPACK_ADDITIONAL_EXAMPLE_APP
+      "${CMAKE_SOURCE_DIR}/Examples/Tutorials/Tutorial-HelloWorld/1_packet.pcap"
+      "${CMAKE_SOURCE_DIR}/Examples/Tutorials/Tutorial-HelloWorld/CMakeLists.txt"
```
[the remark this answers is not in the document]
clementperon: Maybe we should remove the ability to compile the example inside the project and should require to install the library no ?

CMakeLists.txt
```diff
@@ -285,3 +281,58 @@ if(PCAPPP_INSTALL)
...
+  set(CPACK_GENERATOR "TGZ")
+
+  if(APPLE)
+    set(CPACK_SYSTEM_NAME
+        "macos-${CURRENT_OSX_VERSION}-${CMAKE_OSX_ARCHITECTURES}-${PCAPPP_TARGET_COMPILER}-${CMAKE_CXX_COMPILER_VERSION}"
+    )
+  elseif(LINUX)
+    # Easier for Packing but requires CMake 3.22
+    cmake_minimum_required(VERSION 3.22)
+    cmake_host_system_information(RESULT DISTRO_ID QUERY DISTRIB_ID)
```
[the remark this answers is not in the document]
clementperon: resolved by 4e9d898d82254be92b4e8e693fc0adfc64e4c842

CMakeLists.txt
```diff
@@ -285,3 +281,58 @@ if(PCAPPP_INSTALL)
...
+    cmake_host_system_information(RESULT DISTRO_VERSION_ID QUERY DISTRIB_VERSION_ID)
+    set(CPACK_SYSTEM_NAME
+        "${DISTRO_ID}-${DISTRO_VERSION_ID}-${PCAPPP_TARGET_COMPILER}-${CMAKE_CXX_COMPILER_VERSION}-${CMAKE_SYSTEM_PROCESSOR}"
+    )
```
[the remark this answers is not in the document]
clementperon: That's to keep coherency Android and MacOS display this information and we could  generate arm64 for Linux and Windows later.

.github/workflows/build_and_test.yml
```diff
@@ -99,7 +99,10 @@ jobs:
           ${{ matrix.python }} -m pytest --interface eth0 --root-path=../../Dist/examples_bin
 
       - name: Check installation
-        run: cmake --install $BUILD_DIR
+        run: |
+          cmake -DPCAPPP_BUILD_COVERAGE=OFF -S . -B "$BUILD_DIR"
+          cmake --build "$BUILD_DIR" -j
```
seladb, line 104: Why do we need to rebuild everything in this step (same in the other jobs)?

cmake/CPackAdditionalInstall.cmake
```diff
@@ -0,0 +1,2 @@
+file(INSTALL ${CPACK_ADDITIONAL_INSTALL_RESOURCES} DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+file(INSTALL ${CPACK_ADDITIONAL_EXAMPLE_APP} DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/example-app")
```
seladb, line 2: Why is this needed?

clementperon: This is the path of the example-app to be shipped in the CPACK package

seladb (approved):

PR #1080, opened 2023-02-23
Fix #1070 - Add missing SIP status codes

### LINKED ISSUE #1070: SIP 608
The 608 sip code is not parsed. 

https://en.wikipedia.org/wiki/List_of_SIP_response_codes

This code was added recently because of [STIR/SHAKEN](https://en.wikipedia.org/wiki/STIR/SHAKEN)

The library is not treating it and you receive SipStatusCodeUnknown

Silviu

seladb:
Addresses #1070 

Add 3 missing SIP status codes:
- `425 Bad Alert Message`
- `555 Push Notification Service Not Supported`
- `608 Rejected`

https://en.wikipedia.org/wiki/List_of_SIP_response_codes


seladb:
@egecetin CodeCov fails the CI because of 0.04% coverage drop... do you know how to adjust this number?

egecetin:
Check #1081. To be honest I haven't used threshold before but it should fix the issue

egecetin (approved):

PR #1082, opened 2023-02-24
Fix compiler warnings

egecetin:


Packet++/src/DnsResourceData.cpp
```diff
@@ -34,7 +34,7 @@ void IDnsResourceData::encodeName(const std::string& decodedName, char* result,
 }
 
 
-StringDnsResourceData::StringDnsResourceData(const uint8_t* dataPtr, size_t dataLen, IDnsResource* dnsResource)
+StringDnsResourceData::StringDnsResourceData(const uint8_t* dataPtr, size_t, IDnsResource* dnsResource)
 {
 	char tempResult[256];
 	decodeName((const char*)dataPtr, tempResult, dnsResource);
```
clementperon, line 40: I think decodeName should have the length of the buffer in param

Packet++/src/DnsResourceData.cpp
```diff
@@ -96,7 +96,7 @@ bool IPv6DnsResourceData::toByteArr(uint8_t* arr, size_t& arrLength, IDnsResourc
 	return true;
 }
 
-MxDnsResourceData::MxDnsResourceData(uint8_t* dataPtr, size_t dataLen, IDnsResource* dnsResource)
+MxDnsResourceData::MxDnsResourceData(uint8_t* dataPtr, size_t, IDnsResource* dnsResource)
```
clementperon, line 99: Ditto I think that we should check that the preference is not ouf of bound and then pass dataLen - sizeof(preference) to decodename()

Packet++/src/SipLayer.cpp
```diff
@@ -846,7 +846,7 @@ void SipResponseFirstLine::setVersion(const std::string& newVersion)
 
 SipResponseLayer::SipResponseStatusCode SipResponseFirstLine::validateStatusCode(char* data, size_t dataLen, SipResponseLayer::SipResponseStatusCode potentialCode)
 {
-	if (data[0] != ' ')
+	if (data && dataLen && data[0] != ' ')
```
clementperon, line 849: dataLen > 0 would be more explicit than implicit convertion to a bool

Packet++/src/DnsResourceData.cpp
seladb: `decodeName` has its own checks to make sure data is not out of bounds:
https://github.com/seladb/PcapPlusPlus/blob/5e262be0ec099f2b845eed64a5a5d8d28baf5481/Packet%2B%2B/src/DnsResource.cpp#L59

This c'tor is mostly used internally (although it's public), but maybe we can check if `dataPtr != nullptr` and `dataLen > 0`?

Packet++/src/DnsResourceData.cpp
```diff
@@ -164,7 +164,7 @@ GenericDnsResourceData::GenericDnsResourceData(const std::string& dataAsHexStrin
 	}
 }
 
-GenericDnsResourceData::GenericDnsResourceData(const GenericDnsResourceData& other)
+GenericDnsResourceData::GenericDnsResourceData(const GenericDnsResourceData& other) : IDnsResourceData()
```
seladb, line 167: why is the compiler complaining here? 🤔

Packet++/src/SomeIpSdLayer.cpp
```diff
@@ -711,7 +711,7 @@ bool SomeIpSdLayer::addOptionIndex(uint32_t indexEntry, uint32_t indexOffset)
 		return true;
 	}
 
-	if (indexFirstOption + lenFirstOption + 1 == indexOffset)
+	if (static_cast<uint8_t>(indexFirstOption + lenFirstOption + 1) == indexOffset)
```
seladb, line 714: `indexOffset` is `uint32_t`, so maybe it's better to cast to `uint32_t`?

It's also safer because we add `uint8_t` numbers which may overflow the size of `uint8_t` but will never overflow `uint32_t`

Packet++/src/SomeIpSdLayer.cpp
```diff
@@ -727,7 +727,7 @@ bool SomeIpSdLayer::addOptionIndex(uint32_t indexEntry, uint32_t indexOffset)
 		return true;
 	}
 
-	if (indexSecondOption + lenSecondOption + 1 == indexOffset)
+	if (static_cast<uint8_t>(indexSecondOption + lenSecondOption + 1) == indexOffset)
```
seladb, line 730: ditto

Packet++/src/DnsResourceData.cpp
egecetin: This warning is a part of Wextra flag. `Base class ‘class pcpp::IDnsResourceData’ should be explicitly initialized in the copy constructor` I disabled extra flag after fix because test suite complains about lots of unused variable and prints lots of warnings.

Packet++/src/DnsResourceData.cpp
clementperon: Yes but I still think it's more proper to always pass a buffer with it's length and not rely on class member.
Another possibility would be to change the pointer to an offset of m_Data.

clementperon (approved):

CMakeLists.txt
```diff
@@ -177,6 +177,8 @@ if(PCAPPP_USE_DPDK)
     message(FATAL_ERROR "Specify DPDK Library with -DDPDK_ROOT=<DPDK_PATH>")
   endif()
   add_definitions(-DUSE_DPDK)
+  # Disable deprecated warnings when DPDK enabled since warnings are treated as errors
+  add_definitions("-Wno-deprecated-declarations")
```
clementperon, line 181: another possibility is to add a pragma when we include the DPDK header

Packet++/src/DnsResourceData.cpp
seladb: Yes I agree, however I think we can do it in a separate PR. Decoding DNS messages is quite tricky and has a lot of edge cases, so it took me some time to get it right. `decodeName()` is a big part of it so any change in this method should be done and tested carefully

Packet++/src/DnsResourceData.cpp
seladb: sounds good 👍

seladb (approved):

Packet++/src/DnsResourceData.cpp
clementperon: Agree

PR #1083, opened 2023-02-27
workflow: add freebsd

clementperon:


.github/workflows/build_and_test.yml
```diff
@@ -168,6 +168,32 @@ jobs:
       - name: Test Tutorials
         run: cd build_examples/tutorials_bin && ./Tutorial-HelloWorld
 
+  freebsd:
```
seladb, line 171: Why do we need FreeBSD in GitHub Actions? We already run it in CirrusCI...

.github/workflows/build_and_test.yml
```diff
@@ -168,6 +168,32 @@ jobs:
...
+            cmake -S . -B Dist
+            cmake --build Dist -j 4
+            ifconfig em0 promisc
+            python3.9 -m ensurepip
+            python3.9 -m pip install -r ci/run_tests/requirements.txt
+            python3.9 ci/run_tests/run_tests.py --interface em0
+            cd Tests/ExamplesTest
+            python3.9 -m ensurepip
+            python3.9 -m pip install -r requirements.txt
+            python3.9 -m pytest --interface em0 --root-path=../../Dist/examples_bin
```
seladb, line 195: Is there any way to separate into different steps? 🤔

.github/workflows/codeql.yml
```diff
@@ -1,11 +1,10 @@
...
 on:
-  push:
-    branches: [ "dev" ]
-  workflow_dispatch:
-  schedule:
-    - cron: '15 14 * * 1'
+  workflow_run:
+    workflows: ["Build and test"]
+    types:
+      - completed
```
seladb, line 7: Why is this change?

.github/workflows/build_and_test.yml
clementperon: I plan to use it allow for release when we need to pack and upload an artifact

.github/workflows/codeql.yml
clementperon: Codeql should be trig when the first workflow is terminated, we need all the report to be upload.

.github/workflows/build_and_test.yml
clementperon: Yes we could but it will requires to respawn the VM each time.
Will check

.github/workflows/codeql.yml
clementperon: How, I think I mixed between codecov and codeql

.github/workflows/build_and_test.yml
seladb: If this is just for release, maybe we don't need to run all the tests?

My main concern is that it runs on a macOS machine and the limit is much lower compares to Linux machines:
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

So if the build runs for a long time it means other macOS builds need to wait in the queue and the overall build time goes up

clementperon: Yes this should be only used for packaging, I will move this to a separate pipeline without any test

PR #1087, opened 2023-03-02
Add detection of possible encrypted handshake message

### LINKED ISSUE #1085: Incorrect Parsing of Encrypted Handshake Message
Hello!

in some cases, Encrypted Handshake Messages are parsed incorrectly, when the first encrypted bytes of the message match an existing handshake type. In the attached pcap file in [bug2.zip](https://github.com/seladb/PcapPlusPlus/files/10862561/bug2.zip) you can see that the encrypted handshake message begins with leading zeroes. Running the following code snippet shows us that these zeroes are interpreted as two Hello Request messages (handshake type 0).
```c++
#include <PcapFileDevice.h>
#include <Packet.h>
#include <SSLLayer.h>
#include <SSLHandshake.h>
#include <iostream>

int main() {

    pcpp::PcapFileReaderDevice pcapReader("bug2.pcap");
    pcapReader.open();

    pcpp::RawPacket rawPacket;
    while(pcapReader.getNextPacket(rawPacket)){
        pcpp::Packet parsedPacket(&rawPacket);

        // get the 2nd ssl layer of the packet (encrypted handshake message)
        pcpp::SSLLayer *sslLayer = dynamic_cast<pcpp::SSLLayer*>(parsedPacket.getLayerOfType(pcpp::SSL, 1));
        pcpp::SSLHandshakeLayer *handshakeLayer = dynamic_cast<pcpp::SSLHandshakeLayer *>(sslLayer);
        uint64_t numHandshakeMessages = handshakeLayer->getHandshakeMessagesCount();

        std::cout << "We have " << numHandshakeMessages << " handshake messages:" << std::endl;
        for (uint64_t i = 0; i < numHandshakeMessages; ++i) {
            pcpp::SSLHandshakeMessage *handshakeMessage = dynamic_cast<pcpp::SSLHandshakeLayer *>(sslLayer)->getHandshakeMessageAt(i);
            std::cout << i <<". " << handshakeMessage->toString();
            std::cout << ", MessageLength: " << handshakeMessage->getMessageLength() << std::endl;
        }
    }
}
```
The incorrect interpretation of these zeroes in `SSLHandshakeMessage::createHandshakeMessage` and the following updates of indices w.r.t. `getMessageLength()` in the constructor of `SSLHandshakeLayer` lead to more falsely recognized messages.
All in all, the code snippet produces the following output:
```
We have 3 handshake messages:
0. Hello Request message, MessageLength: 4
1. Hello Request message, MessageLength: 4
2. Certificate message, MessageLength: 32
```
The application data record gets parsed correctly.

I couldn't figure out a a quick way to solve this issue, it might be difficult.

axmahr:
Resolves #1085 

To detect an encrypted handshake message, the following heuristic is used:
- handshake layer size is at least 16 bytes and
- the first 5 bytes of the handshake message are zeroes or the parsed message length is over 64K
- or the parsed message type makes no sense


axmahr:
Apparently, the test case ```SSLMultipleRecordParsingTest```fails which uses SSL-MultipleRecords1.pcap

Looking at the test case and the pcap file, I come to the conclusion that the test treats the handshake message in the third record layer as a Hello Request although Wireshark tells me this is an encrypted handshake message...

Packet++/src/SSLHandshake.cpp
```diff
@@ -1228,6 +1228,19 @@ SSLHandshakeMessage* SSLHandshakeMessage::createHandshakeMessage(uint8_t* data,
 		return nullptr;
 
 	ssl_tls_handshake_layer* hsMsgHeader = (ssl_tls_handshake_layer*)data;
+
+	if (dataLen >= 16)
+	{
+		if (*(uint64_t*)data <= 0xFFFFFF || hsMsgHeader->length1 >= 1)
```
seladb, line 1234: Are you sure this is correct `*(uint64_t*)data <= 0xFFFFFF`? 🤔 

Don't we want to take the 5 first bytes of the message data (that comes after the message header), e.g `data + sizeof(ssl_tls_handshake_layer)` and check if all of them are equal to 0?

seladb, line 1234: We can potentially merge these 2 if statements into one if statement

Packet++/src/SSLHandshake.cpp
```diff
@@ -1228,6 +1228,19 @@ SSLHandshakeMessage* SSLHandshakeMessage::createHandshakeMessage(uint8_t* data,
...
+			// possibly Encrypted Handshake Message
+			// used heuristic:
+			// - first 5 bytes of the handshake message are zeroes
+			// - or wrong message length is over 64K
+			// - handshake layer of more than 16 byte
+			// - or message type makes so sense (handled through the switch statement)
+			return new SSLUnknownMessage(data, dataLen, container);
```
seladb, line 1241: We currently don't have `SSLEncryptedMessage`. I'm contemplating if we need one... what do you think?
Or we can keep is `SSLUnknownMessage` for now...

Packet++/src/SSLHandshake.cpp
```diff
@@ -1228,6 +1228,19 @@ SSLHandshakeMessage* SSLHandshakeMessage::createHandshakeMessage(uint8_t* data,
 		return nullptr;
 
 	ssl_tls_handshake_layer* hsMsgHeader = (ssl_tls_handshake_layer*)data;
```
seladb, line 1230: it'd be great to add a test with your pcap file. 

We should also fix the test you mentioned in the PR comment: https://github.com/seladb/PcapPlusPlus/pull/1087#issuecomment-1452378651

Packet++/src/SSLHandshake.cpp
axmahr: How I understood Wireshark's approach, we want the first 5 bytes of the handshake message - including ```ssl_tls_handshake_layer``` in our case - to be zero.
```ssl_tls_handshake_layer``` is namely the additional 4-byte header that comes after the SSL Record Layer Header and specifies the handshake type and payload length. We want to check if these 4 bytes plus the first byte after the header is zero. The expression ```be64toh(*(uint64_t*)data) <= 0xFFFFFF``` should actually do that. I didn't consider the correct byte order, which is now done in the current commit [ea9c0be](https://github.com/seladb/PcapPlusPlus/pull/1087/commits/ea9c0be4dd6cf8b314694612459960d312049cc9) through ```be64toh```.

Packet++/src/SSLHandshake.cpp
axmahr: I would leave it at ```SSLUnknownMessage``` for now. The most important thing to me is that encrypted handshake messages are not mistakenly interpreted as a different handshake type. Our approach should fix that for the most part.

Packet++/src/SSLHandshake.cpp
axmahr: I added a test with my pcap file of #1085 in the commit [fa1fb91](https://github.com/seladb/PcapPlusPlus/pull/1087/commits/fa1fb9149b61b20fd545fb59413e25ae8d009dc9)

The other test should be fixed now as well.

Packet++/src/SSLHandshake.cpp
seladb: You're right, my bad! 👍

seladb:
@axmahr the reason test new test fails on some platforms is because [`SSL-MultipleRecords5.dat`](https://github.com/seladb/PcapPlusPlus/pull/1087/files#diff-7accef88b418c43c85790f9b2893f1b0ee6696f19a5a620f0d5445720fd1f304) has  `\n` at the end which is read as an additional byte to the packet. If you remove this `\n` the test will pass on all platforms

axmahr:
Yeah, I don't know how to fix the tests in ```test_tlsfingerprinting.py```

seladb (approved):

PR #1088, opened 2023-03-04
Refactor SIP method and status code parsing

seladb:


seladb:
@egecetin codecov fails because coverage is up by too much 🤷‍♂️

egecetin:
@seladb Yes its about patch coverage. Maybe we can decrease a bit to pass. I'll check as soon as possible

seladb:
> @seladb Yes its about patch coverage. Maybe we can decrease limit a bit to pass. I'll check as soon as possible

@egecetin I added tests for the methods I refactored but the patch coverage still fails, I'm not sure why 🤔

egecetin:
@seladb It is because of nature of std functions. Even if you call all elements std containers (like unordered_map etc.) shows only partial coverage because of internal invokes. For example compiler generates reallocation instruction sometimes and if you can't invoke this instruction it says this is partial. To be honest invoking all internal things like this is extremely hard for C/C++. What do you think about disabling patch check and only check the project overall coverage? It will be simpler for all of us and contributors.

egecetin:
And also i noticed some problem with icpx container coverage. For some reason report is unparsable for codecov. I'll check it too

seladb:
> @seladb It is because of nature of std functions. Even if you call all elements std containers (like unordered_map etc.) shows only partial coverage because of internal invokes. For example compiler generates reallocation instruction sometimes and if you can't invoke this instruction it says this is partial. To be honest invoking all internal things like this is extremely hard for C/C++. What do you think about disabling patch check and only check the project overall coverage? It will be simpler for all of us and contributors.

Yes, I think disabling patch check is a good idea. I agree it'll be simpler for everyone 👍

clementperon:
> @seladb It is because of nature of std functions. Even if you call all elements std containers (like unordered_map etc.) shows only partial coverage because of internal invokes. For example compiler generates reallocation instruction sometimes and if you can't invoke this instruction it says this is partial. To be honest invoking all internal things like this is extremely hard for C/C++. What do you think about disabling patch check and only check the project overall coverage? It will be simpler for all of us and contributors.

maybe because the libC has not been built with --coverage no ?

egecetin:
@clementperon 🤔. TBH it is too deep knowledge for me, and I don't have any idea but possible

egecetin (approved):

PR #1101, opened 2023-03-12
Build Shared Libs (.dll) on Windows with MSVC

traversebitree:
Dynamic library on Windows cannot be generated before. So I made some changes to do that.
Major changes:
1. In every target's `CMakeLists.txt`, add properties `WINDOWS_EXPORT_ALL_SYMBOLS`, `POSITION_INDEPENDENT_CODE` for building `.dll` on Windows, and add macro `PCPP_tgt_BUILD_DLL` for determining if build shared libs (`tgt` is target's name). Following is an example.
```cmake
if(WIN32 AND BUILD_SHARED_LIBS)
  set_property(TARGET Common++ PROPERTY WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
  set_property(TARGET Common++ PROPERTY POSITION_INDEPENDENT_CODE TRUE)
  target_compile_definitions(Common++ PUBLIC "PCPP_COMMON_BUILD_DLL")
endif()
```
2. In every target's header folder, add a header file named `tgtExport.h` (`tgt` is target's name). For example: `CommonExport.h`. For building `.dll` shared libs, we need to declare `__declspec(dllexport)` and `__declspec(dllimport)` at the beginning of the declarations that are being exported. And define the macro `PCAPPP_tgt_API` to it (When using linux or build static libs, will ignore the macro). And set `tgtExport.h` in public headers in `CMakeLists.txt`.

```cpp
#ifndef PCAPPP_COMMON_EXPORT
#define PCAPPP_COMMON_EXPORT

#if defined(_WIN32) && defined(Common___EXPORTS) && defined(PCPP_COMMON_BUILD_DLL)
#define PCAPPP_COMMON_API __declspec(dllexport)
#elif defined(_WIN32) && defined(PCPP_COMMON_BUILD_DLL)
#define PCAPPP_COMMON_API __declspec(dllimport)
#else
#define PCAPPP_COMMON_API
#endif


#endif // PCAPPP_COMMON_EXPORT
```
3. In nearly every header file, add `#include "tgtExport.h"` at begin, and add `PCAPPP_tgt_API` ahead of each static member of class, for example:`static const IPv4Address Zero;` change to `PCAPPP_COMMON_API static const IPv4Address Zero;`. refer to [https://www.kitware.com//create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/](https://www.kitware.com//create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/). 

4. Add "copy `.dll` to build directory and install `.dll` to install directory" for every example target. Only if `BUILD_SHARED_LIBS` and `WIN32`, will copy `.dll` to the targets folder. Following the example:
```cmake
if(BUILD_SHARED_LIBS AND WIN32)
  add_custom_command(
    TARGET ArpSpoofing
    POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_RUNTIME_DLLS:ArpSpoofing> $<TARGET_FILE_DIR:ArpSpoofing>
    COMMAND_EXPAND_LISTS)
endif()

  if(BUILD_SHARED_LIBS AND WIN32)
    install(FILES $<TARGET_RUNTIME_DLLS:ArpSpoofing> DESTINATION ${PCAPPP_INSTALL_BINDIR})
  endif()

```

Minor changes:
1. In `.gitignore`, add `compile_commands.json` and `.cache` for programmers who use `CMake` and `Clangd`.

Then we can build shared libs (.dll) on Windows using MSVC with option `BUILD_SHARED_LIBS` is `ON`.


clementperon:
Cool thank your for the MR :)

I'm not in the Windows env and really don't know but would it not be easier to do something like this directly in the Cmake

if WIN32 and Compile_Shared and Export 
ADD_DEFINITIONS("PCAPPP_EXPORT=__declspec(dllexport)")
elif WIN32 and Compile_Shared  
else 
ADD_DEFINITIONS("PCAPPP_EXPORT=__declspec(dllimport)")
endif

Leave the PACPP_EXPORT empty for everybody else

And use PACPP_EXPORT in the source code

traversebitree:
@clementperon Hello, here is the reference [https://stackoverflow.com/questions/27429732/cmake-adds-dlibname-exports-compile-definition#:~:text=cmake%20add%20%3Clibname%3E_EXPORTS%20macros%20only%20for%20shared%20libraries.%20It%27s%20useful%20when%20exporting%20API%27s%20in%20Windows%20DLL.](https://stackoverflow.com/questions/27429732/cmake-adds-dlibname-exports-compile-definition#:~:text=cmake%20add%20%3Clibname%3E_EXPORTS%20macros%20only%20for%20shared%20libraries.%20It%27s%20useful%20when%20exporting%20API%27s%20in%20Windows%20DLL.). CMake will automatic add `tgt_Export`, and when building `.dll`, the `tgt_Export` automatically be `true` (to be `__declspec(dllexport)`), and when being linked (other targets link this `.dll`),  the `tgt_Export` automatically be `false` (to be `__declspec(dllimport)`).

clementperon:
Yes I agree you're code is correct, I'm just looking for something simpler.

Just found this why looking :) 
https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html

traversebitree:
@clementperon  Yes, you are right. `GENERATE_EXPORT_HEADER()` is a simpler and cleaner way and I'll have a test on it.

Common++/header/MacAddress.h
```diff
@@ -27,34 +29,37 @@ namespace pcpp
...
-		 * (as MAC address is 6-byte long)
-		 * @todo there is no verification array length >= 6. If this is not the case, address will read uninitialized memory
+		 * A constructor that creates an instance of the class out of a byte array. The byte array length must be equal
+		 * or greater to 6 (as MAC address is 6-byte long)
+		 * @todo there is no verification array length >= 6. If this is not the case, address will read uninitialized
+		 * memory
 		 * @param[in] addr A pointer to the byte array containing 6 bytes representing the MAC address
 		 */
-		MacAddress(const uint8_t* addr) : m_IsValid(true) { memcpy(m_Address, addr, sizeof(m_Address)); }
+		MacAddress(const uint8_t *addr) : m_IsValid(true) { memcpy(m_Address, addr, sizeof(m_Address)); }
```
egecetin, line 46: Not very important but since PR already has many changes cleaning this type of unnecessary changes might be better before merging to ease review, track history etc.

PS: Shared library support for windows would be great.

.gitignore
```diff
@@ -73,3 +73,9 @@ Examples/Tutorials/**/*.exe
 #venv
 .venv/**
 venv/**
+
+#CMake
+compile_commands.json
```
egecetin, line 78: This is more related with VS Code instead of CMake

clementperon: This the list of all compile commands CMake will execute.

traversebitree: It's something about CMake, Clangd, VS Code together. CMake Generates `compile_commands.json` by "CMAKE_EXPORT_COMPILE_COMMANDS" [https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html](https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html) Clangd consumes `compile_commands.json` by set `--compile-commands-dir=` [https://clangd.llvm.org/design/compile-commands](https://clangd.llvm.org/design/compile-commands). And VS Code provides plugins such as `clangd` and `CMake Tools` to use `compile_commands.json` and analyze the whole project.

egecetin: Thanks for info

Common++/header/MacAddress.h
traversebitree: Thank you, I have corrected it.

clementperon:
@traversebitree also would be nice to add a windows job with DLL built in CI, thanks.

Don't hesitate if you need help !

traversebitree:
@clementperon OK, I'll have a try.

traversebitree:
- I tried to add build shared libs (.dll) jobs in the workflow. But these jobs will fail when at stage "TestPcapPlusPlus" -> `python ci\run_tests\run_tests_windows.py --coverage`.
image
image

- I try to manually execute `PcapPlusPlus\Tests\Packet++Test\Bin\Packet++Test.exe` in HyperV. The error occurs here:
image
- Then I try to execute using **static** libs (.lib), no the above error:
image

- Maybe the error occurs on the test case `EthPacketPointerCreation`. I try to manually run the test case `EthPacketPointerCreation` but failed. Can you help to check this? I guess the problem happens at this point. @clementperon @egecetin .

- To add: 
I manually execute `Pcap++Test\Bin\Pcap++Test.exe` and there is no error like in `Packet++Test.exe`.
image

egecetin:
For `Cannot find an interface to run tests on!`, can you try to run python manually on an administrator cmd/powershell? It might be an only privilege issue.

seladb:
@traversebitree is there a way to generate `.def` files using CMake instead of adding `__declspec(dllexport)` everywhere?

Also - you add `#include "tgtExport.h"` in every `.h` file but you don't use the macros in most of them. Is it on purpose?

traversebitree:
@seladb I refered to project `SFML` [https://github.com/SFML/SFML/blob/master/include/SFML/Config.hpp#:~:text=export%20and%20import-,%23,(dllexport),-%23](https://github.com/SFML/SFML/blob/master/include/SFML/Config.hpp#:~:text=export%20and%20import-,%23,(dllexport),-%23) to add `__declspec(dllexport)` ahead of static member of class (which CMake's `WINDOWS_EXPORT_ALL_SYMBOLS` can't cover).
And I `#include "tgtExport.h` in every `.h` file is prepare for future if any header file may add a static member. Of course, for now, it's redundant in many `.h` file.

traversebitree:
@egecetin I add a print for debugging `run_tests_windows.py`.
image

I use admin to run it again, and still no interface in my HyperV Win10.
gif1

seladb:
> @seladb I refered to project `SFML` https://github.com/SFML/SFML/blob/master/include/SFML/Config.hpp#:~:text=export%20and%20import-,%23,(dllexport),-%23 to add `__declspec(dllexport)` ahead of static member of class (which CMake's `WINDOWS_EXPORT_ALL_SYMBOLS` can't cover). And I `#include "tgtExport.h` in every `.h` file is prepare for future if any header file may add a static member. Of course, for now, it's redundant in many `.h` file.

From this PR It looks that `#include "tgtExport.h` is only needed for `public static const` properties, maybe we can limit the changes to only files that contain them?

traversebitree:
@seladb Yes, and I'll recommit and limit the changes.

traversebitree:
I've found the problem. It's the problem with `MemPlumber`. when  set `bool skipMemLeakCheck = true;`. `Packet++Test` won't stuck at the Runtime Error.
I manually delete the test cases, and I found the test cases (in `Packet++Test`) `EthPacketPointerCreation`, `TcpChecksumInvalidRead`, `InsertDataToPacket`, `CreatePacketFromBuffer`, `IcmpV6EditTest` will be stuck at the Runtime Error.

seladb:
> I guess `MSVC`+`dll` incompatible with `MemPlumber`? When set `bool skipMemLeakCheck = true;`. `Packet++Test` won't stuck at the Runtime Error.
> I manually delete the test cases, and I found the test cases (in `Packet++Test`) `EthPacketPointerCreation`, `TcpChecksumInvalidRead`, `InsertDataToPacket`, `CreatePacketFromBuffer`, `IcmpV6EditTest` will be stuck at the Runtime Error.

@traversebitree are these the only tests which fail in `Packet++Test` and `Pcap++Test`?
Do they fail even if you set `skipMemLeakCheck = false`?

seladb:
@traversebitree I think I tried to build shared libs for Windows many years ago and stumbled into a somewhat similar issue. If I remember correctly, there was an issue when a DLL and an external app share memory - e.g the DLL gives direct access to its memory to the external app, or the app gives direct access to its memory to the DLL.

This happens quite a lot in PcapPlusPlus.

For example here the DLL shares a pointer to its internal memory with the external app (in this case the test app):
https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Packet++Test/Tests/IPv4Tests.cpp#L50-L57

**_I'm not at all sure these are the issues you see_**, I just remember them vaguely from many years ago.

If this is indeed the issue, it'd be very hard to solve because PcapPlusPlus uses this memory sharing in many places in its API.

egecetin:
@seladb I'm not 100% sure but looks like similar with situation explained _Memory allocated in Memory Manager A and freed in Memory Manager B_ section of [this](https://stackoverflow.com/a/18904939)

traversebitree:
@egecetin Yes, I also noticed this problem, so I added this `set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>" CACHE STRING "MTd")` CMakeLists.txt. It changes the compile flag from `-MDd` to `-MTd`. But the results haven't changed, and the problem remains.
@seladb Indeed, this problem is really very difficult, it seems to take a long time to solve it.

traversebitree:
@seladb The fact is that building .dll is not compatible with MemPlumber. When the memory check is skipped, the building .dll is valid and passes the test. So, I don't know whether to go ahead and modify the MemPlumber or terminate this PR.

seladb:
> @seladb The fact is that building .dll is not compatible with MemPlumber. When the memory check is skipped, the building .dll is valid and passes the test. So, I don't know whether to go ahead and modify the MemPlumber or terminate this PR. I've been working on my paper recently, so I haven't been able to continue working on this PR lately, sorry about that.

@traversebitree don't we have this issue as well? https://github.com/seladb/PcapPlusPlus/pull/1101#issuecomment-1469593201 that is unrelated to MemPlumber?

traversebitree:
I'm sorry, but after trying too many times, it still doesn't pass the memory leak test.

Issue #1009, opened 2022-12-10
Better support for subnets


seladb:
As discussed here: https://www.reddit.com/r/cpp/comments/yqngc1/comment/ivrmm9u/?utm_source=share&utm_medium=web2x&context=3

We can improve our support for subnets by creating an `IPv4Network` and `IPv6Network` classes that will represent subnets like `192.168.0.0/16` or `2001:db00::0/ffff:ff00::` and then use them to match subnets in `IPv4Address` and `IPv6Address`. This is somewhat similar to the Python approach: https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Network

seladb:
Done in PRs #1103, #1106, #1114

PR #1103, opened 2023-03-16
Add IPv4 networks

seladb:
Addresses issue: https://github.com/seladb/PcapPlusPlus/issues/1009

Inspired by the Python implementation for subnets: https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Network


Common++/header/IpAddress.h
```diff
@@ -447,6 +463,102 @@ namespace pcpp
...
+		 */
+		std::string getSubnetMask() const { return IPv4Address(m_Mask).toString(); }
+
+		/**
+		 * @return The network prefix, for example: the network prefix of 10.10.10.10/16 is 10.10.0.0
+		*/
+		IPv4Address getNetworkPrefix() const { return IPv4Address(m_NetworkPrefix); }
+
+		/**
+		 * @return The lowest IPv4 address in this network, for example: the lowest address in 10.10.l0.10/16 is
```
clementperon, line 524: typo 10.10.l0.10

seladb: Good catch, thanks!

seladb: Fixed in https://github.com/seladb/PcapPlusPlus/pull/1103/commits/96ebb3f4dbe61d8511c75a446110ed7da15b153e

egecetin (approved):

Common++/header/IpAddress.h
```diff
@@ -7,6 +7,17 @@
 #include <algorithm>
 #include <ostream>
 
+#ifndef PCPP_DEPRECATED
+#if defined(__GNUC__) || defined(__clang__)
+#define PCPP_DEPRECATED __attribute__((deprecated))
+#elif defined(_MSC_VER)
+#define PCPP_DEPRECATED __declspec(deprecated)
+#else
+#pragma message("WARNING: DEPRECATED feature is not implemented for this compiler")
```
clementperon, line 16: maybe #error else we will just ignore this message and never catch it in the CI no ?

Common++/header/IpAddress.h
```diff
@@ -447,6 +463,108 @@ namespace pcpp
...
+
+		/**
+		 * @return The prefix length, for example: the prefix length of 10.10.10.10/255.0.0.0 is 8
+		 */
+		uint8_t getPrefixLen() const;
+
+		/**
+		 * @return The subnet mask, for example: the subnet mask of 10.10.10.10/8 is 255.0.0.0
+		 */
+		std::string getSubnetMask() const { return IPv4Address(m_Mask).toString(); }
```
clementperon, line 516: why SubnetMask and not just NetMask ?

Common++/header/IpAddress.h
```diff
@@ -447,6 +463,108 @@ namespace pcpp
...
+
+		/**
+		 * @return The network prefix, for example: the network prefix of 10.10.10.10/16 is 10.10.0.0
```
clementperon, line 519: I would expect network prefix to return 10.10.0.0/16 no ?

Common++/header/IpAddress.h
```diff
@@ -447,6 +463,108 @@ namespace pcpp
...
+		IPv4Address getLowestAddress() const;
+
+		/**
+		 * @return The highest IPv4 address in this network, for example: the highest address in 10.10.10.10/16 is
+		 * 10.10.255.255
+		 */
+		IPv4Address getHighestAddress() const;
+
+		/**
+		 * @return The number of addresses in this network, for example: the number of addresses in 10.10.0.0/24 is 256
```
clementperon, line 536: 0 and 255 are reserved,  i'm not sure this is really useful, what do you think ?

Common++/header/IpAddress.h
```diff
@@ -447,6 +463,108 @@ namespace pcpp
...
+
+		/**
+		 * @param address An IPv4 address
+		 * @return True is the address belongs to the network, false otherwise or if the address isn't valid
+		 */
+		bool includes(const IPv4Address& address) const;
+
+		/**
+		 * @param network An IPv4 network
+		 * @return True is the input network is included within this network, false otherwise, for example:
```
clementperon, line 548: I would explicit if the network should be totally included or should overlap.

True is the input network is **totally/completely** included within this network

Common++/src/IpAddress.cpp
```diff
@@ -178,4 +180,174 @@ namespace pcpp
 	}
 
 
+	// ~~~~~~~~~~~
+	// IPv4Network
+	// ~~~~~~~~~~~
+
+
+	bool IPv4Network::isValidSubnetMask(const std::string& subnetMask)
```
clementperon, line 188: I would make this belong to IPv4address

Common++/header/IpAddress.h
seladb: I didn't mark this as an error because I don't want to break the API. If someone uses this API they should get a warning and hopefully switch to the new API. In later versions we can remove it altogether

Common++/header/IpAddress.h
seladb: Good point! I'll change that

Common++/header/IpAddress.h
seladb: you're right, they are reserved... I just checked what happens in python and the behavior is a bit weird: when you ask for the number of addresses you get 256, but the list of hosts returns only those who are not reserved:

```text
>>> from ipaddress import IPv4Network
>>> n = IPv4Network("192.168.40.0/24")
>>> n.num_addresses
256
>>> sum(1 for _ in n.hosts())
254
>>> next(n.hosts())
IPv4Address('192.168.40.1')
>>> *_, last = n.hosts()
>>> last
IPv4Address('192.168.40.254')
```

@clementperon do you think we should follow the same convention?

Common++/src/IpAddress.cpp
seladb: I'm not sure what you mean... do you mean that `subnetMask` should be an `IPv4Address` and not a `std::string`?

seladb: yes, that is correct. The reasons this method takes a string and not `IPv4Address` are:
- This is a private method so it's not exposed to the user
- The input we get from the user is a string, like [here](https://github.com/seladb/PcapPlusPlus/blob/4a5eb6f8d52215b490b92dfefde53365f6640491/Common%2B%2B/src/IpAddress.cpp#L247) and [here](https://github.com/seladb/PcapPlusPlus/blob/4a5eb6f8d52215b490b92dfefde53365f6640491/Common%2B%2B/src/IpAddress.cpp#L263)
- I thought about making it an `IPv4Address` but this might be a bit confusing because, as you mentioned, a subnet mask is a restricted IP address and for most users it's easier to just type "255.255.0.0"

Common++/header/IpAddress.h
clementperon: I was more thinking PCAPP_DEPRECATED not defined is an error.

Common++/header/IpAddress.h
clementperon: Both are fine but we should explain it in the API comment. Or drop the API if the name is not explicit.

clementperon (approved):

Common++/header/IpAddress.h
seladb: I don't think it should happen that `PCAPP_DEPRECATED`, but just in case - we define it and output this warning...

Common++/header/IpAddress.h
seladb: I modified the behavior here: https://github.com/seladb/PcapPlusPlus/pull/1103/commits/b2be4649d5cb16242fcba996e2f61d546ec58cc5

PR #1114, opened 2023-04-03
Add IP network

seladb:
Addresses issue: https://github.com/seladb/PcapPlusPlus/issues/1009

Adds an `IPNetwork` class that generalizes `IPv4Network` and `IPv6Network`, in a similar way that `IPAddress` generalizes `IPv4Address` and `IPv6Address`

### DIFF
--- a/Common++/header/IpAddress.h
+++ b/Common++/header/IpAddress.h
@@ -760,10 +760,30 @@ namespace pcpp
 				m_IPv4Network = new IPv4Network(addressAndNetmask);
 				m_IPv6Network = nullptr;
 			}
-			catch (const std::invalid_argument& e)
+			catch (const std::invalid_argument&)
 			{
 				m_IPv6Network = new IPv6Network(addressAndNetmask);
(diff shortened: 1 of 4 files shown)


Common++/header/IpAddress.h
```diff
@@ -687,6 +687,322 @@ namespace pcpp
...
+		IPNetwork(const std::string& addressAndNetmask)
+		{
+			try
+			{
+				m_IPv4Network = new IPv4Network(addressAndNetmask);
+				m_IPv6Network = nullptr;
+			}
+			catch (const std::invalid_argument&)
+			{
+				m_IPv6Network = new IPv6Network(addressAndNetmask);
```
clementperon, line 765: Can the IPv6Network also raise a signal ?

seladb: yes, it can raise an exception, but it means the string is not a valid IPv4 nor an IPv6 network, hence the exception should be thrown outside to the user

clementperon (approved):

PR #1121, opened 2023-04-16
Add PAT

egecetin:
For triggering workflow actions automated PRs requires PAT (Mentioned [here](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs)). After merge all of the workflows should be triggered and coverage report should be fixed

### DIFF
--- a/.github/workflows/auto_update.yml
+++ b/.github/workflows/auto_update.yml
@@ -24,6 +24,7 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v4
         with:
+          token: ${{ secrets.PAT }}
           author: GitHub <noreply@github.com>
           add-paths: .pre-commit-config.yaml
           commit-message: Auto pre-commit update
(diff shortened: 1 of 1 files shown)


seladb:
> @seladb Just a reminder. Did you add PAT to repo secrets?

@egecetin I'm sorry, I didn't add a PAT yet. Does it need to be a secret with write access to the repo?
Any idea which permissions it needs?

image

seladb:
@egecetin I'm not sure which of the permissions in the screenshot above is "repo scoped write access". Do you have any idea?

egecetin:
@seladb I'm not sure too but only the public_repo should be enough I think. Check this https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps

seladb:
@egecetin I created the token and put it in a new `PAT` secret, let me know if anything else is needed

PR #1123, opened 2023-04-19
WIP: Improve Fuzzing infrastructure

R3x:
Hello, 

The current fuzzing harness has multiple issues such as File I/O overhead, and not being multiprocess safe. So I am trying to improve it to get better coverage and more executions per second.

Please feel free to tell me suggestions! and I can try to tweak the code accordingly

I am also planning to add Fuzzers for different layers, as well based on the existing test cases that I can see.


Tests/Fuzzers/FuzzPacketParse.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+	if (Size < 2)
+		return 0;
+
+	// read the first (and only) packet from the file
+	pcpp::RawPacket rawPacket(((const uint8_t *)Data + 1), Size - 1, ts, false, pcpp::LinkLayerType(Data[0]));
+	// parse the raw packet into a parsed packet
+	pcpp::Packet parsedPacket(&rawPacket);
+	// Convert the parsed packet to string to trigger all the parsing formats
+	parsedPacket.toString();
```
seladb, line 22: I'm not sure we really need to trigger `toString()` (although we do it now). Creating a `pcpp::Packet` out of a `pcpp::RawPacket` already does all the parsing:

It calls `setRawPacket()`: https://github.com/seladb/PcapPlusPlus/blob/1d29c00f441dc50650b500bb6b6fe72efce28990/Packet%2B%2B/src/Packet.cpp#L127

Which in turn creates all the layers: https://github.com/seladb/PcapPlusPlus/blob/1d29c00f441dc50650b500bb6b6fe72efce28990/Packet%2B%2B/src/Packet.cpp#L58

Tests/Fuzzers/FuzzPacketParse.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include <RawPacket.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
+{
+    struct timespec ts;
+    timespec_get(&ts, TIME_UTC);
+
+	if (Size < 2)
+		return 0;
+
```
seladb, line 16: I think we want to suppress logs like we do today:
https://github.com/seladb/PcapPlusPlus/blob/1d29c00f441dc50650b500bb6b6fe72efce28990/Tests/Fuzzers/FuzzTarget.cpp#L47

To prevent redundant prints

Tests/Fuzzers/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
-add_executable(FuzzTarget FuzzTarget.cpp)
+add_executable(FuzzPacketParse FuzzPacketParse.cpp)
```
seladb, line 1: I think the fact we're changing the executable name makes the regression tests fail:
https://cirrus-ci.com/task/5398413525647360

Tests/Fuzzers/FuzzPacketParse.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include <IPv6Layer.h>
+#include <unistd.h>
+#include <Packet.h>
+#include <PcapFileDevice.h>
+#include <RawPacket.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
+{
+    struct timespec ts;
+    timespec_get(&ts, TIME_UTC);
```
seladb, line 12: nit: replace spaces with tabs (`\t`)

Tests/Fuzzers/FuzzPacketParse.cpp
R3x: I saw the toString functions call a bunch of functions that aren't called otherwise. Seemed to be the most straightforward way to improve coverage. I will double check

Tests/Fuzzers/FuzzPacketParse.cpp
R3x: Noted! will fix

Tests/Fuzzers/FuzzPacketParse.cpp
seladb: yes, in some protocols `toString()` does call a bunch of other functions (and in some not), so it really depends on how much of the PcapPlusPlus API we want to invoke with fuzzing testing. Probably the more the better 😃 but we should think what makes sense...

R3x:
@seladb - Regarding coverage and which functions to use in our harnesses. I am thinking to create fuzzers for each Layer in Packet++. 

My general approach is to classify functions into 2 types -> ones that consume some sort of user data (consumers) and ones that process the data (processors). I will try to write the harness in such a way that, each of the consumers is given random input of random size from the fuzzer and each of the processors is called.

For example - [ArpLayer](https://github.com/seladb/PcapPlusPlus/blob/master/Packet%2B%2B/src/ArpLayer.cpp) has 3 functions defined for the Layer.  Based on [Tests](https://github.com/seladb/PcapPlusPlus/blob/1d29c00f441dc50650b500bb6b6fe72efce28990/Tests/Packet%2B%2BTest/Tests/EthAndArpTests.cpp#L106), I can figure out which all functions I need to call. 

In this case only the constructor needs to be called with random input and I just need to call the other two functions -> which should give us 100% coverage. And moving forward if someone adds new functions, they can just update the harness for that layer as such.

Ofc, this is based on a preliminary reading of the code, I just started looking at this repository a while back - I am open to suggestions/ideas.

seladb:
@R3x I'm not against writing more detailed fuzzing tests, but I'm concerned that calling every method of every layer will create a massive codebase that will be hard to maintain and will basically look like a clone of the current tests. Maybe we should focus on specific layers (the most popular ones) and add more calls there? 🤔

R3x:
Sure, we can start with that - Which layers do you think are the most commonly used?

seladb:
> Sure, we can start with that - Which layers do you think are the most commonly used?

Hmm, maybe we can start with Ethernet, IPv4, IPv6, TCP and UDP? 🤔

seladb:
@R3x this PR is stale for a long time, are you planning to work on it or should we close it?

R3x:
Sorry, forgot about this due to some other work. I will try to get on it this sometime this week. Maybe we can merge the initial fuzzer first and then I can work in parallel on other fuzzers.

seladb:
The CI fails for this PR so we can't merge it as is...

seladb:
@R3x this PR is stale for a long time. Should we close it?

PR #1131, opened 2023-05-07
Allow building fuzzers independently from tests

sashashura:
Building tests takes time every time I need to rebuild the fuzzing harness. It should be possible to build them independently.

Adding subdirectory of subdirectory maybe not elegant. Another option is to move the whole fuzzers directory one level up.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,12 +81,7 @@ cmake_dependent_option(
   OFF)
 option(PCAPPP_BUILD_TESTS "Build Tests" ${PCAPPP_MAIN_PROJECT})
 option(PCAPPP_BUILD_COVERAGE "Generate Coverage Report" OFF)
-cmake_dependent_option(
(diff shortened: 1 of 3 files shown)


clementperon:
Change the tests/CMakeLists.txt to be 

```
if(PCAPPP_BUILD_TESTS)
  add_subdirectory(Packet++Test)
  add_subdirectory(Pcap++Test)
  add_subdirectory(PcppTestFramework)
endif()

if(PCAPPP_BUILD_EXAMPLES)
  add_subdirectory(ExamplesTest)
endif()

if(PCAPPP_BUILD_FUZZERS)
  add_subdirectory(Fuzzers)
endif()
```

and remove the if in the upper CMakeLists

clementperon:
@sashashura what do you think about my proposition ?

clementperon (approved):

seladb (approved):

PR #1143, opened 2023-05-30
Fix out of bounds memory read in pcpp::SomeIpSdLayer::SomeIpSdLayer

sashashura:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53099

### DIFF
--- a/Packet++/src/SomeIpSdLayer.cpp
+++ b/Packet++/src/SomeIpSdLayer.cpp
@@ -631,6 +631,20 @@ uint32_t SomeIpSdLayer::addEntry(const SomeIpSdEntry &entry)
 	return getNumEntries() - 1;
 }
 
+bool SomeIpSdLayer::isDataValid(const uint8_t* data, size_t dataLen)
+{
+	if (!data)
+		return false;
+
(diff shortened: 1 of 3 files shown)


Packet++/src/SomeIpSdLayer.cpp
```diff
@@ -631,6 +631,20 @@ uint32_t SomeIpSdLayer::addEntry(const SomeIpSdEntry &entry)
...
+bool SomeIpSdLayer::isDataValid(const uint8_t* data, size_t dataLen)
+{
+	if (!data)
+		return false;
+
+	if (dataLen < (sizeof(someipsdhdr) + sizeof(uint32_t)))
+		return false;
+
+	if (dataLen < (sizeof(someipsdhdr) + sizeof(uint32_t) + getLenEntries(data)))
+		return false;
```
seladb, line 643: nit: we can make this code slightly shorter:
```cpp
if (!data || dataLen < sizeof(someipsdhdr) + sizeof(uint32_t) || dataLen < sizeof(someipsdhdr) + sizeof(uint32_t) + getLenEntries(data))
{
	return false;
}

return true;
```

seladb (approved):

PR #1148, opened 2023-06-15
Auto precommit update

seladb:
Update pre-commit hooks to latest

Auto generated


egecetin:
@seladb Updated the ignore list since the new warnings are false positives. Also I noticed some suppressions were for deprecated functions already deleted so I removed these suppressions.

seladb:
> @seladb Updated the ignore list since the new warnings are false positives. Also I noticed some suppressions were for deprecated functions already deleted so I removed these suppressions.

I'm sorry for the late response. Sure, thank you for taking care of this! 🙏

PR #1150, opened 2023-07-04
ARP add isRequest and isReply

tigercosmos:
support ARP Layer to check if the packet is a reply or request directly

### DIFF
--- a/Packet++/header/ArpLayer.h
+++ b/Packet++/header/ArpLayer.h
@@ -131,6 +131,16 @@ namespace pcpp
 		 */
 		void computeCalculateFields();
 
+		/**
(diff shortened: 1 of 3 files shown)


seladb:
Thank you for opening the PR @tigercosmos !

Can you please add tests for these methods?
You can probably add it here: https://github.com/seladb/PcapPlusPlus/blob/dc21edb988409ecb798b033e766be6f49a1f6541/Tests/Packet%2B%2BTest/Tests/EthAndArpTests.cpp#L87

tigercosmos:
@seladb  sure, please take a look

seladb (approved):

PR #1159, opened 2023-07-15
Auto OUI Database Update

seladb:
Update OUI database to latest

Auto generated


egecetin:
@seladb Vendor names were wrong due to whitespaces after split. I updated the python code to delete these. It should fix the issue. Can you take a look?

egecetin (approved):

seladb:
@egecetin looks good to me 👍

Issue #1137, opened 2023-05-23
Add DNSKEY answer crash 


aengusjiang:
try to add dnskey answer, it will cause memory access overflow, becuase the tmp buffer only have 256 len, but the DNSKEY len could be 264

aengusjiang:
![image](https://github.com/seladb/PcapPlusPlus/assets/22406984/88621225-6c05-4b6b-bdfa-fa30f30a5656)

aengusjiang:
![image](https://github.com/seladb/PcapPlusPlus/assets/22406984/9cf5dd3c-88b9-40d9-b3e9-35273c0372d6)

aengusjiang:
recommend to change it to 2048 or 4096

seladb:
Thank you @aengusjiang , this indeed seems like a bug. Regardless of the array size, there should be a check to make sure there is no buffer overflow

seladb:
Thank you so much @aengusjiang for providing a fix, much appreciated! 🙏 🙏

PR #1165, opened 2023-07-26
Fix dnslayer memory overflow when add Resource

aengusjiang:
relative issue: https://github.com/seladb/PcapPlusPlus/issues/1137

### DIFF
--- a/Packet++/src/DnsLayer.cpp
+++ b/Packet++/src/DnsLayer.cpp
@@ -475,7 +475,7 @@ DnsResource* DnsLayer::addResource(DnsResourceType resType, const std::string& n
 		uint32_t ttl, IDnsResourceData* data)
 {
 	// create new query on temporary buffer
-	uint8_t newResourceRawData[256];
+	uint8_t newResourceRawData[4096];
 	memset(newResourceRawData, 0, sizeof(newResourceRawData));
 
 	DnsResource* newResource = new DnsResource(newResourceRawData, resType);
@@ -561,7 +561,7 @@ DnsResource* DnsLayer::addResource(DnsResourceType resType, const std::string& n
 DnsQuery* DnsLayer::addQuery(const std::string& name, DnsType dnsType, DnsClass dnsClass)
 {
 	// create new query on temporary buffer
-	uint8_t newQueryRawData[256];
+	uint8_t newQueryRawData[4096];
 	DnsQuery* newQuery = new DnsQuery(newQueryRawData);
 
 	newQuery->setDnsClass(dnsClass);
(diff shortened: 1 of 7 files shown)


egecetin:
@aengusjiang I have a general question about the buffer size. You recommended buffer size 2048 or 4096 bytes. I checked the RFC4034 but I can't find any information about the maximum size. It only mentions, it depends on the algorithm (SHA1 etc.) but do you know is there any theoretical/practical limit for the DNSKEY field?

egecetin:
I understand now, thanks.

egecetin:
Looks like the tests are failing after change.
image

aengusjiang:
I understand. I remove the invalid check dataLength. In fact, it just pass 0.

Packet++/src/DnsLayer.cpp
```diff
@@ -561,7 +561,7 @@ DnsResource* DnsLayer::addResource(DnsResourceType resType, const std::string& n
 DnsQuery* DnsLayer::addQuery(const std::string& name, DnsType dnsType, DnsClass dnsClass)
 {
 	// create new query on temporary buffer
-	uint8_t newQueryRawData[256];
+	uint8_t newQueryRawData[4096];
```
seladb, line 564: isn't DNSKey a resource record and not a query?
Are there queries that exceed 256 bytes?

Packet++/src/DnsLayer.cpp
seladb, line 1: Is it possible to add a test that adds DNSKey with a length of more than 256 bytes?

Packet++/src/DnsLayer.cpp
aengusjiang: DNSKey  is an answer record, here maybe needn't change to 4096. I will try revert it.

Packet++/src/DnsLayer.cpp
aengusjiang: Good idea. I will add later.

seladb:
@aengusjiang please let me know if you are planning to address the comments on this PR so we can merge it

Packet++/src/DnsResourceData.cpp
```diff
@@ -218,7 +218,7 @@ bool GenericDnsResourceData::toByteArr(uint8_t* arr, size_t& arrLength, IDnsReso
 {
 	if (m_DataLen == 0 || m_Data == nullptr)
 	{
-		PCPP_LOG_ERROR("Input data is null or illegal");
+		PCPP_LOG_ERROR("Input data is null or illegal" << "|m_DataLen:" << m_DataLen);
```
seladb, line 221: This change is probably not needed, but we can fix it later

Issue #1167, opened 2023-07-26
Add support for choosing NFLOG groups when opening to NFLOG device


MrPeck:
NFLOG packets are sent to the device "nflog" in a specified NFLOG Group. When not specified which group, 0 is the default.
There is no support for choosing which group to listen to on the device.

**Example**
`tcpdump -i nflog:123`

When running this command, you will listen to nflog on NFLOG Group 123.

I'll later make a pull request with a suggestion for the feature :)

seladb:
Thank you @MrPeck for fixing this issue, much appreciated!  🙏

Issue #1166, opened 2023-07-26
Fix NFLOG error: Operation not permitted


MrPeck:
When opening a NFLOG device we receive the following error:

`[ERROR: /<MY-PATH>/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp: doOpen:267] Can't listen on group group index: Operation not permitted`

This error is cause because in function `bool PcapLiveDevice::open(const DeviceConfiguration& config)`, the device is opened twice. one descriptor is used to capture packets while the second one is to send packets, as described in the source code:

PcapLiveDevice.h
```
// This is a second descriptor for the same device. It is needed because of a bug
// that occurs in libpcap on Linux (on Windows using WinPcap/Npcap it works well):
// It's impossible to capture packets sent by the same descriptor
pcap_t* m_PcapSendDescriptor;
````

PcapLiveDevice.cpp
```
bool PcapLiveDevice::open(const DeviceConfiguration& config)
{
    [...]

    m_PcapDescriptor = doOpen(config);
    m_PcapSendDescriptor = doOpen(config);

    if (m_PcapDescriptor == nullptr || m_PcapSendDescriptor == nullptr)
    {
        m_DeviceOpened = false;
        return false;
    }
    
    [...]
}
```

This cannot be done for NFLOG as a NFLOG device (with a given NFLOG Group) cannot have more than one open instance at a time.

Issue is reproducible by simply opening an nflog device:

```
dev = pcpp::PcapLiveDeviceList::getInstance().getPcapLiveDeviceByName("nflog");
if (dev == nullptr)
{
	std::cerr << "Cannot find nflog interface" << std::endl;
	return 1;
}

if (!dev->open())
{
	std::cerr << "Cannot open device" << std::endl;
	return 1;
}
```

**General Info**
I'm running this in a Linux arm64 emulation, but the issue also happens in ubuntu machines. Let me know if there is any other data required. I will soon be making a pull request with a suggestion for a fix.

seladb:
Thank you @MrPeck for fixing this issue, much appreciated! 🙏

PR #1169, opened 2023-07-26
Add support for NFLOG groups and "Operation not permitted" error

MrPeck:
This pull requests comes to solve two issues: #1166 and #1167

### DIFF
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -212,10 +212,10 @@ namespace pcpp
 			int snapshotLength;
 
 			/**
-			 * Set extra information. For some types of device, extra information might be required. Such a case is when opening a NFLOG
-			 * device and setting a different NFLOG group. If no extra information is required, this configuration is ignored.
+			 * Set NFLOG group. Which NFLOG group to be listened to when connecting to NFLOG device. If device is not of type NFLOG this 
+			 * attribute is ignored.
 			*/
-			unsigned int extra;
+			unsigned int nflogGroup;
 
 			/**
 			 * A c'tor for this struct
@@ -230,18 +230,17 @@ namespace pcpp
 			 * A snapshot length of 262144 should be big enough for maximum-size Linux loopback packets (65549) and some USB packets
 			 * captured with USBPcap (> 131072, < 262144). A snapshot length of 65535 should be sufficient, on most if not all networks,
 			 * to capture all the data available from the packet.
-			 * @param[in] extra Extra information for specific devices. Default value is 0. This value is used when a device needs to be
-			 * provided with some extra information. If such information is not required, this configuration shall be ignored.
+			 * @param[in] nflogGroup NFLOG group for NFLOG devices. Default value is 0.
 			*/
 			explicit DeviceConfiguration(DeviceMode mode = Promiscuous, int packetBufferTimeoutMs = 0, int packetBufferSize = 0,
-				                PcapDirection direction = PCPP_INOUT, int snapshotLength = 0, unsigned int extra = 0)
+				                PcapDirection direction = PCPP_INOUT, int snapshotLength = 0, unsigned int nflogGroup = 0)
 			{
 				this->mode = mode;
 				this->packetBufferTimeoutMs = packetBufferTimeoutMs;
 				this->packetBufferSize = packetBufferSize;
 				this->direction = direction;
 				this->snapshotLength = snapshotLength;
-				this->extra = extra;
+				this->nflogGroup = nflogGroup;
 			}
 		};
 
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -212,7 +212,7 @@ namespace pcpp
 			int snapshotLength;
 
 			/**
-			 * Set extra information. For some types of device, extra information might be required. Such a case is when opening a NFLOG 
+			 * Set extra information. For some types of device, extra information might be required. Such a case is when opening a NFLOG
 			 * device and setting a different NFLOG group. If no extra information is required, this configuration is ignored.
 			*/
 			unsigned int extra;
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -212,7 +212,7 @@ namespace pcpp
 			int snapshotLength;
 
 			/**
-			 * Set NFLOG group. Which NFLOG group to be listened to when connecting to NFLOG device. If device is not of type NFLOG this 
+			 * Set NFLOG group. Which NFLOG group to be listened to when connecting to NFLOG device. If device is not of type NFLOG this
 			 * attribute is ignored.
 			*/
 			unsigned int nflogGroup;
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -230,6 +230,8 @@ namespace pcpp
 			 * A snapshot length of 262144 should be big enough for maximum-size Linux loopback packets (65549) and some USB packets
 			 * captured with USBPcap (> 131072, < 262144). A snapshot length of 65535 should be sufficient, on most if not all networks,
 			 * to capture all the data available from the packet.
+			 * @param[in] extra Extra information for specific devices. Default value is 0. This value is used when a device needs to be
+			 * provided wih some extra information. If such information is not required, this configuration shall be ignored.
 			*/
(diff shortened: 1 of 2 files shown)


Pcap++/header/PcapLiveDevice.h
```diff
@@ -226,13 +232,14 @@ namespace pcpp
 			 * to capture all the data available from the packet.
 			*/
 			explicit DeviceConfiguration(DeviceMode mode = Promiscuous, int packetBufferTimeoutMs = 0, int packetBufferSize = 0,
-				                PcapDirection direction = PCPP_INOUT, int snapshotLength = 0)
+				                PcapDirection direction = PCPP_INOUT, int snapshotLength = 0, unsigned int extra = 0)
```
egecetin, line 235: Doxy gives error since there is an undocumented new parameter, the documentation of the function should be also updated.

Pcap++/header/PcapLiveDevice.h
```diff
@@ -211,6 +211,12 @@ namespace pcpp
 			*/
 			int snapshotLength;
 
+			/**
+			 * Set extra information. For some types of device, extra information might be required. Such a case is when opening a NFLOG 
+			 * device and setting a different NFLOG group. If no extra information is required, this configuration is ignored.
+			*/
+			unsigned int extra;
```
egecetin, line 218: Not important but if it is only related with NFLOG, the name of the parameter might include NFLOG. For example, like nflogFlags or nflogInfo etc

MrPeck: The idea was that sometime in the future, some device might require some extra information as well. So to make it as generic as possible so any other device could use a piece of extra information if needed, I called it extra instead of NflogGroup.

MrPeck: With this in mind, do you think I still should have the name as "nflogGroup"? Or should I keep it as "extra"?

MrPeck:
Hey! I don't understand what failed on the workflow and how I can fix it. Besides that, are there any additional steps I need to take to merge the changes?

egecetin:
@MrPeck It was because the workflow is blocked for first time contributors. Approved the workflow the run all checks. If everything passes it looks good to me. For merging @seladb will check the PR and decide to merge or request additional changes but since its holiday session it may take a while.

Pcap++/header/PcapLiveDevice.h
egecetin: Both options are ok for me

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -44,6 +44,7 @@
 #define LIBPCAP_OPEN_LIVE_TIMEOUT -1
 #endif
 
+static const char *NFLOG_IFACE = "nflog";
```
seladb, line 47: I'm not familiar with nflog... Is the interface name must be `nflog`? Or is it configurable?

Pcap++/header/PcapLiveDevice.h
seladb: I vote for being more specific and calling this parameter `nflogGroup`. In the parameter doxy we can mention this is only relevant for nflog interfaces

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -202,7 +203,14 @@ void PcapLiveDevice::statsThreadMain()
 pcap_t* PcapLiveDevice::doOpen(const DeviceConfiguration& config)
 {
 	char errbuf[PCAP_ERRBUF_SIZE] = {'\0'};
-	pcap_t* pcap = pcap_create(m_Name.c_str(), errbuf);
+	std::string device_name = m_Name;
+
+	if (device_name == NFLOG_IFACE)
+	{
+		device_name += ":" + std::to_string(config.extra & 0xffff);
```
seladb, line 210: What should be the default value of the group name if not provided by the user? is it possible to capture packets from `nflog:0`?

Pcap++/header/PcapLiveDevice.h
seladb, line 1: I have a more general question: is it possible to add tests for this feature? Our Linux CI currently uses docker containers so I guess this would be hard because nflog is driven by the kernel, am I wrong? 🤔

Pcap++/src/PcapLiveDevice.cpp
MrPeck: By default, if the NFLOG group is not defined, it is 0, so connecting to `nflog` is the same as `nflog:0`

Pcap++/header/PcapLiveDevice.h
MrPeck: Alright. Done!

Pcap++/src/PcapLiveDevice.cpp
MrPeck: Yes. The interface name must be `nflog:<NFLOG_GROUP>` or simply `nflog`, in which case NFLOG_GROUP is defaulted to 0.

MrPeck: Also, that's how `libpcap` interacts with this interface.

Pcap++/header/PcapLiveDevice.h
MrPeck: I think it is possible to add tests for this feature. But we would need a few things:

- Run with elevated privileges so we can open `nflog`

`nflog` events are generated by `netfilter` when an IPTables rule is triggered:

- Create an IPTables rule inside the container
- Trigger the rule

I am not familiar with the testing infrastructure in the project, do you think it would be possible to implement the aforementioned steps in the CI workflow?

seladb: PcapPlusPlus uses GitHub Actions as the CI platform. You can see the workflow file here: https://github.com/seladb/PcapPlusPlus/blob/master/.github/workflows/build_and_test.yml

Currently all Linux jobs run on a docker container, but maybe we can add a new job that runs in a VM and not in a docker container. This should be pretty easy, [here are some docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) that explain how to do it. When we create this job we can add the IP table rules etc. This should be the easy part...

The more complex part is how to run the test only in the new "nflog" job and not in the other tests. In order to do that you need to understand the PcapPlusPlus testing infrastructure. Each test has 1 or more "tags" that can be included or excluded in test runs. You can see these tags here: https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Pcap%2B%2BTest/main.cpp
What we can do is add a new "nflog" tag for this test, and run all other tests with `-x "nflog"`. Then in this job we can run it with `-t "nflog"`.

As I think more about it, it seems like a lot of work... maybe we can skip tests for now and keep it as a future enhancement. @egecetin @MrPeck what do you think?

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -306,8 +314,14 @@ bool PcapLiveDevice::open(const DeviceConfiguration& config)
...
 
 	m_PcapDescriptor = doOpen(config);
-	m_PcapSendDescriptor = doOpen(config);
-	if (m_PcapDescriptor == nullptr || m_PcapSendDescriptor == nullptr)
+
+	// It's not possible to have two open instances of the same NFLOG device:group
+	if (m_Name == NFLOG_IFACE)
+		m_PcapSendDescriptor = nullptr;
+	else
+		m_PcapSendDescriptor = doOpen(config);
```
seladb, line 322: nit: I'd add curly brackets here:

```cpp
if (m_Name == NFLOG_IFACE)
{
	m_PcapSendDescriptor = nullptr;
}
else
{
	m_PcapSendDescriptor = doOpen(config);
}
```

Pcap++/header/PcapLiveDevice.h
egecetin: I think separate PR for tests is a better idea. So, we can merge this PR to fix error. Testing the workflow locally is not easy, so it will probably take some time, you have to push and wait again, again and again...

Pcap++/src/PcapLiveDevice.cpp
MrPeck: Done!

Pcap++/header/PcapLiveDevice.h
MrPeck: I agree that keeping tests for the future would be a better idea. And I would gladly take that task upon myself whenever it comes up :)

seladb: Sounds good, thanks! 👍

seladb (approved):

seladb:
Thank you @MrPeck for adding this great feature to PcapPlusPlus, very much appreciated! 🙏 ❤️

Issue #1129, opened 2023-04-27
Heap-buffer-overflow(read) in TLVRecordReader::getNextTLVRecord() with two positions


bladchan:
**Describe the bug**
Bad .pcap files which can lead **TLVRecordReader::getNextTLVRecord** to heap-buffer-overflow (read) issues with two positions: one is in NflogTlv::getType(), the other is in NflogTlv::assign().

Pocs here:
 
[heap_buffer_overflow_pocs.zip](https://github.com/seladb/PcapPlusPlus/files/11339043/heap_buffer_overflow_pocs.zip)

**To Reproduce**
1. Build the whole library with **ASAN**;
2. Driver program (compile it with **ASAN** too):
```c++
// fuzz.cpp
#include<stdio.h>
#include <Packet.h>
#include <PcapFileDevice.h>

using namespace pcpp;

int
main(int argc, char* argv[])
{
  char *file = argv[1];
  // open a pcap file for reading
  pcpp::PcapFileReaderDevice reader(file);
  if (!reader.open())
  {
    printf("Error opening the pcap file\n");
    return 1;
  }

  // read the first (and only) packet from the file
  pcpp::RawPacket rawPacket;
  if (!reader.getNextPacket(rawPacket))
  {
    printf("Couldn't read the first packet in the file\n");
    return 1;
  }

  while (reader.getNextPacket(rawPacket)) {
      // parse the raw packet into a parsed packet
      pcpp::Packet parsedPacket(&rawPacket);
  }

  // close the file
  reader.close();

  return 0;
}
```
3. Run pocs:
```bash
$ ./fuzz ./poc_assign
$ ./fuzz ./poc_gettype
```

**Expected behavior**
The code snippet where the issue happened should avoid the out-bounds read operation.

**Environment (please complete the following information):**
 - System and Version : Ubuntu 20.04.1 + Clang 11
 - commit version: [fb3a560](https://github.com/seladb/PcapPlusPlus/commit/fb3a560482cbc3d5f12c2fa0b49e55cb3e0ab139)

**Additional context**
For `poc_assign`, ASAN says:
```text
$ ./fuzz poc_assign 
=================================================================
==3150137==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x606000004680 at pc 0x7f8fa8b80535 bp 0x7fffc291de50 sp 0x7fffc291de48
READ of size 1 at 0x606000004680 thread T0
[...]

PR #1175, opened 2023-08-04
Fixes poc_assign from #1129

sashashura:
~~I didn't find a better way than removing this odd hack. A zero byte in data is not necessarily an indicator of padding. I think it is better to remove it in order to fix the security issue - out of bounds read in poc_assign from #1129, and, if a poc is available for the padding issue in NFULA_PREFIX, investigate a better way to handle it.~~ Fixed next TLV calculation as in https://github.com/the-tcpdump-group/libpcap/blob/766b607d60d8038087b49fc4cf433dac3dcdb49c/pcap-util.c#L371-L374

### DIFF
--- a/Packet++/header/TLVData.h
+++ b/Packet++/header/TLVData.h
@@ -293,14 +293,6 @@ namespace pcpp
 			if (record.isNull())
 				return resRec;
 
-			// record pointer is out-bounds of the TLV records memory
-			if ((record.getRecordBasePtr() - tlvDataBasePtr) < 0)
-				return resRec;
-
-			// record pointer is out-bounds of the TLV records memory
-			if (record.getRecordBasePtr() - tlvDataBasePtr + (int)record.getTotalSizeAligned() >= (int)tlvDataLen)
-				return resRec;
-
 			resRec.assign(record.getRecordBasePtr() + record.getTotalSizeAligned());
 			if (resRec.getTotalSize() == 0)
 				resRec.assign(NULL);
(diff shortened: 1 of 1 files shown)


Packet++/header/NflogLayer.h
```diff
@@ -103,6 +103,18 @@ namespace pcpp
 		 */
 		size_t getTotalSize() const { return m_Data->recordLen; }
 
+		/**
+		 * @return The total size of the TLV record (in bytes) aligned by 4
+		 */
+		size_t getTotalSizeAligned() const
```
seladb, line 109: Why do we need an additional method `getTotalSizeAligned()` instead of modifying `getTotalSize()` in `NflogTlv` to be aligned by 4?

In the packet example [`nflogPacket.pcap`](https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Packet%2B%2BTest/PacketExamples/nflogPacket.pcap) we have such TLV record which length is 5 but the actual length is 8:

image

seladb: I see your point. However in PcapPlusPlus we usually don't have "gaps" in the data, meaning a TLV record has to start immediately after the previous one.

I looked at the code again, `NflogTlv` doesn't inherit from `TLVRecord` so it doesn't inherit its equality operator. We can implement an equality operator for `NflogTlv` that compares just the "non garbage" part instead of the whole TLV buffer. Also, I don't think `TLVRecordReader` actually uses the equality operator so it's kind of optional...

NFLOG is quite unique in its approach of allowing TLV records to be of unaligned size, but require that the next TLV record starts at an aligned position, so I think it's better to make a "local" fix that is only relevant for NFLOG.

seladb:
@sashashura not related to this PR, but since you're contributing a lot to this project, would you be interested in joining a Slack channel we have for this project where we can discuss issues, ideas, improvements etc.?
If you're interested please send me an email to pcapplusplus@gmail.com and I can send you an invite link

seladb:
@sashashura can we close this PR?

PR #1180, opened 2023-08-11
implement HttpResponseStatusCode toString & toInt

### LINKED ISSUE #1157: implement HttpResponseStatusCode to string
there is no `toString()` function for `HttpResponseStatusCode`

it would be nice to have:

```cpp
HttpResponseStatusCode code = HttpResponseStatusCode::Http200OK;
code.toString(); // "HTTP 200 OK"
```

We already have something like `StatusCodeEnumToString`, however it's only in cpp code and the messages are not clear enough, e.g. `Http200OK` -> `OK` but it's better to have `HTTP 200 OK`

tigercosmos:
fix #1157

- added new `HttpResponseStatusCode` enum class, and provided `toString` & `toInt` methods
- updated status code according to the [latest spec](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml)


Packet++/header/HttpLayer.h
```diff
@@ -207,201 +207,230 @@ namespace pcpp
...
 	public:
-		/**
-		 * Enum for HTTP response status codes
-		 */
-		enum HttpResponseStatusCode
+		enum Value: int
 		{
 			/** 100 Continue*/
-			Http100Continue,
+			Http100Continue = 100,
```
tigercosmos, line 221: let the enum item to have the same value as it's present code value

Packet++/header/HttpLayer.h
```diff
@@ -207,201 +207,230 @@ namespace pcpp
...
+			HttpStatusCodeUnknown = 999999
+		};
+
+		HttpResponseStatusCode() = default;
+		constexpr HttpResponseStatusCode(Value statusCode) : m_value(statusCode) { }
+		
+		constexpr operator Value() const { return m_value; }
+		explicit operator bool() const = delete;
+
+		struct HttpResponseStatusCodeHash {
```
tigercosmos, line 390: for `unordered_map`

Packet++/src/HttpLayer.cpp
```diff
@@ -961,11 +781,10 @@ HttpResponseFirstLine::HttpResponseFirstLine(HttpResponseLayer* httpResponse,  H
 	m_StatusCode = statusCode;
 	m_Version = version;
 
-	std::ostringstream statusCodeAsString;
```
tigercosmos, line 964: we don't even need `ostringstream` now

tigercosmos:
@seladb  please review, I didn't fix the CI, would like to get your agreement on the change first

egecetin:
When I was reviewing your PR, I noticed some of the HTTP status codes are missing (not in your PR, in library). Probably the list is a bit outdated. For example, 103 Early Hints (RFC 8297 from 2017), 421 Misredirected Request and maybe more. @seladb I think the list of all status codes should also be updated. Not checked all of that but you can find the latest list [here](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml)

tigercosmos:
@egecetin  already updated the code list

Packet++/src/HttpLayer.cpp
```diff
@@ -448,260 +448,76 @@ void HttpRequestFirstLine::setVersion(HttpVersion newVersion)
...
+    {HttpResponseStatusCode::Http503ServiceUnavailable, "Service Unavailable"},
+    {HttpResponseStatusCode::Http504GatewayTimeout, "Gateway Timeout"},
+    {HttpResponseStatusCode::Http505HTTPVersionNotSupported, "HTTP Version Not Supported"},
+    {HttpResponseStatusCode::Http506VariantAlsoNegotiates, "Variant Also Negotiates"},
+    {HttpResponseStatusCode::Http507InsufficientStorage, "Insufficient Storage"},
+    {HttpResponseStatusCode::Http508LoopDetected, "Loop Detected"},
+    {HttpResponseStatusCode::Http509BandwidthLimitExceeded, "Bandwidth Limit Exceeded"},
+    {HttpResponseStatusCode::Http510NotExtended, "Not Extended"},
+    {HttpResponseStatusCode::Http511NetworkAuthenticationRequired, "Network Authentication Required"},
+    {HttpResponseStatusCode::HttpStatus1xxCodeUnknown, "1XX Status Code Unknown"},
```
tigercosmos, line 513: added new `HttpStatus1xxCodeUnknown` etc. for better information
code not started from 1xx~5xx should be an error code

Packet++/header/HttpLayer.h
```diff
@@ -207,201 +207,217 @@ namespace pcpp
...
 			/** 417 Expectation Failed */
-			Http417ExpectationFailed,
-			/** 418 I'm a teapot */
-			Http418ImATeapot,
-			/** 419 Authentication Timeout */
-			Http419AuthenticationTimeout,
-			/** 420 (various messages) */
-			Http420,
+			Http417ExpectationFailed = 417,
+			/** 418 Unused **/
```
tigercosmos, line 308: don't add "unused" and "unassigned" code number into the list

Packet++/header/HttpLayer.h
```diff
@@ -207,201 +207,217 @@ namespace pcpp
...
 			/** 300 Multiple Choices */
-			Http300MultipleChoices,
+			Http300MultipleChoices = 300,
 			/** 301 Moved Permanently */
-			Http301MovedPermanently,
-			/** 302 (various messages) */
-			Http302,
+			Http301MovedPermanently = 301,
+			/** 302 Found */
+			Http302Found = 302,
```
seladb, line 258: previously `302` could also mean `Moved Temporarily`, do we want to keep supporting it?

Packet++/header/HttpLayer.h
seladb: `418` is `I'm a teapot` message: https://en.wikipedia.org/wiki/Hyper_Text_Coffee_Pot_Control_Protocol

Packet++/header/HttpLayer.h
```diff
@@ -207,201 +207,217 @@ namespace pcpp
...
-			Http417ExpectationFailed,
-			/** 418 I'm a teapot */
-			Http418ImATeapot,
-			/** 419 Authentication Timeout */
-			Http419AuthenticationTimeout,
-			/** 420 (various messages) */
-			Http420,
+			Http417ExpectationFailed = 417,
+			/** 418 Unused **/
+			/** 419-420 unassigned **/
```
seladb, line 309: `419` and `420` are sometimes used:
https://http.dev/419
https://www.holisticseo.digital/technical-seo/status-code/400s/420/

Packet++/header/HttpLayer.h
```diff
@@ -207,201 +207,217 @@ namespace pcpp
...
-			/** 495 Cert Error */
-			Http495CertError,
-			/** 496 No Cert */
-			Http496NoCert,
-			/** 497 HTTP to HTTPS */
-			Http497HTTPtoHTTPS,
-			/** 498 Token expired/invalid */
-			Http498TokenExpiredInvalid,
-			/** 499 (various messages) */
-			Http499,
```
seladb, line 361: Why did we remove these values?

Packet++/header/HttpLayer.h
```diff
@@ -207,201 +207,217 @@ namespace pcpp
...
-			/** 522 Connection timed out */
-			Http522ConnectionTimedOut,
-			/** 523 Proxy Declined Request */
-			Http523ProxyDeclinedRequest,
-			/** 524 A timeout occurred */
-			Http524aTimeoutOccurred,
-			/** 598 Network read timeout error */
-			Http598NetworkReadTimeoutError,
-			/** 599 Network connect timeout error */
-			Http599NetworkConnectTimeoutError,
```
seladb, line 399: Why did we remove these values?

Packet++/header/HttpLayer.h
```diff
@@ -207,201 +207,217 @@ namespace pcpp
...
-	 * Since it's not an "ordinary" HTTP field, it requires a special treatment and gets a class of it's own: HttpResponseFirstLine.
-	 * Unlike most L2-4 protocols, an HTTP response header can spread over more than 1 packet. PcapPlusPlus currently doesn't support a header
-	 * that is spread over more than 1 packet so in such cases: 1) only the first packet will be parsed as HttpResponseLayer (the other packets
-	 * won't be recognized as HttpResponseLayer) and 2) the HTTP header for the first packet won't be complete (as it continues in the following
-	 * packets), this why PcapPlusPlus can indicate that HTTP response header is complete or not (doesn't end with "\r\n\r\n" or "\n\n") using
-	 * HttpMessage#isHeaderComplete()
+	 * Class for the enum of HTTP response status codes
 	 */
-	class HttpResponseLayer : public HttpMessage
+	class HttpResponseStatusCode
```
seladb, line 215: Ideally we should strive to keep backward compatibility as much as possible, meaning users' code should break once they build their application with this new code.

In order to do this we can:
- Keep `HttpResponseStatusCode` a sub-class of `HttpResponseLayer`
- Keep all the enum values that were previously supported
- Maybe implement the cast to int operator so it's easier to do `static_cast<int>(value)`? 🤔

Packet++/header/HttpLayer.h
```diff
@@ -207,201 +207,217 @@ namespace pcpp
...
 	{
-		friend class HttpResponseFirstLine;
 	public:
-		/**
-		 * Enum for HTTP response status codes
-		 */
-		enum HttpResponseStatusCode
+		enum Value: int
```
seladb, line 218: shouldn't this be an `enum class`?

Packet++/header/HttpLayer.h
tigercosmos: @seladb  answer about all comments for adding and removing code in the list:
as @egecetin  mentioned, those values are outdated, according to https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml, so I just updated them according to the latest spec (2022)
The question is should we support the old code or not? IMO, we should avoid supporting old specs, since it takes more effort to maintain and increases the complexity of the code base.
We probably can make this version a breaking change version?

Packet++/header/HttpLayer.h
tigercosmos: as the doc says "an unofficial client error that is Laravel-specific", that's why not listed in the list.
we can figure out how to support user-defined or unofficial HTTP code, but maybe not in this PR?

Packet++/header/HttpLayer.h
tigercosmos: if we make it `enum class`, then we will need to call `HttpResponseStatusCode::Value::Http417ExpectationFailed`, I think it's okay to be `enum`, since the scope of the enum is already restricted in `HttpResponseStatusCode`

Packet++/header/HttpLayer.h
tigercosmos: answered above, let's discussed in the above thread

Packet++/header/HttpLayer.h
tigercosmos: answered above, let's discuss in the above thread

Packet++/header/HttpLayer.h
tigercosmos: answered above, let's discussed in the above thread

Packet++/header/HttpLayer.h
tigercosmos: 1. okay
2. okay, but what is the code meaning that has changed in the new spec? 
3. it's already supported, currently, it's fine to call `static_cast<int>(HttpResponseStatusCode::Http417ExpectationFailed)` (get `417`)

Packet++/header/HttpLayer.h
egecetin: I think the unofficial ones should also be supported too as @seladb mentioned because some of them widely used. @tigercosmos so there is no need removing old codes only adding missing ones and apply changes enough. But the only problem what will we do when there is a conflict for a status code message (if there are some codes have this situation). Since they are unofficial, different apps can use same code for different messages. What do you think about this @seladb? Should all messages be displayed somehow (maybe comma separated etc), display only most popular one or display a generic message like "various messages" or ...?

seladb: The essence of PcapPlusPlus is parsing packets. People use it to parse different kinds of packets, some of which are very old and contain outdated data, such as old HTTP messages. That's why I think we should support outdated/unofficial HTTP data.
It's true that applications can use HTTP status codes for any purpose, even if that purpose doesn't meet the official specifications. However, for most status codes there is only one usage (classic examples are `200 OK`, `404 Not Found`, `401 Unauthorized`, `500 Internal Server Error`). If there are cases where there is more than one usage, we can probably say "<Official usage> or other". Please let me know what you think

Packet++/header/HttpLayer.h
seladb: > 2\. okay, but what if the code meaning that has changed in the new spec?

Let's discuss specific use-cases, hopefully there aren't too many...



> 3\. it's already supported, currently, it's fine to call `static_cast<int>(HttpResponseStatusCode::Http417ExpectationFailed)` (get `417`)

Awesome! 👍

Packet++/header/HttpLayer.h
tigercosmos: I agree we should support any status code.
actually, even for unknown status codes (`1xx`-`5xx`, exclusive other else, e.g. `999`), I think we should also support it, e.g. `599 My Customized`, because users may want to analyze their internal usage packets.

the current behavior for parsing a packet limits the possibility of the user's customized status code.
how about this, we can add a new `std::string m_StatusCodeCustomizedDefination` in `HttpResponseFirstLine`, to allow the user's customized code.
- if the code is `200`, and it's already in our `HttpResponseFirstLine`, and the definition in the packet is the same as our definition, so  `m_StatusCodeCustomizedDefination` is empty
- if the code is `599`, and the user defines it's `My Customized`, then `m_StatusCodeCustomizedDefination`  is `My Customized`.
- if the code is `302`, it's `Found` in the spec, but the user defines it as `Hello World`,  then `m_StatusCodeCustomizedDefination`  is `Hello World`. 

later on, whenever we need to print or return the definition of the status code, we first see if `m_StatusCodeCustomizedDefination` is defined; otherwise, we return the default definition.

Packet++/header/HttpLayer.h
tigercosmos: yeah, it's valid.
The `enum Value : int` syntax specifies that the enumeration constants should have an underlying type of int, and that's why `static<int>(statusCode)` works

Packet++/header/HttpLayer.h
seladb: `HttpResponseFirstLine` is probably not a great place to put this definition because it has a private constructor and cannot be modified by the user. Besides, it seems that customized codes should be part of `HttpResponseStatusCode`.

In order to keep this PR simple and focused, I'd suggest supporting the existing status codes and maybe add customized code later. Please let me know what you think

Packet++/header/HttpLayer.h
seladb: yes this is correct, however since we currently support it, I'd prefer we keep backward compatibility as much as possible. If someone uses these values we wouldn't want their code to break because of this change...

Packet++/header/HttpLayer.h
seladb: ok, I didn't know that. In that case it's ok

Packet++/header/HttpLayer.h
tigercosmos: 1. done
2. done
3. done

Packet++/header/HttpLayer.h
tigercosmos: sure, let's do it in another PR

Packet++/header/HttpLayer.h
```diff
@@ -207,200 +207,264 @@ namespace pcpp
...
+	 * won't be recognized as HttpResponseLayer) and 2) the HTTP header for the first packet won't be complete (as it continues in the following
+	 * packets), this why PcapPlusPlus can indicate that HTTP response header is complete or not (doesn't end with "\r\n\r\n" or "\n\n") using
+	 * HttpMessage#isHeaderComplete()
+	 */
+	class HttpResponseLayer : public HttpMessage
+	{
+		friend class HttpResponseFirstLine;
+	public:
+		// backward compatibility
+		using HttpResponseStatusCode = pcpp::HttpResponseStatusCode;
```
tigercosmos, line 467: i think it's better to put `HttpResponseStatusCode` outside, since it's big, and it's not a good idea to be a subclass.
however, here we still have backward compatibility

Packet++/header/HttpLayer.h
```diff
@@ -207,200 +207,280 @@ namespace pcpp
...
+			HttpStatus2xxCodeUnknown = 900002, // 2xx: Success - The action was successfully received, understood, and accepted
+			HttpStatus3xxCodeUnknown = 900003, // 3xx: Redirection - Further action must be taken in order to complete the request
+			HttpStatus4xxCodeUnknown = 900004, // 4xx: Client Error - The request contains bad syntax or cannot be fulfilled
+			HttpStatus5xxCodeUnknown = 900005, // 5xx: Server Error - The server failed to fulfill an apparently valid request
+			HttpStatusCodeError = 999999, // other arbitrary number
+		};
+
+		HttpResponseStatusCode() = default;
+		// cppcheck-suppress noExplicitConstructor
+		constexpr HttpResponseStatusCode(Value statusCode) : m_value(statusCode) { }
```
tigercosmos, line 418: skip cppcheck, this is by design to allow `HttpResponseStatusCode::Value` can implicitly cast to `HttpResponseStatusCode`

Packet++/header/HttpLayer.h
```diff
@@ -207,200 +207,280 @@ namespace pcpp
...
+				return std::hash<int>()(static_cast<int>(status));
+			}
 		};
 
+		/**
+		 * @brief A pointer to the first line instance for this message
+		 */
+		std::string toString() const {
+			return std::to_string(m_value);
+		}
```
seladb, line 438: I'm wondering if we also want to expose the code message and not only the value.
Currently this method will return the code as string, for example: `"200"` but there's no method to return `"OK"`.
If we expose a method to return the message, we can also support customer messages. For example, you can do:
```cpp
HttpResponseStatusCode statusCode1(HttpResponseStatusCode::Http200OK);
HttpResponseStatusCode statusCode2(HttpResponseStatusCode::Http200OK, "Custom message");

std::cout << statusCode1.getMessage();
// prints "OK"
std::cout << statusCode2.getMessage();
// prints "Custom message"
```

But maybe we can leave it to a separate PR?

Packet++/src/HttpLayer.cpp
```diff
@@ -890,30 +719,51 @@ void HttpResponseFirstLine::setVersion(HttpVersion newVersion)
...
-	const char* statusCodeData = data + 9;
+	std::string statusCodeDataString(data + 9, 3);
 
-	auto codeAsEnum = StatusCodeStringToEnumMap.find(std::string(statusCodeData, 3));
-	if (codeAsEnum == StatusCodeStringToEnumMap.end())
+	for(const auto& pair : statusCodeExplanationStringMap) {
+		if(int(pair.first) == std::stoi(statusCodeDataString)) {
+			return pair.first;
+		}
+	}
```
seladb, line 736: It's not efficient to scan the entire map to find the code, especially because it happens during packet parsing. We should make it more efficient. Maybe convert `statusCodeExplanationStringMap` to `std::unordered_map<int, std::string>`? 🤔

Packet++/src/HttpLayer.cpp
```diff
@@ -890,30 +719,51 @@ void HttpResponseFirstLine::setVersion(HttpVersion newVersion)
...
+	case '4':{
+		return HttpResponseStatusCode::HttpStatus4xxCodeUnknown;
+	}
+	case '5':{
+		return HttpResponseStatusCode::HttpStatus5xxCodeUnknown;
+	}
+	default:
 	{
-		return HttpResponseLayer::HttpStatusCodeUnknown;
+		return HttpResponseStatusCode::HttpStatusCodeError;
```
seladb, line 756: In all these cases it is better to keep `HttpStatusCodeUnknown` than `HttpStatusCodeError`

Packet++/header/HttpLayer.h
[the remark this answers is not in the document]
tigercosmos: yeah, I just searched that. seems that constexpr can do some computation during compilation time, so we don't need to construct `HttpResponseStatusCode` during runtime for some statement like ` if(status ==  HttpResponseStatusCode::Http302)`

Packet++/header/HttpLayer.h
```diff
@@ -207,200 +207,280 @@ namespace pcpp
...
+			Http599NetworkConnectTimeoutError = 599,
+
 			/** Unknown status code */
-			HttpStatusCodeUnknown
+			HttpStatus1xxCodeUnknown = 900001, // 1xx: Informational - Request received, continuing process
+			HttpStatus2xxCodeUnknown = 900002, // 2xx: Success - The action was successfully received, understood, and accepted
+			HttpStatus3xxCodeUnknown = 900003, // 3xx: Redirection - Further action must be taken in order to complete the request
+			HttpStatus4xxCodeUnknown = 900004, // 4xx: Client Error - The request contains bad syntax or cannot be fulfilled
+			HttpStatus5xxCodeUnknown = 900005, // 5xx: Server Error - The server failed to fulfill an apparently valid request
+			HttpStatusCodeError = 999999, // other arbitrary number
```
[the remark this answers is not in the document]
tigercosmos: maybe we can have both `HttpStatusCodeError` and `HttpStatusCodeUnknown`?
`HttpStatusCodeUnknown` for the initial value, and ``HttpStatusCodeError` for error?

Packet++/header/HttpLayer.h
tigercosmos: I like the idea. let's add this `getMessage` function in this PR

Packet++/src/HttpLayer.cpp
tigercosmos: let me think about a better way

Packet++/header/HttpLayer.h
```diff
@@ -207,200 +207,280 @@ namespace pcpp
...
+		 */
+		bool isUnsupportedCode() const {
+			return m_value > 599;
+		}
+
+		constexpr bool operator==(const HttpResponseStatusCode &other) const { return m_value == other.m_value; }
+		constexpr bool operator!=(const HttpResponseStatusCode &other) const { return m_value != other.m_value; }
+
+		constexpr bool operator==(const Value &otherValue) const { return m_value == otherValue; }
+		constexpr bool operator!=(const Value &otherValue) const { return m_value != otherValue; }
```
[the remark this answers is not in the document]
tigercosmos: yeah, I guess they are not needed.

Packet++/header/HttpLayer.h
```diff
@@ -207,200 +207,280 @@ namespace pcpp
...
+		constexpr HttpResponseStatusCode(Value statusCode) : m_value(statusCode) { }
+
+		constexpr operator Value() const { return m_value; }
+		explicit operator bool() const = delete;
+
+		/**
+	 	 * @struct HttpResponseStatusCodeHash
+		 * @brief The helper structure for hash HttpResponseStatusCode while using std::unordered_map
+		 */
+		struct HttpResponseStatusCodeHash {
```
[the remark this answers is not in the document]
tigercosmos: > Why is this needed for the unordered_map? Can't it hash by enum value?

unfortunately, it's needed, because we need to tell unordered_map how to hash `HttpResponseStatusCode` (even though `HttpResponseStatusCode` can implicitly cast to int)

Packet++/src/HttpLayer.cpp
tigercosmos: sure

Packet++/header/HttpLayer.h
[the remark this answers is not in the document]
seladb: oh ok, so let's keep it!

Packet++/header/HttpLayer.h
seladb: > > Why is this needed for the unordered_map? Can't it hash by enum value?
> 
> unfortunately, it's needed, because we need to tell unordered_map how to hash `HttpResponseStatusCode` (even though `HttpResponseStatusCode` can implicitly cast to int)

In that case, let's move the hash function to `HttpLayer.cpp`? I don't think it should be part of `HttpResponseStatusCode `

Packet++/header/HttpLayer.h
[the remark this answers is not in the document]
tigercosmos: I think we can distinguish for different cases:

- `HttpStatusCodeInitailzed`: initial state
- `HttpStatusCodeError`: when the user gives crazy numbers like 999?
- `HttpStatus4xxCodeUnknown`: 4xx not supported by us
- 
what do you think?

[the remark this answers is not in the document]
seladb: It seems like initial state and crazy numbers can both be `HttpStatusCodeUnknown`, no? 🤔

Packet++/header/HttpLayer.h
```diff
@@ -207,200 +207,280 @@ namespace pcpp
...
+			Http415UnsupportedMediaType = 415,
 			/** 416 Requested Range Not Satisfiable */
-			Http416RequestedRangeNotSatisfiable,
+			Http416RequestedRangeNotSatisfiable = 416,
 			/** 417 Expectation Failed */
-			Http417ExpectationFailed,
-			/** 418 I'm a teapot */
-			Http418ImATeapot,
+			Http417ExpectationFailed = 417,
+			/** 418 I'm a teapot **/
```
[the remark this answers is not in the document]
tigercosmos: done

Packet++/header/HttpLayer.h
```diff
@@ -207,200 +207,280 @@ namespace pcpp
...
+			HttpStatus5xxCodeUnknown = 900005, // 5xx: Server Error - The server failed to fulfill an apparently valid request
+			HttpStatusCodeError = 999999, // other arbitrary number
+		};
+
+		HttpResponseStatusCode() = default;
+		// cppcheck-suppress noExplicitConstructor
+		constexpr HttpResponseStatusCode(Value statusCode) : m_value(statusCode) { }
+
+		constexpr operator Value() const { return m_value; }
+		explicit operator bool() const = delete;
```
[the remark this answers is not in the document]
tigercosmos: added back

Packet++/header/HttpLayer.h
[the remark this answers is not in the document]
tigercosmos: changed

Packet++/header/HttpLayer.h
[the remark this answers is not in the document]
tigercosmos: actually, they are all not needed, since we have `operator Value()`

Packet++/src/HttpLayer.cpp
tigercosmos: add a `HttpResponseStatusCode` constructor from int, which requires a `unordered_map<int, HttpResponseStatusCode>` in cpp, but I think the performance gain is worth it.

tigercosmos:
@seladb  please take a look

Packet++/src/HttpLayer.cpp
```diff
@@ -448,259 +448,225 @@ void HttpRequestFirstLine::setVersion(HttpVersion newVersion)
...
-	{"599", HttpResponseLayer::HttpResponseStatusCode::Http599NetworkConnectTimeoutError }
+/**
+ * @struct HttpResponseStatusCodeHash
+ * @brief The helper structure for hash HttpResponseStatusCode while using std::unordered_map
+ */
+struct HttpResponseStatusCodeHash
+{
+	size_t operator()(const HttpResponseStatusCode& status) const
+	{
+		return std::hash<int>()(static_cast<int>(status));
```
seladb, line 575: I don't think we need to use `std::hash` here, it's enough to do:

```cpp
size_t operator()(const HttpResponseStatusCode& status) const
{
	return static_cast<int>(status);
}

File: Tests/Packet++Test/Tests/HttpTests.cpp [hunk 23]
```diff
@@ -339,7 +348,7 @@ PTF_TEST_CASE(HttpResponseLayerParsingTest)
 	pcpp::HttpResponseLayer* responseLayer = httpPacket.getLayerOfType<pcpp::HttpResponseLayer>();
 	PTF_ASSERT_NOT_NULL(responseLayer);
 
-	PTF_ASSERT_EQUAL(responseLayer->getFirstLine()->getStatusCode(), pcpp::HttpResponseLayer::Http200OK, enum);
+	PTF_ASSERT_EQUAL(responseLayer->getFirstLine()->getStatusCode(), pcpp::HttpResponseStatusCode::Http200OK, enum);
```
seladb, line 351: We don't have to change this line, it will work even if we don't change it.

Same as for the other changes below this line. I think we should preserve it as is to ensure backward compatibility

Packet++/src/HttpLayer.cpp
```diff
@@ -890,30 +847,24 @@ void HttpResponseFirstLine::setVersion(HttpVersion newVersion)
...
 
-	const char* statusCodeData = data + 9;
-
-	auto codeAsEnum = StatusCodeStringToEnumMap.find(std::string(statusCodeData, 3));
-	if (codeAsEnum == StatusCodeStringToEnumMap.end())
-	{
-		return HttpResponseLayer::HttpStatusCodeUnknown;
-	}
-	return codeAsEnum->second;
+	int statusCodeInt = std::stoi(std::string(data + 9, 3));
```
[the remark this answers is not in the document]
tigercosmos: sure, let me change alogorithm

Packet++/src/HttpLayer.cpp
```diff
@@ -828,19 +787,19 @@ std::string HttpResponseFirstLine::getStatusCodeString() const
...
-bool HttpResponseFirstLine::setStatusCode(HttpResponseLayer::HttpResponseStatusCode newStatusCode, std::string statusCodeString)
+bool HttpResponseFirstLine::setStatusCode(HttpResponseStatusCode newStatusCode, std::string statusCodeString)
 {
-	if (newStatusCode == HttpResponseLayer::HttpStatusCodeUnknown)
+	if (m_StatusCode.isUnsupportedCode())
 	{
-		PCPP_LOG_ERROR("Requested status code is HttpStatusCodeUnknown");
+		PCPP_LOG_ERROR("Requested status code is " + m_StatusCode.toString() + statusCodeExplanationStringMap.at(m_StatusCode));
 		return false;
 	}
```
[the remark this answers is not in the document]
tigercosmos: it's a mistake. thanks

Packet++/src/HttpLayer.cpp
tigercosmos: ok

Tests/Packet++Test/Tests/HttpTests.cpp
tigercosmos: previous Http200OK is a global enum in `HttpResponseLayer` namespace; however, unfortunately, now the enum type is actually `HttpResponseStatusCode::Value` (or `HttpResponseLayer::HttpResponseStatusCode::Value`, we cannot use the enum without `HttpResponseStatusCode::`.

it's okay to do the following:
- `pcpp::HttpResponseLayer::Http200OK`
- `pcpp::HttpResponseLayer::HttpResponseStatusCode::Http200OK`
- `pcpp::HttpResponseLayer::HttpResponseStatusCode::Value::Http200OK`

IMO, it's also clearer to use `pcpp::HttpResponseStatusCode::Http200OK` rather than `pcpp::HttpResponseLayer::Http200OK`. what do you think @seladb ?

tigercosmos: however, I can think of one way, to expose `HttpResponseStatusCode::Value` under `HttpResponseLayer` ...
I personally really don't like this ...

Packet++/src/HttpLayer.cpp
```diff
@@ -890,30 +847,32 @@ void HttpResponseFirstLine::setVersion(HttpVersion newVersion)
...
-		return HttpResponseLayer::HttpStatusCodeUnknown;
+		return HttpResponseStatusCode::HttpStatusCodeUnknown;
 	}
 
-	const char* statusCodeData = data + 9;
+	std::string codeString = std::string(data + 9, 3);
 
-	auto codeAsEnum = StatusCodeStringToEnumMap.find(std::string(statusCodeData, 3));
-	if (codeAsEnum == StatusCodeStringToEnumMap.end())
+	if(!codeString.empty() && std::find_if(codeString.begin(), codeString.end(), [](unsigned char c){ return !std::isdigit(c); }) == codeString.end())
```
seladb, line 860: This should be `||`: if the string is empty **OR** it doesn't contain a digit, return `HttpResponseStatusCode::HttpStatusCodeUnknown`

Packet++/src/HttpLayer.cpp
```diff
@@ -890,30 +847,32 @@ void HttpResponseFirstLine::setVersion(HttpVersion newVersion)
...
 	{
-		return HttpResponseLayer::HttpStatusCodeUnknown;
+		return HttpResponseStatusCode::HttpStatusCodeUnknown;
 	}
-	return codeAsEnum->second;
+
+	// cppcheck-suppress containerOutOfBounds
+	int statusCodeInt = codeString[0] * 100 + codeString[1] * 10 + codeString[2];
```
seladb, line 866: This code doesn't work, because `codeString[x]` is of type `char` so multiplying it by a number will give the ascii value x the number. For example: if `codeString[0] == '1'` then `codeString[0] * 100 == 4900` (because the ascii value of `'1'` is `49`).

Here it's actually safe to use `std::stoi(codeString)` because we know the string only contains digits. It will also help us get rid of `// cppcheck-suppress containerOutOfBounds`

We can just do:
```cpp
return HttpResponseStatusCode(std:stoi(codeString));
```

BTW, not sure if you saw but the tests fail in CI:
https://github.com/seladb/PcapPlusPlus/actions/runs/5963344911?pr=1180

tigercosmos: ok, I see your point, you want to check the string first and call `std:stoi`

Packet++/src/HttpLayer.cpp
tigercosmos: fixed.

Packet++/src/HttpLayer.cpp
tigercosmos: fixed. now, it check if it is a number first and then call `stoi`

Tests/Packet++Test/Tests/HttpTests.cpp
seladb: I generally agree with you, my only problem is that it breaks the backward compatibility that we tried to solve with this:
https://github.com/seladb/PcapPlusPlus/pull/1180/files#diff-30f6094c6b01047985499a354425a4f628eadc54a0b3b822c513ec95f72d26a6R485

Meaning, that if someone is already using `pcpp::HttpResponseLayer::Http200OK` their code breaks once they update to the latest code

Packet++/src/HttpLayer.cpp
```diff
@@ -857,14 +857,12 @@ HttpResponseStatusCode HttpResponseFirstLine::parseStatusCode(const char* data,
 
 	std::string codeString = std::string(data + 9, 3);
 
-	if(!codeString.empty() && std::find_if(codeString.begin(), codeString.end(), [](unsigned char c){ return !std::isdigit(c); }) == codeString.end())
+	if(!codeString.empty() || std::find_if(codeString.begin(), codeString.end(), [](unsigned char c){ return !std::isdigit(c); }) == codeString.end())
```
seladb, line 860: You should remove the `!` in `!codeString.empty()`

tigercosmos: actually, I found that the original version is correct ...
`!codeString.empty() && std::find_if(codeString.begin(), codeString.end(), [](unsigned char c){ return !std::isdigit(c); }) == codeString.end()`
or 
`codeString.empty() || std::find_if(codeString.begin(), codeString.end(), [](unsigned char c){ return !std::isdigit(c); }) != codeString.end()`

are the same

seladb: yes you're right, sorry about that...

I personally find using `||` a bit more clear, but maybe it's just me

Tests/Packet++Test/Tests/HttpTests.cpp
seladb: @tigercosmos is there anything we can do with that? ⬆️

seladb:
@tigercosmos is there anything we can do with this? https://github.com/seladb/PcapPlusPlus/pull/1180#discussion_r1305226652

Tests/Packet++Test/Tests/HttpTests.cpp
tigercosmos: @seladb 

we can still achieve backward compatibility by using the following code in `HttpResponseLayer` :
```
using Http100Continue = pcpp::HttpResponseStatusCode::Http100Continue;
using Http101SwitchingProtocols = pcpp::HttpResponseStatusCode::Http101SwitchingProtocols;|
```

It's possible, of course; however, the code will look very redundant and trivial.

the current latest version is v22.11, I think we can check in this patch in v23 (or apply the above code first and remove in the v23), so that people can have a common sense that users may need to migrate(update) the code themself due to the breaking change.

seladb: Yes, I guess you're right. Even though it breaks backward compatibility the fix is pretty simple, basically replace `pcpp::HttpResponseLayer::Http200OK` with `pcpp::HttpResponseStatusCode:Http200OK` or `pcpp::HttpResponseLayer::HttpResponseStatusCode:Http200OK`.

Let's keep it like this

seladb (approved):

seladb:
@tigercosmos thank you so much for your contribution to this project, much appreciated!! 🙏 🙏 

If you're interested maybe you can work on: https://github.com/seladb/PcapPlusPlus/pull/1180#discussion_r1296572283

tigercosmos:
@seladb  thank you for taking time reviewing! :)

PR #1182, opened 2023-08-16
update document to remove doxygen warning

tigercosmos:
mostly about removing "No return value", which is redundant when the function is void.


seladb (approved):

seladb:
@tigercosmos thank you so much for this fix! 🙏 

Just one small comment: when opening PRs, please open them from the `dev` branch so we can merge them to `dev` and not to `master`

PR #1185, opened 2023-08-21
create S7COMM protocol and tests

wivien19:
I did not add it to the readme, because I do not know exactly, where to put it.


Packet++/src/S7commLayer.cpp
```diff
@@ -0,0 +1,93 @@
...
+		s7commHdr->msg_type = msg_type;
+		s7commHdr->reserved = 0x0000;
+		s7commHdr->pdu_ref = htobe16(pdu_ref);
+		s7commHdr->param_length = htobe16(param_length);
+		s7commHdr->data_length = htobe16(data_length);
+		m_Protocol = S7COMM;
+	}
+
+    std::string S7commLayer::toString() const {
+        std::ostringstream msgTypeStream;
```
tigercosmos, line 30: can we have only one `ostringstream`, there is no need to have four here.

Packet++/header/ProtocolType.h
```diff
@@ -327,6 +327,11 @@ namespace pcpp
 	 */
 	const ProtocolType SLL2 = 0x8000000000000;
 
+	/**
+	 * S7COMM protocol
+	 */
+	const ProtocolType S7COMM = 0x9000000000000;
```
seladb, line 333: This is a bit array represented by a hex value. Each protocol show have only 1 bit set. So the next value after `0x8000000000000` is `0x10000000000000`

seladb:
@wivien19 I read a little bit about S7COMM (I didn't know this protocol before), mostly here: http://gmiru.com/article/s7comm/

Apparently this protocol has 3 parts: Header, Parameters and Data.
The protocol implementation in this PR only refers to the Header but doesn't consider the Parameters and the Data. Do you know if it is on purpose?

wivien19:
I have read the articles on this blog.
I haven not had that purpose yet, because I have not had a lot information of them. I think, with these (parameter and date) can be completed.
Is this header not enough, right?

seladb:
> I have read the articles on this blog. I haven not had that purpose yet, because I have not had a lot information of them. I think, with these (parameter and date) can be completed. Is this header not enough, right?

@wivien19 it looks like the Parameter section doesn't have a fixed structure, and so as the Data section. Maybe we can have methods to at least read them as byte arrays?

By the way, even the header structure can change per message:
> The header is 10-12 bytes long, the Acknowledgement messages contain two extra error code bytes

wivien19:
If I want optional parameters (error code), do I need to create another header that I can fit to?

Where should I read the Parameters and the Data values as a byte array? I get its size from fitting it to the header, then I know that it takes up so many bytes. Should I extract a value of variable size from the payload like this?

seladb:
> 

> If I want optional parameters (error code), do I need to create another header that I can fit to?

I'm not sure I understand your question - do you mean you'd like to extract the error code from the Parameter data? As far as I know, not all Parameter types have an error code. Some of them have, some don't...

> 
> Where should I read the Parameters and the Data values as a byte array? I get its size from fitting it to the header, then I know that it takes up so many bytes. Should I extract a value of variable size from the payload like this?

I think we can have a structure like this, please let me know what you think:
```cpp

// S7commLayer.h

class S7commLayer : public Layer
{
public:
	...
	...
	class S7CommParameter
	{
		friend class S7commLayer;
	public:
		S7CommParameter(
		uint8_t* getData() { return m_Data; }
		size_t getDataLength() const { return m_DataLen; }
	private:
		S7CommParameter(uint8_t* data, size_t dataLen) : m_Data(data), m_DataLen(dataLen) {}
		uint8_t* m_Data;
		size_t m_DataLen;
	};

	S7CommParameter* getParameter() const;
...
private:
	S7CommParameter* m_Parameter;
};

// S7commLayer.cpp

S7commLayer::getParameter() const
{
	// If m_Parameter is still nullptr - get the parameter data and index and create it
	// m_Parameter = new S7CommParameter(...);
	return m_Parameter;
}
```

wivien19:
image

I think about this error class and error code. These are optional in header and I cannot parse always. I would like to know, that I should create another header to this option or something else. I have no idea to  

> I think we can have a structure like this, please let me know what you think

I think, it seems to be ok.

seladb:
> image
> 
> I think about this error class and error code. These are optional in header and I cannot parse always. I would like to know, that I should create another header to this option or something else. I have no idea to
> 

My understanding based on this: http://gmiru.com/article/s7comm/#21-header is that only **Ack-Data** messages will have these extra bytes. In that case I think what we can do is create 2 methods inside `S7commLayer`:

```cpp
#pragma pack(push, 1)
typedef struct s7comm_ack_data_hdr : s7comm_hdr
{
	uint8_t errorClass;
	uint8_t errorCode;
}
#pragma pack(pop)

class S7commLayer : public Layer
{
public:
	s7comm_hdr* getS7commHeader() const { return (s7commhdr *)m_Data; } // this method already exists
	s7comm_ack_data_hdr *getS7commAckDataHeader() const; // this method will return `nullptr` if the message is not of type `Ack-Data`, otherwise it will return the struct
}
```

This solution might not be the cleanest, but I think it's good enough since there is only one message type with a different structure.

Please notice that you need to take the header length into account when implementing the `getHeaderLen()` method. It should include the total length of the Header, Parameter and Data

wivien19:
In this commit (c2f1f2c) I have some problem with get parameter values. I have implemented some of the functions, but it does not work good. 

Can you help me  what i am doing wrong?
I would like to get the value of the parameter and check whether it contains the correct value.

seladb:
> In this commit ([c2f1f2c](https://github.com/seladb/PcapPlusPlus/commit/c2f1f2c10112afe237785db7103f9bc6bd446370)) I have some problem with get parameter values. I have implemented some of the functions, but it does not work good.
> 
> Can you help me what i am doing wrong? I would like to get the value of the parameter and check whether it contains the correct value.

I think the `getParameter()` method should be a member of the `S7commLayer` class, and not of `S7CommParameter`. We should use it like this:
```cpp
S7commLayer* myLayer = ...;
S7CommParameter* param = myLayer->getParameter();
...
```

wivien19:
I moved the getParameter() method to the other class. It seems to me that I am returning a completely wrong value. How can such a sequence of bytes be returned? I could tell from its size that it was not good, but its value was empty. I tried to change the size, but the value of the data data member was still not correct.

Can you see where I went wrong?

Thank you in advance for your help.

seladb:
> I moved the getParameter() method to the other class. It seems to me that I am returning a completely wrong value. How can such a sequence of bytes be returned? I could tell from its size that it was not good, but its value was empty. I tried to change the size, but the value of the data data member was still not correct.
> 
> Can you see where I went wrong?
> 
> Thank you in advance for your help.

I'm not sure what you mean by wrong size and empty value. Can you please elaborate?
I think you should be able to debug you code, either using a debugger or by adding prints (`std::cout`) and check the value of variables...

wivien19:
So `s7commLayer->getParameter()->getData()` looks like empty. I have found 224 in `s7commLayer->getParameter()->getDataLength()`, which number is the sum of the parameter length and the data length, at the end of the packet. 

My question is  how can I get the bytes from this to print the paramerter data bytes? I do not understand, why the data 
is empty. I think, I do not parse in the right way, but I cannot solve this problem.

I would like to get the parameters byte array, but it has not a constant structure.


```
...
VrrpCreateAndEditTest              : PASSED                                                                            
CotpLayerTest                      : PASSED                                                                             
224                                                              
                                                                                                                                                                               
S7commLayerTest                    : PASSED
...  
```

seladb:
> So `s7commLayer->getParameter()->getData()` looks like empty. I have found 224 in `s7commLayer->getParameter()->getDataLength()`, which number is the sum of the parameter length and the data length, at the end of the packet.
> 
> My question is how can I get the bytes from this to print the paramerter data bytes? I do not understand, why the data is empty. I think, I do not parse in the right way, but I cannot solve this problem.
> 
> I would like to get the parameters byte array, but it has not a constant structure.

@wivien19 I pushed a commit to your branch with a fix, please take a look: https://github.com/seladb/PcapPlusPlus/pull/1185/commits/6502dcb7d313078d914fce7ceb76a4ca796e43f0

seladb:
@wivien19 can you please fix the failures in CI?
https://github.com/seladb/PcapPlusPlus/actions/runs/6486734315/job/17615529156?pr=1185
https://github.com/seladb/PcapPlusPlus/actions/runs/6486734315/job/17615528181?pr=1185

seladb:
@wivien19 is this PR ready for review?

wivien19:
> @wivien19 is this PR ready for review?

Yes

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+	  private:
+		S7CommParameter(uint8_t *data, size_t dataLen) : m_Data(data), m_DataLen(dataLen) {}
+		uint8_t *m_Data;
+		size_t m_DataLen;
+	};
+	/**
+	 * @class S7commLayer
+	 * Represents a S7COMM (S7 Communication) protocol
+	 */
+	class S7commLayer : public Layer
```
seladb, line 76: nit: maybe change the class name to `S7CommLayer` (with captial-C?)

Packet++/header/S7commLayer.h
seladb, line 1: Maybe rename the file to `S7CommLayer.h` and `S7CommLayer.cpp`?

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+		/** parameter length */
+		uint16_t param_length;
+		/** data length */
+		uint16_t data_length;
+	} s7commhdr;
+#pragma pack(pop)
+
+/**
+ * @struct s7comm_ack_data_hdr
+ * Represents a S7COMM protocol with Ack-Data header
```
seladb, line 33: Update the comment: "Represents a S7COMM protocol header with Ack-Data header"

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+		/** error class */
+		uint8_t error_class;
+		/** error code */
+		uint8_t error_code;
+	};
+#pragma pack(pop)
+
+	/**
+	 * @class S7CommParameter
+	 * Represents a S7COMM (S7 Communication7) protocol Parameter
```
seladb, line 47: Should be "S7 Communication" (without the "7" in the end)

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+	 */
+	class S7CommParameter
+	{
+		friend class S7commLayer;
+
+	  public:
```
seladb, line 53: please remove the redundant spaces (leave only tabs)

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+		/**
+		 * @return The data of the Parameter
+		 */
+		uint8_t *getData() const { return m_Data; }
+		/**
+		 * @return The length of the Parameter data
+		 */
+		size_t getDataLength() const { return m_DataLen; }
+
+	  private:
```
seladb, line 67: please remove the redundant spaces (leave only tabs)

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+			if (getS7commHeader()->msg_type == 0x03)
+			{
+				return (s7comm_ack_data_hdr *)m_Data;
+			}
+			return nullptr;
+		}
+
+		size_t getS7commHeaderLength() const;
+
+		S7CommParameter *m_Parameter = nullptr;
```
seladb, line 230: the `= nullptr` is not needed I think 🤔

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+		/**
+		 * A constructor that allocates a new S7comm header
+		 * @param[in] msg_type The general type of the message
+		 * @param[in] pdu_ref Link responses to their requests
+		 * @param[in] param_length The length of the parameter field
+		 * @param[in] data_length The length of the data field
+		 * @param[in] error_class The value of the error class
+		 * @param[in] error_code The value of the error code
+		 */
+		S7commLayer(uint8_t msg_type, uint16_t pdu_ref, uint16_t param_length, uint16_t data_length,
```
seladb, line 88: Let's update all argument names, both in this method and in the other methods of this class from `kebab_case` to `camelCase` (for example: `msgType` instead of `msg_type`)?

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+		 * @param[in] packet A pointer to the Packet instance where layer will be stored in
+		 */
+		S7commLayer(uint8_t *data, size_t dataLen, Layer *prevLayer, Packet *packet)
+			: Layer(data, dataLen, prevLayer, packet)
+		{
+			m_Protocol = S7COMM;
+			m_Parameter = nullptr;
+		}
+
+		virtual ~S7commLayer() { delete m_Parameter; }
```
seladb, line 105: We should only delete `m_Parameter` is it's not `nullptr`

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+
+		/**
+		 * @return S7comm message type
+		 */
+		uint8_t getMsgType() const;
+
+		/**
+		 * @return S7comm reserved
+		 */
+		uint16_t getReserved() const;
```
seladb, line 120: do we need to expose this field? 🤔

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+		/**
+		 * @return S7comm error code
+		 */
+		uint8_t getErrorCode() const;
+		/**
+		 * @return S7comm error class
+		 */
+		uint8_t getErrorClass() const;
+
+		const S7CommParameter *getParameter();
```
seladb, line 146: the signature should be:
```cpp
const S7CommParameter* getParameter() const;
```

Also - please add doxygen documentation?

Packet++/src/S7commLayer.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+	uint8_t S7commLayer::getProtocolId() const { return getS7commHeader()->protocol_id; }
+
+	uint8_t S7commLayer::getMsgType() const { return getS7commHeader()->msg_type; }
+
+	uint16_t S7commLayer::getReserved() const { return htobe16(getS7commHeader()->reserved); }
+
+	uint16_t S7commLayer::getParamLength() const { return htobe16(getS7commHeader()->param_length); }
+
+	uint16_t S7commLayer::getPduRef() const { return htobe16(getS7commHeader()->pdu_ref); }
```
seladb, line 85: This and the other getters should use `be16toh` - big-endian to host.
Only the setters should use `htobe16` - host to big-endian

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+		 */
+		void setErrorCode(uint8_t error_code) const;
+		/**
+		 * Set the value of the error class
+		 * @param[in] error_class The value of the error class
+		 */
+		void setErrorClass(uint8_t error_class) const;
+
+		/**
+		 * @return Size of @ref s7commhdr
```
seladb, line 184: we should update the doxygen documentation: it's not always the size of `s7commhdr`, it's the size of the S7 Comm layer

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+			return sizeof(s7commhdr) + getParamLength() + getDataLength();
+		}
+
+		/**
+		 * Does nothing for this layer
+		 */
+		void computeCalculateFields() override {}
+
+		/**
+		 * Does nothing for this layer
```
seladb, line 200: Let's mention that this is always the last layer?
`Does nothing for this layer (S7CommLayer is always last)`

Packet++/header/S7commLayer.h
```diff
@@ -0,0 +1,234 @@
...
+
+		/**
+		 * @return Size of @ref s7commhdr
+		 */
+		size_t getHeaderLen() const override
+		{
+			if (getS7commHeader()->msg_type == 0x03)
+				return sizeof(s7comm_ack_data_hdr) + getParamLength() + getDataLength();
+
+			return sizeof(s7commhdr) + getParamLength() + getDataLength();
```
seladb, line 191: Since S7 comm is the last layer, can't we do:
```cpp
size_t getHeaderLen() const { return m_DataLen; }
```

Packet++/src/S7commLayer.cpp
```diff
@@ -0,0 +1,131 @@
...
+			error =
+				", error class: " + std::to_string(getErrorClass()) + ", error code: " + std::to_string(getErrorCode());
+		}
+		str << "S7comm Layer, "
+			<< "msg_type: " << std::to_string(getMsgType()) << ", pdu_ref: " << std::to_string(getPduRef())
+			<< ", param_length: " << std::to_string(getParamLength())
+			<< ", data_length: " << std::to_string(getDataLength()) << error;
+
+		return str.str();
+	}
```
seladb, line 67: This string is a bit long... maybe it should be shorter, like:
```text
S7Comm Layer, Job Request
```

Packet++/header/S7commLayer.h
wivien19: I'm not sure it can be const. In the S7Comm.cpp file, I give m_Paramer a value.

What do you think?

seladb:
@wivien19 the Visual Studio build fails in CI: https://github.com/seladb/PcapPlusPlus/actions/runs/6558348298/job/17811739778?pr=1185

Packet++/header/S7commLayer.h
seladb: Ok sure 👍

Packet++/src/S7commLayer.cpp
seladb: What I meant is the string to reflect the message type which can be one of:
image

Tests/Packet++Test/main.cpp
seladb, line 1: Can you revert these changes?

The only line we need to add is:
```cpp
PTF_RUN_TEST(S7CommLayerTest, "s7comm");
```

Tests/Packet++Test/Tests/S7CommTests.cpp
```diff
@@ -0,0 +1,72 @@
...
+	PTF_ASSERT_TRUE(S7CommLayerTest.isPacketOfType(pcpp::S7COMM));
+	auto *S7CommLayer = S7CommLayerTest.getLayerOfType<pcpp::S7CommLayer>();
+	PTF_ASSERT_NOT_NULL(S7CommLayer);
+
+	PTF_ASSERT_EQUAL(S7CommLayer->getProtocolId(), 0x32);
+	PTF_ASSERT_EQUAL(S7CommLayer->getMsgType(), 0x07);
+	PTF_ASSERT_EQUAL(S7CommLayer->getPduRef(), 0xfd0b);
+	PTF_ASSERT_EQUAL(S7CommLayer->getParamLength(), 12);
+	PTF_ASSERT_EQUAL(S7CommLayer->getDataLength(), 212);
+	PTF_ASSERT_EQUAL(S7CommLayer->getHeaderLen(), 0xea);
```
seladb, line 24: Let's use a decimal value and not hex?

Tests/Packet++Test/Tests/S7CommTests.cpp
```diff
@@ -0,0 +1,72 @@
...
+	PTF_ASSERT_EQUAL(S7CommLayer->toString(), "S7Comm Layer, Job Request");
```
seladb, line 25: This packet is not of type `Job Request`, it's `Userdata`

Tests/Packet++Test/Tests/S7CommTests.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+	PTF_ASSERT_EQUAL(S7CommLayer->getParameter()->getDataLength(), 12);
+	uint8_t expectedParameterData[] = {0x00, 0x01, 0x12, 0x08, 0x12, 0x84, 0x01, 0x01, 0x00, 0x00, 0x00, 0x00};
+	PTF_ASSERT_BUF_COMPARE(S7CommLayer->getParameter()->getData(), expectedParameterData, 12);
+
+	pcpp::S7CommLayer newS7commPacket(0x09, 0xfd0c, 13, 213);
```
seladb, line 31: Is it possible to separate this to a different test `S7CommLayerCreationTest`?

We can rename this test to `S7CommLayerParsingTest`

Packet++/src/S7CommLayer.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include "TcpLayer.h"
+#include <iostream>
+
+#include <cstring>
+#include <sstream>
+
+namespace pcpp
+{
+
+	S7CommLayer::S7CommLayer(uint8_t msgType, uint16_t pduRef, uint16_t paramLength, uint16_t dataLength,
```
seladb, line 13: We don't set `m_Parameter` anywhere in this constructor. I think we can set is to `nullptr`

Packet++/src/S7CommLayer.cpp
```diff
@@ -0,0 +1,117 @@
...
+							 uint8_t errorClass, uint8_t errorCode)
+	{
+		size_t headerLen;
+		if (msgType == 0x03)
+		{
+			headerLen = sizeof(s7comm_ack_data_hdr);
+			m_DataLen = headerLen;
```
seladb, line 20: The data len is not correct I think, because the actual length should be: `sizeof(s7comm_ack_data_hdr) + paramLength + dataLength`. We should also set `m_Data` to this size

Packet++/src/S7CommLayer.cpp
```diff
@@ -0,0 +1,117 @@
...
+			ack_d->paramLength = htobe16(paramLength);
+			ack_d->dataLength = htobe16(dataLength);
+			ack_d->errorClass = errorClass;
+			ack_d->errorCode = errorCode;
+		}
+		else
+		{
+			headerLen = sizeof(s7commhdr);
+			m_DataLen = headerLen;
+			m_Data = new uint8_t[headerLen];
```
seladb, line 37: Ditto: the data len is not correct I think, because the actual length should be: `sizeof(s7commhdr) + paramLength + dataLength`. We should also set `m_Data` to this size

Tests/Packet++Test/Tests/S7CommTests.cpp
```diff
@@ -0,0 +1,72 @@
...
+
```
seladb, line 32: I'd add test for `newS7commPacket.getParamter()`

Packet++/src/S7CommLayer.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+	void S7CommLayer::setMsgType(uint8_t msgType) const { getS7commHeader()->msgType = msgType; }
+
+	uint8_t S7CommLayer::getErrorCode() const { return getS7commAckDataHeader()->errorCode; }
+
+	uint8_t S7CommLayer::getErrorClass() const { return getS7commAckDataHeader()->errorClass; }
+
+	void S7CommLayer::setParamLength(uint16_t paramLength) const
+	{
+		getS7commHeader()->paramLength = htobe16(paramLength);
```
seladb, line 86: If the user changes `paramLength`, we should also change the length of the parameter data, no?
Otherwise the packet will become invalid

The problem is what should happen if there is existing parameter data and the user changes its length (increases or decreases the length)?

I see 2 options:
1. Let the user set the parameter data, something like `s7CommLayer->setParameter(data, dataLen)`
2. Don't let the user set the parameter data, and remove this `setParamLength()` method

The second option is obviously easier than the first one, but I think both options are ok. Which one do you prefer?

Packet++/src/S7CommLayer.cpp
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	void S7CommLayer::setPduRef(uint16_t pduRef) const { getS7commHeader()->pduRef = htobe16(pduRef); }
+
+	void S7CommLayer::setDataLength(uint16_t dataLength) const { getS7commHeader()->dataLength = htobe16(dataLength); }
```
seladb, line 91: Same here: if data len is changed, the packet data should change. Maybe we should just remove it?

Tests/Packet++Test/PacketExamples/s7comm_error_code.dat
seladb, line 1: This is not an error code packet, we should rename to `s7comm_ack_data.dat`

Tests/Packet++Test/PacketExamples/s7comm_error_code.pcap
seladb, line 1: This is not an error code packet, we should rename to `s7comm_ack_data.pcap`

Tests/Packet++Test/Tests/S7CommTests.cpp
```diff
@@ -0,0 +1,72 @@
...
+	READ_FILE_AND_CREATE_PACKET(2, "PacketExamples/s7comm_error_code.dat");
+
+	pcpp::Packet S7CommLayerErrorTest(&rawPacket2);
+	PTF_ASSERT_TRUE(S7CommLayerErrorTest.isPacketOfType(pcpp::S7COMM));
+	auto *s7commErrorLayer = S7CommLayerErrorTest.getLayerOfType<pcpp::S7CommLayer>();
+	PTF_ASSERT_NOT_NULL(s7commErrorLayer);
+	PTF_ASSERT_EQUAL(s7commErrorLayer->getProtocolId(), 0x32);
+	PTF_ASSERT_EQUAL(s7commErrorLayer->getMsgType(), 0x03);
+	PTF_ASSERT_EQUAL(s7commErrorLayer->getPduRef(), 0x0000);
```
seladb, line 56: nit: we can just use `0` instead of `0x0000`

Packet++/src/S7commLayer.cpp
seladb: @wivien19 I think you misunderstood what I meant. I think the at string should represent the message type. For example:
- If the message is of type `0x01 Job Request` the string should be: `S7Comm Layer, Job Request`
- If the message is of type `0x02 Ack` the string should be: `S7Comm Layer, Ack`
- If the message is of type `0x03 Ack-Data` the string should be: `S7Comm Layer, Ack-Data`
- If the message is of type `0x07 Userdata` the string should be: `S7Comm Layer, Userdata`
- Else the string could be: `S7Comm Layer, Unknown message`

Packet++/src/S7CommLayer.cpp
seladb: @wivien19 you didn't fix this one

Packet++/src/S7CommLayer.cpp
seladb: @wivien19 you didn't fix this one

Tests/Packet++Test/Tests/S7CommTests.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+	PTF_ASSERT_EQUAL(s7commErrorLayer->toString(), "S7Comm Layer");
+
+	s7commErrorLayer->setErrorCode(6);
+	s7commErrorLayer->setErrorClass(7);
+	PTF_ASSERT_EQUAL(s7commErrorLayer->getErrorClass(), 7);
+	PTF_ASSERT_EQUAL(s7commErrorLayer->getErrorCode(), 6);
+	PTF_ASSERT_EQUAL(s7commErrorLayer->getParameter()->getDataLength(), 2);
+	uint8_t expectedErrorParameterData[] = {4, 1};
+	PTF_ASSERT_BUF_COMPARE(s7commErrorLayer->getParameter()->getData(), expectedErrorParameterData, 2);
```
seladb, line 74: These lines aren't related to packet creation, they do packet parsing... they should probably be moved to `S7CommLayerParsingTest`

Tests/Packet++Test/Tests/S7CommTests.cpp
```diff
@@ -0,0 +1,75 @@
...
+	newS7commPacket.setPduRef(64778);
+	newS7commPacket.setDataLength(215);
+
+	PTF_ASSERT_EQUAL(newS7commPacket.getMsgType(), 6);
+	PTF_ASSERT_EQUAL(newS7commPacket.getPduRef(), 64778);
+	PTF_ASSERT_EQUAL(newS7commPacket.getDataLength(), 215);
+
+	READ_FILE_AND_CREATE_PACKET(2, "PacketExamples/s7comm_ack_data.dat");
+
+	pcpp::Packet S7CommLayerErrorTest(&rawPacket2);
```
seladb, line 53: nit: variable names should begin with lowecase and I'd rename it to: `s7CommLayer`

Tests/Packet++Test/Tests/S7CommTests.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "S7CommLayer.h"
+#include "SystemUtils.h"
+
+PTF_TEST_CASE(S7CommLayerParsingTest)
+{
+	timeval time;
+	gettimeofday(&time, nullptr);
+	READ_FILE_AND_CREATE_PACKET(1, "PacketExamples/S7comm.dat");
+
+	pcpp::Packet S7CommLayerTest(&rawPacket1);
```
seladb, line 14: nit: variable names should begin with lowecase and I'd rename it to: `s7CommLayer`

Packet++/src/S7CommLayer.cpp
seladb: @wivien19 you didn't address this comment

seladb:
@wivien19 here are the remaining unresolved comments, can you please address them so we can merge this PR:
- https://github.com/seladb/PcapPlusPlus/pull/1185#discussion_r1377084679
- https://github.com/seladb/PcapPlusPlus/pull/1185#discussion_r1377086697
- https://github.com/seladb/PcapPlusPlus/pull/1185#discussion_r1377086927

wivien19:
> @wivien19 here are the remaining unresolved comments, can you please address them so we can merge this PR:
> 
> * [create S7COMM protocol and tests #1185 (comment)](https://github.com/seladb/PcapPlusPlus/pull/1185#discussion_r1377084679)
> * [create S7COMM protocol and tests #1185 (comment)](https://github.com/seladb/PcapPlusPlus/pull/1185#discussion_r1377086697)
> * [create S7COMM protocol and tests #1185 (comment)](https://github.com/seladb/PcapPlusPlus/pull/1185#discussion_r1377086927)

Sorry, I overlooked them. It was my mistake.

seladb (approved):

seladb:
Thank you @wivien19 for working on this, much appreciated! 🙏 🙏

PR #1186, opened 2023-08-22
Compatibility  with DPDK 22.11, 23.03 and 23.07

SDMProjects:
Hi,
I use an Intel E810 NIC with the ice driver. I also use DPDK 23.03 or higher..

I have adapted PcapPlusPlus to be compatible with DPDK versions 22.11, 23.03 and  23.07.

I've attached my build script to show the options I use

Sincerelu yours
[PcapPlusPlus_build_script.txt](https://github.com/seladb/PcapPlusPlus/files/12410223/PcapPlusPlus_build_script.txt)

### DIFF
--- a/Examples/CMakeLists.txt
+++ b/Examples/CMakeLists.txt
@@ -44,7 +44,10 @@ endif()
 if(PCAPPP_USE_DPDK)
   add_subdirectory(DpdkExample-FilterTraffic)
   add_subdirectory(DpdkBridge)
-  add_subdirectory(KniPong)
+
+  if(DPDK_VERSION VERSION_LESS "22.11")
+    add_subdirectory(KniPong)
+  endif()
 endif()
 
 if(PCAPPP_USE_PF_RING)
--- a/Examples/CMakeLists.txt
+++ b/Examples/CMakeLists.txt
@@ -44,10 +44,7 @@ endif()
 if(PCAPPP_USE_DPDK)
   add_subdirectory(DpdkExample-FilterTraffic)
   add_subdirectory(DpdkBridge)
-
-  if(DPDK_VERSION VERSION_LESS "23.11")
-    add_subdirectory(KniPong)
-  endif()
+  add_subdirectory(KniPong)
 endif()
 
 if(PCAPPP_USE_PF_RING)
--- a/Examples/CMakeLists.txt
+++ b/Examples/CMakeLists.txt
@@ -45,7 +45,7 @@ if(PCAPPP_USE_DPDK)
   add_subdirectory(DpdkExample-FilterTraffic)
   add_subdirectory(DpdkBridge)
 
-  if(DPDK_VERSION VERSION_LESS "22.11")
+  if(DPDK_VERSION VERSION_LESS "23.11")
     add_subdirectory(KniPong)
   endif()
 endif()
--- a/Pcap++/CMakeLists.txt
+++ b/Pcap++/CMakeLists.txt
@@ -1,9 +1,6 @@
-add_library(
-  Pcap++
+set(SOURCES
   $<$<BOOL:${PCAPPP_USE_DPDK}>:src/DpdkDevice.cpp>
   $<$<BOOL:${PCAPPP_USE_DPDK}>:src/DpdkDeviceList.cpp>
-#  $<$<BOOL:${PCAPPP_USE_DPDK}>:src/KniDevice.cpp>
-#  $<$<BOOL:${PCAPPP_USE_DPDK}>:src/KniDeviceList.cpp>
   $<$<BOOL:${LINUX}>:src/LinuxNicInformationSocket.cpp>
   $<$<BOOL:${PCAPPP_USE_DPDK}>:src/MBufRawPacket.cpp>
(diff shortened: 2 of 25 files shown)


Pcap++/CMakeLists.txt
```diff
@@ -2,8 +2,8 @@ add_library(
   Pcap++
   $<$<BOOL:${PCAPPP_USE_DPDK}>:src/DpdkDevice.cpp>
   $<$<BOOL:${PCAPPP_USE_DPDK}>:src/DpdkDeviceList.cpp>
-  $<$<BOOL:${PCAPPP_USE_DPDK}>:src/KniDevice.cpp>
-  $<$<BOOL:${PCAPPP_USE_DPDK}>:src/KniDeviceList.cpp>
+#  $<$<BOOL:${PCAPPP_USE_DPDK}>:src/KniDevice.cpp>
```
clementperon, line 5: why dropping these files ? at least it should be compatible with 22.11

Pcap++/src/DpdkDevice.cpp
```diff
@@ -24,6 +24,10 @@
 #include <stdint.h>
 #include <unistd.h>
 
+#if (RTE_VER_YEAR > 22) || (RTE_VER_YEAR == 22 && RTE_VER_MONTH >= 11)
+#include "dev_driver.h"
```
clementperon, line 28: why not include the one from ./lib/eal/include/dev_driver.h for newer DPDK ?

Pcap++/CMakeLists.txt
seladb: @SDMProjects yes, I agree with @clementperon - why do we need to remove KNI? 

According to this: https://doc.dpdk.org/guides/rel_notes/deprecation.html KNI will only be removed in DPDK 23.11 🤔

Pcap++/src/DpdkDevice.cpp
```diff
@@ -518,7 +526,12 @@ void DpdkDevice::getLinkStatus(LinkStatus& linkStatus) const
 	rte_eth_link_get((uint8_t) m_Id, &link);
 	linkStatus.linkUp = link.link_status;
 	linkStatus.linkSpeedMbps = (unsigned) link.link_speed;
+
+#if (RTE_VER_YEAR < 22) || (RTE_VER_YEAR == 22 && RTE_VER_MONTH < 11)
 	linkStatus.linkDuplex = (link.link_duplex == ETH_LINK_FULL_DUPLEX) ? LinkStatus::FULL_DUPLEX : LinkStatus::HALF_DUPLEX;
+#else
+	linkStatus.linkDuplex = (link.link_duplex == RTE_ETH_LINK_FULL_DUPLEX) ? LinkStatus::FULL_DUPLEX : LinkStatus::HALF_DUPLEX;
```
seladb, line 533: For all these consts that have changes in newer DPDK versions, we usually a "PcapPlusPlus" version of them once at the beginning of `DpdkDevice.cpp` which might change per version, like this: https://github.com/seladb/PcapPlusPlus/blob/master/Pcap%2B%2B/src/DpdkDevice.cpp#L52-L58

Then we don't have to do these `#if..#else` everywhere, and instead we just used our own const.

As an example: we define `DPDK_CONFIG_MQ_RSS` which can be `ETH_RSS` in older DPDK versions or `RTE_ETH_MQ_RX_RSS` in newer DPDK versions. In the code we use  `DPDK_CONFIG_MQ_RSS` everywhere

Examples/CMakeLists.txt
```diff
@@ -44,7 +44,10 @@ endif()
 if(PCAPPP_USE_DPDK)
   add_subdirectory(DpdkExample-FilterTraffic)
   add_subdirectory(DpdkBridge)
-  add_subdirectory(KniPong)
+
+  if(DPDK_VERSION VERSION_LESS "23.11")
```
clementperon, line 48: let's add this when DPDK 23.11 is released

Pcap++/CMakeLists.txt
```diff
@@ -21,6 +18,18 @@ add_library(
   # Force light pcapng to be link fully static
   $<TARGET_OBJECTS:light_pcapng>)
 
+if(DPDK_VERSION VERSION_LESS "23.11")
```
clementperon, line 21: I would say the same keep this MR small and we will add this when 23.11 is realeased

cmake/modules/FindDPDK.cmake
```diff
@@ -69,6 +69,27 @@ if(DPDK_FOUND)
     message("Version: ${DPDK_VERSION}")
     message("-----------")
   endif()
+
+  find_package_handle_standard_args(
```
clementperon, line 73: this code is common between pkg-config and a find search no? Why duplicating it ?

seladb:
@SDMProjects please notice CI still fails due to compilation issues:
https://github.com/seladb/PcapPlusPlus/actions/runs/6086589705/job/16532427707?pr=1186

seladb:
DPDK 22.11 support was added in this PR by @clementperon : https://github.com/seladb/PcapPlusPlus/pull/1222

We can close this PR now

Issue #247, opened 2019-09-30
Relicense PcapPlusPlus to MIT?


seladb:
As you all know I created PcapPlusPlus as an open source project, with a goal in mind that everyone will be able to use it for any purpose (personal, academic, commercial or other). I'm always keeping this goal in mind and trying to make sure people are not blocked using this software because of licensing issues.

The current license is [Unilicense](https://unlicense.org/). I'm not a lawyer and have very little understanding of open source licensing. I chose Unilicense because I was looking for the most permissive open source license I could find, and this seemed like a good option. I used this web-site which has an easy-to-understand comparison between the different licenses: https://choosealicense.com/licenses/

I recently came to know that Unilicense has some detractors. For example: it hasn't been fully approved by the OSI (Open Source Initiative). Also it is much less common and popular comparing to other open source licenses (like GPL, MIT, Apache, Mozilla, etc.), which might make lawyers less comfortable in approving software licensed that way.

There is a discussion going on in PcapPlusPlus Google group about this topic: https://groups.google.com/forum/#!topic/pcapplusplus-support/vR5csiyUY1Y

Gareth who started this thread suggested to dual license PcapPlusPlus, but I think it might make things more complex for the average user. So my suggestion is to relicense PcapPlusPlus to a more standard and lawyer-friendly open source license, such as [MIT](https://opensource.org/licenses/MIT). From some reading I've done (and again - I'm not a lawyer) it seems to be the most permissive open source license out of the popular ones.

As you can imagine I've never went through a process of relicensing a project and my knowledge is very limited so I decided to ask you - PcapPlusPlus contributors and users. I'd really appreciate your thoughts and opinions about this.

If we decide to relicense we can probably do it from the next release of PcapPlusPlus as [this thread](https://opensource.stackexchange.com/questions/33/how-can-a-project-be-relicensed) suggests.

Please let me know what you think.

@echo-Mike @Dysl3xik @vicenterb @gx740 @Lapshin @solvingj @eteran @rpanah @luigino @krepver @gvanem @f-squirrel @bpagon13 @xloem @tomerb @sinall @ncrumbak @max197616 @christinaa @axasoft @MrSiz @Lasorda @AndreyBronin

luigino:
Looks like a good idea.

krepver:
Looks good to me

christinaa:
[...]

PR #1191, opened 2023-09-06
Updated unlicense link.

jpcofr:
This is just a small fix to the doc as the original Unlicense URL seems inactive. You may drop it as it does not add much.
Also I want to check if I am doing proper merge requests.


LICENSE
```diff
@@ -21,4 +21,4 @@ OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
 ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 OTHER DEALINGS IN THE SOFTWARE.
 
-For more information, please refer to <http://unlicense.org>
+For more information, please refer to <https://choosealicense.com/licenses/unlicense>
```
seladb, line 24: I'm not sure I want to touch the license file itself because I don't know what are the legal implications for it... Maybe just change from `http` to `https`, as it appears here? https://choosealicense.com/licenses/unlicense

jpcofr: Perhaps it is better to drop this fix as I am not familiar with licensing. I'll follow #247 instead and wait for the community.

seladb: We could still merge the changes in the README.md file...

clementperon: @seladb I would say as the original unlicense give all the credit/copyright to anybody. Basically anyone can create a new repo and relicense this with another license without any issue as you have all credit/copyright without the needs to notice someone.

clementperon (approved):

seladb (approved):

PR #1192, opened 2023-09-07
Skip network tests in CirrusCI Linux

seladb:
For some reason, CirrusCI Linux Zstd tests started to fail with `socket: Operation not permitted` error, probably related to something that changes in CirrusCI docker runners. Here is an example: https://cirrus-ci.com/task/5529878364160000?logs=test_pcapplusplus#L272

I opened a ticket for CirrusCI and waiting for their response: https://github.com/cirruslabs/cirrus-ci-docs/issues/1229

In the meantime, since we run network tests in many other CI tests, I think we can skip it.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -60,6 +60,7 @@ jobs:
             config-zstd: OFF
           - image: centos7
             python: python3.8
+            # TODO: TestLoggerMultiThread sometimes fail in CI so we exclude it for now, need to look into it
             test-flags: --pcap-test-args="-s -x TestLoggerMultiThread"
             config-zstd: OFF
           - image: fedora34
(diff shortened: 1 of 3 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -60,7 +60,7 @@ jobs:
             config-zstd: OFF
           - image: centos7
             python: python3.8
-            test-flags: --test-args="-s"
+            test-flags: --pcap-test-args="-s -x TestLoggerMultiThread"
```
seladb, line 63: This is not directly related, but CentOS sometimes fails in this test, so we can disable it for now until we have time to look more into it

clementperon: Could be nice to have a comment in yaml file directly to explain why this test is disabled.

clementperon (approved):

.github/workflows/build_and_test.yml
seladb: Added a comment here: https://github.com/seladb/PcapPlusPlus/pull/1192/commits/90ed7e130d6b2ef660efd09b502dfa797d87783e

PR #1194, opened 2023-09-07
timeval to timespec

Teles1:
Let me know if that works for you. I barely touched the code. I tried to keep it as close as possible to original. I could also make it be std::chrono but for that I'd need to know how you want that to be done.

### DIFF
--- a/Packet++/header/TcpReassembly.h
+++ b/Packet++/header/TcpReassembly.h
@@ -89,9 +89,9 @@ struct ConnectionData
 	/** A 4-byte hash key representing the connection */
 	uint32_t flowKey;
 	/** Start TimeStamp of the connection */
-	timeval startTime;
+	timespec startTime;
 	/** End TimeStamp of the connection */
-	timeval endTime;
+	timespec endTime;
 
 	/**
 	 * A c'tor for this struct that basically zeros all members
@@ -102,13 +102,13 @@ struct ConnectionData
 	 * Set startTime of Connection
 	 * @param[in] startTimeValue integer value
 	 */
-	void setStartTime(const timeval &startTimeValue) { startTime = startTimeValue; }
+	void setStartTime(const timespec &startTimeValue) { startTime = startTimeValue; }
 
 	/**
 	 * Set endTime of Connection
 	 * @param[in] endTimeValue integer value
 	 */
-	void setEndTime(const timeval &endTimeValue) { endTime = endTimeValue; }
+	void setEndTime(const timespec &endTimeValue) { endTime = endTimeValue; }
 };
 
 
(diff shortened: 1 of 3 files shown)


Teles1:
Not sure why the tests didn't run.

tigercosmos:
I also vote for `std::chrono`, should be easier to maintain.

Packet++/header/TcpReassembly.h
```diff
@@ -89,9 +89,9 @@ struct ConnectionData
 	/** A 4-byte hash key representing the connection */
 	uint32_t flowKey;
 	/** Start TimeStamp of the connection */
-	timeval startTime;
+	timespec startTime;
 	/** End TimeStamp of the connection */
-	timeval endTime;
+	timespec endTime;
```
seladb, line 94: I'm afraid this might be a breaking change for everyone using `ConnectionData`. Maybe we should keep `startTime` and `endTime` as `timeval` , add a deprecation warning to them (like we do [here](https://github.com/seladb/PcapPlusPlus/blob/22c6c85130ececa0cd4f9367332b8142687722c4/Common%2B%2B/header/IpAddress.h#L151)) and add 2 new properties:
```cpp
timespec startTimestamp;
timespec endTimestamp;
```

Packet++/header/TcpReassembly.h
```diff
@@ -102,13 +102,13 @@ struct ConnectionData
...
 	 */
-	void setStartTime(const timeval &startTimeValue) { startTime = startTimeValue; }
+	void setStartTime(const timespec &startTimeValue) { startTime = startTimeValue; }
 
 	/**
 	 * Set endTime of Connection
 	 * @param[in] endTimeValue integer value
 	 */
-	void setEndTime(const timeval &endTimeValue) { endTime = endTimeValue; }
+	void setEndTime(const timespec &endTimeValue) { endTime = endTimeValue; }
```
seladb, line 111: Same here: we can probably keep those with `timeval` (that will update both the old and new properties), and add an override for both methods that accept `timespec`

Packet++/header/TcpReassembly.h
```diff
@@ -166,18 +166,25 @@ class TcpStreamData
 	 */
 	const ConnectionData& getConnectionData() const { return m_Connection; }
 
-	/**
+	/** 
+	 * DEPRECATED
```
seladb, line 170: We can use the deprecation warning we do here: https://github.com/seladb/PcapPlusPlus/blob/22c6c85130ececa0cd4f9367332b8142687722c4/Common%2B%2B/header/IpAddress.h#L151

seladb:
> I also vote for `std::chrono`, should be easier to maintain.

Yeah, maybe we can use `std::chrono`. @Teles1 what do you think?

seladb:
@Teles1 what should we do with this PR?

seladb:
@Teles1 this PR is stale for a long time so I'm closing it

PR #1201, opened 2023-09-17
SMTP Protocol Support

egecetin:


Packet++/header/SmtpLayer.h
```diff
@@ -0,0 +1,340 @@
...
+			/// Close conversation
+			QUIT = ('Q') | ('U' << 8) | ('I' << 16) | ('T' << 24),
+			/// Receiver indication
+			RCPT = ('R') | ('C' << 8) | ('P' << 16) | ('T' << 24),
+			/// Abort transaction
+			RSET = ('R') | ('S' << 8) | ('E' << 16) | ('T' << 24),
+			/// Identify user
+			VRFY = ('V') | ('R' << 8) | ('F' << 16) | ('Y' << 24),
+			/// Start TLS handshake
+			STARTTLS = ('S') | ('T' << 8) | ('L' << 16) | ('S' << 24),
```
egecetin, line 94: There is bug for STARTTLS packets since the command is longer than four characters. I'll fix it.

egecetin:
@seladb I'll double check parsing with different SMTP packets but it's ready for review.

seladb:
> @seladb I'll double check parsing with different SMTP packets but it's ready for review.

Thank you @egecetin ! It might take me a few days to review but I'll get to it ASAP

Packet++/src/SmtpLayer.cpp
```diff
@@ -0,0 +1,228 @@
...
+			std::stringstream ss;
+			std::string field = getCommandOptionInternal();
+			for (size_t idx = 0; idx < field.size(); ++idx)
+			{
+				if (int(field.c_str()[idx]) < 127 && int(field.c_str()[idx]) > 31) // From SPACE to ~
+					ss << field.c_str()[idx];
+			}
+			return ss.str();
+		}
+		return getCommandOptionInternal();
```
seladb, line 40: I have a slightly simpler way to implement it:
```cpp
std::string SmtpRequestLayer::getCommandOption(bool removeEscapeCharacters) const
{
	std::string option = getCommandOptionInternal();
	if (!removeEscapeCharacters)
	{
		return option;
	}

	std::string optionWithEscapeChars;

	for (char ch : option)
	{
		if (ch < 127 && ch > 31)
		{
			optionWithEscapeChars.push_back(ch);
		}
	}
	
	return optionWithEscapeChars;
}
```

Packet++/header/SmtpLayer.h
```diff
@@ -0,0 +1,341 @@
...
+
+		/**
+		 * Get the command of request message
+		 * @return SmtpCommand Value of the command
+		 */
+		SmtpCommand getCommand() const;
+
+		/**
+		 * Get the command of request message as string
+		 * @return std::string Value of the command as string
```
seladb, line 154: In the `@return` doxygen document we usually don't put the type, so `std::string` is redundant.
The same for the rest of the methods in this file

Packet++/src/SmtpLayer.cpp
```diff
@@ -0,0 +1,228 @@
...
+			char val = (uint64_t(code) >> (8 * idx)) & UINT8_MAX;
+			if (val) // Dont push if it is a null character
+			{
+				oss << val;
+			}
+		}
+		return oss.str();
+	}
+
+	std::string SmtpRequestLayer::toString() const { return "SMTP Request: " + getCommandString(); }
```
seladb, line 114: As I mentioned in a previous comment, `getCommandString()` might return an empty string. We should take it into consideration here

Packet++/src/SmtpLayer.cpp
```diff
@@ -0,0 +1,228 @@
...
+			std::stringstream ss;
+			std::string field = getCommandOptionInternal();
+			for (size_t idx = 0; idx < field.size(); ++idx)
+			{
+				if (int(field.c_str()[idx]) < 127 && int(field.c_str()[idx]) > 31) // From SPACE to ~
+					ss << field.c_str()[idx];
+			}
+			return ss.str();
+		}
+		return getCommandOptionInternal();
```
seladb, line 146: Please see a previous comment with a suggestion for a simpler implementation

Packet++/src/SingleCommandTextProtocol.cpp
```diff
@@ -5,25 +5,31 @@
...
+
+		// Find Hyphen "-" if exists
+		uint8_t *pos = (uint8_t *)memchr(m_Data, ASCII_HYPHEN, maxLen);
 		if (pos)
 			return pos - m_Data;
 
-		// Find Hyphen "-" if exists
-		pos = (uint8_t *)memchr(m_Data, ASCII_HYPHEN, maxLen);
+		// Find <SP> if exists
+		pos = (uint8_t *)memchr(m_Data, ASCII_SPACE, maxLen);
```
seladb, line 32: Maybe we should check what comes sooner - `<SP>` or `-`?
Since we look at the first 9 bytes of the string there could theoretically be a case like this:
`CMD a-b-c`

(`CMD` is some command in some protocol)

In this case, the result will be incorrect because it'll consider the `-` as separator instead of `<SP>`

Packet++/src/TcpLayer.cpp
```diff
@@ -380,6 +381,10 @@ void TcpLayer::parseNextLayer()
 		m_NextLayer = SomeIpLayer::parseSomeIpLayer(payload, payloadLen, this, m_Packet);
 	else if (TpktLayer::isDataValid(payload, payloadLen) && TpktLayer::isTpktPort(portSrc, portDst))
 		m_NextLayer =  new TpktLayer(payload, payloadLen, this, m_Packet);
+	else if (SmtpLayer::isSmtpPort(portSrc) && SmtpLayer::isDataValid(payload, payloadLen))
+		m_NextLayer = new SmtpResponseLayer(payload, payloadLen, this, m_Packet);
+	else if (SmtpLayer::isSmtpPort(portDst) && SmtpLayer::isDataValid(payload, payloadLen))
+		m_NextLayer = new SmtpRequestLayer(payload, payloadLen, this, m_Packet);
```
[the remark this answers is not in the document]
egecetin: But response and request layers are different. Did you miss it, or did I not understand it completely?

Packet++/header/SmtpLayer.h
```diff
@@ -0,0 +1,341 @@
...
+		 * @param[in] code Value to set command
+		 * @return True if the operation is successful, false otherwise
+		 */
+		bool setCommand(SmtpCommand code);
+
+		/**
+		 * Get the command of request message
+		 * @return SmtpCommand Value of the command
+		 */
+		SmtpCommand getCommand() const;
```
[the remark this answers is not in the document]
egecetin: I'll check it

Packet++/src/SmtpLayer.cpp
egecetin: I'm not sure I understand your mention about previous comment. Which one?

Packet++/src/TcpLayer.cpp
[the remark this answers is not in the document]
seladb: oh you're right, my bad 🤦‍♂️

Packet++/header/SmtpLayer.h
```diff
@@ -0,0 +1,342 @@
...
+		/**
+		 * Set the command of request message
+		 * @param[in] code Value to set command
+		 * @return True if the operation is successful, false otherwise
+		 */
+		bool setCommand(SmtpCommand code);
+
+		/**
+		 * Get the command of request message
+		 * @return SmtpCommand Value of the command
```
seladb, line 149: We can remove the return value `SmtpCommand` from here too

Packet++/header/SmtpLayer.h
```diff
@@ -0,0 +1,342 @@
...
+		/**
+		 * Set the status code of response message
+		 * @param[in] code Value to set status code
+		 * @return True if the operation is successful, false otherwise
+		 */
+		bool setStatusCode(SmtpStatusCode code);
+
+		/**
+		 * Get the status code of response message
+		 * @return SmtpStatusCode Value of the status code
```
seladb, line 302: We can remove `SmtpStatusCode`

Packet++/src/SmtpLayer.cpp
seladb: I meant this comment: https://github.com/seladb/PcapPlusPlus/pull/1201#discussion_r1407354460

If the command is empty, `toString()` will be: `"SMTP Request: "` which is weird...

Packet++/header/SmtpLayer.h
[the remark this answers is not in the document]
egecetin: Looks like Wireshark uses session state to determine these packets ([here](https://github.com/wireshark/wireshark/blob/5877ba7fc5468613910c597da66dc44a3910b749/epan/dissectors/packet-smtp.c#L560)) so basically it knows the server asked username or password, so I don't know is there a way to understand these packets by only looking/parsing them without knowledge about the session.

Packet++/src/SmtpLayer.cpp
egecetin: If it is not weird enough it currently prints the username or password for these packets instead of empty string😁

image

I can first check the length of the command string to remove `:` but I should check why I design the algorithm to decide all of the data as command when there is no hyphen or space since I can't remember for now. If I can change to decide data as option it will return empty string so i can remove it🤔

seladb:
> A rare possibility occurs in SMTP example pcap which data packet bounces with ICMP and SMTP layer in one of these ICMP packets exactly end with \r\n so both port check and SingleCommandTextProtocol::isDataValid succeeds.

@egecetin I'm not sure I understand this comment... 🤔

egecetin:
@seladb Normally since the data packets are binary they are not end with CRLF sequence but since they are binary they might 😀. Since the data does not fit into single packet it can be send over multiple packets so there is a still chance for the smtp data packet might end with CRLF sequence. In addition to this ICMP protocol truncates and adds the failed packet in its payload so also there is a chance for even the original SMTP packet does not end with CRLF sequence because of the truncation the SMTP layer after ICMP might end with CRLF and can pass the checks during parsing. Wireshark's SMTP example pcap have an example for this situation. Even though SMTP packet ends with random bytes, one of the ICMP packet exactly cuts the packet from right place so binary data packet is assumed as SMTP command/reply. I'll send the screenshot of the packet later.

seladb:
> es not fit into single p

@egecetin I think I understand now. But if the TCP port is SMTP (25 or 587) and the packet ends with `\r\n` - isn't it a good enough indication that this is an SMTP packet? Are there additional checks we can do? 🤔

egecetin:
Probably nothing we can do instead of checking of the session (at least i dont know)

Packet++/src/SmtpLayer.cpp
```diff
@@ -0,0 +1,248 @@
...
+		size_t val = 0;
+		std::string field = getCommandString();
+
+		for (size_t idx = 0; idx < field.size(); ++idx)
+			val |= (field.c_str()[idx] << (idx * 8));
+
+		return static_cast<SmtpCommand>(val);
+	}
+
+	std::string SmtpRequestLayer::getCommandString() const { return getCommandInternal(); }
```
seladb, line 23: Why can't we call `getCommandAsString(getCommand())` here? 🤔 

It'll be more consistent because in username/password packets the return value will be "Unknown command" and not the username/password value

Packet++/src/SmtpLayer.cpp
```diff
@@ -0,0 +1,248 @@
...
+		}
+		return oss.str();
+	}
+
+	std::string SmtpRequestLayer::toString() const
+	{
+		std::string cmd = getCommandString();
+		if (cmd.length())
+			return "SMTP Request: " + cmd;
+		return "SMTP Request";
```
seladb, line 123: If we implement `getCommandString()` the way I suggested in the previous command we don't need this `if` statement and can always return `"SMTP Request: " + cmd;`

Also - let's be consistent with other protocols and format the string like this:
```
SMTP request layer, command: <CMD>
```

Packet++/src/SmtpLayer.cpp
```diff
@@ -0,0 +1,248 @@
...
+			return "Unknown status code";
+		}
+	}
+
+	std::string SmtpResponseLayer::toString() const
+	{
+		std::string option = getStatusCodeString();
+		if (option.length())
+			return "SMTP Response: " + getStatusCodeString();
+		return "SMTP Response";
```
seladb, line 245: Same here: let's format the string as other layers:
```
SMTP response layer, status code: <CODE>
```

Tests/Packet++Test/Tests/SmtpTests.cpp
```diff
@@ -0,0 +1,240 @@
...
+#include "EndianPortable.h"
+#include "EthLayer.h"
+#include "IPv4Layer.h"
+#include "IPv6Layer.h"
+#include "Packet.h"
+#include "SmtpLayer.h"
+#include "SystemUtils.h"
+#include "TcpLayer.h"
+
+PTF_TEST_CASE(SmtpParsingTests)
```
seladb, line 12: Let's also test `getHeaderLen()`?

Tests/Packet++Test/Tests/SmtpTests.cpp
```diff
@@ -0,0 +1,240 @@
...
+{
+	timeval time;
+	gettimeofday(&time, nullptr);
+
+	// Command
+	READ_FILE_AND_CREATE_PACKET(1, "PacketExamples/smtpCommand.dat");
+	pcpp::Packet smtpPacket1(&rawPacket1);
+	pcpp::SmtpRequestLayer *smtpLayer1 = smtpPacket1.getLayerOfType<pcpp::SmtpRequestLayer>();
```
seladb, line 20: nit: we can use `auto` instead of `pcpp::SmtpRequestLayer *` - here and in other places in this test

Tests/Packet++Test/Tests/SmtpTests.cpp
seladb, line 12: I think we can also add a username/password request packet and check that we see an unknown command

egecetin:
@seladb I saw your comments but still don't have time to fix them. I try to fix your comments and update the PR as soon as possible for protocol type changes

seladb:
> @seladb I saw your comments but still don't have time to fix them. I try to fix your comments and update the PR as soon as possible for protocol type changes. Sorry for the delay

Sure @egecetin no worries. We can keep the PR open until you're available to address the comments

Packet++/src/SmtpLayer.cpp
egecetin: getCommand() also depends getCommandString(). So it will be like an infinite loop 😁

egecetin: Actually this function for the get command as a string not for returning explanation. I mean it returns "EHLO" etc.

egecetin: I understand your command now, updated!

egecetin:
@seladb It's ready for review. You can check the PR when you have time

seladb:
> @seladb It's ready for review. You can check the PR when you have time

@egecetin you can update the branch, the FreeBSD issue should be fixed

image

Tests/Packet++Test/Tests/SmtpTests.cpp
```diff
@@ -0,0 +1,277 @@
...
+	auto *smtpLayer3 = smtpPacket3.getLayerOfType<pcpp::SmtpResponseLayer>();
+
+	PTF_ASSERT_NOT_NULL(smtpLayer3);
+	PTF_ASSERT_EQUAL(smtpLayer3->getHeaderLen(), 181);
+	PTF_ASSERT_EQUAL(smtpLayer3->getStatusCode(), pcpp::SmtpResponseLayer::SmtpStatusCode::SERVICE_READY, enumclass);
+	PTF_ASSERT_EQUAL(smtpLayer3->getStatusCodeString(), "220");
+	PTF_ASSERT_EQUAL(
+		smtpLayer3->getStatusOption(),
+		"xc90.websitewelcome.com ESMTP Exim 4.69 #1 Mon, 05 Oct 2009 01:05:54 -0500 220-We do not authorize "
+		"the use of this system to transport unsolicited, 220 and/or bulk e-mail.");
```
seladb, line 57: In case of multiline, do we want to remove the status code from all the lines?

Meaning, remove `220-` from the second and third lines?

seladb (approved):
Please take a look at one comment below. I'm ok to merge this PR without it

Tests/Packet++Test/Tests/SmtpTests.cpp
egecetin: I remember I also thought about it but to be honest I can't remember why I don't add any support for it. Let me check the code and protocol again before merging.

egecetin: @seladb added support with f2634f4. Both FTP and SMTP follow the same structure for multiline formatting. Can you check again? If its ok for you, we can merge it.

egecetin: @seladb Updated the code for support it in both SMTP and FTP (They use same structure for it) but codecov-action currently broken because of node version (Check https://github.com/codecov/codecov-action/issues/1230 and https://github.com/codecov/codecov-action/pull/1228) so workflow is failing for now.

Packet++/src/SingleCommandTextProtocol.cpp
```diff
@@ -143,8 +143,19 @@ namespace pcpp
 
 			// Remove XXX- and XXX<SP> since they are delimiters of the protocol where XXX is the usually status code
 			// Check RFC821 (SMTP) Section 3.3 and RFC959 (FTP) Section 4.2
-			std::regex reg(std::regex(getCommandInternal() + "-|" + getCommandInternal() + " "));
-			return std::regex_replace(option, reg, std::string());
+			std::vector<std::string> vDelim;
+			vDelim.push_back(getCommandInternal() + "-");
+			vDelim.push_back(getCommandInternal() + " ");
```
seladb, line 148: nit: we could run `getCommandInternal()` only once and create + init the vector in one line:
```cpp
auto option = getCommandInternal();
auto vDelim = std::vector<std::string> { option + " ", option + "-"}
```

seladb (approved):

Issue #1216, opened 2023-10-13
Issues with PIC code on latest release


NewbiZ:
Hello,

I have issues when linking the latest release static libraries (23.09) to a shared library. 
```
/usr/bin/ld: /usr/local/lib/libCommon++.a(SystemUtils.cpp.o): warning: relocation against `_ZN4pcpp23ApplicationEventHandler14handlerRoutineEi' in read-only section `.text'
/usr/bin/ld: /usr/local/lib/libPcap++.a(PcapFileDevice.cpp.o): relocation R_X86_64_PC32 against symbol `_ZN4pcpp22PcapNgFileWriterDevice11writePacketERKNS_9RawPacketE' can not be used when making a shared object; recompile with -fPIC
```

I did not spend too much time investigating yet, but it looks to me the new CMake build system does not generate PIC unless when building shared libraries.
In my case, I would like the pre-built static libraries of PcapPlusPlus to be PIC, as they used to be in 22.11.

I don't think there is any downside to _always_ generate PIC, and it makes life easier for users creating shared libraries.

seladb:
In general, PcapPlusPlus doesn't support shared libraries because some of its APIs don't work on some platforms.

Adding @clementperon who worked on the new CMake build system

clementperon:
@NewbiZ thanks for your report.

You could set this easily with Cmake just add
set(POSITION_INDEPENDENT_CODE TRUE)

And rebuild it, I will double check what are the impacts but I think you re right we should build static librairies as PIC.

clementperon:
@NewbiZ Thanks for the report I have opened a MR

NewbiZ:
@clementperon that looks great, thanks for making the change!

Do you want me to check the build before release?

clementperon:
Fixed with #1219

avallee-eisler:
Hello @clementperon,
Would you be able to create a release with this change included?

PR #1219, opened 2023-10-16
cmake: build static libraries with position independent code

clementperon:
From https://github.com/seladb/PcapPlusPlus/issues/1216 issue during the CMake migration we have forgotten to enable the PIC option that was previously enabled see:
 https://github.com/seladb/PcapPlusPlus/blob/v22.11/mk/PcapPlusPlus.mk.common#L16

This option allow the final user to freely build a position independent library or executable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,8 @@ option(PCAPPP_PACKAGE "Package Pcap++ could require a recent version of CMake" O
 set(CMAKE_CXX_STANDARD 11)
 # popen()/pclose() are not C++ standards
 set(CMAKE_CXX_EXTENSIONS ON)
+# Set Position Independent Code for static libraries
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -103,6 +103,8 @@ option(PCAPPP_PACKAGE "Package Pcap++ could require a recent version of CMake" O
 set(CMAKE_CXX_STANDARD 11)
 # popen()/pclose() are not C++ standards
 set(CMAKE_CXX_EXTENSIONS ON)
+# Set Position Independent Code for static libraries
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
seladb, line 107: Maybe we should have a default value (`ON` or `OFF`?), but allow the user to control it?

clementperon: I don't see a case where you want to disable it, and it was built with fPIC for a very long time and nobody complains.

It's easy to add an option but not sure it worth it.

seladb: ok 👍

seladb (approved):

PR #1226, opened 2023-10-29
Move 3rdParty GetOpt to Examples subdir and add weekly test build with vcpkg

clementperon:
Hi,

In order to regularly check that the vcpkg package manager is not broken add a CI workflow that test to build the Examples with PcapPlusPLus installed using vcpkg.

In order to build out-of-tree on Windows the getopt library needs to be moved to Examples directory

Also bump the checkout action to v4 everywhere.

DRAFT: Until vcpkg 23.09 is merged on vcpkg repo and move to official vcpkg repo instead of clementperon's fork


seladb:
@clementperon `Packet++Test` and `Pcap++Test` also use `getopt_long` and they don't fail with this change. Maybe `getopt` / `getop_long` support was added since VS 2019 so this dependency isn't needed anymore? 🤔

clementperon:
> @clementperon `Packet++Test` and `Pcap++Test` also use `getopt_long` and they don't fail with this change. Maybe `getopt` / `getop_long` support was added since VS 2019 so this dependency isn't needed anymore? 🤔



> @clementperon `Packet++Test` and `Pcap++Test` also use `getopt_long` and they don't fail with this change. Maybe `getopt` / `getop_long` support was added since VS 2019 so this dependency isn't needed anymore? 🤔

No it's because we 
  add_subdirectory(Examples)
before
  add_subdirectory(Tests)

So the GetOpt is present for Tests...

seladb:
> > @clementperon `Packet++Test` and `Pcap++Test` also use `getopt_long` and they don't fail with this change. Maybe `getopt` / `getop_long` support was added since VS 2019 so this dependency isn't needed anymore? 🤔
> 
> No it's because we add_subdirectory(Examples) before add_subdirectory(Tests)
> 
> So the GetOpt is present for Tests...

So if users don't want to build the examples they won't be to build the test either?

clementperon:
> > > @clementperon `Packet++Test` and `Pcap++Test` also use `getopt_long` and they don't fail with this change. Maybe `getopt` / `getop_long` support was added since VS 2019 so this dependency isn't needed anymore? 🤔
> > 
> > 
> > No it's because we add_subdirectory(Examples) before add_subdirectory(Tests)
> > So the GetOpt is present for Tests...
> 
> So if users don't want to build the examples they won't be to build the test either?

Yes so not acceptable as it.

seladb:
> > > > @clementperon `Packet++Test` and `Pcap++Test` also use `getopt_long` and they don't fail with this change. Maybe `getopt` / `getop_long` support was added since VS 2019 so this dependency isn't needed anymore? 🤔
> > > 
> > > 
> > > No it's because we add_subdirectory(Examples) before add_subdirectory(Tests)
> > > So the GetOpt is present for Tests...
> > 
> > 
> > So if users don't want to build the examples they won't be to build the test either?
> 
> Yes so not acceptable as it.

Yes, I think it's not a good idea to move GetOpt to `Examples` if the tests also use them...

PR #1227, opened 2023-11-01
Improve fuzzing coverage

sashashura:
Improves fuzzing coverage and fixes a memory leak in SnoopFileReader.


seladb:
@sashashura I'm contemplating if we should add all of the protocols to the fuzzing tests:
- It makes the fuzzing test longer and more complex
- When we add new protocols we need to remember to update the fuzzing tests

Is it worth adding this complexity? 🤔

sashashura:
> I'm contemplating if we should add all of the protocols to the fuzzing tests:

Do you mean you have an idea adding something? Could you elaborate on that? Or is it about the changes done in the pull request?

The current fuzzing covers all protocols already because it parses binary blobs. If a new protocol is added it is automatically discovered by fuzzer (New pcap test files are automatically added to the seed corpus by the build.sh script, but I don't know if oss-fuzz automatically picks them up or needs some manual bump to add them to the accumulated corpus files it maintains itself. Anyway at least eventually nothing stops the fuzzer from discovering the protocol after randomly mutating existing corpus files).

This pull request however adds new fuzzers for ngpcap and snoop file containers. Each file format is fuzzed by a separately built fuzzer that is run in parallel by oss-fuzz, so it doesn't slow the fuzzing.
Also the pull request adds a new fuzzer for pcap and ngpcap writers. They are added as separate fuzz targets exactly because otherwise they would significantly slow the existing fuzz targets. Since oss-fuzz runs all fuzz targets in parallel this is the way to write fuzzers.

The goal of the pull request is to improve the fuzzing coverage from the existing ~25% to 50-52%. More coverage - more fixed bugs.

seladb:
@sashashura I agree that we can add pcapng, but I doubt if we need to add snoop which is an old and rarely used format.
Just to make sure I understand: the only reason to add pcapng to the fuzz tests is to discover issues in pcapng format parsing, is that correct? Otherwise I'm not sure pcapng brings a lot more value than pcap. As you mentioned, the fuzzer creates binary blobs that mostly test packet parsing, the file format itself doesn't really matter.

Also, I wanted to ask about the main change in this PR which is in this file: `Tests/Fuzzers/FuzzTarget.cpp`. It goes over all the protocols that exist on the blob/packet generated by the fuzzer, and tries to extract some data for them. My question is: is this change really necessary?
Going back to what you mentioned - the fuzzer will eventually discover new protocols and find "dark corners" in existing protocols, so why do we need to add specific treatment for each protocol?

The main concerns that I have with this PR:
- Adding specific treatment for each protocol is something that contributors need to remember when adding a new protocol. A lot of contributors are new to PcapPlusPlus (some are even new to C++ programming) and aren't familiar with fuzzing tests. This change will make it harder for them to contributre...
- The `Tests/Fuzzers/FuzzTarget.cpp` file grows from 87 lines to 582 lines which makes it more complex and harder to understand. Is this change worth the complexity it brings?

sashashura:
> I agree that we can add pcapng, but I doubt if we need to add snoop which is an old and rarely used format.

If it is supported it should be fuzzed. Otherwise we should delete it.

> Just to make sure I understand: the only reason to add pcapng to the fuzz tests is to discover issues in pcapng format parsing, is that correct?

Technically yes, but with a caveat. In ideal world we would have a set of pcap files that covered all possible cases, so fuzzing with pcapng would be unnecessary overlap. My coverage tests show however that the seed for pcap and the seed for pcapng have non-overlapping parts.

But I hear you and made the changes to test only format parsing for .snoop files.

> Also, I wanted to ask about the main change in this PR which is in this file: Tests/Fuzzers/FuzzTarget.cpp. It goes over all the protocols that exist on the blob/packet generated by the fuzzer, and tries to extract some data for them. My question is: is this change really necessary?
Going back to what you mentioned - the fuzzer will eventually discover new protocols and find "dark corners" in existing protocols, so why do we need to add specific treatment for each protocol?

Yes, it is. It increases functions called by fuzzer from roughly 25% to 50%.
There is a small misunderstanding. Fuzzer may mutate files and eventually discover new code paths blocked by some `if` condition. But the code must be theoretically callable. Say, the old fuzzer calls `getNext`, it triggers other functions. The calls I added are not triggered neither by `getNext`, nor by `toString`. They must be called explicitly. It is not a specific treatment for each protocol. It is calling parsed packet functions that are not otherwise reachable by general parsing.

> Adding specific treatment for each protocol is something that contributors need to remember when adding a new protocol. A lot of contributors are new to PcapPlusPlus (some are even new to C++ programming) and aren't familiar with fuzzing tests. This change will make it harder for them to contributre...

I think it is misunderstanding that it brings contribution complexity. Fuzzing harness is essentially one more unit test. I have picked up many of the calls from existing tests. Expanding the fuzzing harness doesn't require anything neither from existing nor from new contributors. If a new Layer public method is added it doesn't require the contributor to add it to the harness. If the method signature changes (Side note, I think you don't allow changing of public API methods) existing unit tests will stop compiling and the contributor must update all places to make it compile anyway. If the function is deleted, the contributor is welcome to delete it from the fuzzer too. Most of the functions that are added are very simple as `getContent()` or `getContentLength()`.

> The Tests/Fuzzers/FuzzTarget.cpp file grows from 87 lines to 582 lines which makes it more complex and harder to understand. Is this change worth the complexity it brings?

I see your point and extracted the additional calls into a separate function to improve readability.

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -205,13 +205,13 @@ bool SnoopFileReaderDevice::getNextPacket(RawPacket& rawPacket)
...
-	char* packetData = new char[packetSize];
-	m_snoopFile.read(packetData, packetSize);
+	std::unique_ptr<char[]> packetData(new char[packetSize]);
+	m_snoopFile.read(packetData.get(), packetSize);
 	if(!m_snoopFile) {
 		return false;
 	}
 	timespec ts = { static_cast<time_t>(be32toh(snoop_packet_header.time_sec)), static_cast<long>(be32toh(snoop_packet_header.time_usec)) * 1000 };
-	if (!rawPacket.setRawData((const uint8_t*)packetData, packetSize, ts, static_cast<LinkLayerType>(m_PcapLinkLayerType)))
+	if (!rawPacket.setRawData((const uint8_t*)packetData.release(), packetSize, ts, static_cast<LinkLayerType>(m_PcapLinkLayerType)))
```
seladb, line 214: Maybe we should move this fix to a separate PR and dedicate this PR just to improve fuzzing coverage?

Tests/Fuzzers/FuzzTarget.cpp
```diff
@@ -1,87 +1,76 @@
...
 
-	written = fwrite(data, 1, size, fd);
-	if (static_cast<size_t>(written) != size)
+	if (dumpDataToPcapFile(data, size, tmpFile.c_str()) < 0)
 	{
-		std::cerr << "Error writing pcap file\n";
-		fclose(fd);
-		return -1;
+		std::cerr << "Can't Dump buffer to the '" << tmpFile << "' file!!!!\n";
+		return 1;
```
seladb, line 21: If I remember correctly the return value matters, but I don't remember how. Can you please explain when it's recommended to return `0` and when we should return non-zero value like `1`?

Tests/Fuzzers/FuzzTarget.cpp
```diff
@@ -1,87 +1,76 @@
...
-	pcpp::PcapFileReaderDevice reader(TMP_FILEPATH);
-	if (!reader.open())
+	if (auto ngReader = dynamic_cast<pcpp::PcapNgFileReaderDevice*>(reader.get()))
 	{
-		std::cerr << "Error opening the pcap file\n";
-		return 0;
+		std::cout << "OS is '" << ngReader->getOS() << "'; Hardware is '" << ngReader->getHardware() << "'"
+				  << "'; CaptureApplication is '" << ngReader->getCaptureApplication()
+				  << "'; CaptureFileComment is '" << ngReader->getCaptureFileComment()
+				  << "'" << std::endl;
```
seladb, line 43: As you probably know, pcapng parsing is done using `LightPcapNg` which is a 3rd party library PcapPlusPlus is using. My main concern is that these lines will uncover issues in `LightPcapNg` which we're not familiar with, and not regularly maintain. Please let me know what you think

Tests/Fuzzers/ReadParsedPacket.h
```diff
@@ -0,0 +1,522 @@
...
+#include <FtpLayer.h>
+#include <GreLayer.h>
+#include <GtpLayer.h>
+#include <SomeIpSdLayer.h>
+
+// Call some pcpp::Packet methods that are not invoked from general virtual methods
+// as `pcpp::Packet::toString` or `pcpp::Packet::computeCalculateFields` to trigger possible crashes.
+// The general rule is the functions do not modify the `parsedPacket`.
+// If you made changes to PcapPlusPlus and the code doesn't compile - fix the method call as any other unit test
+static void readParsedPacket(pcpp::Packet parsedPacket, pcpp::Layer* layer)
```
seladb, line 28: This file covers ~20 protocols out of ~45 supported in PcapPlusPlus. My questions are:
- How did you choose which protocols to include and which not to include?
- When we add a new protocol (we recently added [S7comm](https://github.com/seladb/PcapPlusPlus/pull/1185)), how should we, as maintainers / PR reviewers, guide contributors on whether to add their new protocol to the fuzzing tests or not? Or maybe contributors shouldn't worry about it and we (maintainers) will add it after the fact?
- If a contributor wants to add code to the fuzzing tests, how do they test fuzzing to make sure they don't break anything? Maybe we should add some documentation on how to contribute to this part of the code. So far we didn't need it because the fuzzing tests were simple and very static (we made very little changes in them over the past few years)

Tests/Fuzzers/default.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+close_fd_mask=3
```
seladb, line 2: this is probably related to oss-fuzz? 🤔 
What does it do?

Tests/Fuzzers/ossfuzz.sh
```diff
@@ -0,0 +1,34 @@
...
+    cp $(ldd $OUT/${fuzzer} | cut -d" " -f3) $OUT
+    cp Tests/Fuzzers/default.options $OUT/${fuzzer}.options
+done
+
+# Copy corpora
+find $SRC/ -iname "*.pcap"   | xargs zip $OUT/FuzzTarget_seed_corpus.zip
+find $SRC/ -iname "*.pcapng" | xargs zip $OUT/FuzzTargetNg_seed_corpus.zip
+find $SRC/ -iname "*.snoop"  | xargs zip $OUT/FuzzTargetSnoop_seed_corpus.zip
+find $SRC/ -iname "*.pcap"   | xargs zip $OUT/FuzzWriter_seed_corpus.zip
+find $SRC/ -iname "*.pcapng" | xargs zip $OUT/FuzzWriterNg_seed_corpus.zip
```
seladb, line 34: What does this do?

Tests/Fuzzers/ossfuzz.sh
seladb, line 1: Can you please elaborate on who will run this script? How does oss-fuzz know it should run it?

Tests/Fuzzers/FuzzTarget.cpp
```diff
@@ -1,87 +1,76 @@
...
+		tmpName = tmpnam (NULL);
 
-	fd = fopen(TMP_FILEPATH, "wb");
-	if (fd == NULL)
-	{
-		std::cerr << "Error opening pcap file for writing\n";
-		return -1;
-	}
+	if (tmpFile.empty())
+		tmpFile = tmpName + FILE_EXT;
```
seladb, line 16: Why are these lines needed? 🤔

Tests/Fuzzers/ossfuzz.sh
sashashura: It zips all files it can find in the source directories with the specified extension into a [seed archive for oss-fuzz](https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/#seed-corpus).

seladb: How does oss-fuzz know these zip files have the seed corpus?

Tests/Fuzzers/ossfuzz.sh
sashashura: OSS-FUZZ will do. Once this pull request is merged I'll change https://github.com/google/oss-fuzz/pull/11198 from draft.
Adding new fuzzers or making other changes as zipping seeds now requires two pull requests - one in pcapplusplus repository and the other in oss-fuzz. It is also a recommended practice by oss-fuzz.

Tests/Fuzzers/default.options
sashashura: Yes, this is a [libFuzzer flag](https://llvm.org/docs/LibFuzzer.html). It says we are not interested errors and warnings to stdout/stderr. Nothing changes here, the file is just moved from https://github.com/google/oss-fuzz/pull/11198

Tests/Fuzzers/ReadParsedPacket.h
sashashura: The current coverage report for pcapplusplus is [here](https://storage.googleapis.com/oss-fuzz-coverage/pcapplusplus/reports/20231106/linux/report.html). It is also possible to generate it [locally according to instructions](https://google.github.io/oss-fuzz/advanced-topics/code-coverage/) if, say, you have local changes.

I was looking into the report seeking functions that are marked as uncovered. For example
`getEventgroupId` - https://storage.googleapis.com/oss-fuzz-coverage/pcapplusplus/reports/20231106/linux/src/PcapPlusPlus/Packet++/src/SomeIpSdLayer.cpp.html#L378
Then I was looking why it is not called. What are the other functions that call it?
As you can see [here for example](https://github.com/search?q=repo%3Aseladb%2FPcapPlusPlus%20getEventgroupId&type=code) the function is called only explicitly from a unit test.

> When we add a new protocol (we recently added https://github.com/seladb/PcapPlusPlus/pull/1185), how should we, as maintainers / PR reviewers, guide contributors on whether to add their new protocol to the fuzzing tests or not? Or maybe contributors shouldn't worry about it and we (maintainers) will add it after the fact?

New protocol contributors should not worry about adding anything to the harness. After some time a new coverage check can be done and the functions added.

> If a contributor wants to add code to the fuzzing tests, how do they test fuzzing to make sure they don't break anything?

Fuzzers are built and run in CI. Oss-fuzz also has [instructions](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#testing-locally) how to build and test locally.

Tests/Fuzzers/FuzzTarget.cpp
sashashura: Yes, I looked up and it seems the source code is not there anymore, i.e. PcapPlusPlus has it's own fork of the pcapng parser.
Here is the dilemma: we want to assure our user the library is safe. Users can open pcapng with the library. If there is a bug in the third party library our options are:
a. Fix it in the third party library (not possible in this case?)
b. Fix it in the local fork.
c. Disable pcapng support by default (make a setting that the user has explicitly enable) and bug a big banner that the feature is insecure and experimental.
d. Switch to another library for pcapng handling.

Tests/Fuzzers/FuzzTarget.cpp
sashashura: Good point, I checked the documentation and it looks like anything other than 0 and -1 is reserved. https://llvm.org/docs/LibFuzzer.html#id23 https://llvm.org/docs/LibFuzzer.html#id41
I have pushed a fix.

The rule of thumb I used: if situation is expected (`getNextPackets` fails, but doesn't crash) - return 0. In abnormal situation - return an error code.

Pcap++/src/PcapFileDevice.cpp
sashashura: Since it was immediately breaking a new fuzzer I bundled it here, but as a separate commit. But if you want I can extract it.

Tests/Fuzzers/ossfuzz.sh
sashashura: It is a convention it uses: fuzzer name + _seed_corpus.zip in the same directory.

Tests/Fuzzers/FuzzTarget.cpp
sashashura: The file name was hardcoded when we had only one fuzzer. This change allows each fuzzer to have a separate file name. However for performance reasons it is initialized only once. `tmpFile` is static and empty in the beginning. `LLVMFuzzerTestOneInput` is called millions of times in the same process.

Tests/Fuzzers/FuzzTarget.cpp
```diff
@@ -1,87 +1,76 @@
...
-	pcpp::RawPacket rawPacket;
-	if (!reader.getNextPacket(rawPacket))
+	pcpp::RawPacketVector packets;
+	if (reader->getNextPackets(packets, 1) != 1)
 	{
 		std::cerr << "Couldn't read the first packet in the file\n";
 		return 0;
 	}
 
+	pcpp::RawPacket& rawPacket = *packets.front();
```
seladb, line 53: Why do we need to call `reader->getNextPackets()` before calling `reader.getNextPacket()` in a loop?

Pcap++/src/PcapFileDevice.cpp
seladb: sure, we can leave this fix 👍

Tests/Fuzzers/FuzzTarget.cpp
seladb: I guess we can start with the code as is, if we see too many issues with pcapng we can decide what to do - either options (b) or (c), or just disable opening pcapng files in the fuzzer

Tests/Fuzzers/ReadParsedPacket.h
seladb: This is very useful information! I think we should add a README with this info and useful instructions. I'd include:
- Some general information on what are fuzzing tests
- What fuzzing tests are included here for PcapPlusPlus
- Useful links: 
  - oss-fuzz project
  - The latest coverage report
  - Open issues for PcapPlusPlus
  - How to build locally
  - How to create a coverage report locally
- Anything else you can think of?

Tests/Fuzzers/FuzzTarget.cpp
sashashura: It is just a little trick to increase variety of functions that the fuzzer calls to increase the coverage.

Tests/Fuzzers/ReadParsedPacket.h
sashashura: I have added the readme.

Tests/Packet++Test/PacketExamples/gtp-u.pcap
seladb, line 1: Is this pcap file relevant for this PR?

sashashura: Yes, it improves the coverage. That was the case when code was reachable, but the fuzzer didn't manage to reach the code without a help.

seladb (approved):

seladb:
Thank you so much @sashashura for working on this! 🙏 🙏

sashashura:
Marked https://github.com/google/oss-fuzz/pull/11198 as ready for review, but this pr wasn't merged to master yet.

seladb:
@sashashura I merged your PR to `master`

PR #1228, opened 2023-11-02
Introduce XDP device

seladb:
This PR introduces support for AF_XDP sockets (XSK): https://www.kernel.org/doc/html/next/networking/af_xdp.html
XDP stands for eXpress Data Path and it is part of [eBPF](https://ebpf.io/).

The functionality included in this PR is pretty minimal:
- Configure and initialize an AF_XDP socket (including UMEM initialization which is a region of virtual contiguous memory used by the socket)
- Receive and send packets (RX/TX)
- Gather statistics

To build this code I used a few existing AF_XDP examples provided in the following projects:
- https://github.com/xdp-project/bpf-examples
- https://github.com/xdp-project/xdp-tutorial/tree/master/advanced03-AF_XDP
- https://github.com/Packet-Batch/PB-AF-XDP (which does only TX)


Tests/Pcap++Test/Tests/XdpTests.cpp
```diff
@@ -0,0 +1,304 @@
...
+#include "PcapLiveDeviceList.h"
+#include "XdpDevice.h"
+#include "PcapFileDevice.h"
+#include "Packet.h"
+#include "Logger.h"
+
+
+extern PcapTestArgs PcapTestGlobalArgs;
+
+#if USE_XDP
```
clementperon, line 12: better to use ifdef as we set USE_XDP with add_definitions(-DUSE_XDP) 
And we don't explictly set it to 1

seladb: Good catch! Fixed in https://github.com/seladb/PcapPlusPlus/pull/1228/commits/8ee9c83b433f644b89b5bbafc37a4c1620c28d3c

Pcap++/CMakeLists.txt
```diff
@@ -93,6 +98,7 @@ target_link_libraries(
          Packet++
          $<$<BOOL:${PCAPPP_USE_PF_RING}>:PF_RING::PF_RING>
          $<$<BOOL:${PCAPPP_USE_DPDK}>:DPDK::DPDK>
+         $<$<BOOL:${PCAPPP_USE_XDP}>:bpf>
```
clementperon, line 101: It would be nice to have a find_package(bpf)

So in case bpf is not found it fails at the configuration step. 
Also this make explicit where we have to look for lib and headers and allo user to have a custom BPF library with -DBPF_ROOT=<path_to_custom_bpf>

seladb: How should I do that?

I found a few examples of `FindLibbpf.cmake`, is this what you mean?

- https://github.com/iovisor/bcc/blob/master/cmake/FindLibBpf.cmake
- https://fossies.org/linux/nghttp2/cmake/FindLibbpf.cmake

clementperon: The one from fossies seems nice but I would export a bpf::bpf target

seladb: @clementperon I updated the PR and added `FindBPF` which exports a `BPF::BPF` target which is used in Pcap++ CMakeLists.txt. Please let me know what you think

cmake/modules/FindBPF.cmake
```diff
@@ -0,0 +1,38 @@
...
+#  BPF_FOUND        - System has libbpf
+#  BPF_INCLUDE_DIRS - The libbpf include directories
+#  BPF_LIBRARIES    - The libraries needed to use libbpf
+# ~~~
+
+find_package(PkgConfig QUIET)
+pkg_check_modules(PC_LIBBPF libbpf)
+
+find_path(
+  BPF_INCLUDE_DIR
```
clementperon, line 15: BPF_INCLUDE_DIRS

cmake/modules/FindBPF.cmake
```diff
@@ -0,0 +1,38 @@
...
+  NAMES bpf/bpf.h
+  HINTS ${PC_LIBBPF_INCLUDE_DIRS})
+
+find_library(
+  BPF_LIBRARY
```
clementperon, line 20: BPF_LIBRARIES

cmake/modules/FindBPF.cmake
```diff
@@ -0,0 +1,38 @@
...
+  VERSION_VAR BPF_VERSION
+  FAIL_MESSAGE "libbpf not found!")
+
+if(BPF_FOUND AND NOT TARGET BPF::BPF)
+  add_library(BPF::BPF INTERFACE IMPORTED)
+  set_target_properties(
+    BPF::BPF
+    PROPERTIES INTERFACE_LINK_LIBRARIES "${BPF_LIBRARY}"
+               INTERFACE_INCLUDE_DIRECTORIES "${BPF_INCLUDE_DIR}"
+               INTERFACE_COMPILE_OPTIONS "${BPF_CFLAGS_OTHER}")
```
clementperon, line 37: There is no specific Compile flags to set

cmake/modules/FindBPF.cmake
```diff
@@ -0,0 +1,38 @@
...
+
+find_library(
+  BPF_LIBRARY
+  NAMES bpf
+  HINTS ${PC_LIBBPF_LIBRARY_DIRS})
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(
+  BPF
+  REQUIRED_VARS BPF_LIBRARY BPF_INCLUDE_DIR
```
clementperon, line 27: Libraries and include_dirs

cmake/modules/FindBPF.cmake
seladb: Fixed in https://github.com/seladb/PcapPlusPlus/pull/1228/commits/1a5b29beef3781272bf0c5d5f1ef8574e90b30b7

cmake/modules/FindBPF.cmake
seladb: Fixed in https://github.com/seladb/PcapPlusPlus/pull/1228/commits/1a5b29beef3781272bf0c5d5f1ef8574e90b30b7

cmake/modules/FindBPF.cmake
seladb: Fixed in https://github.com/seladb/PcapPlusPlus/pull/1228/commits/1a5b29beef3781272bf0c5d5f1ef8574e90b30b7

cmake/modules/FindBPF.cmake
seladb: Fixed in https://github.com/seladb/PcapPlusPlus/pull/1228/commits/1a5b29beef3781272bf0c5d5f1ef8574e90b30b7

clementperon (approved):

PR #1229, opened 2023-11-04
Ported TcpReassembly example to C++11

jpcofr:


Examples/TcpReassembly/main.cpp
jpcofr, line 1: Hi @seladb - The windows-2019 target seems not to parse the method `void listInterfaces()` correctly and argues about an incomplete struct.

```
C2059: syntax error: '<class-head>'
C2143: syntax error: missing ';' before '{'
C2332: 'struct': missing tag name
C2226: syntax error: unexpected type 'listInterfaces::<unnamed-tag>'
C2332: 'struct': missing tag name
```
Have you seen these before in this compilation target?

seladb: Can you try `for (auto interface : devList)` instead of `for (const auto& interface : devList)`?

Examples/TcpReassembly/main.cpp
```diff
@@ -329,7 +328,7 @@ void listInterfaces()
 
 	std::cout << std::endl << "Network interfaces:" << std::endl;
 
-	for (const auto& interface : devList)
+	for (auto& interface : devList)
```
jpcofr, line 331: @seladb The problem is still there. How do you check if the code compiles for different systems? Using Docker or directly on GitHub with its fast CI? I need a faster way to test changes; right now, I can test only 2 or 3 times in an hour.

seladb: I usually use VMs to test on different platforms. For unix/linux-based platforms you can also use docker.

For Windows you can use these free VMs offered by Microsoft:
https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/

You can install VS 2019 / 2022 and test whatever you need

seladb: @jpcofr I think the issue is with the variable name... `interface` is probably a reserved word in Visual Studio 🤦‍♂️ 

I changed `interface` to `dev` and it seems to work now...

Examples/TcpReassembly/main.cpp
```diff
@@ -374,7 +373,7 @@ static void tcpReassemblyMsgReadyCallback(int8_t sideIndex, const pcpp::TcpStrea
 		if (result == 1)
 		{
 			// find the connection from the flow key
-			TcpReassemblyConnMgrIter iter2 = connMgr->find(flowKeyToCloseFiles);
+			auto iter2 = connMgr->find(flowKeyToCloseFiles);
```
seladb, line 376: should we rename the variable to `flow2`, the same way we renamed `iter` to `flow`?

seladb: Fixed in https://github.com/seladb/PcapPlusPlus/pull/1229/commits/48be196b84314896d8c55de42ac68a39470e5428

seladb (approved):

PR #1230, opened 2023-11-05
add regresssion_samples explanation

sashashura:


.pre-commit-config.yaml
```diff
@@ -5,6 +5,7 @@ repos:
     rev: v4.5.0
     hooks:
       - id: trailing-whitespace
+        args: [--markdown-linebreak-ext=md]
```
seladb, line 8: Why is this change needed?

sashashura: Because two spaces at the end is a valid markdown line-break syntax. Pre-commit was too aggressive for markdown files.

seladb: While two spaces at the end is a valid markdown line-break, it's not considered a best practice: https://www.markdownguide.org/basic-syntax/#line-break-best-practices

sashashura: Hmm, The best practice is to mix html with markdown?? Just saying - https://grep.app/search?q=--markdown-linebreak-ext
Anyway, as you wish, I have reverted it.

seladb (approved):

PR #1236, opened 2023-11-12
 Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64110 

sashashura:


Packet++/src/IPv4Layer.cpp
```diff
@@ -290,9 +290,15 @@ void IPv4Layer::parseNextLayer()
...
+		{
+			if (GREv0Layer::isDataValid(payload, payloadLen))
+				m_NextLayer = new GREv0Layer(payload, payloadLen, this, m_Packet);
+		}
 		else if (greVer == GREv1)
-			m_NextLayer = new GREv1Layer(payload, payloadLen, this, m_Packet);
+		{
+			if (GREv1Layer::isDataValid(payload, payloadLen))
+				m_NextLayer = new GREv1Layer(payload, payloadLen, this, m_Packet);
+		}
```
seladb, line 301: It's a bit nicer to do:
```cpp
if (greVer == GREv0 && GREv0Layer::isDataValid(payload, payloadLen))
	m_NextLayer = new GREv0Layer(payload, payloadLen, this, m_Packet);
else if (greVer == GREv1 && GREv1Layer::isDataValid(payload, payloadLen))
	m_NextLayer = new GREv1Layer(payload, payloadLen, this, m_Packet);
else
```

Packet++/src/IPv6Layer.cpp
```diff
@@ -242,9 +242,15 @@ void IPv6Layer::parseNextLayer()
...
+		{
+			if (GREv0Layer::isDataValid(payload, payloadLen))
+				m_NextLayer = new GREv0Layer(payload, payloadLen, this, m_Packet);
+		}
 		else if (greVer == GREv1)
-			m_NextLayer = new GREv1Layer(payload, payloadLen, this, m_Packet);
+		{
+			if (GREv1Layer::isDataValid(payload, payloadLen))
+				m_NextLayer = new GREv1Layer(payload, payloadLen, this, m_Packet);
+		}
```
seladb, line 253: same here

Packet++/src/IPv4Layer.cpp
sashashura: But it means if it is `GREv0` and invalid, it will go to `new PayloadLayer`.

seladb: yes, we always want to have a next layer... the default is `PayloadLayer` if there is no better option...

seladb (approved):

PR #1237, opened 2023-11-12
 Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64111 

sashashura:


Packet++/src/BgpLayer.cpp
```diff
@@ -713,7 +713,21 @@ bool BgpUpdateMessageLayer::setNetworkLayerReachabilityInfo(const std::vector<pr
...
+		{
+			bool res = extendLayer(headerLen, sizeof(uint16_t));
+			if (!res)
+			{
+				PCPP_LOG_ERROR("Couldn't extend BGP update layer to include the additional NLRI data");
+				return res;
+			}
+			*(uint16_t*)(m_Data + sizeof(bgp_common_header) + sizeof(uint16_t)) = 0;
+			getBasicHeader()->length = htobe16(be16toh(getBasicHeader()->length) + sizeof(uint16_t));
+		}
```
seladb, line 728: Instead of trying to extend the layer twice, we can just return an error if `headerLen < minLen` because it's not really a valid `BgpUpdateMessageLayer`

sashashura: I dug deeper and introduced `isDataValid` to prevent creation of the invalid object instead.

Packet++/src/BgpLayer.cpp
```diff
@@ -703,6 +703,26 @@ void BgpUpdateMessageLayer::getNetworkLayerReachabilityInfo(std::vector<prefix_a
...
 }
 
+bool BgpUpdateMessageLayer::isDataValid(const uint8_t *data, size_t dataSize)
+{
+	if (data == nullptr || dataSize < sizeof(bgp_common_header) + 2*sizeof(uint16_t))
+		return false;
+
+	uint16_t messageLen = be16toh(((bgp_common_header*)data)->length);
+	if (dataSize < messageLen)
+		return false;
```
seladb, line 713: Some of the checks here can move to `BgpLayer::parseBgpLayer`, for example:
- Check if `data == nullptr`
- Check if `dataSize < messageLen`

seladb (approved):

PR #1240, opened 2023-11-15
 Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64190 

sashashura:


Packet++/src/SomeIpSdLayer.cpp
```diff
@@ -636,11 +636,25 @@ bool SomeIpSdLayer::isDataValid(const uint8_t* data, size_t dataLen)
...
+	while (len < lenOptions)
+	{
+		if (len + sizeof(uint16_t) + 3 * sizeof(uint8_t) > lenOptions)
+			return false;
+
+		uint32_t lenOption = be16toh(*((uint16_t *)(data + offsetOption + len))) + 3 * sizeof(uint8_t);
+		len += lenOption;
+		if (len > lenOptions) // the last must be equal to lenOptions
+			return false;
+	}
```
seladb, line 656: it's a bit weird to parse all the options in `isDataValid()`. Maybe we should fix `countOptions()` and `findOption()` to avoid going into non-allocated memory?

sashashura:
I thought the pattern is to reject invalid options instead of allowing to
create them and then having multiple checks all over different get methods
(and not only getters. Add* methods should also handle the corrupted data).

seladb:
Yes, usually we prefer to reject invalid layers using the checks in `isDataValid()`. But in this case it feels like we do too much parsing to understand if the layer is valid, even before we actually create the layer...

sashashura:
`countOptions` is called in constructor. As I understand it is a deliberate design of the library not to throw exceptions in constructors and have `isDataValid`. The question is why `countOptions` should try to get sense from non-valid data?

OK, ignoring the corrupted tail sounds easy, but there are all kind of fields in the beginning that describe the size of remaining data. Handling such data means to remember to use the "trimmed" sizes instead.

Also this "tolerate and fix" led to https://github.com/seladb/PcapPlusPlus/pull/1237. In short the construction of `BgpUpdateMessageLayer` tolerated incorrect withdrawnRoutes size. Adding new withdrawnRoutes correctly discarded the garbage data and added the new withdrawnRoutes. Clearing the routes correctly trimmed them, except that the final object became without the space for the attributes data size (which was previously assumed to be in the non-parsed data after routes). To summarize "tolerate and fix" leads to construction of an object with ticking bombs where some clear or add operation may lead to invalid memory access.

seladb:
> `countOptions` is called in constructor. As I understand it is a deliberate design of the library not to throw exceptions in constructors and have `isDataValid`. The question is why `countOptions` should try to get sense from non-valid data?

I agree it's probably wrong to do this calculation in the c'tor. A lazy calculation might have been better + making `countOptions` more resilient to bad data

> Also this "tolerate and fix" led to #1237. In short the construction of `BgpUpdateMessageLayer` tolerated incorrect withdrawnRoutes size. Adding new withdrawnRoutes correctly discarded the garbage data and added the new withdrawnRoutes. Clearing the routes correctly trimmed them, except that the final object became without the space for the attributes data size (which was previously assumed to be in the non-parsed data after routes). To summarize "tolerate and fix" leads to construction of an object with ticking bombs where some clear or add operation may lead to invalid memory access.

I agree that when possible we shouldn't use the "tolerate and fix" approach because of the reasons you mentioned. However, protocol parsing is not always easy, and for some protocols verifying that the data is fully valid would require a very complex logic in `isDataValid()` which would become as complex as parsing the entire layer, which is what the layer itself is doing.
So we need to find the right balance on which checks are simple enough to include in `isDataValid()` and which are not, and we should "tolerate" them after the layer is created.

Going back to this specific case - I think the addition to `isDataValid()` is quite complex and hard to read. So I see 2 options:
- Remove it, don't calculate `countOptions` in the c'tor and make sure the layer "tolerates" bad data
- Keep this validation in `isDataValid()` but find a way to make it simpler and more readable, not only by just using comments, but by making the code itself easier to understand. For example: extract pieces to smaller functions and/or reuse some of the code between `isDataValid()`, `countOptions()` and `findOption()`, 

sashashura:
Please take a look.

Packet++/src/SomeIpSdLayer.cpp
```diff
@@ -636,28 +636,35 @@ bool SomeIpSdLayer::isDataValid(const uint8_t* data, size_t dataLen)
...
+		!countOptions(nullptr, data))
 	{
 		return false;
 	}
 
 	return true;
 }
 
-uint32_t SomeIpSdLayer::countOptions()
+bool SomeIpSdLayer::countOptions(uint32_t* count, const uint8_t* data)
```
seladb, line 648: Instead of `uint32_t*` I'd put a reference `uint32_t&`, that's a more standard way in c++ to return a value.

Then in `isDataValid()` we can create a dummy variable to pass to this method

seladb (approved):

PR #1242, opened 2023-11-18
Fix OUI update

egecetin:
Since Wireshark changed the manuf file with static C file python script should be updated for new structure


3rdParty/OUIDataset/create_oui_data.py
```diff
@@ -4,9 +4,13 @@
 from typing import Optional
 import urllib.request
 
-MANUF_URL = "https://gitlab.com/wireshark/wireshark/-/raw/master/manuf"
+MANUF_URL = "https://gitlab.com/wireshark/wireshark/-/raw/master/epan/manuf-data.c"
 MANUF_FILENAME = "manuf.dat"
```
seladb, line 8: I think that this const is not used anymore

3rdParty/OUIDataset/create_oui_data.py
```diff
@@ -4,9 +4,13 @@
...
 
+MA_L = 0
+MA_M = 1
+MA_S = 2
```
seladb, line 12: We these consts used anywhere?

egecetin: They are required for `eval` function. During the processing for lines https://gitlab.com/wireshark/wireshark/-/blob/master/epan/manuf-data.c?ref_type=heads#L18 it raises an error `Undefined variable MA_S`. To keep changes minimal instead of parsing this array to detect mask value I used MAC length (character length shows mask for 24/28/36) based detection

seladb: got it. That's the downside of using `eval()`, it's not easy to track what the code is really doing... maybe we should be more explicit, parse these lines as strings and convert them to python lists?

egecetin: It is possible, I tried a fast solution. I can take a look again

3rdParty/OUIDataset/create_oui_data.py
```diff
@@ -150,13 +157,12 @@ def main() -> None:
...
-        if line[0] in ["#", "\n"]:
+        if line.startswith(("//", "/*", " *", "*/", "\n", "\r\n", "static", "};")):
             continue
 
-        if line_elements := parse_mac_and_vendor(
-            [word for word in line.split("\t") if word.strip()]
-        ):
-            update_oui_dataset(oui_dataset, line_elements)
+        if line_elements := convert_line(line):
+            if line_elements := parse_mac_and_vendor(convert_line(line)):
```
seladb, line 164: Something seems to be wrong here: you have nested `if` statements that use the same variable name (`line_elements`) so they override each other. Can you explain what these `if` statements are supposed to check?

egecetin: Actually, it was for preventing the error caused when `convert_line` return `None` but looks like I forgot to change variable of second `if` with `parse_mac_and_vendor(line_elements)`. But since first if successed it worked luckily so I missed it. The problem with `None` is there is no `len` of `None` so it raises an error in https://github.com/egecetin/PcapPlusPlus/blob/df622a8a1de05afda72d6849d6fd4607399cc22c/3rdParty/OUIDataset/create_oui_data.py#L77-L79

Maybe I can move this none check to inside of `parse_mac_and_vendor` it would be better for readability 🤔

seladb (approved):

PR #761, opened 2021-11-30
Add CIFuzz integration

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.
This is a service offered by OSS-Fuzz, on which PcapPlusPlus already runs.

PR #1243, opened 2023-11-18
Prevent regressions

sashashura:


sashashura:
This pr uncomments memory sanitizer regressions tests and adds a quick fuzzing tests. I see [you had concerns in the past that](https://github.com/seladb/PcapPlusPlus/pull/761) that CIFuzz may give alerts not related to the pr and make for people harder to contribute. However it is written in the documentation:
If CIFuzz doesn’t find a crash during the allotted time, the CI test passes (green check). If CIFuzz finds a crash, it reports the crash only if both of following are true:
* The crash is reproducible (on the PR/commit build).
* The crash does not occur on older OSS-Fuzz builds. (If the crash does occur on older builds, then it was not introduced by the PR/commit being tested.)

So if it is not something that can be reproduced with the old build it won't fail.

seladb:
> This pr uncomments memory sanitizer regressions tests and adds a quick fuzzing tests. I see [you had concerns in the past that](https://github.com/seladb/PcapPlusPlus/pull/761) that CIFuzz may give alerts not related to the pr and make for people harder to contribute. However it is written in the documentation: If CIFuzz doesn’t find a crash during the allotted time, the CI test passes (green check). If CIFuzz finds a crash, it reports the crash only if both of following are true:
> 
>     * The crash is reproducible (on the PR/commit build).
> 
>     * The crash does not occur on older OSS-Fuzz builds. (If the crash does occur on older builds, then it was not introduced by the PR/commit being tested.)
> 
> 
> So if it is something that can be reproduced with the old build it won't fail.

What does an "older build" mean? I'd expect that CIFuzz will try to reproduce the crash on the branch the PR will be merged to (in our case - it's `dev` most of the time). Is that the case?

sashashura:
I guess so. Is there anything wrong about it? It is quite effective
strategy - even if the crash wasn't known before, just rerun the single
reproducer case. If it reliably crashes with the target branch, then it
wasn't introduced with the pr and there is no need to fail the check.

seladb:
> I guess so. Is there anything wrong about it? It is quite effective strategy - even if the crash wasn't known before, just rerun the single reproducer case. If it reliably crashes with the target branch, then it wasn't introduced with the pr and there is no need to fail the check.

I'd like to make sure that if the CIFuzz run fails it's only because of something related to the changes in the PR. The only way to verify is to run the same scenario on the branch the PR will be merged to and make sure it doesn't crash. Is there any way to make sure?

What I'd like to avoid is a scenario where CIFuzz discovered a crash in a certain PR that is also reproducible in `dev`, which means the crash is not related to the changes in that PR. This will make the contributor frustrated because the PR cannot be merged due to things not related to their changes

clementperon:
@selab they are lot of projects using CiFuzz and not small ones.

I would say let's try it and if this kind of issue happens we can always disable it.

My understanding is that in case of new fuzzing discovery they redo the test with the previous commit just to be sure it's not an improvement of the fuzzing engine.

.github/workflows/build_and_test.yml
```diff
@@ -568,3 +568,63 @@ jobs:
...
+        run: |
+          apt-get update && apt-get install -y cmake autoconf flex bison
+          cd ..
+          cp -r PcapPlusPlus/ $SRC/PcapPlusPlus
+          git clone --depth=1 https://github.com/the-tcpdump-group/libpcap.git $SRC/libpcap
+          cd $SRC/PcapPlusPlus/
+      - name: Compile fuzzer
+        run: |
+          export FUZZING_LANGUAGE=c
+          export ARCHITECTURE=x86_64
```
clementperon, line 594: Nitpick but I'm not fan of using environment variable to pass value to a script.

I prefer to properly pass it as parameters, so we can properly check the input, have default and explain what we want.

But not a big deal as it's only for CI and not related to this MR

clementperon (approved):

sashashura:
> The only way to verify is to run the same scenario on the branch the PR will be merged

It seems this is how it already works.
I agree with Clement - let's try it and see. We can always disable it.

seladb (approved):
Ok, let's go for it!

Issue #1220, opened 2023-10-19
support `pcap_get_selectable_fd`


tigercosmos:
as discussed with @seladb 
> we can expose the fd from `PcapLiveDevice` for users to use with poll() and select(). Please note that pcap_get_selectable_fd has a few limitations explained [here](https://man7.org/linux/man-pages/man3/pcap_get_selectable_fd.3pcap.html) so we probably only want to expose it for certain platforms 

so, Windows is not supported for sure, seems macOS 10.7 after can be supported?
Linux/Unix should be all supported.

seladb:
Thanks for opening the issue @tigercosmos !
I think we can assume it's supported on Linux and macOS and not on Windows

seladb:
@tigercosmos when I think about it more, the current [`PcapLiveDevice::startCaptureBlockingMode()`](https://github.com/seladb/PcapPlusPlus/blob/8c4f231cf0432b5ca1ef42cd26322e15f2a78489/Pcap%2B%2B/src/PcapLiveDevice.cpp#L469) already does more or less the same thing:
- It runs in an endless loop that will break only if the user returns `true` in the callback
- In each iteration of the loop it calls `pcap_dispatch` which essentially waits for the next packet (similar to `poll()`), then calls an internal callback `onPacketArrivesBlockingMode`
- This internal callback creates a `RawPacket` object and calls the user's callback. If the user's callback returns `true` it sets a flag to stop capturing packets

I think this covers the scenario you described above without using `poll()` and without opening another thread. Am I missing something?

tigercosmos:
@seladb  True, indeed it's more or less the same.

The current `startCaptureBlockingMode` works fine and it ensures compatibility for all platforms.

However, if a programmer is going to use blocking mode (or serial programing), there is no need to open another thread. That's why I am proposing to add new support for `pcap_get_delectable_fd`.

seladb:
@tigercosmos with the example you described above, how can the program do other logic while waiting for a packet in `poll()`?
When there are no incoming packets, the program is stuck in `poll()` and when there are incoming packets, the program handles them, there is no other logic that can happen in between 🤔

tigercosmos:
True, sorry for confusing you. 
I just read the implementation of `startCaptureBlockingMode` again.
Actually, `startCaptureBlockingMode`  doesn't open any thread (though there are many "thread" keywords, which are confusing); however, it uses *busy looping* while running before timeout.
```cpp
while(1) {
[...]

PR #1245, opened 2023-11-24
implement poll() for `startCaptureBlockingMode`

tigercosmos:
please refer to https://github.com/seladb/PcapPlusPlus/issues/1220

new in PR:
- `timeout` of `startCaptureBlockingMode` keeps the unit as second but changes to use `double` type to allow millisecond precision.
- add `usePoll` option
- main behavior difference: originally, if there are no packets coming, then the timeout will never be triggered due to blocking of `pcap_dispatch`. Use `poll()` to frequently check if it is already timeout.


tigercosmos:
@seladb  please take a look.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -330,10 +348,20 @@ bool PcapLiveDevice::open(const DeviceConfiguration& config)
...
 	}
 
+	m_PcapSelectableFd = pcap_get_selectable_fd(m_PcapSendDescriptor);
+
 	PCPP_LOG_DEBUG("Device '" << m_Name << "' opened");
 
 	m_DeviceOpened = true;
 
+	if(config.usePoll) {
+#if defined(_WIN32)
```
tigercosmos, line 358: use check the platform here

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -488,29 +518,44 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
 	m_cbOnPacketArrivesBlockingMode = onPacketArrives;
 	m_cbOnPacketArrivesBlockingModeUserCookie = userCookie;
 
-	long startTimeSec = 0, startTimeNSec = 0;
-	clockGetTime(startTimeSec, startTimeNSec);
-
-	long curTimeSec = 0;
+	auto startTime = std::chrono::steady_clock::now();
 
+	std::chrono::_V2::steady_clock::time_point currentTime;
```
tigercosmos, line 523: use morden `chrono` here

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -466,6 +466,99 @@ PTF_TEST_CASE(TestPcapLiveDeviceBlockingMode)
...
 } // TestPcapLiveDeviceBlockingMode
 
+PTF_TEST_CASE(TestPcapLiveDeviceBlockingModeWithPoll)
+{
+	// open device
+	pcpp::PcapLiveDevice* liveDev = pcpp::PcapLiveDeviceList::getInstance().getPcapLiveDeviceByIp(PcapTestGlobalArgs.ipToSendReceivePackets.c_str());
+	PTF_ASSERT_NOT_NULL(liveDev);
+
+	pcpp::PcapLiveDevice::DeviceConfiguration defaultConfig;
+	defaultConfig.usePoll = true;
```
tigercosmos, line 476: test with the new flag.

seladb:
@tigercosmos I updated the base branch to `dev`. We always merge PRs to `dev` first before merging them to `master`.

Please also note CI fails on many platforms. Can you please check?

tigercosmos:
> @tigercosmos I updated the base branch to `dev`. We always merge PRs to `dev` first before merging them to `master`.
> 
> Please also note CI fails on many platforms. Can you please check?

@seladb Sure, I will fix the CI. I just like to ask your opinion about this PR. Does the current implementation make sense to you?

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -488,29 +518,44 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
-	clockGetTime(startTimeSec, startTimeNSec);
-
-	long curTimeSec = 0;
+	auto startTime = std::chrono::steady_clock::now();
 
+	auto currentTime = startTime;
 	m_CaptureThreadStarted = true;
 	m_StopThread = false;
 
-	if (timeout <= 0)
```
seladb, line 499: In the previous code we supported "no timeout" mode where we call `pcap_dispatch` forever until the user decides to stop capture. I think we should continue supporting it

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -488,29 +518,44 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
+
+	constexpr int shortIntervalMs = 10;
+
+	while (!m_StopThread)
 	{
-		while (!m_StopThread)
+		if(m_usePoll)
 		{
-			pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
+			int ready = poll(&pcapPollFd, 1, shortIntervalMs); // wait for few milliseconds
```
seladb, line 538: Why do we use `shortIntervalMs` instead of `timeout`? Isn't the whole point is to use `poll` for the timeout provided by the user?

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
seladb: I guess we can only test it on supported platforms?

seladb (commented):
Please see a few initial comments. I will do a more thorough review once CI passes

Pcap++/header/PcapLiveDevice.h
```diff
@@ -402,13 +411,14 @@ namespace pcpp
...
-		 * @param[in] timeout A timeout in seconds for the blocking to stop even if the user didn't return "true" in the onPacketArrives callback
+		 * @param[in] timeout A timeout in seconds for the blocking to stop even if the user didn't return "true" in the onPacketArrives callback.
+		 * The precision of `timeout` is millisecond, e.g. 2.345 seconds means 2345 milliseconds.
 		 * If this timeout is set to 0 or less the timeout will be ignored, meaning the method will keep blocking until the user frees it via
 		 * the onPacketArrives callback
 		 * @return -1 if timeout expired, 1 if blocking was stopped via onPacketArrives callback or 0 if an error occurred (such as device
 		 * not open etc.). When returning 0 an appropriate error message is printed to log
 		 */
-		virtual int startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacketArrives, void* userCookie, int timeout);
+		virtual int startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacketArrives, void* userCookie, const double timeout);
```
clementperon, line 421: struct timespec no ?

Pcap++/src/PcapLiveDevice.cpp
tigercosmos: the logic is like this:
```
while(not timeout) {
     if(check if poll got a packet "in" event within a short period) {
        pcap_function()
     }
     
     // either poll got packets or poll got timeout , we can check if the while got timeout
}
```

ok, now imagine a case, the while loop timeout is 2000ms, and we set the poll timeout also 2000ms.
0. start time is 0 ms.
1. a packet comes at time 1000ms
2. while-loop not timeout, go to the next loop
3. no packets coming anymore
4. poll waits for 2000ms
And you can see that the whole process causes 1000+2000 ms

Pcap++/src/PcapLiveDevice.cpp
tigercosmos: yes, the behavior is still the same. I did a refactoring to remove the duplicate `while()`, please check the new code.

from 
```
if(timeout) {
 while(stop) { pcap() }
else {
 while(stop) { pcap() }
}
```
to
```
while(stop){
  if{ pcap() }
  else{ pcap() } 
  
  if(timeout){ break;}
}
```

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: I agree. In addition, I would like to incorporate a test to ensure that PCPP can correctly terminate the loop when there are no incoming packets. However, I'm currently unsure of how to implement this within the existing testing framework

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -488,29 +517,47 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
+	pcapPollFd.events = POLLIN;
+	constexpr int pollShortIntervalMs = 10;
+#endif
+
+	while (!m_StopThread)
 	{
-		while (!m_StopThread)
+		if(m_usePoll)
 		{
-			pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
```
tigercosmos, line 503: notice here `onPacketArrivesBlockingMode` is renamed to `onPacketArrivesAndStopBlockingMode`

Pcap++/src/PcapLiveDevice.cpp
seladb: you're right I missed that. Thank you for the explanation!

Pcap++/src/PcapLiveDevice.cpp
seladb: You're right again, I forgot what the meaning of `timeout` was 🤦‍♂️ 

But maybe we can build this logic which might be more accurate:
```
startTime = now
pollTimeout = timeout
while (not timeout) {
    poll(pollTimeout)

    if (poll timed out) {
        break and return -1
    }

    pcap()

    pollTimeout -= now - startTime
}
```

That should solve your scenario:

0. start time is 0 ms. poll timeout is 2000ms
1. a packet comes at time 1000ms
2. while-loop not timeout, go to the next loop. poll timeout is now 1000ms
3. no packets coming anymore
4. poll waits for 1000ms and quits the loop

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
seladb: Yes, I don't have a good idea how to test this scenario, especially in CI when we use docker containers 🤔 

If these were real network interfaces maybe we could use `iptables` to block all network traffic like this: https://serverfault.com/questions/880128/how-to-block-all-traffic-between-two-interfaces

But I don't think this will work in containers...

tigercosmos: I tried to use a simple way to test it. setting the timeout to only 1ms, and it should get nothing and timeout correctly.

Pcap++/src/PcapLiveDevice.cpp
tigercosmos: That sounds good, changed.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +379,124 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
+	configs.emplace_back(); // the config used poll
+	configs[1].usePoll = true;
+#endif
 
-	// sanity - test blocking mode returns before timeout
-	int packetCount = 0;
-	PTF_ASSERT_EQUAL(liveDev->startCaptureBlockingMode(packetArrivesBlockingModeNoTimeout, &packetCount, 30), 1);
-	PTF_ASSERT_EQUAL(packetCount, 5);
+	// test the common behaviour for all configs
+	for(const auto & config: configs)
```
tigercosmos, line 391: by doing this, we can test many different configs.

tigercosmos:
don't know why centos7 failed, probably some undefined behavior occurs.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
seladb: Do you plan to add more configs to this list?

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
seladb: Hmm that could work! We just need to make sure it's deterministic. If it's not, maybe we can retry capturing packets a few times until no packets are received

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: correctly no, but I think it's better than duplicating the code twice.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: @seladb  I don't understand, could you provide an example?

tigercosmos: the idea about a very short time doesn't work. 
I use iptables to drop all packets for testing this.

tigercosmos:
@seladb  I think it's ready for review

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -488,41 +499,67 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
+				{
+					pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
+					currentTime = std::chrono::steady_clock::now();
+				}
+				else if(ready < 0)
+				{
+					PCPP_LOG_ERROR("poll() got error '" <<  strerror(errno) << "'");
+					return -1;
+				}
+#endif
```
seladb, line 543: We can add `#else`:
```cpp
#else
				PCPP_LOG_ERROR("Windows doesn't support poll()");
				return 0;
#endif
```

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -488,41 +499,67 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
+			if(m_usePoll)
+			{
+#if !defined(_WIN32)
+				int64_t pollTimeoutMs = timeoutMs - std::chrono::duration_cast<std::chrono::milliseconds>(currentTime  - startTime).count();
+				pollTimeoutMs = std::max(pollTimeoutMs, (int64_t)0); // poll will be in blocking mode if negative value
+				int ready = poll(&pcapPollFd, 1, pollTimeoutMs); // wait the packets until timeout
+				if(ready > 0)
+				{
+					pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
+					currentTime = std::chrono::steady_clock::now();
```
seladb, line 536: We use this line twice - once in `if` and the other in `else`. Instead we can use it after `if..else`:
```cpp
if (m_UsePoll)
{
	...
}
else
{
	...
}
currentTime = std::chrono::steady_clock::now();
```

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +379,133 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
 
-	// a negative test
-	pcpp::Logger::getInstance().suppressLogs();
-	PTF_ASSERT_FALSE(liveDev->startCapture(packetArrives, &packetCount));
-	pcpp::Logger::getInstance().enableLogs();
+	const char* iptablesDeleteInputDrop = "sudo iptables -D INPUT -i eth0 -j DROP";
+	const char* iptablesDeleteOutputDrop = "sudo iptables -D OUTPUT -o eth0 -j DROP";
+	std::system(iptablesDeleteInputDrop);
+	std::system(iptablesDeleteOutputDrop);
+#endif
```
seladb, line 508: Let's add `#else`:
```cpp
#else
	PTF_SKIP_TEST("This test can not run in Windows environment");
#endif
```

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +379,133 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
 	PTF_ASSERT_FALSE(liveDev->isOpened());
 
-	// a negative test
-	pcpp::Logger::getInstance().suppressLogs();
-	PTF_ASSERT_FALSE(liveDev->startCapture(packetArrives, &packetCount));
-	pcpp::Logger::getInstance().enableLogs();
+	const char* iptablesDeleteInputDrop = "sudo iptables -D INPUT -i eth0 -j DROP";
+	const char* iptablesDeleteOutputDrop = "sudo iptables -D OUTPUT -o eth0 -j DROP";
+	std::system(iptablesDeleteInputDrop);
+	std::system(iptablesDeleteOutputDrop);
```
seladb, line 507: Maybe we can derive from `DeviceTeardown` and add this to it?

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +379,133 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
+
+	PTF_ASSERT_TRUE(liveDev->open(defaultConfig));
+	DeviceTeardown devTeardown(liveDev);
+
+	int packetCount = 0;
+	PTF_ASSERT_EQUAL(liveDev->startCaptureBlockingMode(packetArrivesBlockingModeTimeout, &packetCount, 2), -1);
+	PTF_ASSERT_EQUAL(packetCount, 0);
 
 	liveDev->close();
 	PTF_ASSERT_FALSE(liveDev->isOpened());
```
seladb, line 502: I don't think this assertion is needed

Pcap++/header/PcapLiveDevice.h
```diff
@@ -87,6 +87,7 @@ namespace pcpp
 		// that occurs in libpcap on Linux (on Windows using WinPcap/Npcap it works well):
 		// It's impossible to capture packets sent by the same descriptor
 		pcap_t* m_PcapSendDescriptor;
+		int m_PcapSelectableFd = -1; // the selectable fd got from `pcap_get_selectable_fd`
```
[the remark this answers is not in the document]
tigercosmos: I used to leave comments, but sure.

Pcap++/header/PcapLiveDevice.h
```diff
@@ -109,6 +110,7 @@ namespace pcpp
 		RawPacketVector* m_CapturedPackets;
 		bool m_CaptureCallbackMode;
 		LinkLayerType m_LinkType;
+		bool m_usePoll = false; // use poll() for Unix-like system. Set in DeviceConfiguration.
```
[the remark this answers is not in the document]
tigercosmos: actually, this is unsued. removed.

Pcap++/header/PcapLiveDevice.h
```diff
@@ -217,6 +220,11 @@ namespace pcpp
 			*/
 			unsigned int nflogGroup;
 
+
+			/// @brief In Unix-like system, use poll() for blocking mode.
+			bool usePoll = false;
```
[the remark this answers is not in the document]
tigercosmos: ok

Pcap++/header/PcapLiveDevice.h
```diff
@@ -231,16 +239,18 @@ namespace pcpp
 			 * captured with USBPcap (> 131072, < 262144). A snapshot length of 65535 should be sufficient, on most if not all networks,
 			 * to capture all the data available from the packet.
 			 * @param[in] nflogGroup NFLOG group for NFLOG devices. Default value is 0.
+			 * @param[in] usePoll use `poll` implementation for `startCaptureBlockingMode` on Unix-like system. Default value is false.
```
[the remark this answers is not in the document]
tigercosmos: fixed.

Pcap++/header/PcapLiveDevice.h
```diff
@@ -402,13 +412,14 @@ namespace pcpp
 		 * @param[in] userCookie A pointer to a user provided object. This object will be transferred to the onPacketArrives callback
 		 * each time it is called. This cookie is very useful for transferring objects that give context to the capture callback, for example:
 		 * objects that counts packets, manages flow state or manages the application state according to the packet that was captured
-		 * @param[in] timeout A timeout in seconds for the blocking to stop even if the user didn't return "true" in the onPacketArrives callback
+		 * @param[in] timeout A timeout in seconds for the blocking to stop even if the user didn't return "true" in the onPacketArrives callback.
+		 * The precision of `timeout` is millisecond, e.g. 2.345 seconds means 2345 milliseconds.
```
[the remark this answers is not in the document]
tigercosmos: changed.

Pcap++/header/PcapLiveDevice.h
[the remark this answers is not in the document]
tigercosmos: fixed.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -334,6 +336,15 @@ bool PcapLiveDevice::open(const DeviceConfiguration& config)
 
 	m_DeviceOpened = true;
 
+	if(config.usePoll) {
```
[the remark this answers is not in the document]
tigercosmos: nice catch, I forgot to handle open many times.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -334,6 +336,15 @@ bool PcapLiveDevice::open(const DeviceConfiguration& config)
...
+#if defined(_WIN32)
+		PCPP_LOG_ERROR("Windows doesn't support poll(), ignore usePoll=true");
```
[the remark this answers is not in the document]
tigercosmos: ok

Pcap++/src/PcapLiveDevice.cpp
tigercosmos: true

Pcap++/src/PcapLiveDevice.cpp
tigercosmos: should not going to here, but sure, just in case

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: let me remove

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: IMO, the interface and the device are separate concepts. do you still want to do it?

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: fixed.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
seladb: Maybe add another teardown class to delete these `iptables` rules?
It's important to always remove them in teardown because if this test fails and we don't remove them, other tests will fail too

Pcap++/header/PcapLiveDevice.h
```diff
@@ -127,6 +129,7 @@ namespace pcpp
 		static void onPacketArrives(uint8_t* user, const struct pcap_pkthdr* pkthdr, const uint8_t* packet);
 		static void onPacketArrivesNoCallback(uint8_t* user, const struct pcap_pkthdr* pkthdr, const uint8_t* packet);
 		static void onPacketArrivesBlockingMode(uint8_t* user, const struct pcap_pkthdr* pkthdr, const uint8_t* packet);
+		static void onPacketArrivesAndStopBlockingMode(uint8_t* user, const struct pcap_pkthdr* pkthdr, const uint8_t* packet);
```
seladb, line 132: This method declaration is redundant, right?

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: sure, just added `SystemCommandTeardown`

Pcap++/header/PcapLiveDevice.h
tigercosmos: true, I forgot to delete.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -334,6 +336,23 @@ bool PcapLiveDevice::open(const DeviceConfiguration& config)
...
+		PCPP_LOG_ERROR("Windows doesn't support poll(), ignoring the `usePoll` parameter");
+#endif
+	}
+	else
+	{
+#if !defined(_WIN32)
+		m_UsePoll = false;
+		m_PcapSelectableFd = -1;
+#endif
+	}
```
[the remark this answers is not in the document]
tigercosmos: I think this makes more sense
```cpp
	if(!config.usePoll)
	{
		m_UsePoll = false;
		m_PcapSelectableFd = -1;
	}
	else
	{
#if !defined(_WIN32)
		m_UsePoll = true;
		m_PcapSelectableFd = pcap_get_selectable_fd(m_PcapSendDescriptor);
#else
		PCPP_LOG_ERROR("Windows doesn't support poll(), ignoring the `usePoll` parameter");
#endif
	}
```

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +379,133 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
 	}
+} // TestPcapLiveDeviceBlockingMode
 
-	PTF_PRINT_VERBOSE("Total sleep time: " << totalSleepTime << " secs");
+PTF_TEST_CASE(TestPcapLiveDeviceBlockingModePollTimeout)
+{
+#if !defined(_WIN32)
+	// drop all packets on the interface
+	const char* iptablesAddInputDrop = "sudo iptables -A INPUT -i eth0 -j DROP";
+	const char* iptablesAddOutputDrop = "sudo iptables -A OUTPUT -o eth0 -j DROP";
```
seladb, line 485: I just noticed we hard-code the interface `eth0`. This is problematic because the interface name may change per machine. Do you know if it's possible to block traffic with `iptables` using an IP address which is available in `PcapTestGlobalArgs.ipToSendReceivePackets`?

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -488,41 +505,68 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
 		{
-			long curTimeNSec = 0;
-			pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
-			clockGetTime(curTimeSec, curTimeNSec);
+			if(m_UsePoll)
+			{
+#if !defined(_WIN32)
+				int64_t pollTimeoutMs = timeoutMs - std::chrono::duration_cast<std::chrono::milliseconds>(currentTime  - startTime).count();
+				pollTimeoutMs = std::max(pollTimeoutMs, (int64_t)0); // poll will be in blocking mode if negative value
+				int ready = poll(&pcapPollFd, 1, pollTimeoutMs); // wait the packets until timeout
```
seladb, line 538: I ran the test locally and noticed that `poll()` always returns `1` even if there is no network traffic. Did you test it on your device and see `poll()` timing out? 🤔

tigercosmos: I don't know how you tested it, did you make sure the `onPacketArrive` callback returns false?

In addition, I think both you and agree that the logic in test file is correct.

please try to run the following code with and without `system()` lines.

```c++
#include <iostream>
#include <pcapplusplus/PcapLiveDevice.h>
#include <pcapplusplus/PcapLiveDeviceList.h>

static bool packetArrivesBlockingModeTimeout(pcpp::RawPacket* rawPacket, pcpp::PcapLiveDevice* dev, void* userCookie)
{
    (*(int*)userCookie)++;
	return false;
}


{

	std::system("sudo iptables -A INPUT -i lo -j DROP");
	std::system("sudo iptables -A OUTPUT -o lo -j DROP");

    pcpp::PcapLiveDevice *dev = pcpp::PcapLiveDeviceList::getInstance().getPcapLiveDeviceByIpOrName("lo");


	// open device
	pcpp::PcapLiveDevice::DeviceConfiguration defaultConfig;
	pcpp::PcapLiveDevice::DeviceConfiguration newConfig;
	newConfig.usePoll = true;

	dev->open(newConfig);


	int packetCount = 0;

    /// block 2 second
	int result = dev->startCaptureBlockingMode(packetArrivesBlockingModeTimeout, &packetCount, 2);

    std::cout << "result:" << result << std::endl;
    std::cout << "packets:" << packetCount << std::endl;

	std::system("sudo iptables -D INPUT -i lo -j DROP");
	std::system("sudo iptables -D OUTPUT -o lo -j DROP");

    return 0;
}
```

this is what I get, dropping packets:
```
result:-1
packets:0
```
receiving packets:
```
result:-1
packets:34
```

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: yes, it's possible, however, we need to use additional code to find out the interface name of the IP

Pcap++/src/PcapLiveDevice.cpp
tigercosmos: let me investigate a bit

tigercosmos:
don't know why fuzzy failed.

sashashura:
> @sashashura @clementperon any idea why the fuzzing tests fail in this PR? 🤔

It failed to compile the fuzzers: https://github.com/seladb/PcapPlusPlus/actions/runs/7060797305/job/19244714111?pr=1245#step:4:724

```
/src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:181:4: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
                        pcap_dispatch(m_PcapDescriptor, -1, onPacketArrives, (uint8_t*)this);
                        ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:186:4: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
                        pcap_dispatch(m_PcapDescriptor, 100, onPacketArrivesNoCallback, (uint8_t*)this);
                        ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:526:4: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
                        pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
                        ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:548:6: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
                                        pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
                                        ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:568:5: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
                                pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
                                ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
5 errors generated.
```

Pcap++/src/PcapLiveDevice.cpp
tigercosmos: Yes, I did run on the interface with incoming packets.

Did you try to use tcpdump or wireshark to monitor the interface and to see if the packets are really blocked? if `poll` returns `1`, it means there are still packets receiving.

a dummy interface could be fine, but I am afraid that it will become too complicated.

Tests/Pcap++Test/Common/TestUtils.cpp
```diff
@@ -111,3 +118,52 @@ void testSetUp()
 	}
 	#endif
 }
+
+#if !defined(_WIN32)
+std::string findInterfaceNameByIpAddress(const std::string& ipAddress, std::string& errorMessage) {
```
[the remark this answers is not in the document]
tigercosmos: I see. thanks for telling.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +379,140 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
-			break;
+		throw std::runtime_error(errorMessage);
 	}
 
-	PTF_PRINT_VERBOSE("Total sleep time: " << totalSleepTime << " secs");
+	// drop all packets on the interface
+	auto iptablesAddInputDrop = "sudo iptables -A INPUT -i " + interfaceName + " -j DROP";
+	auto iptablesAddOutputDrop = "sudo iptables -A OUTPUT -o " + interfaceName + " -j DROP";
+	std::system(iptablesAddInputDrop.c_str());
+	std::system(iptablesAddOutputDrop.c_str());
```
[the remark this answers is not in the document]
tigercosmos: got it.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -488,41 +505,81 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
 		{
 			pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
 		}
-		curTimeSec = startTimeSec + timeout;
 	}
 	else
 	{
-		while (!m_StopThread && curTimeSec <= (startTimeSec + timeout))
+		std::cerr << "duration before while: " <<  std::chrono::duration_cast<std::chrono::milliseconds>(currentTime  - startTime).count() << std::endl;
+		while (!m_StopThread && std::chrono::duration_cast<std::chrono::milliseconds>(currentTime  - startTime).count() <= timeoutMs )
```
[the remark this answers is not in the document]
tigercosmos: I try to be the same as the previous code. I can make it to `< timeoutMs`, but we still need to notice that if the `onArrivePacket` callback is being called right before the timeout moment, there is no way to stop the callback function even though the time has already expired.

tigercosmos:
> > @sashashura @clementperon any idea why the fuzzing tests fail in this PR? 🤔
> 
> It failed to compile the fuzzers: https://github.com/seladb/PcapPlusPlus/actions/runs/7060797305/job/19244714111?pr=1245#step:4:724
> 
> ```
> /src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:181:4: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
>                         pcap_dispatch(m_PcapDescriptor, -1, onPacketArrives, (uint8_t*)this);
>                         ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:186:4: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
>                         pcap_dispatch(m_PcapDescriptor, 100, onPacketArrivesNoCallback, (uint8_t*)this);
>                         ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:526:4: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
>                         pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
>                         ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:548:6: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
>                                         pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
>                                         ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:568:5: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
>                                 pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
>                                 ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 5 errors generated.
> ```

@seladb  it's weird, the `pcap_dispatch` call doesn't handle the return value even before this PR.
But sure, I can add some handling for this.

Pcap++/src/PcapLiveDevice.cpp
tigercosmos: @seladb I attempted to set up a dummy interface, and while it was successful on Linux, I encountered challenges creating a similar interface on macOS. I propose just using the `eth0` interface for testing purposes. Moreover, if we both agree with the accuracy of the unit test, we can confidently consider the functionality as correct.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -488,41 +505,81 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
 
 	m_CaptureThreadStarted = false;
-
 	m_StopThread = false;
-
 	m_cbOnPacketArrivesBlockingMode = nullptr;
 	m_cbOnPacketArrivesBlockingModeUserCookie = nullptr;
 
-	if (curTimeSec > (startTimeSec + timeout))
+	if (std::chrono::duration_cast<std::chrono::milliseconds>(currentTime  - startTime).count() > timeoutMs )
```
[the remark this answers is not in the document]
tigercosmos: done

seladb:
> > > @sashashura @clementperon any idea why the fuzzing tests fail in this PR? 🤔
> > 
> > 
> > It failed to compile the fuzzers: https://github.com/seladb/PcapPlusPlus/actions/runs/7060797305/job/19244714111?pr=1245#step:4:724
> > ```
> > /src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:181:4: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
> >                         pcap_dispatch(m_PcapDescriptor, -1, onPacketArrives, (uint8_t*)this);
> >                         ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> > /src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:186:4: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
> >                         pcap_dispatch(m_PcapDescriptor, 100, onPacketArrivesNoCallback, (uint8_t*)this);
> >                         ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> > /src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:526:4: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
> >                         pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
> >                         ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> > /src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:548:6: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
> >                                         pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
> >                                         ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> > /src/PcapPlusPlus/Pcap++/src/PcapLiveDevice.cpp:568:5: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
> >                                 pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this);
> >                                 ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> > 5 errors generated.
> > ```
> 
> @seladb it's weird, the `pcap_dispatch` call doesn't handle the return value even before this PR. But sure, I can add some handling for this.

yes, that's because we just merged this PR that added fuzzers to CI: https://github.com/seladb/PcapPlusPlus/pull/1243

Anyway, I opened a PR to fix it, because now both `dev` and `master` are broken 😕 : https://github.com/seladb/PcapPlusPlus/pull/1252

Pcap++/src/PcapLiveDevice.cpp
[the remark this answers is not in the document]
seladb: Yes, that's reasonable. We should call the `onArrivePacket` callback as long as timeout didn't pass. Of course the callback can include `sleep(60)` and then the timeout will expire for sure 😄

Pcap++/src/PcapLiveDevice.cpp
seladb: > I attempted to set up a dummy interface, and while it was successful on Linux, I encountered challenges creating a similar interface on macOS

Yes you're right, this doesn't work on macOS 😕 

> I propose just using the `eth0` interface for testing purposes

If I remember correctly, the interface isn't by default `eth0` on non-Linux OS's (macOS and FreeBSD), and of course people may want to run the tests locally and their default interface won't necessarily be `eth0`. So it's better to do what I suggested here: https://github.com/seladb/PcapPlusPlus/pull/1245#discussion_r1412769322

>  Moreover, if we both agree with the accuracy of the unit test, we can confidently consider the functionality as correct.

I'm not sure `iptables` causes `poll` to time out, I want to double-check that. The problem with the current test is that it passes even if `poll` didn't time out which is not what we want... but I don't know how to improve that and make sure the timeout is because of `poll`

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +379,137 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
 
-	liveDev->stopCapture();
-	PTF_ASSERT_GREATER_THAN(packetCount, 0);
+	// drop all packets on the interface
+	auto iptablesAddInputDrop = "sudo iptables -A INPUT -i " + interfaceName + " -j DROP";
+	auto iptablesAddOutputDrop = "sudo iptables -A OUTPUT -o " + interfaceName + " -j DROP";
+	if(std::system(iptablesAddInputDrop.c_str()) < 0)
+		throw std::runtime_error("failed to run: " + iptablesAddInputDrop);
+	if(std::system(iptablesAddOutputDrop.c_str()) < 0)
+		throw std::runtime_error("failed to run: " + iptablesAddOutputDrop);
```
seladb, line 492: This shouldn't throw an exception. Instead - it should fail:
```cpp
PTF_ASSERT_GREATER_OR_EQUAL_THAN(std::system(iptablesAddInputDrop.c_str()), 0);
PTF_ASSERT_GREATER_OR_EQUAL_THAN(std::system(iptablesAddOutputDrop.c_str()), 0);
```

Pcap++/src/PcapLiveDevice.cpp
seladb: **EDIT:** I tested it again, and indeed: when `iptables` drops all packets, `poll` returns `1`. I ran `tcpdump` and Wireshark - both of them don't capture packets related to that interface. They do capture other packets because the (virtual) interface was in promiscuous mode. When I disabled this mode `tcpdump` and Wireshark capture almost no packets, but `poll` still returns `1` all the time...

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: do you have a better way for the following code? it's in `TestUtils.h`, but I think we shouldn't include `PcppTestFramework.h` into it.
```cpp
~SystemCommandTeardown() noexcept(false)
	{
		if (!m_CancelTeardown)
		{
			if(std::system(m_command.c_str()) < 0)
			{
				// cppcheck-suppress exceptThrowInDestructor
				throw std::runtime_error("failed to run: " + m_command);
			}
		}
	}
```

Pcap++/src/PcapLiveDevice.cpp
tigercosmos: > If I remember correctly, the interface isn't by default eth0 on non-Linux OS's (macOS and FreeBSD),..

Yes, sorry for confusing you, I want to say, use the default interface for doing the test.

> I'm not sure iptables causes poll to time out, I want to double-check that. The problem with the current test is that it passes even if poll didn't time out which is not what we want..

I think I can create a test to prove that it will not timeout without `poll` help. 

>  but poll still returns 1 all the time...

it's intereseting. According to `poll` documents
```
On success, poll() returns a nonnegative value which is the number  of  elements  in  the
       pollfds  whose revents fields have been set to a nonzero value (indicating an event or an
       error).  A return value of zero indicates that the system call timed out before any  file
       descriptors became read.
```

as long as `poll` implementatoin is correct, it shouldn't return `1` if there is no packet.

Since I cannot reproduce your result, could you try to use a different machine or maybe other VMs and see if you can still get the same result?

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
seladb: We currently don't have a graceful way to exit teardown functions. I guess that throwing an exception is ok...

Pcap++/src/PcapLiveDevice.cpp
seladb: > Since I cannot reproduce your result, could you try to use a different machine or maybe other VMs and see if you can still get the same result?

I'll test it on a different VM and keep you posted

>I think I can create a test to prove that it will not timeout without poll help.

Just curious - what implementation do you have in mind?

tigercosmos: @seladb  please refer to the new added `TestPcapLiveDeviceBlockingModePollNotTimeout`

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: sure.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +380,188 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
+    std::future<int> future = task.get_future();
+
+    // Start a new thread to execute `startCaptureBlockingMode` and to test it timeout
+    std::thread thread(std::move(task));
+
+    // Wait for the function to finish or the timeout to occur
+    auto status = future.wait_for(std::chrono::milliseconds(2100)); // a little bit more than 2 seconds
+
+	// the function doesn't timeout
+	PTF_ASSERT_TRUE(status == std::future_status::timeout);
```
tigercosmos, line 556: this line tells us that the function has a timeout.

seladb: Does it tell if the timeout was because of `poll` timeout or because the while loop finished after timeout? 🤔

tigercosmos: please notice that, if `poll` timeout, it should timeout at right after 2000ms, and I set future timeout at 2100ms. If you have concern, I can also set timeout of future to be like 3000ms.

tigercosmos: ops, I think you misunderstood.
This test doesn't use poll.
`TestPcapLiveDeviceBlockingModeNotTimeoutWithoutPoll` is explaining, the original implementation can not successfully timeout. So I use `future` to monitor it.
If the original function is correct,  it should timeout itself. `future` will not get `std::future_status::timeout`.

seladb: I still don't understand... how does `future` ensure the timeout is because of `poll` timeout and not because of something else?

tigercosmos: in this test, we don **NOT** use `poll`; we are using the original implementation.

Recall that the PR is trying to resolve that the `pcap_dispatch` call will never break if there is no packets coming, leading to `while(!timeout)` never being checked(, hence it get stuck)

so, in this test, we disable the interface, so there is no packets coming, and by using the original implementation, `pcap_dispatch` will block and wait for packets. We use another thread (by `future`) to run the blocking call, and the main thread will monitor if the `startCaptureBlockingMode` can succecully timeout.

**if the original implementation is correct**, `startCaptureBlockingMode` will timeout right after 2000ms, and the `futre` will get `std::future_status::ready`. However, **if `startCaptureBlockingMode` is still blocking**, after the `future`  gets timeout, say 3000ms, `future` will get  `std::future_status::timeout`.

Once `future` gets `std::future_status::timeout`, we know that `startCaptureBlockingMode` is still blocking and doen't break, so we enble the interface, and there are packets coming, and `startCaptureBlockingMode` can go to the next loop, now its `while(!timeout)` is execute, and the functoin call finally finished. If we correctly run `thread.join()`, it means that the `startCaptureBlockingMode` thread correctly closes (otherwise it means the thread is still blocking)

seladb: Thanks for the explanation @tigercosmos ! I thing I understand it better now.
However, I'm not sure what the test is trying to test - the whole point of this PR is to use `poll` because `pcap_dispatch` tends to get stuck if there are no incoming packets. I'd expect the test will run the scenario I mentioned:
- we use `poll`
- we block the network traffic
- in the original implementation we'd expect `future` to get `std::future_status::timeout` because `pcap_dispatch` got stuck waiting for packets that don't arrive
- however, in the new implementation `poll` times out so we don't get stuck and we always get `std::future_status::ready`

What am I missing here?

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +380,192 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
+
+    // Wait for the function to finish or the timeout to occur
+    auto status = future.wait_for(std::chrono::milliseconds(3000)); // already wait for 3 seconds, and we know `startCaptureBlockingMode` is still blocking
+
+	// the function doesn't timeout
+	PTF_ASSERT_TRUE(status == std::future_status::timeout);
+
+	// restore the interface to let the callback receive packets so the thread can be jointed.
+	iptablesDeleteInputDrop.~SystemCommandTeardown();
+	iptablesDeleteOutputDrop.~SystemCommandTeardown();
```
seladb, line 560: Calling the teardown d'tor explicitly kind of beats the purpose of the teardown class 🤔

tigercosmos: True, let me call `std::system` directly here

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +380,192 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
+
+	auto func = [&]() -> int
+	{
+		return liveDev->startCaptureBlockingMode(packetArrivesBlockingModeTimeout, &packetCount, 2); // ideally, it should timeout after 2 seconds
+	};
+
+	// test it not timeout after 2 seconds
+	std::packaged_task<int()> task(func);
+
+    std::future<int> future = task.get_future();
```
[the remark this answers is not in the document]
tigercosmos: ops, got it.

Tests/Pcap++Test/main.cpp
```diff
@@ -229,6 +229,7 @@ int main(int argc, char* argv[])
 	PTF_RUN_TEST(TestPcapLiveDeviceNoNetworking, "no_network;live_device");
 	PTF_RUN_TEST(TestPcapLiveDeviceStatsMode, "live_device");
 	PTF_RUN_TEST(TestPcapLiveDeviceBlockingMode, "live_device");
+	PTF_RUN_TEST(TestPcapLiveDeviceBlockingModePollTimeout, "live_device");
```
seladb, line 232: You also need to add `TestPcapLiveDeviceBlockingModeNotTimeoutWithoutPoll` here if you want to run it...

Tests/Pcap++Test/TestDefinition.h
```diff
@@ -36,6 +36,7 @@ PTF_TEST_CASE(TestPcapLiveDeviceClone);
 PTF_TEST_CASE(TestPcapLiveDeviceNoNetworking);
 PTF_TEST_CASE(TestPcapLiveDeviceStatsMode);
 PTF_TEST_CASE(TestPcapLiveDeviceBlockingMode);
+PTF_TEST_CASE(TestPcapLiveDeviceBlockingModePollTimeout);
```
seladb, line 39: You also need to add `TestPcapLiveDeviceBlockingModeNotTimeoutWithoutPoll` here

tigercosmos: ops, I missed it. thanks

tigercosmos:
it's weird, it seems that `iptables` does not work by default, but it succeeded before: https://github.com/seladb/PcapPlusPlus/pull/1245/commits/1a50a03bf946d4a96cf52f80ca65cedcae6c4213

tigercosmos:
@seladb I try to use `sudo` in the `std::system` and also try to use `sudo` to start the test. Neither of them work:
```
iptables v1.6.1: can't initialize iptables table `filter': Permission denied (you must be root)
Perhaps iptables or your kernel needs to be upgraded.
iptables v1.6.1: can't initialize iptables table `filter': Permission denied (you must be root)
Perhaps iptables or your kernel needs to be upgraded.
```

Moreover, currently CI containers complain there is no iptables, so we also need to install it.
However, before a couple of commits merged into `dev`, I already used iptables, and CI was totally fine. 

I also check the changes on `dev`, it seems it should not affect this "cannot use iptables".

seladb:
> @seladb I try to use `sudo` in the `std::system` and also try to use `sudo` to start the test. Neither of them work:
> 
> ```
> iptables v1.6.1: can't initialize iptables table `filter': Permission denied (you must be root)
> Perhaps iptables or your kernel needs to be upgraded.
> iptables v1.6.1: can't initialize iptables table `filter': Permission denied (you must be root)
> Perhaps iptables or your kernel needs to be upgraded.
> ```
> 
> Moreover, currently CI containers complain there is no iptables, so we also need to install it. However, before a couple of commits merged into `dev`, I already used iptables, and CI was totally fine.
> 
> I also check the changes on `dev`, it seems it should not affect this "cannot use iptables".

Maybe it's not possible to run `iptables` inside a docker container: https://forums.docker.com/t/iptables-permission-denied/81211/3 😕

tigercosmos:
> > @seladb I try to use `sudo` in the `std::system` and also try to use `sudo` to start the test. Neither of them work:
> > ```
> > iptables v1.6.1: can't initialize iptables table `filter': Permission denied (you must be root)
> > Perhaps iptables or your kernel needs to be upgraded.
> > iptables v1.6.1: can't initialize iptables table `filter': Permission denied (you must be root)
> > Perhaps iptables or your kernel needs to be upgraded.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Moreover, currently CI containers complain there is no iptables, so we also need to install it. However, before a couple of commits merged into `dev`, I already used iptables, and CI was totally fine.
> > I also check the changes on `dev`, it seems it should not affect this "cannot use iptables".
> 
> Maybe it's not possible to run `iptables` inside a docker container: https://forums.docker.com/t/iptables-permission-denied/81211/3 😕

True... after survey, seems that Github action has some limitations. 
seems a possible solution is to run a docker inside Github action with `--cap-add=NET_ADMIN`

Or, probably we should think of some other ways to test the function.

tigercosmos:
@seladb Maybe we should not make it very complicated... as long as the poll version can work the same as the original version, maybe we don't need to prove that `poll` can timeout with no packets coming (we should believe that `poll` is always correct)

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -500,48 +518,84 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
-			long curTimeNSec = 0;
-			if (pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, reinterpret_cast<uint8_t*>(this)) == -1)
+			if(m_UsePoll)
 			{
-				PCPP_LOG_ERROR("pcap_dispatch returned an error: " << pcap_geterr(m_PcapDescriptor));
-				pcapDispatchError = true;
-				m_StopThread = true;
+#if !defined(_WIN32)
+				// Be careful when you modify the code related to `poll` in this block
+				// Some tests are not supported to run on CIs: `TestPcapLiveDeviceBlockingModePollTimeout`, `TestPcapLiveDeviceBlockingModeNotTimeoutWithoutPoll`
```
tigercosmos, line 557: @seladb  I added some comments here to mention that manual tests are required here

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +380,209 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
 
 PTF_TEST_CASE(TestPcapLiveDeviceBlockingMode)
 {
-	// open device
+	std::vector<pcpp::PcapLiveDevice::DeviceConfiguration> configs;
+	configs.emplace_back(); // the default config
+
+#if !defined(_WIN32)
+	configs.emplace_back(); // the config used poll
+	configs[1].usePoll = true;
```
tigercosmos, line 388: this ensures the behaviour is as same as before.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +380,209 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
+		// a negative test
+		pcpp::Logger::getInstance().suppressLogs();
+		PTF_ASSERT_FALSE(liveDev->startCapture(packetArrives, &packetCount));
+		pcpp::Logger::getInstance().enableLogs();
+	}
+} // TestPcapLiveDeviceBlockingMode
+
+PTF_TEST_CASE(TestPcapLiveDeviceBlockingModePollTimeout)
+{
+#if !defined(_WIN32) and defined(MANUAL_TEST)
```
tigercosmos, line 483: although this test cannot run on CI, we can manually run it locally.
I think it is worth keeping it, so I added a `MANUAL_TEST` flag for it.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -379,91 +380,209 @@ PTF_TEST_CASE(TestPcapLiveDeviceStatsMode)
...
+	PTF_SKIP_TEST("This test can not run in Windows environment");
+#endif
+} // TestPcapLiveDeviceBlockingMode
 
-	// verify it's possible to capture non-blocking after the mess done in previous lines
-	packetCount = 0;
-	PTF_ASSERT_TRUE(liveDev->startCapture(packetArrives, &packetCount));
+PTF_TEST_CASE(TestPcapLiveDeviceBlockingModeNotTimeoutWithoutPoll)
+{
+#if !defined(_WIN32) and defined(MANUAL_TEST)
```
tigercosmos, line 535: same for this test. use `MANUAL_TEST` here.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: I found that it's really difficult to use iptables to block all packets. No matter how I set up rules, I still can see packets on Wireshark. Maybe the easiest way is to use a clean interface to test... since we can ensure no packets on it ...

tigercosmos: @seladb  please let me know your opinion

seladb: Yes, I also encountered problems with `iptables`. I think we should just remove these tests. There is no point in `MANUAL_TEST` which we don't run anywhere...

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
seladb: I think we can remove this test also...

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -500,48 +518,84 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
+					{
+						PCPP_LOG_ERROR("pcap_dispatch returned an error: " << pcap_geterr(m_PcapDescriptor));
+						pcapDispatchError = true;
+						m_StopThread = true;
+					}
+				}
+				else if(ready < 0)
+				{
+					PCPP_LOG_ERROR("poll() got error '" <<  strerror(errno) << "'");
+					return -1;
```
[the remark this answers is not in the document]
tigercosmos: fixed.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -500,48 +518,84 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
+				}
+#else
+				PCPP_LOG_ERROR("Windows doesn't support poll()");
+				return 0;
```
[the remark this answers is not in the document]
tigercosmos: fixed.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -500,48 +518,84 @@ int PcapLiveDevice::startCaptureBlockingMode(OnPacketArrivesStopBlocking onPacke
...
+
 	bool pcapDispatchError = false;
 
-	if (timeout <= 0)
+	if(timeoutMs <= 0)
 	{
 		while (!m_StopThread)
 		{
-			if (pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, reinterpret_cast<uint8_t*>(this)) == -1)
+			if(pcap_dispatch(m_PcapDescriptor, -1, onPacketArrivesBlockingMode, (uint8_t*)this) < 0)
```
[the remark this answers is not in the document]
tigercosmos: sorry, fixed.

tigercosmos:
@seladb  all addressed

seladb (approved):

tigercosmos:
finally  :tada: 
thanks @seladb

seladb:
> finally 🎉 thanks @seladb

Thank you @tigercosmos for your hard work and patience in solving all the issues, very much appreciated! 🙏

Issue #977, opened 2022-10-08
Port the code to C++11


seladb:
We recently moved to C++11 and big parts of the code are still using C++98 style.
This will be an ongoing task to port more and more parts of the code to C++11.

For example:
- Use `nullptr` instead of `NULL`
- Use `auto` where possible
- Use `enum class` more
- Use range-based loops where possible
- Use `delete`, `default`, `override` where needed
- Use `unordered_map` instead of `map` for better performance
- etc.

InfiniteVerma:
I'd like to help with this.

egecetin:
Not directly related with C++11 but also, library have some many C style casts, which can be easily located by using `cppcheck`. Anyone who wants fix them just remove or comment `cstyleCast:*` line from `cppcheckSuppressions.txt` and run `pre-commit run --all-files` or manually run `cppcheck` from command line.

seladb:
> Hi @seladb - I would like to collaborate on this task. Is there any task breakdown already or is there any progress? Also, why porting to C++11 instead of later versions? (is it because the latest port you mention in your first comment was to C++11?
> 
> UPDATE - based on the Release notes (changes from v22.05) it seems this issue/ticket should be closed...

Hi @jpcofr , thanks for offering your help! Actually, the reason this issue is still open is because this is an ongoing effort. We migrated to C++11 about a year ago, but a lot of the code base is still in C++98 style (definitely most of the code that was written more than a year ago). For example: we don't always use `auto` where needed, we still use `std::map` where we could use `std::unordered_map`, we don't use range-based loops in many cases, etc. We could use your help with those.

As I mentioned above - we don't expect a huge PR that fixes all of these things. We actually prefer small PRs that make incremental fixes, and we hope that we can slowly migrate as much of the code as possible.

It'd be really great if you can help. I'd like to thank you again and I'm looking forward to seeing your contributions! 🙏 🙏

jpcofr:
>  be really great if you can help. 

@seladb ok I can collaborate. I am new to this repo so I need to check out how is everything working. What would be the next namespace/class etc. that you would like to be ported? I can start from there...

seladb:
> > be really great if you can help.
> 
> @seladb ok I can collaborate. I am new to this repo so I need to check out how is everything working. What would be the next namespace/class etc. that you would like to be ported? I can start from there...

[...]

PR #1251, opened 2023-12-02
Fix cppcheck postfix operator issues

egecetin:
Most of the postfix warnings comes from old iterator based for loops so replaced them with `auto` to help #977

### DIFF
--- a/Common++/src/TablePrinter.cpp
+++ b/Common++/src/TablePrinter.cpp
@@ -89,9 +89,9 @@ void TablePrinter::printSeparator()
 	}
 
 	int totalLen = 0;
-	for (std::vector<int>::iterator iter = m_ColumnWidths.begin(); iter != m_ColumnWidths.end(); iter++)
+	for (auto iter : m_ColumnWidths)
 	{
-		totalLen += 2 + (*iter) + 1;
+		totalLen += 2 + iter + 1;
 	}
 
 	totalLen++;
--- a/Common++/src/TablePrinter.cpp
+++ b/Common++/src/TablePrinter.cpp
@@ -4,6 +4,7 @@
 #include <sstream>
 #include <iostream>
 #include <iterator>
+#include <numeric>
 #include <utility>
 #include "TablePrinter.h"
 #include "Logger.h"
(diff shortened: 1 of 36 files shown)


Common++/src/TablePrinter.cpp
```diff
@@ -88,13 +89,7 @@ void TablePrinter::printSeparator()
...
 	}
 
-	int totalLen = 0;
-	for (std::vector<int>::iterator iter = m_ColumnWidths.begin(); iter != m_ColumnWidths.end(); iter++)
-	{
-		totalLen += 2 + (*iter) + 1;
-	}
-
-	totalLen++;
+	int totalLen = std::accumulate(m_ColumnWidths.begin(), m_ColumnWidths.end(), m_ColumnWidths.size() * 3) + 1;
```
tigercosmos, line 92: more clearer!

Packet++/src/BgpLayer.cpp
```diff
@@ -416,41 +416,41 @@ size_t BgpUpdateMessageLayer::prefixAndIPDataToByteArray(const std::vector<prefi
 
 	size_t dataLen = 0;
 
-	for (std::vector<prefix_and_ip>::const_iterator iter = prefixAndIpData.begin(); iter != prefixAndIpData.end(); iter++)
+	for (const auto &iter : prefixAndIpData)
```
tigercosmos, line 419: can we rename `iter` to a meaningful name, e.g. `prefixAndIp`? `iter` is hard for readers to understand.

Packet++/src/IPv4Layer.cpp
```diff
@@ -37,9 +37,9 @@ IPv4OptionBuilder::IPv4OptionBuilder(IPv4OptionTypes optionType, const std::vect
 	m_RecValue[curOffset++] = 0; // init pointer value
 
 	bool firstZero = false;
-	for (std::vector<IPv4Address>::const_iterator iter = ipList.begin(); iter != ipList.end(); iter++)
+	for (const auto &iter : ipList)
```
tigercosmos, line 40: I would use `for (const auto &ipAddress: ipList)`

tigercosmos:
@egecetin I would suggest not using `iter` directly.  Instead, it would be better to use a meaningful name for each element.

Common++/src/TablePrinter.cpp
seladb, line 92: nit: maybe use `auto` for `totalLen` and for the `for` loop later?

seladb (approved):
Please see one small comment, otherwise approved!

Common++/src/TablePrinter.cpp
egecetin: @seladb What about this e38dfa1 ? I also removed `for` loop and use string constructor

egecetin:
A not important note, I noticed there are still some postfix operators not generating warnings. For example you mentioned a for loop in https://github.com/seladb/PcapPlusPlus/pull/1251#discussion_r1425046426 which has `index++` as increment operator. I try to fix all of I noticed but probably some of them will raise warning at a point where `cppcheck` or `pre-commit` updated.

PR #1252, opened 2023-12-05
Check `pcap_dispatch` return value

seladb:
We add this because:
1. Checking the return value is the right thing to do 😄 
2. Because the fuzzers build fails, here is an example: https://github.com/seladb/PcapPlusPlus/actions/runs/7092633076/job/19304338003

[According to the documentation ](https://linux.die.net/man/3/pcap_dispatch) only a return value of `-1` is considered an error, the rest of the values are not errors.

### DIFF
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -177,7 +177,7 @@ void PcapLiveDevice::captureThreadMain()
 	{
 		while (!m_StopThread)
 		{
-			if (pcap_dispatch(m_PcapDescriptor, -1, onPacketArrives, (uint8_t*)this) == -1)
+			if (pcap_dispatch(m_PcapDescriptor, -1, onPacketArrives, reinterpret_cast<uint8_t*>(this)) == -1)
 			{
 				PCPP_LOG_ERROR("pcap_dispatch returned an error: " << pcap_geterr(m_PcapDescriptor));
 				m_StopThread = true;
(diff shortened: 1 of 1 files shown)


Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -176,12 +176,24 @@ void PcapLiveDevice::captureThreadMain()
 	if (m_CaptureCallbackMode)
 	{
 		while (!m_StopThread)
-			pcap_dispatch(m_PcapDescriptor, -1, onPacketArrives, (uint8_t*)this);
+		{
+			if (pcap_dispatch(m_PcapDescriptor, -1, onPacketArrives, (uint8_t*)this) == -1)
```
egecetin, line 180: Might be `static_cast/reinterpret_cast` instead of a C style cast

egecetin (approved):

seladb:
> @seladb the CI is still failed. https://github.com/seladb/PcapPlusPlus/actions/runs/7097841699/job/19319122960

Yes, I'm not sure why... maybe it'd because the fuzzers job clones the `master` branch instead of the current branch. I pinged @sashashura to assist

tigercosmos:
@seladb  actually I know why, please see this: https://github.com/seladb/PcapPlusPlus/pull/1245/commits/8125da5e5e9587a96a9d2a84975ab6ddd9b27e79

tigercosmos:
you shouldn't let it return `0` if `pacp_dispatch` got errors; otherwise the behaviour changes.

seladb:
> you shouldn't let it return `0` if `pacp_dispatch` got errors; otherwise the behaviour changes.

I'm not sure I understand... 🤔

PR #1254, opened 2023-12-05
Extract CIFuzz to a separate workflow

sashashura:


.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,33 @@
+name: Fuzz CI
+on:
+  pull_request:
+    branches: ["master", "dev"]
```
seladb, line 4: We always open PRs against `dev`, is there a reason to include `master` also?

sashashura: No specific reason. I have removed the master.

Issue #1248, opened 2023-11-29
introducing clang-format


tigercosmos:
### Formatting applied to:
- [x] Common++
- [x] Packet++ (https://github.com/seladb/PcapPlusPlus/pull/1510)
- [x] Packet++ Tests (#1456)
- [x] Pcap++ (#1516)
- [x] Pcap++ Tests (#1456)
- [x] PcapTestFramework
- [x] Fuzzers
- [x] Examples (#1464)
- [x] Examples Tests (#1456)

seladb:
We can add clang-format if we can create a custom format that doesn't change the entire code base 🤯 

clang-format supports 6 [predefined styles](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#clang-format-style-options): LLVM, Google, Chromium, Mozilla, WebKit, Microsoft.

I think I went over them before and each one will create tons of changes compared to the style we currently have...

egecetin:
I think custom clang-format will be great, manually formatting can be problematic and both reviewers and contributors miss some minor things. A simple example,

![image](https://github.com/seladb/PcapPlusPlus/assets/64282645/3f0ca348-2d9d-45f6-805d-66b73cafd01c)

Mixed use of tab and spaces (4 tabs + spaces equivalent for another 4 tabs).

Another example is from my PRs which I noticed yesterday, the alignment of public private keyword. Some PRs indent them with spaces but most of the code base not align them. Probably both of us miss them during merge @seladb

seladb:
@egecetin I totally agree! Unfortunately I know there are many formatting issues in the current code base 😕 

@egecetin @tigercosmos would you consider working on a custom clang-format style?

tigercosmos:
I will find some time to work on this. Let's see how we can customize the style.

tigercosmos:
@clementperon  
The intent behind incorporating clang-format is to standardize the coding style.

At present, the CI system does not verify the code format, and relying on manual inspection is not consistently reliable. Additionally, with the integration of a code formatter, individuals can effortlessly adhere to the official standard by using IDEs.

While acknowledging that a substantial merge request may ensue, I contend that the advantages outweigh the disadvantages.

clementperon:
Sorry if I was not clear, but I'm 100% in favor to have a clang-format checker in the CI.

I'm not convince to have a custom clang-format coding style.

IMP I would stick with the LLVM default coding style

tigercosmos:
[...]

PR #1257, opened 2023-12-07
Draft: clang-format init

tigercosmos:
For #1248 This is currently a draft for testing `clang-format`.

The `.clang-format` is now the default setting from clang-format 11

Welcome options for the coding styles.

Things to discuss:
- The official clang-format version (based on difference OS, the default clang-format are also different. Once we choose the official version, users may need to go to LLVM website to download the corresponding version. For example, Debian 11 default is clang-format-11, Ubuntu 22 default is clang-format 14.
- The rules in `.clang-format`. Basically we follow LLVM style, but may have some customized styles settings.

People can try to clone this patch, modify the `.clang-format` file, and run `./run_format.bash` to see the result.

main file to check:
- `.clang-format`
- `.github/workflows/format.yml`

cc. @seladb @clementperon  @egecetin @echo-Mike  @sashashura


.github/workflows/format.yml
```diff
@@ -0,0 +1,37 @@
+name: lint
```
egecetin, line 1: `pre-commit` hooks already supports to clang-format check. Maybe instead of adding a new runner enabling it from config file (https://github.com/seladb/PcapPlusPlus/blob/3ffed8250fde57733f9895df4461a7096e21aae3/.pre-commit-config.yaml#L17-L22) might be better. People can run `pre-commit run --all-files` to check locally. As a default it search `.clang-format` in workspace directory

PR #1259, opened 2023-12-07
Fix SingleCommandTextProtocol parsing

egecetin:
Due to misuse of `find_last_of` and `find_first_of` some packets are not parsed correctly. This PR fixes them with replacing `rfind` and `find` to correct behavior.

### DIFF
--- a/Packet++/src/SingleCommandTextProtocol.cpp
+++ b/Packet++/src/SingleCommandTextProtocol.cpp
@@ -6,7 +6,7 @@
 #define ASCII_HYPHEN 0x2d
 #define ASCII_SPACE 0x20
 #define MAX_COMMAND_LENGTH 9 // From SMTP command "STARTTLS" + 1 byte hyphen or space
-#define MIN_PACKET_LENGTH 2	 // CRLF
+#define MIN_PACKET_LENGTH 2 // CRLF
 
 namespace pcpp
 {
--- a/Packet++/src/SingleCommandTextProtocol.cpp
+++ b/Packet++/src/SingleCommandTextProtocol.cpp
@@ -5,6 +5,8 @@
 
 #define ASCII_HYPHEN 0x2d
 #define ASCII_SPACE 0x20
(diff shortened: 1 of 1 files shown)


Packet++/src/SingleCommandTextProtocol.cpp
```diff
@@ -5,29 +5,46 @@
...
+		if (posHyphen == std::string::npos && posSpace == std::string::npos && posCRLF == std::string::npos)
+			return 0;
+		// Hyphen not found while <SP> exists or both found but <SP> detected earlier
+		else if ((posHyphen == std::string::npos && posSpace != std::string::npos) ||
+				 (posSpace != std::string::npos && posHyphen != std::string::npos && posSpace < posHyphen))
+			return posSpace;
+		// <SP> not found while hyphen exists or both found but hyphen detected earlier
+		else if ((posSpace == std::string::npos && posHyphen != std::string::npos) ||
+				 (posSpace != std::string::npos && posHyphen != std::string::npos && posHyphen < posSpace))
+			return posHyphen;
```
seladb, line 42: maybe a slightly simpler way to write it can be:
```cpp
else if (posHyphen != std::string::npos || posSpace != std::string::npos)
{
	return std::min(posSpace, posHyphen);
}
```

This should work since `npos` should be the max value representable by `size_t`

seladb (approved):
See one comment

Packet++/src/SingleCommandTextProtocol.cpp
```diff
@@ -5,29 +5,41 @@
 
 #define ASCII_HYPHEN 0x2d
 #define ASCII_SPACE 0x20
+#define MAX_COMMAND_LENGTH 9 // From SMTP command "STARTTLS" + 1 byte hyphen or space
+#define MIN_PACKET_LENGTH 2	 // CRLF
```
seladb, line 9: very nit: a redundant tab before `// CRLF`

PR #1273, opened 2023-12-25
using std::function for callback

### LINKED ISSUE #1261: Why OnPacketArrivesCallback is not std::function
Hello and thank you for this great project! I wonder why OnPacketArrivesCallback is defined as a C-style function pointer, while there's a C++ std::function.

Sincerely,
Seva D.

tigercosmos:
fixed #1261

using `std::fucntion` for `startCapture` related APIs.

`std::function` itself supports lamba and c-style functions, so no need to worry about backward compatibility.

### DIFF
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -30,19 +30,17 @@ namespace pcpp
 	class PcapLiveDevice;
 
 	/**
-	 * @typedef OnPacketArrivesCallback
 	 * A callback that is called when a packet is captured by PcapLiveDevice
-	 * @param[in] pPacket A pointer to the raw packet
-	 * @param[in] pDevice A pointer to the PcapLiveDevice instance
+	 * @param[in] packet A pointer to the raw packet
+	 * @param[in] device A pointer to the PcapLiveDevice instance
 	 * @param[in] userCookie A pointer to the object put by the user when packet capturing stared
 	 */
 	using OnPacketArrivesCallback = std::function<void(RawPacket*, PcapLiveDevice*, void*)>;
 
 	/**
-	 * @typedef OnPacketArrivesStopBlocking
 	 * A callback that is called when a packet is captured by PcapLiveDevice
-	 * @param[in] pPacket A pointer to the raw packet
-	 * @param[in] pDevice A pointer to the PcapLiveDevice instance
+	 * @param[in] packet A pointer to the raw packet
+	 * @param[in] device A pointer to the PcapLiveDevice instance
 	 * @param[in] userCookie A pointer to the object put by the user when packet capturing stared
 	 * @return True when main thread should stop blocking or false otherwise
 	 */
@@ -50,7 +48,6 @@ namespace pcpp
 
 
 	/**
-	 * @typedef OnStatsUpdateCallback
 	 * A callback that is called periodically for stats collection if user asked to start packet capturing with periodic stats collection
 	 * @param[in] stats A reference to the most updated stats
 	 * @param[in] userCookie A pointer to the object put by the user when packet capturing stared
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -6,6 +6,7 @@
 #include <vector>
 #include <string.h>
 #include <thread>
+#include <functional>
 
 #include "IpAddress.h"
 #include "Packet.h"
@@ -35,7 +36,7 @@ namespace pcpp
 	 * @param[in] pDevice A pointer to the PcapLiveDevice instance
 	 * @param[in] userCookie A pointer to the object put by the user when packet capturing stared
 	 */
-	typedef void (*OnPacketArrivesCallback)(RawPacket* pPacket, PcapLiveDevice* pDevice, void* userCookie);
+	using OnPacketArrivesCallback = std::function<void(RawPacket*, PcapLiveDevice*, void*)>;
 
 	/**
 	 * @typedef OnPacketArrivesStopBlocking
@@ -45,7 +46,7 @@ namespace pcpp
 	 * @param[in] userCookie A pointer to the object put by the user when packet capturing stared
 	 * @return True when main thread should stop blocking or false otherwise
 	 */
-	typedef bool (*OnPacketArrivesStopBlocking)(RawPacket* pPacket, PcapLiveDevice* pDevice, void* userCookie);
+	using OnPacketArrivesStopBlocking = std::function<bool(RawPacket*, PcapLiveDevice*, void*)>;
 
 
 	/**
@@ -54,7 +55,7 @@ namespace pcpp
 	 * @param[in] stats A reference to the most updated stats
 	 * @param[in] userCookie A pointer to the object put by the user when packet capturing stared
 	 */
-	typedef void (*OnStatsUpdateCallback)(IPcapDevice::PcapStats& stats, void* userCookie);
+	using OnStatsUpdateCallback = std::function<void(IPcapDevice::PcapStats&, void*)>;
 
 	// for internal use only
 	typedef void* (*ThreadStart)(void*);
--- a/Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
+++ b/Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
@@ -517,19 +517,9 @@ PTF_TEST_CASE(TestPcapLiveDeviceWithLambda)
 } // TestPcapLiveDeviceWithLambda
 
 
+
 PTF_TEST_CASE(TestPcapLiveDeviceBlockingModeWithLambda)
 {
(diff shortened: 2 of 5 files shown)


tigercosmos:
@seladb  the PR is ready for review.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -65,6 +65,31 @@ static bool packetArrivesBlockingModeStartCapture(pcpp::RawPacket* rawPacket, pc
 	return false;
 }
 
+static bool packetArrivesBlockingModeStartCaptureUsingLambda(pcpp::RawPacket* rawPacket, pcpp::PcapLiveDevice* dev, void* userCookie)
+{
+	pcpp::Logger::getInstance().suppressLogs();
+	if (dev->startCaptureBlockingMode(packetArrivesBlockingModeTimeout, nullptr, 5) != 0)
+		return false;
+
+	auto packetArrivesLambda = [](pcpp::RawPacket* rawPacket, pcpp::PcapLiveDevice* pDevice, void* userCookie)
```
tigercosmos, line 74: just copy the c style function at the beginning of the file.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -65,6 +65,31 @@ static bool packetArrivesBlockingModeStartCapture(pcpp::RawPacket* rawPacket, pc
 	return false;
 }
 
+static bool packetArrivesBlockingModeStartCaptureUsingLambda(pcpp::RawPacket* rawPacket, pcpp::PcapLiveDevice* dev, void* userCookie)
```
tigercosmos, line 68: just copy the test before this test.

seladb, line 68: There is no test that uses this static method, so it doesn't really run anywhere

Pcap++/header/PcapLiveDevice.h
```diff
@@ -35,7 +36,7 @@ namespace pcpp
 	 * @param[in] pDevice A pointer to the PcapLiveDevice instance
 	 * @param[in] userCookie A pointer to the object put by the user when packet capturing stared
 	 */
-	typedef void (*OnPacketArrivesCallback)(RawPacket* pPacket, PcapLiveDevice* pDevice, void* userCookie);
+	using OnPacketArrivesCallback = std::function<void(RawPacket*, PcapLiveDevice*, void*)>;
```
seladb, line 39: I think we should modify the doxygen documentation. The `@typedef` keyword is no longer relevant. I'm not sure how to document a `using` statement, and also how to document its parameters 🤔 

This also applies to the other changes below

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: ops. true. I always forget to put the new test into the list.

Pcap++/header/PcapLiveDevice.h
tigercosmos: `using` is sometimes used for namespace, however here is for the type alias.

There is no specific notation for `using` but I think using `@typedef` is fine here, because the overall idea is the same.

I just checked the doxygen output, I think it's good to keep the current doxygen comment.
image

tigercosmos: another choice here is to use `@var`, but I don't think there is difference.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -439,6 +464,11 @@ PTF_TEST_CASE(TestPcapLiveDeviceBlockingMode)
 		PTF_ASSERT_EQUAL(liveDev->startCaptureBlockingMode(packetArrivesBlockingModeStartCapture, &packetCount, 30), 1);
 		PTF_ASSERT_EQUAL(packetCount, 5);
 
+		// try to start capture from within the callback, verify no error
+		packetCount = 0;
+		PTF_ASSERT_EQUAL(liveDev->startCaptureBlockingMode(packetArrivesBlockingModeStartCaptureUsingLambda, &packetCount, 30), 1);
```
tigercosmos, line 469: test the callback as lambda here.

Pcap++/header/PcapLiveDevice.h
seladb: I tried to remove the `@typedef` lines and it still works, so maybe we can do that?

Also - please notice I changed `pPacket` to `packet` and `pDevice` to `device`:

```cpp
	/**
	 * A callback that is called when a packet is captured by PcapLiveDevice
	 * @param[in] packet A pointer to the raw packet
	 * @param[in] device A pointer to the PcapLiveDevice instance
	 * @param[in] userCookie A pointer to the object put by the user when packet capturing stared
	 */
	using OnPacketArrivesCallback = std::function<void(RawPacket*, PcapLiveDevice*, void*)>;

	/**
	 * A callback that is called when a packet is captured by PcapLiveDevice
	 * @param[in] packet A pointer to the raw packet
	 * @param[in] device A pointer to the PcapLiveDevice instance
	 * @param[in] userCookie A pointer to the object put by the user when packet capturing stared
	 * @return True when main thread should stop blocking or false otherwise
	 */
	using OnPacketArrivesStopBlocking = std::function<bool(RawPacket*, PcapLiveDevice*, void*)>;


	/**
	 * A callback that is called periodically for stats collection if user asked to start packet capturing with periodic stats collection
	 * @param[in] stats A reference to the most updated stats
	 * @param[in] userCookie A pointer to the object put by the user when packet capturing stared
	 */
	using OnStatsUpdateCallback = std::function<void(IPcapDevice::PcapStats&, void*)>;
```

image

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: addressed

Pcap++/header/PcapLiveDevice.h
tigercosmos: addressed

tigercosmos:
@seladb  all issues above are addressed.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -477,6 +477,194 @@ PTF_TEST_CASE(TestPcapLiveDeviceBlockingMode)
...
+
+	PTF_PRINT_VERBOSE("Total sleep time: " << totalSleepTime << " secs");
+
+	liveDev->stopCapture();
+	PTF_ASSERT_GREATER_THAN(packetCount, 0);
+	PTF_ASSERT_GREATER_OR_EQUAL_THAN(numOfTimeStatsWereInvoked, totalSleepTime-2);
+} // TestPcapLiveDeviceWithLambda
+
+
+PTF_TEST_CASE(TestPcapLiveDeviceBlockingModeWithLambda)
```
seladb, line 520: I don't think we should duplicate the entire test to test lambda. We just need to add one lambda test to the existing [`TestPcapLiveDeviceBlockingMode`](https://github.com/seladb/PcapPlusPlus/blob/dd9b3d8aa1b7403d08be7960bf01cc1be94575cd/Tests/Pcap%2B%2BTest/Tests/LiveDeviceTests.cpp#L379) as I suggested [here](https://github.com/seladb/PcapPlusPlus/pull/1273#discussion_r1436690333):

```cpp
PTF_TEST_CASE(TestPcapLiveDeviceBlockingMode)
{
	...

	auto packetArrivesBlockingModeLambda = [](pcpp::RawPacket* rawPacket, pcpp::PcapLiveDevice* dev, void* userCookie)
	{
		int* packetCount = (int*)userCookie;
		if ((*packetCount) == 5)
			return true;

		(*packetCount)++;
		return false;
	};

	int packetCount = 0;
		
	PTF_ASSERT_EQUAL(liveDev->startCaptureBlockingMode(packetArrivesBlockingModeLambda, &packetCount, 30), 1);
	PTF_ASSERT_EQUAL(packetCount, 5);

	...
} // TestPcapLiveDeviceBlockingMode
```

seladb:
@tigercosmos please see this one last comment: https://github.com/seladb/PcapPlusPlus/pull/1273#discussion_r1436722937
Other than that it looks good to me!

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: I misunderstood. okay, let me just keep one `startCaptureBlockingMode` call in `TestPcapLiveDeviceBlockingMode`

tigercosmos:
@seladb  should be fine now.

seladb (approved):

seladb:
Thank you @tigercosmos for working on this, much appreciated! 🙏

PR #1278, opened 2024-01-15
Auto precommit update

seladb:
Update pre-commit hooks to latest

Auto generated


egecetin:
@seladb pre-commit failed because of definitions here (COFIG -> CONFIG),

https://github.com/seladb/PcapPlusPlus/blob/f1b570041c5839cd04cbe470088256068aea54bb/Pcap%2B%2B/src/DpdkDevice.cpp#L46-L51

since they are in `DpdkDevice.cpp` probably we are only using inside from library, right? I assume it won't be a direct breaking change otherwise I can add a ignore rule

seladb:
> @seladb pre-commit failed because of definitions here (COFIG -> CONFIG),

@egecetin yes you're right! Fixed in https://github.com/seladb/PcapPlusPlus/pull/1278/commits/ece55c339c48111a6151514e00a5d45812af97e5

egecetin (approved):

PR #1279, opened 2024-01-16
Remove FreeBSD 12 support

seladb:
FreeBSD 12 reached its EOL: https://www.freebsd.org/security/unsupported/ and its EOL spaces were killed also, and that's probably why the build failed: https://cirrus-ci.com/task/5783173880086528

Cirrus CI doesn't support this version anymore: https://github.com/cirruslabs/cirrus-ci-docs/pull/1258

It's time to remove FreeBSD 12 support.

I tried to add FreeBSD 14 support instead but a few tests failed:
- `TestGetMacAddress`
- `TestPcapFiltersLive`
- `TestDNSResolver::test_hostname_not_exist`

It's probably because all of them need access to the Internet and in the FreeBSD 14 this access may not be available. I need to explore this further, but in the meantime we can just remove FreeBSD 12 in order to unblock other PRs

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,7 +6,7 @@ task:
   name: Test FreeBSD
   freebsd_instance:
     matrix:
-      image_family: freebsd-12-3
+      image_family: freebsd-14-0
       image_family: freebsd-13-1
   install_script:
     - pkg install -y bash cmake git gmake gsed libpcap tcpreplay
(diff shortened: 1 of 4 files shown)


clementperon:
@seladb maybe the hack $> chmod a+rw /dev/bpf* should be updated for FreeBSD 14

clementperon (approved):

egecetin (approved):

seladb:
> chmod a+rw /dev/bpf

@clementperon you mean the change that was done in this PR? https://github.com/seladb/PcapPlusPlus/pull/1046
Yes, maybe that's the issue, thanks for noticing!
Do you think you'll have time to try to fix it?

PR #1281, opened 2024-01-19
CI: bump FreeBSD to 13.2 and 14.0

clementperon:
Test FreeBSD on Cirrus

### DIFF
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -39,9 +39,15 @@
 #include <sys/sysctl.h>
 #endif
 
-// On Mac OS X and FreeBSD timeout of -1 causes pcap_open_live to fail so value of 1ms is set here.
+// TODO: FIX FreeBSD
+// On Mac OS X and FreeBSD timeout of -1 causes pcap_open_live to fail.
+// A value of 1ms first solve the issue but since Jan. 2024 an issue
+// seems to make pcap_breakloop() to not properly break pcap_dispatch()
+// After multiple test a 10ms is the minimum to fix pcap_breakloop().
 // On Linux and Windows this is not the case so we keep the -1 value
-#if defined(__APPLE__) || defined(__FreeBSD__)
+#if defined(__FreeBSD__)
+#define LIBPCAP_OPEN_LIVE_TIMEOUT 10
+#elif defined(__APPLE__)
 #define LIBPCAP_OPEN_LIVE_TIMEOUT 1
 #else
 #define LIBPCAP_OPEN_LIVE_TIMEOUT -1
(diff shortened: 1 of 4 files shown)


Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -42,7 +42,7 @@
 // On Mac OS X and FreeBSD timeout of -1 causes pcap_open_live to fail so value of 1ms is set here.
 // On Linux and Windows this is not the case so we keep the -1 value
 #if defined(__APPLE__) || defined(__FreeBSD__)
-#define LIBPCAP_OPEN_LIVE_TIMEOUT 1
+#define LIBPCAP_OPEN_LIVE_TIMEOUT 10
```
seladb, line 45: I wonder if it works with a lower value like 2ms 🤔 
Also - maybe separate FreeBSD and Apple? Apple could stay 1ms and we can have a different value for FreeBSD

clementperon: I will do the test, but we need to found the proper fix.
This is just to highlight that it's not an issue with the CI or the user privilege that has changed.

clementperon: maybe we should move to a nonblock version ?

clementperon: Timeout 1ms: 3x KO
Timeout 2ms: 3x KO
Timeout 3ms: 3x KO
Timeout 4ms: 3x KO
Timeout 5ms: 3x KO
Timeout 6ms: 3x KO
Timeout 7ms: 3x KO
Timeout 9ms: 3x KO
Timeout 10ms: 3x OKKK
Timeout 11ms: 3x OKKK
Timeout 15ms: 3x OKKK
Timeout 20ms: 3x OKKK
Timeout 30ms: 3x OKKK

seladb: @clementperon I'm not sure what this output means 🤔

clementperon: Sorry not clear, I test each configuration 3 times to be sure it was repeatable

clementperon: To be honest I don't know, I have tested that on a FreeBSD VM Aarch64 on MacOS.

I'm not sure we should merge this without understand why no?

clementperon: I have pushed an update with a proper comment

seladb (approved):

Pcap++/src/PcapLiveDevice.cpp
seladb: > To be honest I don't know, I have tested that on a FreeBSD VM Aarch64 on MacOS.
> 
> I'm not sure we should merge this without understand why no?

Honestly, I'm not sure how we can test it... seems like an internal issue of libpcap on FreeBSD 🤷‍♂️

clementperon: Ideally we would open a bug with  a small sample code to reproduce the issue

seladb: Sure, you can open a bug. But even if they fix it we need to support older libpcap versions, so we'll need this hack anyway...

tigercosmos:
thanks for the PR. waiting for this merged.

seladb:
@clementperon I'm merging this PR because is blocks other PRs. We can try to figure out the issue with FreeBSD 14 later and maybe fix it

PR #1285, opened 2024-01-23
Use std::unordered_map instead of std::map

tigercosmos:
part of #977

changed `std::map` to `std::unordered_map`

there are a few cases that still keep using `std::map` due to the original intention usage of the `std::map` properties.

### DIFF
--- a/Common++/header/LRUList.h
+++ b/Common++/header/LRUList.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#include <map>
+#include <unordered_map>
 #include <list>
 
 #if __cplusplus > 199711L || _MSC_VER >= 1800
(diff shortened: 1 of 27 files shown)


tigercosmos:
@seladb  ready for review. the only failed case is the CI issue.

tigercosmos:
@seladb  still happens, any idea?
image

seladb:
@tigercosmos maybe you need to change the CirrusCI image to `freebsd-version: "13.2"`?
This is done in this PR: https://github.com/seladb/PcapPlusPlus/pull/1281

tigercosmos:
@seladb  should be fine now.

seladb (approved):

seladb:
Thank you @tigercosmos for working on this! I think we finally got rid of `std::map` 😄

PR #1302, opened 2024-01-31
Add PR trigger for package workflow

egecetin:
If the package workflow is modified in a PR, workflow should run for testing. With this PR, if the package.yml modified, packaging workflow triggered without uploading the artifacts.


tigercosmos:
First time see this.. It would be helpful if you could add some explanation and comments.

egecetin:
@clementperon @seladb Do you have any idea why freebsd fails? I'm not sure I catch up with latest FreeBSD PRs

.github/workflows/package.yml
```diff
@@ -54,6 +57,7 @@ jobs:
         run: cmake --build "$BUILD_DIR" --target package
 
       - name: Upload binaries to release
+        if: github.event_name != 'pull_request'
```
clementperon, line 60: Nitpick but I would prefer to have.
In case we introduce another event.

if: github.event_name == 'tags'

.github/workflows/package.yml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     tags:
       - '*'
+  pull_request:
+    paths:
+      - '.github/workflows/package.yml'
```
clementperon, line 9: Add comment
# Also run this workflow when this package.yml is updated by a PR

clementperon:
@egecetin During the update to freebsd-vm@v1  we should have update the runs-on: macos-12 to runs-on: ubuntu-latest

See https://github.com/vmactions/freebsd-vm/issues/78

.github/workflows/package.yml
egecetin: Fixed with a35f531

.github/workflows/package.yml
egecetin: I'm not sure I understand. You mean replacing this line with `if: github.event_name == 'tags'`? And what is 'tags' here a custom definition or a standard GitHub definition

clementperon: Sorry it should be 
if: github.ref_type == 'tag'

egecetin: Fixed with c428429

clementperon (approved):

seladb (approved):

PR #1306, opened 2024-02-05
Added common base class CompositeFilter for shared logic between AndFilter and OrFilter.

Dimi1010:
AndFilter and OrFilter generally implement the same logic excluding the delimiter used when merging the inner filters into a string. This PR extracts their shared logic into a base class CompositeFilter that implements adding/removing/setting/clearing the inner filters of the composite and parsing the CompositeFilter to a string based on set delimiter during filter construction.


Pcap++/src/PcapFilter.cpp
```diff
@@ -297,48 +297,43 @@ void EtherTypeFilter::parseToString(std::string& result)
...
 	}
 }
 
-void AndFilter::setFilters(std::vector<GeneralFilter*>& filters)
+void CompositeFilter::removeFilter(GeneralFilter* filter)
 {
-	m_FilterList.clear();
-
-	for(std::vector<GeneralFilter*>::iterator it = filters.begin(); it != filters.end(); ++it)
+	std::vector<GeneralFilter*>::iterator it = m_FilterList.begin();
```
tigercosmos, line 311: this part can be shortened to:

```cpp
	for (auto it = m_FilterList.begin(); it != m_FilterList.end(); ++it)
	{
		if (*it == filter)
		{
			m_FilterList.erase(it);
			break;
		}
	}
```

Pcap++/header/PcapFilter.h
```diff
@@ -484,79 +484,105 @@ namespace pcpp
...
+		explicit CompositeFilter(std::string compositeDelimiter) : m_CompositeDelimiter(std::move(compositeDelimiter)){};
 
 		/**
-		 * A constructor that gets a list of pointers to filters and creates one filter from all filters with logical "and" between them
+		 * A constructor that gets a list of pointers to filters and creates one filter from all filters
 		 * @param[in] filters The list of pointers to filters
+		 * @param[in] compositeDelimiter A delimiter to add between different filters.
 		 */
-		explicit AndFilter(std::vector<GeneralFilter*>& filters);
+		explicit CompositeFilter(std::vector<GeneralFilter*>& filters, std::string compositeDelimiter);
```
tigercosmos, line 510: let's make it to `const string&`

Pcap++/header/PcapFilter.h
```diff
@@ -484,79 +484,105 @@ namespace pcpp
...
 		std::vector<GeneralFilter*> m_FilterList;
 	public:
-
 		/**
-		 * An empty constructor for this class. Use addFilter() to add filters to the and condition
+		 * A constructor that sets the composite delimiter. Use addFilter() to add filters to the composite filter.
+		 * @param[in] compositeDelimiter A delimiter to add between different filters.
 		 */
-		AndFilter() {}
+		explicit CompositeFilter(std::string compositeDelimiter) : m_CompositeDelimiter(std::move(compositeDelimiter)){};
```
tigercosmos, line 503: let's make it to `const string&`

Pcap++/header/PcapFilter.h
```diff
@@ -484,79 +484,105 @@ namespace pcpp
...
+	 * For deeper understanding of the filter concept please refer to PcapFilter.h
+	 */
+	class AndFilter : public CompositeFilter
+	{
+	public:
+
+		/**
+		 * An empty constructor for this class. Use addFilter() to add filters to the and condition
+		 */
+		AndFilter() : CompositeFilter(" and ") {}
```
tigercosmos, line 554: not sure if we should expose the implementation here, maybe it would be better to put it into cpp?

but I am fine with this.

Pcap++/header/PcapFilter.h
tigercosmos: same for all other `string`

Pcap++/header/PcapFilter.h
Dimi1010: The change to `const string&` will prevent parsing string literals to being passed to the constructor, which is the way AndFilter and OrFilter specify their delimiter. The change will also require either adding a r-value reference constructor being added or wrapping the string literal into a `std::string` before passing it to the constructor as a const reference.
Which solution would you prefer?

Pcap++/header/PcapFilter.h
Dimi1010: Specified it there because I saw some other constructors that just call the base constructors being written like that. Granted, those passed variables instead of constants. Anyway the change is minor and I can have it done if its preferred.

Tests/Pcap++Test/Tests/FilterTests.cpp
```diff
@@ -656,14 +656,42 @@ PTF_TEST_CASE(TestPcapFiltersOffline)
...
+	compositeFilter.setFilters(filterVec);
+	compositeFilter.parseToString(filterAsString);
+	PTF_ASSERT_EQUAL(filterAsString, "(ip and src net 10.0.0.6) and (udp)")
 
 	//-----------------------
 	//And filter - Proto + IP
 	//-----------------------
 
-	pcpp::IPFilter ipFilter("10.0.0.6", pcpp::SRC);
+	ipFilter.setAddr("10.0.0.6");
```
tigercosmos, line 691: let's not reuse the variable, which makes it hard to understand the intention.
please separate the variable for " Composite filter - Proto + IP" and "And filter - Proto + IP"

tigercosmos (changes requested):
please check the comments

Pcap++/header/PcapFilter.h
tigercosmos: r-value reference constructor makes sense.

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: Reused them because `ipFilter` and `filterVec` seemed to be reused from the And Filter test into the Complex Filter test so I figured that was fine.

While on the subject, if variable reuse was discouraged across tests, would it not be better to have each test in a block scope, that way they can't be reused accidentally?

egecetin:
My comment is a more generic, what about changing the name to `LogicFilter` instead of `CompositeFilter`. Since the filter holds `and` and `or` logical operations it might be more suitable. What do you think?

Dimi1010:
Personally, I don't feel like `LogicFilter` is a better name for it as it does not describe what the filter does at a glance. While yes, `and` and `or` logical operations are implemented through the filter, at its core the filter is combining several filters into one, not implementing logical operations specifically.

`NotFilter` also exists and at first glance my general assumption would be that it is also a `LogicFilter` as `not` is a logical operation.

Pcap++/src/PcapFilter.cpp
Dimi1010: Resolved in 7f6dba6 and dde74b9
Iterator type is left as `std::vector<GeneralFilter*>::iterator` instead of `auto` for consistency with the other functions.

Pcap++/header/PcapFilter.h
Dimi1010: Resolved in e3be18e

It ended up not requiring a r-value constructor to pass c-strings, so its just the `string` -> `const string&` change.

Pcap++/header/PcapFilter.h
Dimi1010: Resolved in e3be18e

Pcap++/header/PcapFilter.h
Dimi1010: Resolved in fd1528c

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: Resolved in 34a9f04

tigercosmos (approved):

tigercosmos:
@seladb  could you help review this? thanks

Pcap++/src/PcapFilter.cpp
```diff
@@ -348,11 +338,19 @@ void OrFilter::parseToString(std::string& result)
...
+			result += m_CompositeDelimiter;
 		}
 	}
 }
 
+AndFilter::AndFilter() : CompositeFilter(" and ") {}
+
+AndFilter::AndFilter(std::vector<GeneralFilter*>& filters) : CompositeFilter(filters, " and ") {}
+
+OrFilter::OrFilter() : CompositeFilter(" or ") {}
```
clementperon, line 350: I don't know this part of the code so maybe a stupid remark, but if the only change is this string why not using a Template?

```
enum FilterOPs { OR, AND, XOR, NOT };

Template <FilterOPs operator>
class CompositeFilter : public GeneralFilter

using OrFilter = CompositeFilter<OR>
```

tigercosmos: I also like the way to use template.
note that a template function is required to get the correct delimiter according to the op type.

clementperon: Ok let's be consistent then, fine for me.

Dimi1010: I did play around with the templated version. It actually did not turn out too bad.
Gonna push it to a different branch on my fork and link it here if you want to check it when I'm done.

Dimi1010: This is the potential solution I came up with as a templated version: https://github.com/Dimi1010/PcapPlusPlus/pull/1

Pcap++/src/PcapFilter.cpp
```diff
@@ -348,11 +338,19 @@ void OrFilter::parseToString(std::string& result)
 		result += '(' + innerFilter + ')';
 		if (m_FilterList.back() != *it)
 		{
-			result += " or ";
+			result += m_CompositeDelimiter;
 		}
 	}
 }
 
+AndFilter::AndFilter() : CompositeFilter(" and ") {}
```
seladb, line 346: I also think the operator should be an enum and not a string, probably a protected enum that users can't see, and then `parseToString()` can translate the enum to a string

Tests/Pcap++Test/Tests/FilterTests.cpp
```diff
@@ -656,6 +656,35 @@ PTF_TEST_CASE(TestPcapFiltersOffline)
 	}
 	rawPacketVec.clear();
 
+	//------------------------------
+	// Composite filter - Proto + IP
+	//------------------------------
```
seladb, line 661: If we make `CompositeFilter` abstract / not instantiable, we can probably use the existing test for `OrFilter` and `AndFilter` to test the new `removeFilter()` and `clearAllFilters()` methods

Pcap++/header/PcapFilter.h
```diff
@@ -484,79 +484,105 @@ namespace pcpp
...
 		std::vector<GeneralFilter*> m_FilterList;
 	public:
-
 		/**
-		 * An empty constructor for this class. Use addFilter() to add filters to the and condition
+		 * A constructor that sets the composite delimiter. Use addFilter() to add filters to the composite filter.
+		 * @param[in] compositeDelimiter A delimiter to add between different filters.
 		 */
-		AndFilter() {}
+		explicit CompositeFilter(const std::string& compositeDelimiter) : m_CompositeDelimiter(compositeDelimiter){};
```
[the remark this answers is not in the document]
Dimi1010: Hmm, I am pretty sure the templated version of CompositeFilter discussed in the other comment thread resolves this and the other comments (except the cpp+11 loop).

This is the pull request for it, if you want to have a look.
https://github.com/Dimi1010/PcapPlusPlus/pull/1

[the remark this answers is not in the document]
seladb: The main issue with the other PR is that it breaks backward compatibility. We don't want people to need to change their code to upgrade to the latest version. When we do that, we usually mark the "old" interface as deprecated and remove it in later versions. In this case, I'd argue that the implementation of this PR is probably good enough and is also keep backward compatibility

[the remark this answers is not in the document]
Dimi1010: Does it break? AndFilter and OrFilter are still defined as aliases of CompositeLogicFilter and they keep the same interface, so I would assume code would compile fine without changes when moving versions.

Example:
`using AndFilter = CompositeLogicFilter<CompositeLogicFilterOp::AND>;`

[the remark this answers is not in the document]
seladb: oh, I didn't notice it... maybe you can open a PR to this repo with the template version and then it'll be easier for us to review?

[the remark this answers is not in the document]
Dimi1010: That PR mostly builds uppon this one, so I am not sure about opening a new one. I guess I can add the changes to this one. Worst case can revert the merge and changes if needed.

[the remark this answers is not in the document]
seladb: sure, that sounds good! 🙏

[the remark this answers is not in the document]
tigercosmos: @Dimi1010  you can just merge https://github.com/Dimi1010/PcapPlusPlus/pull/1 into `Dimi1010:feature/composite-filter` and do some cleanup. Just be careful about the backward compacity of the public APIs when you modify the code.

Pcap++/src/PcapFilter.cpp
```diff
@@ -297,48 +297,38 @@ void EtherTypeFilter::parseToString(std::string& result)
...
-	for(std::vector<GeneralFilter*>::iterator it = filters.begin(); it != filters.end(); ++it)
-	{
-		m_FilterList.push_back(*it);
-	}
-}
-
-void AndFilter::parseToString(std::string& result)
-{
-	result.clear();
 	for(std::vector<GeneralFilter*>::iterator it = m_FilterList.begin(); it != m_FilterList.end(); ++it)
```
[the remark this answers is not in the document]
Dimi1010: Should be resolved in parts in 379e06e, 296da5b and 5192c1f.

Another thing for the C++11 cleanup would be marking the `parseToString` as `const` member function, as well as `override` and `final` where applicable, but I feel like that would be better as its own PR, as it is a bit out of the scope of this one.

Pcap++/src/PcapFilter.cpp
Dimi1010: Implemented in 2b3bb92 and polished a bit in the following commits.

Pcap++/header/PcapFilter.h
```diff
@@ -484,82 +484,117 @@ namespace pcpp
...
-		void parseToString(std::string& result);
+		void parseToString(std::string& result)
+		{
+			result.clear();
+			for (auto it = m_FilterList.cbegin(); it != m_FilterList.cend(); ++it)
+			{
+				std::string innerFilter;
+				(*it)->parseToString(innerFilter);
+				result += '(' + innerFilter + ')';
+				if (m_FilterList.back() != *it)
```
Dimi1010, line 573: This has the edge case of technically being false  in the middle of a run if a filter is added twice to the composite filter as it compares the pointer values and not the iterators. Though unsure if the change has a performance impact.

```suggestion
				if (m_FilterList.cend() - 1 != it)
```

tigercosmos: could you give an example demonstrating the issue you are concerned about?

Pcap++/header/PcapFilter.h
```diff
@@ -484,82 +484,117 @@ namespace pcpp
...
+	};
 
+	namespace detail
+	{
+		/* Could potentially be moved into CompositeLogicFilter as a private member function, with if constexpr when C++17 is the minimum supported standard.*/
 		/**
-		 * A constructor that gets a list of pointers to filters and creates one filter from all filters with logical "or" between them
-		 * @param[in] filters The list of pointers to filters
+		 * Returns the delimiter for joining filter strings for the composite logic filter operation.
+		 * @return A string li
```
tigercosmos, line 547: not finshed?

Pcap++/header/PcapFilter.h
```diff
@@ -484,82 +484,117 @@ namespace pcpp
...
+				if (m_FilterList.back() != *it)
+				{
+					result += detail::getCompositeLogicOpDelimiter<op>();
+				}
+			}
+		}
 	};
 
+	/**
+	 * @class AndFilter
```
tigercosmos, line 582: not a class, remove @class.

Pcap++/src/PcapFilter.cpp
```diff
@@ -297,60 +297,23 @@ void EtherTypeFilter::parseToString(std::string& result)
...
-
-	for(std::vector<GeneralFilter*>::iterator it = filters.begin(); it != filters.end(); ++it)
-	{
-		m_FilterList.push_back(*it);
-	}
-}
+CompositeFilter::CompositeFilter(const std::vector<GeneralFilter*>& filters) : m_FilterList(filters) {}
 
-void AndFilter::parseToString(std::string& result)
+void CompositeFilter::removeFilter(GeneralFilter* filter)
```
tigercosmos, line 302: nice improvement

Pcap++/src/PcapFilter.cpp
```diff
@@ -297,60 +297,23 @@ void EtherTypeFilter::parseToString(std::string& result)
...
-	{
-		std::string innerFilter;
-		(*it)->parseToString(innerFilter);
-		result += '(' + innerFilter + ')';
-		if (m_FilterList.back() != *it)
-		{
-			result += " or ";
-		}
-	}
+	m_FilterList = filters;
```
tigercosmos, line 316: question, clearing the origin vector and adding elements v.s. copy assignment, which one performance is better?

Dimi1010: Since C++11 copy assign attempts to reuse memory block of the vector being assigned to, if the allocator is not propagated by the vector being copied from or both vectors use the same allocator. [Link to cpp reference](https://en.cppreference.com/w/cpp/container/vector/operator%3D#:~:text=If,wise%20copy%2Dassignment.)

The default allocator does not propagate on copy so most of the time the vector block will be reused if possible.
Provided no reallocations happen due to insufficient capacity, copy assignment ends up being slightly faster due to less function calls overhead.

Ran a quick benchmark on https://perfbench.com/ and got those results:
NB: Horrible formatting, I know. Might be better actually running the sample on the site as it gives a boxplot result too.

ID                                          Calls    Median       0ns      32ns     128ns     512ns       2us       8us      33us     131us     524us       2ms       8ms      34ms     134ms     537ms
copy_assign_initial_empty                    2000      33us         0         0         0         0         0         0      1996         0         0         0         4         0         0         0
clear\_&\_push_back_initial_empty              2000     131us         0         0         0         0         0         0         0      1938         0         0        62         0         0         0
clear\_&\_push_back_initial_has_data           2000      33us         0         0         0         0         0         0      1980         0         0         0        20         0         0         0
copy_assign_enough_capacity                  2000      33us         0         0         0         0         0         0      1993         0         0         0         7         0         0         0
copy_assign_require_capacity_expansion       2000     131us         0         0         0         0         0         0         0      1969         0         0        31         0         0         0

ID                                          Calls        Min        Avg        Max      StDev
copy_assign_initial_empty                    2000     41.816     68.928    13.05ms    581.028 
clear\_&\_push_back_initial_empty              2000    388.390    795.693    13.40ms     2.25ms 
clear\_&\_push_back_initial_has_data           2000    126.226    256.895    13.13ms     1.29ms 
copy_assign_enough_capacity                  2000     44.616     91.881    13.05ms    768.054 
copy_assign_require_capacity_expansion       2000    193.277    401.726    13.21ms     1.61ms 


Benchmark code here (requires to set timeout to 5secs):
https://gist.github.com/Dimi1010/3cdecaa4a0a94df29d1862add64ea9d5

Pcap++/header/PcapFilter.h
Dimi1010: Essentially that this test will fail with the current implementation.

```
pcpp::IPFilter ipFilter("10.0.0.6", pcpp::SRC);
protoFilter.setProto(pcpp::UDP);
std::vector<pcpp::GeneralFilter*> filterVec;
filterVec.push_back(&ipFilter);
filterVec.push_back(&protoFilter);

pcpp::CompositeLogicFilter<pcpp::CompositeLogicFilterOp::AND> compositeFilter(filterVec);

// Adding the same proto filter again to the composite filter.
compositeFilter.addFilter(&protoFilter);
compositeFilter.parseToString(filterAsString);
PTF_ASSERT_EQUAL(filterAsString, "(ip and src net 10.0.0.6) and (udp) and (udp)")
```

TestPcapFiltersOffline             : FAILED (\<ProjectLocation\>\PcapPlusPlus\Tests\Pcap++Test\Tests\FilterTests.cpp:690). Assert EQUAL failed:
   Actual:   (ip and src net 10.0.0.6) and (udp)(udp)
   Expected: (ip and src net 10.0.0.6) and (udp) and (udp)

Pcap++/src/PcapFilter.cpp
tigercosmos: thanks for explaining, well I didn't see any drawback to have a clean code now :)

btw, next time, you can just have a screen shot than raw data.

Pcap++/src/PcapFilter.cpp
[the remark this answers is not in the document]
tigercosmos: sure, it's fine.

Tests/Pcap++Test/Tests/FilterTests.cpp
tigercosmos: I think it's addressed.

Pcap++/src/PcapFilter.cpp
tigercosmos: seems to be addressed.

Pcap++/header/PcapFilter.h
tigercosmos: I am thinking maybe the correct way to handle this is to check duplicate or not?

Dimi1010: Not sure, if checking for duplicates is the better option, as it would end up being a larger change for dubious benefit over the suggested change. As I said, this is an edge case. I assume most sane programs wouldn't wouldn't be adding the same filter twice regularly.

When checking for duplicates, you either have to change the backing structure from `vector` to `set`, which has an overall slower iteration speed than `vector`. Or you check for duplicates on entry keeping the vector, making `addFilter` an O(n) worst case instead of O(1) amortized.

In my opinion it is better to eat the potential performance penalty of parsing the same filter twice then adding overhead on all iterations/additions of filters by checking if there is a duplicate, since most of the time there won't be any duplicates in sane filters in the first place.

Tests/Pcap++Test/Tests/FilterTests.cpp
```diff
@@ -656,6 +656,35 @@ PTF_TEST_CASE(TestPcapFiltersOffline)
...
+	// Composite logic filter - Proto + IP
+	//------------------------------
+	{
+		pcpp::IPFilter ipFilter("10.0.0.6", pcpp::SRC);
+		protoFilter.setProto(pcpp::UDP);
+		std::vector<pcpp::GeneralFilter*> filterVec;
+		filterVec.push_back(&ipFilter);
+		filterVec.push_back(&protoFilter);
+		pcpp::CompositeLogicFilter<pcpp::CompositeLogicFilterOp::AND> compositeFilter(filterVec);
```
clementperon, line 668: Can't we use directly pcpp:AndFilter here ?

clementperon (approved):

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: Not fully actually. CompositeFilter is abstract now, but the test class just changed to use CompositeLogicFilter.

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: Yes, and probably will be, just haven't gotten to it yet. This comment also references the same suggestion: https://github.com/seladb/PcapPlusPlus/pull/1306#discussion_r1485352091

Pcap++/header/PcapFilter.h
Dimi1010: question, how does the documentation generator work with aliases?

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: Resolved e86630e

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: Resolved e86630e

Pcap++/header/PcapFilter.h
Dimi1010: Resolved 4586c5f

Pcap++/header/PcapFilter.h
tigercosmos: here is  an example
```
	/**
	 * A callback that is called when a packet is captured by PcapLiveDevice
	 * @param[in] packet A pointer to the raw packet
	 * @param[in] device A pointer to the PcapLiveDevice instance
	 * @param[in] userCookie A pointer to the object put by the user when packet capturing stared
	 * @return True when main thread should stop blocking or false otherwise
	 */
	using OnPacketArrivesStopBlocking = std::function<bool(RawPacket*, PcapLiveDevice*, void*)>;
	```

File: Pcap++/header/PcapFilter.h (hunk 11 above)
Comment (tigercosmos) [reply Pcap++/header/PcapFilter.h:573]: sure, I think it's fine, I think your current solution makes sense.
could you try to add more edge cases to verify the correctness of the behavior?

File: Pcap++/header/PcapFilter.h [hunk 17]
```diff
@@ -484,82 +484,117 @@ namespace pcpp
...
-		AndFilter() {}
+		CompositeFilter() {};
 
 		/**
-		 * A constructor that gets a list of pointers to filters and creates one filter from all filters with logical "and" between them
+		 * A constructor that gets a list of pointers to filters and creates one filter from all filters
 		 * @param[in] filters The list of pointers to filters
 		 */
-		explicit AndFilter(std::vector<GeneralFilter*>& filters);
+		explicit CompositeFilter(const std::vector<GeneralFilter*>& filters);
```
seladb, line 505: Although `CompositeFilter` is abstract and cannot be instantiated, I'd still move the c'tors to be `protected` instead of `public` to be even more explicit

Pcap++/header/PcapFilter.h
```diff
@@ -484,82 +484,117 @@ namespace pcpp
...
+	protected:
 		std::vector<GeneralFilter*> m_FilterList;
 	public:
-
 		/**
-		 * An empty constructor for this class. Use addFilter() to add filters to the and condition
+		 * An empty constructor for this class. Use addFilter() to add filters to the composite filter.
 		 */
-		AndFilter() {}
+		CompositeFilter() {};
```
seladb, line 499: We can use `= default` here:
```cpp
CompositeFilter() = default;
```

Pcap++/header/PcapFilter.h
seladb: I think the current implementation is good enough 👍

Pcap++/header/PcapFilter.h
Dimi1010: The change of the ctors from `public` to `protected` would require `CompositeLogicFilter` to have its ctors be explicitly redeclared as `public` instead of the current `using CompositeFilter::CompositeFilter;` as the ctors declared with a using declaration retain their access level from the base class.

Pcap++/header/PcapFilter.h
Dimi1010: Resolved 8ac748b

Pcap++/header/PcapFilter.h
Dimi1010: Resolved 2b0abbb

Pcap++/header/PcapFilter.h
Dimi1010: Fix implemented ebd8466

Pcap++/header/PcapFilter.h
```diff
@@ -484,82 +484,115 @@ namespace pcpp
...
 		 */
-		explicit AndFilter(std::vector<GeneralFilter*>& filters);
+		explicit CompositeFilter(const std::vector<GeneralFilter*>& filters);
 
 		/**
-		 * Add filter to the and condition
+		 * Add filter to the composite filter
 		 * @param[in] filter The filter to add
 		 */
 		void addFilter(GeneralFilter* filter) { m_FilterList.push_back(filter); }
```
Dimi1010, line 511: On the note of edge cases. Should not adding a `nullptr` to the filter be the caller's responsibility or the library's?
Since if `addFilter` is passed a `nullptr` the program burns in flames when parseToString is called.
(same issue with setFilters and the constructor from vector of pointers)

Ideally in C++11, I would resolve that issue by:
- using a `GeneralFilter const&` for `addFilter` & `removeFilter`.
- using a `std::vector<std::reference_wrapper<GeneralFilter>>` instead of `vector<GeneralFilter*>`
but those are breaking changes in the API.

tigercosmos: should be user's responsibility,  but I think we can throw an exception if it is a nullptr?

Dimi1010: I mean that works, but is there a specific reason pointers are used in places where passing nullptr is an invalid value?

IMO, a reference better describes an API that should not receive a nullptr, since the compiler would not allow it in the first place, allowing the error to be caught at compile time instead of runtime.

Perhaps a gradual effort to move to references instead of pointers for sections where nullptr value is invalid might be beneficial?

tigercosmos: It's just legacy code. I agree with your statement. However, we consider backward compatibility the most important. There isn't much benefit to changing to use reference by breaking API/deprecating API in this case.

Dimi1010: Alrighty then. As for the potential exception throw, `std::invalid_argument` or derivative of it could work, I guess?
Or just ignore the nullptr and do not throw anything?

tigercosmos: I guess `std::invalid_argument` make sense, please follow the way how other code throws

Pcap++/header/PcapFilter.h
seladb: ok, I agree 👍

Pcap++/header/PcapFilter.h
seladb: Yes, we can throw an exception in this case, although I don't think it's critical because as you mentioned, the program will anyway crash if `nullptr` is being passed to the composite filter

Dimi1010: Alright, leaving it as it is.

Tests/Pcap++Test/Tests/FilterTests.cpp
```diff
@@ -668,6 +667,29 @@ PTF_TEST_CASE(TestPcapFiltersOffline)
...
+		pcpp::OrFilter externalFilter;
+		andFilter.removeFilter(&externalFilter);
+		PTF_ASSERT_EQUAL(filterAsString, "(udp) and (ip and src net 10.0.0.6)");
+	}
+
+	andFilter.clearAllFilters();
+	andFilter.parseToString(filterAsString);
+	PTF_ASSERT_EQUAL(filterAsString, "");
+
+	andFilter.setFilters(filterVec);
```
tigercosmos, line 690: this seems redundant?  line 668 already does the same thing.

Tests/Pcap++Test/Tests/FilterTests.cpp
```diff
@@ -668,6 +667,29 @@ PTF_TEST_CASE(TestPcapFiltersOffline)
...
+	andFilter.addFilter(&ipFilter);
+	andFilter.parseToString(filterAsString);
+	PTF_ASSERT_EQUAL(filterAsString, "(ip and src net 10.0.0.6) and (udp) and (ip and src net 10.0.0.6)");
+
+	andFilter.removeFilter(&ipFilter);
+	andFilter.parseToString(filterAsString);
+	PTF_ASSERT_EQUAL(filterAsString, "(udp) and (ip and src net 10.0.0.6)");
+
+	{
+		pcpp::OrFilter externalFilter;
```
tigercosmos, line 681: can we hava a case that uses `OrFilter` to parse into string?

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: 668 tests that the vector ctor works correctly. This one tests `setFilters` specifically.

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: Added a6411a1

Tests/Pcap++Test/Tests/FilterTests.cpp
tigercosmos: ops, sorry, I didn't notice that.

tigercosmos (approved):
LGTM

tigercosmos:
@seladb  please take a look.

seladb (approved):

seladb:
Thank you @Dimi1010 for working on this change, highly appreciated! 🙏 ❤️ 
I'm sorry for my delayed response and that it took a long time to merge the PR

PR #1308, opened 2024-02-07
support macos-14 CI

tigercosmos:
add macos-14 CI support

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -215,11 +215,11 @@ jobs:
       - name: Test Tutorials
         run: cd build_examples/tutorials_bin && ./Tutorial-HelloWorld
 
-  macos:
+  macos-x86:
     runs-on: ${{ matrix.os-version }}
     strategy:
       matrix:
-        os-version: [macos-11, macos-12]
+        os-version: [macos-12, macos-13]
         arch: [x86_64, arm64]
         # Handle ZSTD build by Cirrus CI
         # config-zstd: [ON, OFF]
@@ -235,7 +235,8 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.8.x"
+          # support version: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
+          python-version: "3.12.x"
 
       - name: Configure PcapPlusPlus
         # Ensure user have access to network devices instead of giving super-user right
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -215,11 +215,11 @@ jobs:
       - name: Test Tutorials
         run: cd build_examples/tutorials_bin && ./Tutorial-HelloWorld
 
-  macos:
+  macos-x86:
     runs-on: ${{ matrix.os-version }}
     strategy:
       matrix:
-        os-version: [macos-11, macos-12]
+        os-version: [macos-12, macos-13]
```
tigercosmos, line 222: 11 is too old, update to 12, 13

.github/workflows/build_and_test.yml
```diff
@@ -235,7 +235,8 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.8.x"
+          # support version: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
+          python-version: "3.12.x"
```
tigercosmos, line 239: macos-14 only supports Python from 3.10, so let's use 3.12 for all macos versions.

tigercosmos:
@seladb  please take a look, thanks

.github/workflows/build_and_test.yml
clementperon: Regarding the documentation https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-input
I would write
    python-version: '3.12' instead of '3.12.x'

.github/workflows/build_and_test.yml
```diff
@@ -297,6 +298,80 @@ jobs:
           fail_ci_if_error: false
           verbose: true
 
+  macos-m1:
+    runs-on: ${{ matrix.os-version }}
+    strategy:
+      matrix:
+        os-version: [macos-14]
```
clementperon, line 305: so there will no MacOS x86 runner on macos-14 ?

.github/workflows/build_and_test.yml
```diff
@@ -297,6 +298,80 @@ jobs:
...
+      - name: Check installation
+        run: |
+          cmake -DPCAPPP_BUILD_COVERAGE=OFF -S . -B "$BUILD_DIR"
+          cmake --build "$BUILD_DIR" -j
+          cmake --install $BUILD_DIR
+
+      - name: Build Tutorials
+        run: |
+          mkdir -p build_examples
+          cmake -DPCAPPP_BUILD_TUTORIALS=ON -DCMAKE_OSX_ARCHITECTURES=${{ matrix.arch }} -S Examples -B build_examples
```
clementperon, line 356: As you're not cross compiling here no need to define CMAKE_OSX_ARCHITECTURES

.github/workflows/build_and_test.yml
```diff
@@ -297,6 +298,80 @@ jobs:
...
+          cd Tests/ExamplesTest
+          python -m pip install -U pip
+          python -m pip install -r requirements.txt
+          python -m pytest --interface en0 --use-sudo --root-path=../../Dist/examples_bin
+
+      - name: Check installation
+        run: |
+          cmake -DPCAPPP_BUILD_COVERAGE=OFF -S . -B "$BUILD_DIR"
+          cmake --build "$BUILD_DIR" -j
+          cmake --install $BUILD_DIR
```
clementperon, line 351: Better to put path between " " in case there is some spaces in the path

```suggestion
          cmake --install "$BUILD_DIR"
```

.github/workflows/build_and_test.yml
```diff
@@ -297,6 +298,80 @@ jobs:
...
+          # support version: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
+          python-version: "3.12.x"
+
+      - name: Configure PcapPlusPlus
+        # Ensure user have access to network devices instead of giving super-user right
+        # Seems macos-14 has more strict file permission
+        run: |
+          sudo chmod a+rw /dev/bpf*
+          sudo chmod a+rw /usr/local
+          cmake -DCMAKE_OSX_ARCHITECTURES=${{ matrix.arch }} -DPCAPPP_BUILD_COVERAGE=ON -S . -B "$BUILD_DIR"
```
clementperon, line 323: As you're not cross compiling here no need to define CMAKE_OSX_ARCHITECTURES

.github/workflows/build_and_test.yml
tigercosmos: yes, according to github official blog, macos-14 is m1 only
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

.github/workflows/build_and_test.yml
tigercosmos: fixed.

.github/workflows/build_and_test.yml
tigercosmos: fixed.

.github/workflows/build_and_test.yml
tigercosmos: fixed.

.github/workflows/build_and_test.yml
tigercosmos: sure, let me fix it.

.github/workflows/build_and_test.yml
clementperon: Really strange because they have runner images for macos-14 nad macos-14-arm64
https://github.com/actions/runner-images/tree/main/images/macos

clementperon (approved):

.github/workflows/build_and_test.yml
tigercosmos: I am also confused, but since macOS 14 is designed for arm64 architecture, shouldn't see an x86 runner

clementperon: Supported by both x86 and arm64. 
There are still plenty of Mac having an x86 CPUs

.github/workflows/build_and_test.yml
```diff
@@ -297,6 +298,79 @@ jobs:
...
+        run: |
+          sudo chmod a+rw /dev/bpf*
+          sudo chmod a+rw /usr/local
+          cmake -DPCAPPP_BUILD_COVERAGE=ON -S . -B "$BUILD_DIR"
+
+      - name: Build PcapPlusPlus
+        run: cmake --build "$BUILD_DIR" -j
+
+      - name: Check architecture
+        run: lipo $BUILD_DIR/Pcap++/libPcap++.a -verify_arch ${{ matrix.arch }}
```
clementperon, line 328: you need to remove this check or hard code the arch

```suggestion
        run: lipo $BUILD_DIR/Pcap++/libPcap++.a -verify_arch arm64
       ```

tigercosmos: ops, forgot this.

seladb (approved):

PR #1312, opened 2024-02-10
Cmake use optional git

clementperon:
To have branch and commit in build

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -17,11 +17,14 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
 
+      # Checkout is performed out of the container and doesn't match our user
+      - name: Fix checkout ownership
+        run: chown -R "$(id -u):$(id -g)" "$GITHUB_WORKSPACE"
+
       - name: Install dependencies
         run: |
           apk update && apk add cppcheck python3-dev
           python3 -m pip install cmake-format
-          git config --global --add safe.directory "$GITHUB_WORKSPACE"
 
       - uses: pre-commit/action@v3.0.0
 
@@ -39,7 +42,7 @@ jobs:
           git diff --exit-code
 
   linux:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     container: seladb/${{ matrix.image }}
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -72,6 +72,13 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
 
+      - name: Debug
```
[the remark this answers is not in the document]
clementperon: yes i'm investigating why the safe directory doesn't work. 
It seems that the userid that clone the repo is a random uid which is not root .

CMakeLists.txt
```diff
@@ -203,8 +203,22 @@ if(PCAPPP_USE_DPDK)
 endif()
 
 # Git Commit and Branch
-add_compile_definitions(GIT_COMMIT="${GIT_HASH}")
-add_compile_definitions(GIT_BRANCH="${GIT_BRANCH}")
+find_package(Git)
+if(Git_FOUND)
+  execute_process(
```
tigercosmos, line 208: wondering if the result was your expectation in CI?
```log
-- Found Git: /usr/bin/git (found version "2.39.2") 
-- Building from commit:a67f7c3 on branch:HEAD
```

btw, just curious why you want to add this?

egecetin: It really helps with debugging. For example, since you don't have to use the packaged version of the library and when something is wrong checking the version (for this case Git commit) is really important because the problem might be already fixed with a later version, or you can just check your local version is really up to date.

The other applications/projects can just check and print the version using https://github.com/seladb/PcapPlusPlus/blob/a48ff9219db24e81ac76cf2a11bab796da48652e/Common%2B%2B/src/PcapPlusPlusVersion.cpp#L6-L12 so you can be sure about the version of PcapPlusPlus during runtime.

tigercosmos: ops, I didn't know we had this file, then it makes sense.

I though we just print the hash during cmake config.

clementperon: Yes except the branch Head but this is due to how checkout is configured in CI.
I think it doesn't matter and I don't want to increase the checkout time as branch information is not really a lot of value

tigercosmos (approved):

egecetin (approved):

seladb (approved):

PR #1314, opened 2024-02-13
Give a option to toggle the dpdk init verification OFF

MatteO-Matic:
Disable/Enable initial verification of hugespages and loaded kernel modules by initDpdk(). Letting the dpdk EAL handle it instead.

This will make it possible to pass parameters to rte_eal_init that otherwise would be blocked.

example:
./program --no-pci --legacy-mem --vdev=net_pcap0,iface=enp1s0 --vdev=net_pcap1,iface=enp2s0 --


Pcap++/src/DpdkDeviceList.cpp
```diff
@@ -87,9 +93,12 @@ bool DpdkDeviceList::initDpdk(CoreMask coreMask, uint32_t mBufPoolSizePerDevice,
 		}
 	}
 
-	if (!verifyHugePagesAndDpdkDriver())
+	if (m_DpdkInitVerify)
```
clementperon, line 96: Nitpick but maybe more readable like this no ?

```suggestion
	if (m_DpdkInitVerify && !verifyHugePagesAndDpdkDriver())
		return false;
```

clementperon (approved):

Pcap++/src/DpdkDeviceList.cpp
MatteO-Matic: good good

Pcap++/src/DpdkDeviceList.cpp
```diff
@@ -56,6 +56,7 @@ namespace pcpp
 bool DpdkDeviceList::m_IsDpdkInitialized = false;
 CoreMask DpdkDeviceList::m_CoreMask = 0;
 uint32_t DpdkDeviceList::m_MBufPoolSizePerDevice = 0;
+bool DpdkDeviceList::m_DpdkInitVerify = true;
```
seladb, line 59: Why do we need it as a private member of `DpdkDeviceList`? Instead, we can add it as an additional argument to `DpdkDeviceList::initDpdk()` because it's only used there... 🤔

seladb:
@MatteO-Matic do you think you'll have time to address the comments in this PR so we can merge it?

Pcap++/header/DpdkDeviceList.h
```diff
@@ -173,6 +174,13 @@ namespace pcpp
 		 */
 		bool writeDpdkLogToFile(FILE* logFile);
 
+		/**
+		 * Disable/Enable initial verification of hugespages and loaded kernel modules by initDpdk()
+		 * Will not disable DPDK EAL checks
+		 * @param[in] verify boolean value
+		 */
+		static void setDpdkInitVerify(bool verify);
```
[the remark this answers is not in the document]
tigercosmos: I think keep it as non-const is fine as well

egecetin:
@MatteO-Matic Did you see the last comment?

seladb:
@MatteO-Matic I implemented my suggestion in commits https://github.com/seladb/PcapPlusPlus/pull/1314/commits/87d237e73d44c8ffd14e5ca15fd662e3faa98a16 and https://github.com/seladb/PcapPlusPlus/pull/1314/commits/8e7c536bd81db1ac962b03beedafed636a9c2671. I think it's much simpler, please let me know what you think

egecetin (approved):

tigercosmos (approved):

seladb (approved):

PR #1318, opened 2024-02-16
fix windows CI

tigercosmos:


ci/install_winpcap.bat
```diff
@@ -1,7 +1,8 @@
 :: Install WinPcap Runtime
 git submodule update --init --recursive
-git clone https://github.com/mfontanini/winpcap-installer.git
-winpcap-installer\winpcap-boundary-meter-4.1.3.exe /S
+:: use our fork to prevent the repository from being deleted
+git clone https://github.com/tigercosmos/winpcap-installer
```
seladb, line 4: I forked it under the PcapPlusPlus organization here: https://github.com/PcapPlusPlus/winpcap-installer
Maybe it's better to have it there instead of one of our personal accounts? 🤔

seladb (approved):
Please see my comment below, otherwise LGTM!

ci/install_winpcap.bat
tigercosmos: ok

PR #1319, opened 2024-02-17
Fix android package

clementperon:
With the bump to upload-artifact@v4 we don't support merging artifact with the same name and we should explicit it during the download artifact process.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -283,16 +283,16 @@ jobs:
           find . -name example-app -type d -exec rm -r {} +
           find . -name cmake -type d -exec rm -r {} +
           find . -name pkgconfig -type d -exec rm -r {} +
-          mv ${COMBINED_PACKAGE_DIR}/include/pcapplusplus/* ${COMBINED_PACKAGE_DIR}/include/
-          rmdir ${COMBINED_PACKAGE_DIR}/include/pcapplusplus/
-          mkdir -p android-package
-          mv ${COMBINED_PACKAGE_DIR} android-package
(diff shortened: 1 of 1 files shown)


tigercosmos (approved):

.github/workflows/package.yml
```diff
@@ -283,16 +283,16 @@ jobs:
...
+          mkdir -p "android-package"
+          mv "${COMBINED_PACKAGE_DIR}" "android-package"
 
       - uses: actions/upload-artifact@v4
         with:
-          name: android-package
-          if-no-files-found: error
           path: android-package
+          name: android-package-${{ matrix.target }}-${{ matrix.api-version }}
+          if-no-files-found: error
```
seladb, line 295: This is a breaking change because until now we provided all archs in the same package. This is also the way we recommend integrating PcapPlusPlus in our example: https://github.com/seladb/ToyVpn-PcapPlusPlus

Maybe we can use `@actions/upload-artifact/merge` to merge the packages after they have been created?
https://github.com/actions/upload-artifact/blob/main/merge/README.md

clementperon: This is still the case we will merge and reupload année artifact combined

seladb: Oh, this is what `merge-multiple` does?

clementperon: This allow to download multiple artifacts and to merge them in the same folder.

Then we repack it.

seladb: ok sure 👍

seladb (approved):

PR #1322, opened 2024-02-23
MacAddress interface refactoring

tigercosmos:
**New Edit**
As discussion, we give up using `const char*` constructor, and remove `inValid` method for `MacAddress`. Now if the input MAC address is invalid, the constructor of `MacAddress` will throw directly.

**Old content**
> Currently, we just assume people give 6-bytes long MAC address always in the format as `unsigned char*`
>
> However, it's very common that people use `char*` to represent bytes MAC address, for example, in linux implementation, `ifreq::ifr_hwaddr::sa_data`, which stores the MAC address in `char*`.
>
> This change doesn't break the backward compacity to take string MAC format, but also embrace the bytes format MAC address for the function `MacAddress(const char* addr)`


Tests/Packet++Test/Tests/EthAndArpTests.cpp
```diff
@@ -119,7 +119,7 @@ PTF_TEST_CASE(EthAndArpPacketParsing)
 PTF_TEST_CASE(ArpPacketCreation)
 {
 	pcpp::MacAddress srcMac("6c:f0:49:b2:de:6e");
-	pcpp::MacAddress dstMac("ff:ff:ff:ff:ff:ff:");
```
tigercosmos, line 122: also find a bug in previous test.

Tests/Pcap++Test/Tests/IpMacTests.cpp
```diff
@@ -211,6 +211,11 @@ PTF_TEST_CASE(TestMacAddress)
 	PTF_ASSERT_TRUE(macAddr4.isValid());
 	PTF_ASSERT_EQUAL(macAddr1, macAddr4);
 
+	char addrAsArr2[6] = { 0x11, 0x2, 0x33, 0x4, 0x55, 0x6 };
```
tigercosmos, line 214: test for taking bytes format MAC address in `char*`

Common++/header/MacAddress.h
```diff
@@ -42,9 +42,12 @@ namespace pcpp
 		MacAddress(const uint8_t* addr) : m_IsValid(true) { memcpy(m_Address, addr, sizeof(m_Address)); }
 
 		/**
-		 * A constructor that creates an instance of the class out of a (char*) string.
-		 * If the string doesn't represent a valid MAC address, instance will be invalid, meaning isValid() will return false
-		 * @param[in] addr A pointer to the (char*) string
+		 * @brief Constructor that initializes an instance of the class from a (char*) string representing a MAC address.
```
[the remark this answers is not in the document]
tigercosmos: fixed.

tigercosmos:
@seladb  PR is ready, I would like to have your review  :)

seladb:
@tigercosmos we already have [this c'tor](https://github.com/seladb/PcapPlusPlus/blob/e1f8668f23f3ad784e8b2dd974b5c605f6747762/Common++/header/MacAddress.h#L42) which does the exact same thing, but accepts a `uint8_t*` which is a "better hint" that it's a byte array

tigercosmos:
> @tigercosmos we already have [this c'tor](https://github.com/seladb/PcapPlusPlus/blob/e1f8668f23f3ad784e8b2dd974b5c605f6747762/Common++/header/MacAddress.h#L42) which does the exact same thing, but accepts a `uint8_t*` which is a "better hint" that it's a byte array

As I mentioned in the PR, it's common for people to use `char*` to store bytes data, and PCPP doesn't (and shouldn't) stop people from doing that.

seladb:
> > @tigercosmos we already have [this c'tor](https://github.com/seladb/PcapPlusPlus/blob/e1f8668f23f3ad784e8b2dd974b5c605f6747762/Common++/header/MacAddress.h#L42) which does the exact same thing, but accepts a `uint8_t*` which is a "better hint" that it's a byte array
> 
> @seladb As I mentioned in the PR, it's common for people to use `char*` to store bytes data, and PCPP doesn't (and shouldn't) stop people from doing that.

But `char*` and `uint8_t*` are equivalent... 🤔

tigercosmos:
> > > @tigercosmos we already have [this c'tor](https://github.com/seladb/PcapPlusPlus/blob/e1f8668f23f3ad784e8b2dd974b5c605f6747762/Common++/header/MacAddress.h#L42) which does the exact same thing, but accepts a `uint8_t*` which is a "better hint" that it's a byte array
> > 
> > 
> > @seladb As I mentioned in the PR, it's common for people to use `char*` to store bytes data, and PCPP doesn't (and shouldn't) stop people from doing that.
> 
> But `char*` and `uint8_t*` are equivalent... 🤔

Initially, unless the user carefully examines the PCPP documentation, there is a likelihood that individuals might assume that using both `uint8_t*` and `int8_t*`(`char*`) is acceptable for extracting MAC address bytes.

Furthermore, converting bytes from `int8_t*` to `uint8_t*` seems counterintuitive given the declaration "it's bytes".

seladb:
> > > > @tigercosmos we already have [this c'tor](https://github.com/seladb/PcapPlusPlus/blob/e1f8668f23f3ad784e8b2dd974b5c605f6747762/Common++/header/MacAddress.h#L42) which does the exact same thing, but accepts a `uint8_t*` which is a "better hint" that it's a byte array
> > > 
> > > 
> > > @seladb As I mentioned in the PR, it's common for people to use `char*` to store bytes data, and PCPP doesn't (and shouldn't) stop people from doing that.
> > 
> > 
> > But `char*` and `uint8_t*` are equivalent... 🤔
> 
> Initially, unless the user carefully examines the PCPP documentation, there is a likelihood that individuals might assume that using both `uint8_t*` and `int8_t*`(`char*`) is acceptable for extracting MAC address bytes.
> 
> Furthermore, converting bytes from `int8_t*` to `uint8_t*` seems counterintuitive given the declaration "it's bytes". Plus, even in linux, use `char*` to store bytes.

I'm not sure I understand... `MacAddress` has 2 c'tors: one for a string representation (e.g `"xx:xx:xx:xx:xx:xx"`) and one for byte representation. Bytes are represented as `uint8_t*` everywhere across PcapPlusPlus so this is consistent with the rest of the API and should hopefully be clear to users. I don't remember I heard any complaints about it...

tigercosmos:
> > > > > @tigercosmos we already have [this c'tor](https://github.com/seladb/PcapPlusPlus/blob/e1f8668f23f3ad784e8b2dd974b5c605f6747762/Common++/header/MacAddress.h#L42) which does the exact same thing, but accepts a `uint8_t*` which is a "better hint" that it's a byte array
> > > > 
> > > > 
> > > > @seladb As I mentioned in the PR, it's common for people to use `char*` to store bytes data, and PCPP doesn't (and shouldn't) stop people from doing that.
> > > 
> > > 
> > > But `char*` and `uint8_t*` are equivalent... 🤔
> > 
> > 
> > Initially, unless the user carefully examines the PCPP documentation, there is a likelihood that individuals might assume that using both `uint8_t*` and `int8_t*`(`char*`) is acceptable for extracting MAC address bytes.
> > Furthermore, converting bytes from `int8_t*` to `uint8_t*` seems counterintuitive given the declaration "it's bytes". Plus, even in linux, use `char*` to store bytes.
> 
> I'm not sure I understand... `MacAddress` has 2 c'tors: one for a string representation (e.g `"xx:xx:xx:xx:xx:xx"`) and one for byte representation. Bytes are represented as `uint8_t*` everywhere across PcapPlusPlus so this is consistent with the rest of the API and should hopefully be clear to users. I don't remember I heard any complaints about it...

I agree that maintaining consistency by using `uint8_t*` to represent bytes in PCPP. However, a potential challenge arises for users employing PCPP as libraries. They might encounter issues when providing `char[6]` and anticipating correct byte parsing, only to discover that this approach is deemed illegal upon consulting the documentation.
Moreover, this feature is backward compatible. This implies that whether the user provides a "xx:xx:xx:xx:xx:xx" format or bytes, both will function correctly. I think it's more intuitive for usage.

seladb:
> > > > > > @tigercosmos we already have [this c'tor](https://github.com/seladb/PcapPlusPlus/blob/e1f8668f23f3ad784e8b2dd974b5c605f6747762/Common++/header/MacAddress.h#L42) which does the exact same thing, but accepts a `uint8_t*` which is a "better hint" that it's a byte array
> > > > > 
> > > > > 
> > > > > @seladb As I mentioned in the PR, it's common for people to use `char*` to store bytes data, and PCPP doesn't (and shouldn't) stop people from doing that.
> > > > 
> > > > 
> > > > But `char*` and `uint8_t*` are equivalent... 🤔
> > > 
> > > 
> > > Initially, unless the user carefully examines the PCPP documentation, there is a likelihood that individuals might assume that using both `uint8_t*` and `int8_t*`(`char*`) is acceptable for extracting MAC address bytes.
> > > Furthermore, converting bytes from `int8_t*` to `uint8_t*` seems counterintuitive given the declaration "it's bytes". Plus, even in linux, use `char*` to store bytes.
> > 
> > 
> > I'm not sure I understand... `MacAddress` has 2 c'tors: one for a string representation (e.g `"xx:xx:xx:xx:xx:xx"`) and one for byte representation. Bytes are represented as `uint8_t*` everywhere across PcapPlusPlus so this is consistent with the rest of the API and should hopefully be clear to users. I don't remember I heard any complaints about it...
> 
> I agree that maintaining consistency by using `uint8_t*` to represent bytes in PCPP. However, a potential challenge arises for users employing PCPP as libraries. They might encounter issues when providing `char[6]` and anticipating correct byte parsing, only to discover that this approach is deemed illegal upon consulting the documentation. Moreover, this feature is backward compatible. This implies that whether the user provides a "xx:xx:xx:xx:xx:xx" format or bytes, both will function correctly. I think it's more intuitive for usage.

I agree this is backward-compatible, but I don't think converting from `char[6]` to `uint8_t[6]` is a big issue for C++ developers. The documentation is pretty clear about the parameters each c'tor takes. Maybe the right solution is to replace `char*` with `std::string` which will make things more clear. I wrote this code a long time ago and don't remember why I chose `char*` 🤷‍♂️ 
We can do is another c'tor that takes `std::string` and mark the `char*` c'tor as deprecated. What do you think?

seladb:
> > > > > > > > @tigercosmos we already have [this c'tor](https://github.com/seladb/PcapPlusPlus/blob/e1f8668f23f3ad784e8b2dd974b5c605f6747762/Common++/header/MacAddress.h#L42) which does the exact same thing, but accepts a `uint8_t*` which is a "better hint" that it's a byte array
> > > > > > > 
> > > > > > > 
> > > > > > > @seladb As I mentioned in the PR, it's common for people to use `char*` to store bytes data, and PCPP doesn't (and shouldn't) stop people from doing that.
> > > > > > 
> > > > > > 
> > > > > > But `char*` and `uint8_t*` are equivalent... 🤔
> > > > > 
> > > > > 
> > > > > Initially, unless the user carefully examines the PCPP documentation, there is a likelihood that individuals might assume that using both `uint8_t*` and `int8_t*`(`char*`) is acceptable for extracting MAC address bytes.
> > > > > Furthermore, converting bytes from `int8_t*` to `uint8_t*` seems counterintuitive given the declaration "it's bytes". Plus, even in linux, use `char*` to store bytes.
> > > > 
> > > > 
> > > > I'm not sure I understand... `MacAddress` has 2 c'tors: one for a string representation (e.g `"xx:xx:xx:xx:xx:xx"`) and one for byte representation. Bytes are represented as `uint8_t*` everywhere across PcapPlusPlus so this is consistent with the rest of the API and should hopefully be clear to users. I don't remember I heard any complaints about it...
> > > 
> > > 
> > > I agree that maintaining consistency by using `uint8_t*` to represent bytes in PCPP. However, a potential challenge arises for users employing PCPP as libraries. They might encounter issues when providing `char[6]` and anticipating correct byte parsing, only to discover that this approach is deemed illegal upon consulting the documentation. Moreover, this feature is backward compatible. This implies that whether the user provides a "xx:xx:xx:xx:xx:xx" format or bytes, both will function correctly. I think it's more intuitive for usage.
> > 
> > 
> > I agree this is backward-compatible, but I don't think converting from `char[6]` to `uint8_t[6]` is a big issue for C++ developers. The documentation is pretty clear about the parameters each c'tor takes. Maybe the right solution is to replace `char*` with `std::string` which will make things more clear. I wrote this code a long time ago and don't remember why I chose `char*` 🤷‍♂️ We can add another c'tor that takes `std::string` and mark the `char*` c'tor as deprecated. What do you think?
> > **EDIT**: I just noticed we already have a c'tor that takes `std::string`, so maybe it's time to deprecate the `char*` ctor
> 
> sounds good, keeping only `uint8_t*` and `std::string` c'tor make the API clearer. I agree we can just deprecate the `char*` ctor, but we should throw an error if the length of `char*` mismatches the length of "xx:xx:xx:xx:xx:xx" (18), otherwise it will take the user a well to discover the reason. Also, I think we should get rid of using `isValid()`, if a MAC address cannot be parsed, we should throw it directly.

In general, I agree it'd be good to throw an error if the length doesn't match the length of "xx:xx:xx:xx:xx:xx", but this might not be backward compatible, for example, if someone uses the API and assumes it never throws an exception, instead he reiles on `isValid()` to check the address validity...

tigercosmos:
(the thread has been too long, I will just not copy above comments)

@seladb what do you think we deprecate the `char*` c'tor and `isValid()` method, and add a warning log if the parse of MAC address fails. In the future, we should just throw directly.

seladb:
> (the thread has been too long, I will just not copy above comments)
> 
> @seladb what do you think we deprecate the `char*` c'tor and `isValid()` method, and add a warning log if the parse of MAC address fails. In the future, we should just throw directly.

deprecating `isValid()` is a bit weird if we don't throw an exception because we encourage users not to use it, but don't provide a better way to understand if the instance is valid or not.
I'm wondering if we should make a breaking change and just throw an exception, and then it'll make sense to deprecate or even remove `isValid()`. I'm curious what @clementperon and @egecetin think about it.

By the way, we have the same issue in `IPAddress`, IPv4Address`, `IPv6Address` classes, so whatever we decide - maybe we should make the same change there too

egecetin:
@seladb To be honest (and if I understand correctly from the thread sorry if I missed a point) I'm against the completely removing of `isValid()` function because it might come handy in some situations (check without throw). But we can separate the function from `MacAddress` itself. We can convert it to a static function with regex to check at least the format of the string matches (for binary or `uint8_t` we will just assume it is valid unfortunately). So, for class implementation we can directly throw an exception in constructor but there will be still a check function which returns true/false without throwing. So,

- Probably most people are aware of constructors can throw something so this won't hurt.
- Because of static implementation someone can check the address without constructing/deconstructing a `MacAddress` for faster/simpler usage.
- Non static implementation always returns true (since no throw from constructor) with a deprecation notice, and a static function to replace/encouragement for only format checks. 

Everything will be backward compatible as much as possible I think 🤔.

seladb:
> @seladb To be honest (and if I understand correctly from the thread sorry if I missed a point) I'm against the completely removing of `isValid()` function because it might come handy in some situations (check without throw). But we can separate the function from `MacAddress` itself. We can convert it to a static function with regex to check at least the format of the string matches (for binary or `uint8_t` we will just assume it is valid unfortunately). So, for class implementation we can directly throw an exception in constructor but there will be still a check function which returns true/false without throwing. So,
> 
>     * Probably most people are aware of constructors can throw something so this won't hurt.
> 
>     * Because of static implementation someone can check the address without constructing/deconstructing a `MacAddress` for faster/simpler usage.
> 
>     * Non static implementation always returns true (since no throw from constructor) with a deprecation notice, and a static function to replace/encouragement for only format checks.
> 
> 
> Everything will be backward compatible as much as possible I think 🤔.

@egecetin I'm not sure what you mean by static vs non-static implementation. Can you describe the interface you're proposing?

egecetin:
@seladb To give you a quick idea

Current implementation,

```
class MacAddress
{
	private:
		uint8_t m_Address[6];
		bool m_IsValid;
	
    ...

	public:
		bool isValid() const { return m_IsValid; }
    
    ...
}
```

If I understand correctly instead of using a boolean `m_IsValid` we want to throw exception from c'tor. So I propose like this,

```
class MacAddress
{
    // c'tor throws directly if something went wrong
    // so no need for m_IsValid
	private:
		uint8_t m_Address[6];

    ...

    public:
        // Which you mentioned we can deprecate c'tor for char*
        PCPP_DEPRECATED MacAddress(const char* addr)

        // Add notice for deprecation always return true for a one or two version. 
        // Recommend using static function below since deprecated
        PCPP_DEPRECATED bool isValid() { return true; }

        // A new function which is not deprecated so people can use it to just verify format for future use
        // So we keep the similar functionality
        static bool isValid(std::string &addr) {
            const std::regex pattern("^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$");
            return std::regex_match(mac, pattern);
        }
    
    ...
}
```

seladb:
> @seladb To give you a quick idea
> 
> Current implementation,
> 
> ```
> class MacAddress
> {
> 	private:
> 		uint8_t m_Address[6];
> 		bool m_IsValid;
> 	
>     ...
> 
> 	public:
> 		bool isValid() const { return m_IsValid; }
>     
>     ...
> }
> ```
> 
> If I understand correctly instead of using a boolean `m_IsValid` we want to throw exception from c'tor. So I propose like this,
> 
> ```
> class MacAddress
> {
>     // c'tor throws directly if something went wrong
>     // so no need for m_IsValid
> 	private:
> 		uint8_t m_Address[6];
> 
>     ...
> 
>     public:
>         // Which you mentioned we can deprecate c'tor for char*
>         PCPP_DEPRECATED MacAddress(const char* addr)
> 
>         // Add notice for deprecation always return true for a one or two version. 
>         // Recommend using static function below since deprecated
>         PCPP_DEPRECATED bool isValid() { return true; }
> 
>         // A new function which is not deprecated so people can use it to just verify format for future use
>         // So we keep the similar functionality but of course not compatible
>         static bool isValid(std::string &addr) {
>             const std::regex pattern("^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$");
>             return std::regex_match(mac, pattern);
>         }
>     
>     ...
> }
> ```
> 
> And same goes for IP addresses.

Ok got it, thank you for the explanation!
If we're already breaking the interface by throwing an exception in the c'tor, maybe just remove the non-static `isValid()`? That way, if someone relies on it, they'll know the interface has changed and will expect an exception in the c'tor 🤔

egecetin:
> If we're already breaking the interface by throwing an exception in the c'tor, maybe just remove the non-static `isValid()`? That  way, if someone relies on it, they'll know the interface has changed and will expect an exception in the c'tor 🤔

Good idea

tigercosmos:
@seladb  @egecetin  so do we have agreement to change the interface as:

```cpp
class MacAddress
{
    // c'tor throws directly if something went wrong
    // so no need for m_IsValid
	private:
		uint8_t m_Address[6];

    ...

    public:

        // "xx:xx:xx:xx:xx:xx", may throw if invalid
        MacAddress(const std::string& macAddress);

        // 6-bytes array, may throw if invalid
        MacAddress(const uint8_t* addr);

        // Which you mentioned we can deprecate c'tor for char*
        PCPP_DEPRECATED MacAddress(const char* addr)

        // Add notice for deprecation always return true for a one or two version. 
        // Recommend using static function below since deprecated
        PCPP_DEPRECATED bool isValid() { return true; }
}
```

egecetin:
@tigercosmos It is ok for me. Let's wait for @clementperon also, he might give a better idea.

clementperon:
I don't know how this code is used, but I agree that having an object and check if it's valid is not really convinient.
If the object exist it should be valid.

In other way, In my coding style I'm not use to Throwing/Catching exception.

What I don't like is that we have a parser in the constructor, but fine for me

isValid should return a bool no ?
`static bool isValid(std::string &addr)`

egecetin:
@clementperon You are right, it was typo sorry

clementperon:
I think that in case someone doesn't want to use Throw/Catch exception we would like to have 

```
MacAddress mac; // value of mac is 00-00-00-00-00

if (mac.fromString("00:11:22:33:44") != SUCESS)
  return -1;
  
```

tigercosmos:
> I think that in case someone doesn't want to use Throw/Catch exception we would like to have
> 
> ```
> MacAddress mac; // value of mac is 00-00-00-00-00
> 
> if (mac.fromString("00:11:22:33:44") != SUCESS)
>   return -1;
>   
> ```

could also be good, if we want to provide an option with a non-throw API. how about this @seladb  @egecetin  @clementperon 

```cpp
class MacAddress
{
    // c'tor throws directly if something went wrong
    // so no need for m_IsValid
	private:
		uint8_t m_Address[6];

    ...

    public:

        // "xx:xx:xx:xx:xx:xx", may throw if invalid
        MacAddress(const std::string& macAddress);

        // 6-bytes array, may throw if invalid
        MacAddress(const uint8_t* addr);

        // keep the current parsing behavior, but also throw if invalid?
        PCPP_DEPRECATED MacAddress(const char* addr)
        
        // remove the function `isValid()`, since we will throw directly now
        
        // return true if valid
        bool fromString(const std::string& otherAddress);
        
        // return true if valid
        bool fromBytes(const uint8_t* bytesAddress);
}
```

clementperon:
Nitpick but I'm not fan of passing a buffer without a size (const uint8_t* bytesAddress).

Either we should have (const uint8_t* bytesAddress, size_t bytesLength)
Or a std::array<uint8_t, 6>

tigercosmos:
> Nitpick but I'm not fan of passing a buffer without a size (const uint8_t* bytesAddress).
> 
> Either we should have (const uint8_t* bytesAddress, size_t bytesLength) Or a std::array<uint8_t, 6>

mac address is always 6 bytes, I think that's why we don't need the length. even we provide the length parameter, it will always be 6.

clementperon:
> > Nitpick but I'm not fan of passing a buffer without a size (const uint8_t* bytesAddress).
> > Either we should have (const uint8_t* bytesAddress, size_t bytesLength) Or a std::array<uint8_t, 6>
> 
> mac address is always 6 bytes, I think that's why we don't need the length. even we provide the length parameter, it will always be 6.

(const uint8_t* bytesAddress, size_t bytesLength = 6)

Yes we could default to 6, but assuming the size of a buffer is never a good idea IMO

tigercosmos:
> > > Nitpick but I'm not fan of passing a buffer without a size (const uint8_t* bytesAddress).
> > > Either we should have (const uint8_t* bytesAddress, size_t bytesLength) Or a std::array<uint8_t, 6>
> > 
> > 
> > mac address is always 6 bytes, I think that's why we don't need the length. even we provide the length parameter, it will always be 6.
> 
> (const uint8_t* bytesAddress, size_t bytesLength = 6)
> 
> Yes we could default to 6, but assuming the size of a buffer is never a good idea IMO

`(const uint8_t* bytesAddress, const size_t bytesLength = 6)` is fine to me. However, I don't think `bytesLength` has any chance of being another value. Maybe it's more readable(?)

seladb:
I think it's very convenient to have c'tors that take a string / byte array, even if it might throw an exception.

Maybe we can combine both approaches?
- Add an empty c'tor that creates a MAC address of `00-00-00-00-00`
- Add c'tors that accept `std::string` and `uint8_t*`
- Add **static** methods to create a MAC address from string and byte array, and have a flag of whether they might throw an exception

```cpp
class MacAddress
{
	// c'tor throws directly if something went wrong
	// so no need for m_IsValid
private:
	uint8_t m_Address[6];

	...

public:

	// create a MAC address of 00-00-00-00-00
	MacAddress();

	// "xx:xx:xx:xx:xx:xx", may throw if invalid
	MacAddress(const std::string& macAddress);

	// 6-bytes array, may throw if invalid
	MacAddress(const uint8_t* addr);

	// keep the current parsing behavior, but also throw if invalid?
	PCPP_DEPRECATED MacAddress(const char* addr);

	// remove the function `isValid()`, since we will throw directly now

	// if throwExceptionIfInvalid == false, return an object of value 00-00-00-00-00
	static bool fromString(const std::string& addressAsString, throwExceptionIfInvalid = true);

	// if throwExceptionIfInvalid == false, return an object of value 00-00-00-00-00
	static bool fromBytes(const uint8_t* bytesAddress, throwExceptionIfInvalid = true);
}
```

Then you can use it in one of multiple options:
```cpp
// a "zero" mac address
pcpp::MacAddress mac1;

// might throw an exception if the string is not valid
pcpp::MacAddress mac2("00:11:22:33:44");

// might throw an exception if the string is not valid
pcpp::MacAddress mac3 = pcpp::MacAddress::fromString("00:11:22:33:44");

// will not throw an exception, the return value might be 00-00-00-00-00 if the string is not valid
pcpp::MacAddress mac3 = pcpp::MacAddress::fromString("00:11:22:33:44", false);
```

tigercosmos:
> I think it's very convenient to have c'tors that take a string / byte array, even if it might throw an exception.
> 
> Maybe we can combine both approaches?
> 
> * Add an empty c'tor that creates a MAC address of `00-00-00-00-00`
> * Add c'tors that accept `std::string` and `uint8_t*`
> * Add **static** methods to create a MAC address from string and byte array, and have a flag of whether they might throw an exception
> 
> ```c++
> class MacAddress
> {
> 	// c'tor throws directly if something went wrong
> 	// so no need for m_IsValid
> private:
> 	uint8_t m_Address[6];
> 
> 	...
> 
> public:
> 
> 	// create a MAC address of 00-00-00-00-00
> 	MacAddress();
> 
> 	// "xx:xx:xx:xx:xx:xx", may throw if invalid
> 	MacAddress(const std::string& macAddress);
> 
> 	// 6-bytes array, may throw if invalid
> 	MacAddress(const uint8_t* addr);
> 
> 	// keep the current parsing behavior, but also throw if invalid?
> 	PCPP_DEPRECATED MacAddress(const char* addr);
> 
> 	// remove the function `isValid()`, since we will throw directly now
> 
> 	// if throwExceptionIfInvalid == false, return an object of value 00-00-00-00-00
> 	static bool fromString(const std::string& addressAsString, throwExceptionIfInvalid = true);
> 
> 	// if throwExceptionIfInvalid == false, return an object of value 00-00-00-00-00
> 	static bool fromBytes(const uint8_t* bytesAddress, throwExceptionIfInvalid = true);
> }
> ```
> 
> Then you can use it in one of multiple options:
> 
> ```c++
> // a "zero" mac address
> pcpp::MacAddress mac1;
> 
> // might throw an exception if the string is not valid
> pcpp::MacAddress mac2("00:11:22:33:44");
> 
> // might throw an exception if the string is not valid
> pcpp::MacAddress mac3 = pcpp::MacAddress::fromString("00:11:22:33:44");
> 
> // will not throw an exception, the return value might be 00-00-00-00-00 if the string is not valid
> pcpp::MacAddress mac3 = pcpp::MacAddress::fromString("00:11:22:33:44", false);
> ```

I am fine with the new interface. if @clementperon  @egecetin all agree, I can start on it.

clementperon:
I think if everybody is fine with throwing exception, we can remove the "throwExceptionIfInvalid = true" param.

tigercosmos:
> I think if everybody is fine with throwing exception, we can remove the "throwExceptionIfInvalid = true" param.
> 
> Because if we start to have this logic, then we should keep coherency with the rest of the code.

If I remember correctly, part of our code already adopts the throwing way, so indeed, we can just keep the consistency of throw?

egecetin:
I'm ok with removing the option throwExceptionIfInvalid = true. Exception also works for me but the current version offers a way to boolean check without throwing so I suggest adding a non-throwing version to keep same functionality only. So, if some people have a code without exception check they can still easily update their code for new `fromString` or `fromBytes` functions.

seladb:
I think I made a small mistake in the sample code above. I meant that `fromString` and `fromBytes` to return a `MacAddress` instance and not `bool`...

But if we all agree that we always throw an exception if the string is invalid, we probably don't need `fromString` and `fromBytes`. We can just have c'tors that take `std::string` and `uint8_t` (like we have now) that can throw an exception, as well as remove `isValid` and deprecate the c'tor that take `char*`. Should be rather easy to implement

tigercosmos:
> I think I made a small mistake in the sample code above. I meant that `fromString` and `fromBytes` to return a `MacAddress` instance and not `bool`...
> 
> But if we all agree that we always throw an exception if the string is invalid, we probably don't need `fromString` and `fromBytes`. We can just have c'tors that take `std::string` and `uint8_t` (like we have now) that can throw an exception, as well as remove `isValid` and deprecate the c'tor that take `char*`. Should be rather easy to implement

it seems we are on the same page now. Let me start to work on it.

tigercosmos:
@seladb  It's ready for review. As our discussion, make all constuctors throw, and delete const char* c'tor and `invalid()` for `MacAddress`. To keep the PR simple, I haven't touch `IpAddress`, that will be the next PR.

Common++/header/MacAddress.h
```diff
@@ -29,60 +29,60 @@ namespace pcpp
...
 		 */
-		MacAddress(const uint8_t* addr) : m_IsValid(true) { memcpy(m_Address, addr, sizeof(m_Address)); }
-
-		/**
-		 * A constructor that creates an instance of the class out of a (char*) string.
-		 * If the string doesn't represent a valid MAC address, instance will be invalid, meaning isValid() will return false
-		 * @param[in] addr A pointer to the (char*) string
-		 */
-		MacAddress(const char* addr) { init(addr); }
+		explicit MacAddress(const uint8_t* addr) { memcpy(m_Address, addr, sizeof(m_Address)); }
```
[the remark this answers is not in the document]
tigercosmos: @seladb I intentionally make it `explicit`. I believe employing implicit casting, as demonstrated in`layer->setTargetAddr("00:90:27:85:cf:01");,` is not a recommended practice. In my opinion, we should address this concern; otherwise, individuals might mistakenly assume that `setTargetAddr` accepts a string as a parameter initially. Moreover, the PR is already a **breaking change** regardless.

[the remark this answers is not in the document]
seladb: I agree this PR already contains breaking changes, but I believe this change is not required. Why do you think implicit casting isn't recommended in this case?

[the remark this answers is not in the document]
seladb: You're right, maybe that's the reason I kept the `char*` constructor and didn't remove it a long time ago... 🤔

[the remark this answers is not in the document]
tigercosmos: @seladb  Certainly, and we've previously talked about the reasons for rejecting the `char*` constructor. I think prioritizing a clear interface design is more valuable than relying on the convenience of implicit casting.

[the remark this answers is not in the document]
seladb: I'm curious to hear what @egecetin and @clementperon think about it. This PR introduces breaking changes anyway, but this one might be a bit more painful than the others. On the other hand, I agree with @tigercosmos that we should have a more clear interface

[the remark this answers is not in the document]
egecetin: Implicit casting is not recommended because it can lead bugs because of it _might_ cast from another type accidentally if the developer misunderstand something. I agree in this situation it gives the flexibility to developer and we should prevent too many changes somehow but to be honest I'm not sure about should we add explicit or not. Maybe with this PR we can only make other changes (without adding `explicit` keyword and related changes), then introduce a `Fix explicit issues` PR later?

I remember the same issue from one of my previous PR, reverted explicit from some class constructors because of this breaking change. I can't remember which one exactly (probably one of pre-commit or maybe cppcheck related PRs) but added this ignore rule because of them,

https://github.com/seladb/PcapPlusPlus/blob/e5bcc908e921ce336c621f721a6ed3ecbbe45b06/cppcheckSuppressions.txt#L9-L13

If we fix the issue, we can also remove this ignore rules.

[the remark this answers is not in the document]
tigercosmos: Avoiding cppcheck checks appears to resemble an anti-pattern. In addition, `explicit` is the cause of many lines of changes, the main reason is the char* constructor.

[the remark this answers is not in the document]
tigercosmos: @seladb @egecetin  well, I think if we want to have `explicit`, it's the best chance now. However, if removing `explicit` can make this PR merged, I can compromise.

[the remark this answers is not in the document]
seladb: I tried throwing this problem at ChatGPT and it offered the following solution that seems to work to convert a string to `MacAddress` explicitly`:

```cpp
template<typename T>
MacAddress(const T& addr, typename std::enable_if<std::is_convertible<T, std::string>::value>::type* = nullptr)
{
    ...
}
```

Do you think it might work? 🤔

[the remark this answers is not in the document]
tigercosmos: @seladb  smart! just added a template constructor. It's clearer now.

```
explicit MacAddress(const std::string& addr);

template<typename T, typename = typename std::enable_if<std::is_convertible<T, std::string>::value>::type>
MacAddress(const T& addr) : MacAddress(static_cast<std::string>(addr)) {};
```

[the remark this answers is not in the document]
tigercosmos: @clementperon  we already disable the cppcheck for implicit casting. We were discussing whether to keep `explicit` or not. Now I think we have the answer, we can still keep the explicit constructor for std::string, but at the same time, have a template constructor (for std::string convertible string) do implicit casting.

Common++/header/MacAddress.h
```diff
@@ -29,60 +29,69 @@ namespace pcpp
...
-		 * A constructor that creates an instance out of the initializer list. The length of the list must be equal to 6 (as MAC address is 6-byte long)
+		 * A constructor that creates an instance out of the initializer list. The length of the list must be equal to 6 (as MAC address is 6-byte long).
+		 * If the list is invalid, the constructor throws.
 		 * @param[in] addr An initializer list containing the values of type uint8_t representing the MAC address
 		 */
-		MacAddress(std::initializer_list<uint8_t> octets) : m_IsValid { octets.size() == sizeof(m_Address) }
+		MacAddress(std::initializer_list<uint8_t> octets)
 		{
-			if(m_IsValid)
+			if(octets.size() != sizeof m_Address)
```
egecetin, line 86: Nit

```suggestion
			if(octets.size() != sizeof(m_Address))
```

Common++/header/MacAddress.h
```diff
@@ -103,18 +112,21 @@ namespace pcpp
 #if __cplusplus > 199711L || _MSC_VER >= 1800
 		/**
 		 * Overload of the assignment operator
+		 * @param[in] octets An initializer list containing the values of type uint8_t representing the MAC address, the length of the list must be equal to 6
 		 */
 		MacAddress& operator=(std::initializer_list<uint8_t> octets)
 		{
-			m_IsValid = (octets.size() == sizeof m_Address);
-			if(m_IsValid)
+			if(octets.size() != sizeof m_Address)
```
egecetin, line 119: same here

Common++/src/MacAddress.cpp
```diff
@@ -15,32 +15,21 @@ std::string MacAddress::toString() const
...
 }
 
-void MacAddress::init(const char* addr)
+MacAddress::MacAddress(const std::string& address)
 {
-	const unsigned int addrLen = sizeof m_Address;
-	unsigned int i = 0;
-
-	for(; *addr != 0 && i < addrLen; ++i)
+	constexpr size_t validMacAddressLength = 17;
```
egecetin, line 20: nit: alignment different from following line

Common++/header/MacAddress.h
tigercosmos: fixed.

Common++/header/MacAddress.h
tigercosmos: fixed.

Common++/src/MacAddress.cpp
tigercosmos: fixed.

tigercosmos:
@seladb  I think the PR is ready for your review.

Common++/header/MacAddress.h
```diff
@@ -29,60 +29,69 @@ namespace pcpp
...
+		{
+			m_Address[0] = firstOctet;
+			m_Address[1] = secondOctet;
+			m_Address[2] = thirdOctet;
+			m_Address[3] = fourthOctet;
+			m_Address[4] = fifthOctet;
+			m_Address[5] = sixthOctet;
+		}
 
 #if __cplusplus > 199711L || _MSC_VER >= 1800
```
seladb, line 78: I think we can remove this `#if` now, as PcapPlusPlus supports C++11

Common++/src/MacAddress.cpp
```diff
@@ -15,32 +15,21 @@ std::string MacAddress::toString() const
...
-			m_IsValid = false;
-			return;
-		}
+            m_Address[i] = values[i];
+        }
+    }
+	else
+	{
+		throw std::invalid_argument("Invalid MAC address format, should be xx:xx:xx:xx:xx:xx");
 	}
```
seladb, line 32: nit: that's a bit nicer:
```cpp
constexpr size_t validMacAddressLength = 17;
unsigned int values[6];
if (address.size() != validMacAddressLength || sscanf(address.c_str(), "%x:%x:%x:%x:%x:%x", &values[0], &values[1], &values[2], &values[3], &values[4], &values[5]) != 6)
{
	throw std::invalid_argument("Invalid MAC address format, should be xx:xx:xx:xx:xx:xx");
}

for (int i = 0; i < 6; ++i)
{
	m_Address[i] = values[i];
}
```

Common++/src/MacAddress.cpp
```diff
@@ -15,32 +15,21 @@ std::string MacAddress::toString() const
...
-
-		// The strtol function returns zero value in two cases: when an error occurs or the string '00' is converted.
-		// This code verifies that it's the second case.
-		if(m_Address[i] == 0 && (byte[0] != '0' || byte[1] != '0'))
+        for (int i = 0; i < 6; ++i)
 		{
-			m_IsValid = false;
-			return;
-		}
+            m_Address[i] = values[i];
```
seladb, line 26: Please use tabs instead of spaces

tigercosmos: fixed

Common++/src/MacAddress.cpp
tigercosmos: fixed

Common++/header/MacAddress.h
tigercosmos: fixed

Common++/header/MacAddress.h
```diff
@@ -29,60 +29,69 @@ namespace pcpp
...
+		 * Initializes the address as 00:00:00:00:00:00.
 		 */
-		MacAddress() : m_IsValid(true) { memset(m_Address, 0, sizeof(m_Address)); }
+		MacAddress() {}
 
 		/**
-		 * A constructor that creates an instance of the class out of a byte array. The byte array length must be equal or greater to 6
-		 * (as MAC address is 6-byte long)
-		 * @todo there is no verification array length >= 6. If this is not the case, address will read uninitialized memory
+		 * A constructor that creates an instance of the class out of a byte array. The byte array length 6 (as MAC address is 6-byte long),
```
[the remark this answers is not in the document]
tigercosmos: fixed

Common++/header/MacAddress.h
```diff
@@ -29,60 +29,69 @@ namespace pcpp
 	public:
 		/**
 		 * Default constructor for this class.
-		 * Initializes object to me MacAddress::Zero
+		 * Initializes the address as 00:00:00:00:00:00.
 		 */
-		MacAddress() : m_IsValid(true) { memset(m_Address, 0, sizeof(m_Address)); }
+		MacAddress() {}
```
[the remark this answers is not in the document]
tigercosmos: fixed

tigercosmos:
@seladb  all addressed.

Common++/header/MacAddress.h
seladb, line 1: I think we can also remove the `#if __cplusplus > 199711L || _MSC_VER >= 1800` in line 7

Common++/src/MacAddress.cpp
seladb, line 1: I think `#include <stdlib.h>` (line 2) is not needed anymore

seladb (approved):
Please see a few additional comments, but otherwise LGTM!

Common++/header/MacAddress.h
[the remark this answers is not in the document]
clementperon: Then what's the difference between an implicit constructor vs explicit constructor + implicit cast ?

Issue #1288, opened 2024-01-27
IPFilter not for IPv6?


JasMetzger:
Documentation indicates that IPFilter works for IPv4. Ultimately, it creates a BPF using string such as "ip and src net <address>" used to filter the raw packet.  Is there any reason this is just suitable for IPv4 and not IPv6? Using the BPF Wrapper I believe I have created what would be the same thing for IPv6 which remains to be tested. It seems natural that IPFilter would apply to IPv6 addressing but the documentation is specific about parameters being IPv4.

seladb:
@JasMetzger I wrote this a long time ago so I don't remember all the details but I think that libpcap/WinPcap/Npcap don't support netmask for IPv6. Please see this comment: https://github.com/seladb/PcapPlusPlus/blob/f49f009ecfac28f9978e8030834e7bd5477e6367/Pcap%2B%2B/src/PcapFilter.cpp#L164

We could (and maybe should) support IPv6 without a mask, or maybe have both `IPv4Filter` (with mask support), `IPv6Filter` (without mask support), `IPFilter` (IPv4/6 without mask support)

Dimi1010:
I think that solution can be done as a (somewhat lengthy) two step process.

Step 1 would be:
The current `IPFilter` being renamed to `IPv4Filter` as it already has mask support and works fine.
The `IPFilter` is kept as a deprecated alias for 1 release.
A new `IPv6Filter` is added (without mask support).

Step 2 would need to proceed after the old IPFilter has been removed, and would be adding a new `IPFilter` that works with IPv4/v6 without mask support. This can be an alias to `IPv6Filter` if that one is made to work with IPv4/v6 without mask support or be a facade class that exposes only the shared functionality between the v4 and v6 filters and internally chooses which one to use based on the IP.

JasMetzger:
@Dimi1010 Step 1 is a good idea. I agree that the current filter should be renamed IPv4Filter which comports with the naming conventions used elsewhere. Step 2 may be a good step to take if and when IPv6Filter can support masking.

seladb:
First I'd check if libpcap, WinPcap, Npcap indeed don't support IPv6 masks. If they do, it's easy to transform this class to support both IPv4 and IPv6.

If they don't, we can still transform this class to support both IPv4 and IPv6, but throw an exception when trying to set a mask when `IPFilter` is instantiated with IPv6.

[...]

PR #1324, opened 2024-02-25
Refactor IPFilter to support IPv6 where possible.

Dimi1010:
The PR aims to address #1288, supporting IPv6 filtering where possible, and raise errors where it is not supported.


Pcap++/header/PcapFilter.h
```diff
@@ -243,19 +243,26 @@ namespace pcpp
 	class IPFilter : public IFilterWithDirection
 	{
 	private:
-		std::string m_Address;
+		static constexpr int NO_Len_VALUE = 0;
```
tigercosmos, line 246: the way of mixing uppercase and lowercase looks weird. please check how other code wrote.

Pcap++/header/PcapFilter.h
```diff
@@ -243,19 +243,26 @@ namespace pcpp
...
+		static constexpr const char* NO_IPv4Mask_VALUE = "";
```
tigercosmos, line 247: `static` and `constexpr` should not be used together
another way you can write `constexpr std::string_view DEFAULT_IPV4MASK_EMPTY = "";` here.

Dimi1010: > `static` and `constexpr` should not be used together

Huh? Why? Since they do different things? `static` declaring that the variable is shared between all the instances and `constexpr` that the variable can be computed at compile time. Also removing `static` from the expression leads to a compiler error about `constexpr` being invalid there.

> `constexpr std::string_view DEFAULT_IPV4MASK_EMPTY = "";` here.

`string_view` is part of the C++17 standard, not C++11.

tigercosmos: Yeah, I forgot we are C++11, however as this string is just a "empty" string, I don't see any benefit to keep it static, the empty string should be already inlined during compilation by constexpr.

Dimi1010: You can't not make it a member constexpr field and not be static. The compiler throws a fit.

image

Pcap++/header/PcapFilter.h
Dimi1010: Fixed 00e7408

Pcap++/header/PcapFilter.h
tigercosmos: you could put `constexpr const char* EMTPY_STRING = "";` outside the class. again, no point in using `static` for an empty string here, I would rather you fill `""` in the code directly.

Dimi1010: > you could put constexpr const char* EMTPY_STRING = ""; outside the class.

Yes, but that throws away the entire point of the change. The primary point was to centralize the definition of the magic value used for "no mask" so you don't have random unspecified magic values sprinkled in the code. Not just to have a `EMPTY_STRING` variable, and I agree that having an `EMPTY_STRING` is pointless and you might as well use `""` where you need an empty string.

A thing that popped up is that apparently `inline` variables are C++17 (and `constexpr` implies inline in C++17 and after), so I will revert the change due to this as pre-C++17 `constexpr` variables in header files have ODR problems.

But that still does not answer my question of what your reasoning is that `static` and `constexpr` should not be used together inside a class scope?

tigercosmos: >  pre-C++17 `constexpr` variables in header files have ODR problems.

nice catch. seems it may cause an undefined behavior.

> static and constexpr should not be used together inside a class scope?

program-wide, yes it is okay. I didn't say clearly before, but I already explained why I think `static` is unnecessary here.

Dimi1010: > nice catch. seems it may cause an undefined behavior.

Not undefined actually, just that every translation unit will get its own copy. Doesn't matter much for a constant but it is still extra memory. This SO [answer](https://stackoverflow.com/a/50489130/11922936) explains it in more detail.

tigercosmos: sure, seems you are correct.

Pcap++/header/PcapFilter.h
```diff
@@ -266,7 +270,9 @@ namespace pcpp
 		 * @param[in] dir The address direction to filter (source or destination)
 		 * @param[in] ipv4Mask The mask to use. Mask should also be in a valid IPv4 format (i.e x.x.x.x), otherwise parsing this filter will fail
 		 */
-		IPFilter(const std::string& ipAddress, Direction dir, const std::string& ipv4Mask) : IFilterWithDirection(dir), m_Address(ipAddress), m_IPv4Mask(ipv4Mask), m_Len(0) {}
+		IPFilter(const std::string& ipAddress, Direction dir, const std::string& ipv4Mask) : IPFilter(IPv4Address(ipAddress), dir, ipv4Mask) {}
+
+		IPFilter(IPv4Address ipAddress, Direction dir, const std::string &ipv4Mask) : IFilterWithDirection(dir), m_Address(std::move(ipAddress)), m_IPv4Mask(ipv4Mask), m_Len(0) {}
```
tigercosmos, line 275: can we have `const IPv4Address&`? then you can do the copy constructor in the initialized list.

Pcap++/header/PcapFilter.h
```diff
@@ -286,19 +294,46 @@ namespace pcpp
...
+		}
+
+		void clearMask() { m_IPv4Mask = ""; }
 
 		/**
 		 * Set the subnet
 		 * @param[in] len The subnet to use (e.g "/24")
 		 */
-		void setLen(int len) { m_IPv4Mask = ""; m_Len = len; }
+		void setLen(int len) { this->clearMask(); m_Len = len; }
```
tigercosmos, line 334: nit: `const int len`

Pcap++/header/PcapFilter.h
```diff
@@ -286,19 +294,46 @@ namespace pcpp
...
 		 * Set the IPv4 mask
 		 * @param[in] ipv4Mask The mask to use. Mask should also be in a valid IPv4 format (i.e x.x.x.x), otherwise parsing this filter will fail
 		 */
-		void setMask(const std::string& ipv4Mask) { m_IPv4Mask = ipv4Mask; m_Len = 0; }
+		void setMask(const std::string& ipv4Mask)
+		{
+			if (!m_Address.isIPv4())
+			{
+				// TODO: What exception type to throw here?
+				throw std::runtime_error("Attempting to set an IPv4 mask on non-IPv4 address. "
```
tigercosmos, line 320: I think `runtime_error` is fine. there is a extra space at the end of the message.

Pcap++/header/PcapFilter.h
```diff
@@ -286,19 +294,46 @@ namespace pcpp
 		 * @param[in] ipAddress The IPv4 address to build the filter with. If this address is not a valid IPv4 address an error will be
 		 * written to log and parsing this filter will fail
 		 */
-		void setAddr(const std::string& ipAddress) { m_Address = ipAddress; }
+		void setAddr(const std::string& ipAddress) { this->setAddr(IPAddress(ipAddress)); }
+
+		void setAddr(const IPAddress &ipAddress)
```
tigercosmos, line 299: nit: -> `const IPAddress& ipAddress`

Pcap++/header/PcapFilter.h
```diff
@@ -286,19 +294,46 @@ namespace pcpp
...
+		}
 
 		/**
 		 * Set the IPv4 mask
 		 * @param[in] ipv4Mask The mask to use. Mask should also be in a valid IPv4 format (i.e x.x.x.x), otherwise parsing this filter will fail
 		 */
-		void setMask(const std::string& ipv4Mask) { m_IPv4Mask = ipv4Mask; m_Len = 0; }
+		void setMask(const std::string& ipv4Mask)
+		{
+			if (!m_Address.isIPv4())
```
tigercosmos, line 317: maybe have a branch for `isIPv6`?

Pcap++/header/PcapFilter.h
Dimi1010: Extra space is to have a space between this and the next sentence.
Because the way C++ compiler concatenates string literals the code `"This is " "a multipart string literal"` will produce "This is a multipart string literal" after compilation.

The entire error message is:
"Attempting to set an IPv4 mask on non-IPv4 address. Please set an IPv4 address before setting the mask."

Pcap++/header/PcapFilter.h
Dimi1010: What do you mean?

Pcap++/header/PcapFilter.h
Dimi1010: Fixed 8c70753

Pcap++/header/PcapFilter.h
Dimi1010: Fixed f55a679

Pcap++/header/PcapFilter.h
Dimi1010: Fixed f55a679

Pcap++/header/PcapFilter.h
```diff
@@ -243,19 +243,23 @@ namespace pcpp
...
 		/**
 		 * The basic constructor that creates the filter from an IPv4 address and direction (source or destination)
 		 * @param[in] ipAddress The IPv4 address to build the filter with. If this address is not a valid IPv4 address an error will be
 		 * written to log and parsing this filter will fail
 		 * @param[in] dir The address direction to filter (source or destination)
 		 */
-		IPFilter(const std::string& ipAddress, Direction dir) : IFilterWithDirection(dir), m_Address(ipAddress), m_IPv4Mask(""), m_Len(0) {}
+		IPFilter(const std::string& ipAddress, Direction dir) : IPFilter(IPAddress(ipAddress), dir) {}
+
+		IPFilter(const IPAddress& ipAddress, Direction dir) : IFilterWithDirection(dir), m_Address(ipAddress), m_IPv4Mask(""), m_Len(0) {}
```
seladb, line 262: Please add doxygen documentation here and in the other new c'tors

Pcap++/header/PcapFilter.h
```diff
@@ -286,19 +294,46 @@ namespace pcpp
...
 		 */
-		void setAddr(const std::string& ipAddress) { m_Address = ipAddress; }
+		void setAddr(const std::string& ipAddress) { this->setAddr(IPAddress(ipAddress)); }
+
+		void setAddr(const IPAddress& ipAddress)
+		{
+			if (!ipAddress.isIPv4() && hasMask())
+			{
+				// TODO: What exception type to throw here?
+				throw std::runtime_error("Attempting to set non-IPv4 address while an IPv4 mask is set. "
```
seladb, line 304: Maybe throw `std::invalid_argument`?

Pcap++/header/PcapFilter.h
seladb: You can use `std::invalid_argument` here also

Pcap++/header/PcapFilter.h
tigercosmos: I meant maybe better to also have an if branch for `m_Address.isIPv6()`

Dimi1010: If you are talking about the `not isIPv4` to be replaced with `isIPv6`, I have it as `not IPv4` because I think it better describes the requirement for the mask. It requires a IPv4 explicitly, so the rationale is "fail everything that is not IPv4". At the moment IPv6 is the only other IP version that is not v4, but however unlikely, that does not mean it will always be the case.

tigercosmos: I believed we could offer a more specific message, such as "it's an IPv6 address, not an IPv4 address." However, perhaps it's not worth the effort.

Dimi1010:
Hmm... so I seem to have hit a bit of a roadblock. When the filter is being compiled by Winpcap, I get the message "bogus IPv6 address". This persists for even the simplest IPv6 addresses like `::1`.

Full error message:
```
[ERROR: ...\PcapPlusPlus\Pcap++\src\PcapDevice.cpp: pcpp::IPcapDevice::setFilter:31] Error compiling filter. Error message is: bogus IPv6 address ::1
TestPcapFiltersOffline             : FAILED (E:\Users\dimi1\Development\C++\PcapPlusPlus\Tests\Pcap++Test\Tests\FilterTests.cpp:530). Assert TRUE failed:
   [fileReaderDev5.setFilter(ipFilterWithMask)]
```

seladb:
> [fileReaderDev5.setFilter

@Dimi1010 did you try the full version of an ipv6 address, for example: instead of `::1` use `0000:0000:0000:0000:0000:0000:0000:0001`?

Dimi1010:
> @Dimi1010 did you try the full version of an ipv6 address, for example: instead of `::1` use `0000:0000:0000:0000:0000:0000:0000:0001`?

@seladb Yes, tried these so far:
- `2001:db9:0:12::1` -> `Error compiling filter. Error message is: bogus IPv6 address 2001:db9:0:12::1`
- `::1` -> `Error compiling filter. Error message is: bogus IPv6 address ::1`
- `0000:0000:0000:0000:0000:0000:0000:0001` -> `Error compiling filter. Error message is: bogus IPv6 address ::1`

seladb:
> > @Dimi1010 did you try the full version of an ipv6 address, for example: instead of `::1` use `0000:0000:0000:0000:0000:0000:0000:0001`?
> 
> @seladb Yes, tried these so far:
> 
>     * `2001:db9:0:12::1` -> `Error compiling filter. Error message is: bogus IPv6 address 2001:db9:0:12::1`
> 
>     * `::1` -> `Error compiling filter. Error message is: bogus IPv6 address ::1`
> 
>     * `0000:0000:0000:0000:0000:0000:0000:0001` -> `Error compiling filter. Error message is: bogus IPv6 address ::1`

I looked in [WinPcap source code](https://www.winpcap.org/install/bin/WpcapSrc_4_1_3.zip) and found this error string in `wpcap/libpcap/scanner.c`. It looks like it run `getaddrinfo(pcap_text, NULL, &hints, &res)` to find the address from string, so I'm not sure why it doesn't work 🤔 

Does it work in Npcap and libpcap?

seladb:
@Dimi1010 did you manage to figure out what's the problem with WinPcap or Npcap (does it happen in both)?

Dimi1010:
@seladb Haven't really had much time for it, yet. Had some trouble getting npcap to work coz I was using vcpkg and it only had winpcap, then got hit with a lot of stuff from work to deal with.

Dimi1010:
@seladb The main issue I have with atm is getting the tests to run. For some reason without vcpkg, the project builds fine, and then when I run the test exe, I get this.

image

egecetin:
@Dimi1010 It might be related with libpcap version. Original winpcap (which can be downloaded from their site) supports libpcap version 1.4.2 (I checked their SDK. Inside the SDK package the pcap.h library mentions this version) but there is another version in sourceforge currently which supports libpcap 1.7.4 (personally not tested this fork). From your screenshot the `pcap_open_offline_with_timestamp_precision` function introduced with libpcap 1.5.1. Is there any possibility of more than one libpcap version in your system or maybe some similar problem to create confusion for compiler/linker?

seladb:
> @Dimi1010 It might be related with libpcap version. Original winpcap (which can be downloaded from their site) supports libpcap version 1.4.2 (I checked their SDK. Inside the SDK package the `pcap.h` library mentions this version) but there is another version in sourceforge currently which supports libpcap 1.7.4 (personally not tested this fork). From your screenshot the `pcap_open_offline_with_timestamp_precision` function introduced with libpcap 1.5.1. Is there any possibility of more than one libpcap version in your system or maybe some similar problem to create confusion for compiler/linker?
> 
> My suggestion, check the linked libraries to your exe and search `pcap.h` headers in your system paths

In addition, I'd advise using [Npcap](https://npcap.com/#download) instead on WinPcap, it's maintained and is likely to include a much newer libpcap version under the hood

Dimi1010:
ok, apparently it was a random `wpcap.dll` that somehow found itself in the tests Bin folder shadowing the one in `system32`. No idea how it got itself there, and the rebuilds did not remove it apparently. Deleted it and now the tests run fine.

Dimi1010:
> > @seladb Yes, tried these so far:
> > ```
> > * `2001:db9:0:12::1` -> `Error compiling filter. Error message is: bogus IPv6 address 2001:db9:0:12::1`
> > 
> > * `::1` -> `Error compiling filter. Error message is: bogus IPv6 address ::1`
> > 
> > * `0000:0000:0000:0000:0000:0000:0000:0001` -> `Error compiling filter. Error message is: bogus IPv6 address ::1`
> > ```
> 
> I looked in [WinPcap source code](https://www.winpcap.org/install/bin/WpcapSrc_4_1_3.zip) and found this error string in `wpcap/libpcap/scanner.c`. It looks like it run `getaddrinfo(pcap_text, NULL, &hints, &res)` to find the address from string, so I'm not sure why it doesn't work 🤔
> 
> Does it work in Npcap and libpcap?

@seladb Npcap also errors on `pcap_compile` but with the message `Error compiling filter. Error message is: expression rejects all packets` instead for the following expressions `ip and src net 2001:db9:0:12::1` and `ip and src net ::1`.

Pcap++/header/PcapFilter.h
Dimi1010: Fixed bb69dcb

Pcap++/header/PcapFilter.h
Dimi1010: Fixed bb69dcb

Pcap++/header/PcapFilter.h
Dimi1010: Fixed bb69dcb

Pcap++/header/PcapFilter.h
```diff
@@ -243,62 +243,145 @@ namespace pcpp
...
+		 * "filter only IP addresses that matches the subnet 10.0.0.x"
+		 * @param[in] ipAddress The IPv4 address to use. Only the part of the address that is not masked will be
+		 * matched. For example: if the address is "1.2.3.4" and the mask is "255.255.255.0" than the part of the
+		 * address that will be matched is "1.2.3.X".
+		 * @param[in] dir The address direction to filter (source or destination)
+		 * @param[in] ipv4Mask The mask to use. Mask should also be in a valid IPv4 format (i.e x.x.x.x), otherwise
+		 * parsing this filter will fail
 		 */
-		IPFilter(const std::string& ipAddress, Direction dir, const std::string& ipv4Mask) : IFilterWithDirection(dir), m_Address(ipAddress), m_IPv4Mask(ipv4Mask), m_Len(0) {}
+		IPFilter(const IPv4Address& ipAddress, Direction dir, const std::string& ipv4Mask) : IFilterWithDirection(dir), m_Address(ipAddress), m_IPv4Mask(ipv4Mask), m_Len(0) {}
```
tigercosmos, line 290: if the IP address is already `IPv4Address`, why not let the mask also be `IPv4Address`?

Pcap++/header/PcapFilter.h
```diff
@@ -243,62 +243,145 @@ namespace pcpp
...
 		 * the subnet 10.0.0.x"
 		 * @param[in] ipAddress The IPv4 address to use. Only the part of the address that is not masked will be matched. For example: if the address
-		 * is "1.2.3.4" and the mask is "255.255.255.0" than the part of the address that will be matched is "1.2.3.X". If this address is not a
-		 * valid IPv4 address an error will be written to log and parsing this filter will fail
+		 * is "1.2.3.4" and the mask is "255.255.255.0" than the part of the address that will be matched is "1.2.3.X".
 		 * @param[in] dir The address direction to filter (source or destination)
 		 * @param[in] ipv4Mask The mask to use. Mask should also be in a valid IPv4 format (i.e x.x.x.x), otherwise parsing this filter will fail
+		 * @throws std::invalid_argument The provided address is not a valid IPv4 address.
+		 */
+		IPFilter(const std::string& ipAddress, Direction dir, const std::string& ipv4Mask) : IPFilter(IPv4Address(ipAddress), dir, ipv4Mask) {}
```
tigercosmos, line 278: we may not need this constructor, as IPv4Address has a string implicit constructor.

Pcap++/header/PcapFilter.h
```diff
@@ -243,62 +243,145 @@ namespace pcpp
...
 	{
 	private:
-		std::string m_Address;
+		IPAddress m_Address;
 		std::string m_IPv4Mask;
 		int m_Len;
 		void convertToIPAddressWithMask(std::string& ipAddrmodified, std::string& mask) const;
 		void convertToIPAddressWithLen(std::string& ipAddrmodified) const;
+
+		bool hasMask() const { return !m_IPv4Mask.empty(); }
```
tigercosmos, line 252: maybe `!m_IPv4Mask.empty() && m_Address.isIPv4()` ?

Pcap++/header/PcapFilter.h
```diff
@@ -243,62 +243,145 @@ namespace pcpp
 	class IPFilter : public IFilterWithDirection
 	{
 	private:
-		std::string m_Address;
+		IPAddress m_Address;
 		std::string m_IPv4Mask;
```
tigercosmos, line 247: btw, we have `IPNetwork`, maybe we should use it?

Pcap++/src/PcapFilter.cpp
```diff
@@ -221,24 +216,54 @@ void IPFilter::convertToIPAddressWithLen(std::string& ipAddrmodified) const
...
+			mask[fullyMaskedBytes] = static_cast<uint8_t>(-1) << (8 - partialMaskBits);
+		}
+		for (size_t i = 0; i < addrAsBytes.size(); i++)
+		{
+			addrAsBytes[i] &= mask[i];
+		}
+
+		ipAddrmodified = IPv6Address(addrAsBytes.data()).toString();
+	}
+	else
```
tigercosmos, line 245: maybe remove the last `else{}` and move the `throw` outside, then we can save three lines.

Pcap++/src/PcapFilter.cpp
```diff
@@ -221,24 +216,54 @@ void IPFilter::convertToIPAddressWithLen(std::string& ipAddrmodified) const
...
-	result = "ip and " + dir + " net " + ipAddr;
+	result = ipProto + " and " + dir + " net " + ipAddr;
 	if (m_IPv4Mask != "")
 		result += " mask " + mask;
 	else if (m_Len > 0)
 	{
-		std::ostringstream stream;
-		stream << m_Len;
-		result += '/' + stream.str();
+		result += '/' + std::to_string(m_Len);
```
tigercosmos, line 266: `'/' + std::to_string(m_Len);` creates one more string, we can reuse `result` :
```
result += '/';
result += std::to_string(m_Len);
```
but I am not sure if this really matters the performance.

Tests/Pcap++Test/Tests/FilterTests.cpp
```diff
@@ -4,6 +4,7 @@
 #include "EthLayer.h"
 #include "VlanLayer.h"
 #include "IPv4Layer.h"
+#include "IPv6Layer.h"
```
tigercosmos, line 7: are there all new contractors in the test?

Pcap++/src/PcapFilter.cpp
Dimi1010: Are we tho? We need to add `return` statements at the end of each valid block, if we pull the `throw` outside the `if`/`else` chain.

Another thing to thing about, the `if`/`else` chain could potentially be refactored into a `switch`.

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: What does 'new contractors' mean?

tigercosmos: ops, sorry, I meant the all constructors added in this PR.

Pcap++/src/PcapFilter.cpp
tigercosmos: you are correct, then it is fine.

Pcap++/header/PcapFilter.h
seladb: yes, I agree - we can probably refactor some of the c'tors to accept `IPNetwork` and use it instead of a string

seladb: Perhaps we can do a slightly larger refactor of this class and use `IPNetwork` to represent masks - it already supports IPv4/IPv6 netmasks so should have a lot of the functionality that currently exists in the filter class

Pcap++/header/PcapFilter.h
Dimi1010: True. Documentation slightly differs tho, as the string constructors can throw `invalid_argument` when constructing the IP from string. Should I move the `@throws std::invaild_argument` documentation to IPv4Address' and IPv6Address' documentations?

Pcap++/header/PcapFilter.h
Dimi1010: It can be done with `IPNetwork` but its a bigger change. 

The setters `setAddr`, `setMask` and `setLen` would also have to be replaced with `setNetwork`/`getNetwork`. The `IPFilter` would still need to keep the data set from the deprecated setters for the deprecation period as the setters can't create a new `IPNetwork` object by themselves to update the filter.

Parsing to a filter string would be simpler tho, as the new filter would generally always use the `net <netAddr>/<prefixLen>` syntax, instead of mixing between it and `net <netAddr> mask <maskAddr>` syntax.

Pcap++/src/PcapFilter.cpp
Dimi1010: Implemented in 120fb7d. Added a reserve call to result before string assignment and removed extra string constructions made by chaining `+` operators.

Pcap++/header/PcapFilter.h
tigercosmos: yes, sounds good.

Pcap++/header/PcapFilter.h
tigercosmos: Since we already have many breaking changes, perhaps it's okay to deprecate functions, and even remove the function if necessary. Need to confirm with @seladb.

Dimi1010: @seladb Seems good.

Pcap++/header/PcapFilter.h
Dimi1010: Implemented in aa7ae76 and f2f5fa7

Dimi1010: And reverted f2f5fa7 because it broke when you have an IP address written as a C-string literal.

tigercosmos: sure.

Pcap++/header/PcapFilter.h
Dimi1010: @seladb @tigercosmos 
Ok, refactored `IPFilter` to use `IPNetwork`. Did not deprecate any of the old setters although switching IPv6 to IPv4 address essentially gives a network of 1 address if the prefix length is over 32 as it caps it at 32.

General overview of the changes include:
- `IPNetwork` (+ v4 and v6) getting a new explicit 1 parameter constructor that makes a 1 address network.
- `IPv4Network` and `IPv6Network` having their exception messages about `invalid netmask` expanded.
- The `PCPP_DEPRECATED` macro being moved from 'HttpLayer.h' to 'DeprecationUtils.h'. I thought I might use it, but I ended up not needing it.
- All of `IPFilter`'s methods now work for IPv4 and IPv6 (`setMask` is no longer IPv4 exclusive, yay). Netmasks still need to be matched IPv4 address with IPv4 netmask format, and vice versa for IPv6.

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: Should be although I haven't written any specific tests for the `IPAddress` ctors. The old `string` ctors delegate to the `IPAddress` ctors, so covering the string ones should cover the new ones.

With the `IPNetwork` refactor, there is a new test about constructing an `IPFilter` from `IPNetwork` object.

Pcap++/src/PcapFilter.cpp
```diff
@@ -153,93 +154,20 @@ std::string IFilterWithOperator::parseOperator()
...
-}
-
 void IPFilter::parseToString(std::string& result)
 {
 	std::string dir;
-	std::string ipAddr = m_Address;
-	std::string mask = m_IPv4Mask;
-	convertToIPAddressWithMask(ipAddr, mask);
-	convertToIPAddressWithLen(ipAddr);
+	std::string ipAddr = m_Network.toString();
```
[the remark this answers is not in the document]
Dimi1010: That was how it uses to be, but it adds a branch condition as `toString` always generates with the prefix length. Is there a performance difference between filtering just an `net IP` and `net IP/Prefix`?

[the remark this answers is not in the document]
seladb: I'm not sure if there is a performance difference, there might be 🤷‍♂️ 
I'm less worried about adding a branch condition to `toString()` because it's usually not called in the critical path, it's just parsing and setting the filter. Using `net IP` is nicer than `net IP/prefix` in my opinion...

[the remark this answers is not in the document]
Dimi1010: Well, I am not talking about adding a branch to `toString()` but around it, as it would need to get just the network prefix if the subnet is full size, instead of calling the `IPNetwork`'s `toString()`.
IMO tho, it makes it more consistent and predictable if the filter always generates the format `net IP/Prefix` instead of "generates `net IP/Prefix` unless the prefix is full, then its just `net IP`" ¯\_(ツ)_/¯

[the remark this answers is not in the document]
seladb: > Well, I am not talking about adding a branch to `toString()` but around it in the call site, as it would need to get just the network prefix if the subnet is full size, instead of calling the `IPNetwork`'s `toString()`.

I'm not sure I understand this sentence...



> IMO tho, it makes it more consistent and predictable if the filter's `parseToString()` always generates the format `net IP/Prefix` instead of "generates `net IP/Prefix` unless the prefix is full, then its just `net IP`" ¯\_(ツ)_/¯

I don't have a strong opinion here so we can go with your preference 🙂

Pcap++/header/PcapFilter.h
Dimi1010: Hmm, possibly can be done. 

A couple of things to note tho:
- After the change to use `IPNetwork`, the `IPNetwork` ctor uses `string` for netmask too, so that has to be changed too.
- Changing `string` to `IPAddress` will break any C-strings that are currently used as parameters, as it becomes a double conversion (`C-string`->`string`->`IPAddress`) so the new ctors can be added but changing the type directly will break backwards compat.

Pcap++/header/PcapFilter.h
Dimi1010: Resolving this as the current version does not have that method anymore.

Common++/header/IpAddress.h
```diff
@@ -668,13 +687,21 @@ namespace pcpp
 	class IPNetwork
 	{
 	public:
+		/**
+		 * A constructor that creates an instance of the class out of an IP address and a full prefix length,
+	     * essentially making a network of consisting of only 1 address.
+	     *
+	     * @param address An address representing the network prefix.
+		 */
+		explicit IPNetwork(const IPAddress& address) : IPNetwork(address, address.isIPv4() ? 32u : 128u) {}
```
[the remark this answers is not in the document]
Dimi1010: Fixed 4011821

Pcap++/header/PcapFilter.h
```diff
@@ -243,62 +243,140 @@ namespace pcpp
...
 
 		/**
-		 * Set the IPv4 address
-		 * @param[in] ipAddress The IPv4 address to build the filter with. If this address is not a valid IPv4 address an error will be
-		 * written to log and parsing this filter will fail
+		 * Set the network to build the filter with.
+		 * @param[in] network The IP Network object to be used when building the filter.
 		 */
-		void setAddr(const std::string& ipAddress) { m_Address = ipAddress; }
+		void setNetwork(const IPNetwork& network) {
```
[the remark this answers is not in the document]
Dimi1010: Fixed in 4011821 (I think I got all of them)

Pcap++/src/PcapFilter.cpp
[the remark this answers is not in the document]
Dimi1010: > I'm not sure I understand this sentence...

Eh, it was not that important anyway.

> I don't have a strong opinion here so we can go with your preference 🙂

Alrighty then, so I guess we are keeping it as is currently, unless there is a significant performance impact, which I don't think there should be, but you never know. ¯\\\_(ツ)\_/¯

Mark this as resolved?

Pcap++/header/PcapFilter.h
seladb: I think we can keep it as is

seladb (approved):
LGTM

@tigercosmos can you also re-review?

Pcap++/header/PcapFilter.h
```diff
@@ -243,62 +243,141 @@ namespace pcpp
...
 		 * A constructor that enable to filter only part of the address by using a mask (aka subnet). For example: "filter only IP addresses that matches
 		 * the subnet 10.0.0.x"
-		 * @param[in] ipAddress The IPv4 address to use. Only the part of the address that is not masked will be matched. For example: if the address
-		 * is "1.2.3.4" and the mask is "255.255.255.0" than the part of the address that will be matched is "1.2.3.X". If this address is not a
-		 * valid IPv4 address an error will be written to log and parsing this filter will fail
+		 * @param[in] ipAddress The IP address to use. Only the part of the address that is not masked will be matched. For example: if the address
+		 * is "1.2.3.4" and the mask is "255.255.255.0" than the part of the address that will be matched is "1.2.3.X".
 		 * @param[in] dir The address direction to filter (source or destination)
-		 * @param[in] ipv4Mask The mask to use. Mask should also be in a valid IPv4 format (i.e x.x.x.x), otherwise parsing this filter will fail
+		 * @param[in] netmask The mask to use. Mask should be in a valid IPv4 format (i.e x.x.x.x) or IPv6 format (i.e. xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx).
```
tigercosmos, line 270: nit: I think "i.e." here is incorrect. "i.e." is for "that is", but here we should use "for example", which is "i.g."

suggest to change to
```
@param[in] netmask The mask to use. Mask should be in a valid format (e.g., x.x.x.x for IPv4 or xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx for IPv6)."

// or

* @param[in] netmask The mask to use. The mask should be a valid IP address in either IPv4 dotted-decimal format (e.g., 255.255.255.0) or IPv6 colon-separated hexadecimal format (e.g., FFFF:FFFF:FFFF:FFFF::).


```

tigercosmos (approved):
LGTM, just a comment that needs to be addressed. @Dimi1010

Pcap++/header/PcapFilter.h
Dimi1010: Fixed in 8e63b9d. I chose the second version.

Issue #1179, opened 2023-08-09
Bluetooth HCI packet trace support


dhruvkakadiya:
Hello,

I can't find documentation if this library supports capturing/parsing Bluetooth HCI packets, so it would be great to add support for the same if possible. If it supports that already, then can someone point me to the right direction?

Thank you!

seladb:
Hi @dhruvkakadiya ! Bluetooth HCI is not currently supported in PcapPlusPlus. I'm not even sure what is the format of such packets... can you attached an example pcap file?
Can you also share a link to the RFC or some info about the packet structure?

dhruvkakadiya:
Thank you, @seladb for quick response!

Reference Links: [Wireshark Bluetooth](https://wiki.wireshark.org/Bluetooth), [RFC](https://www.ietf.org/rfc/rfc7668.txt), [Bluetooth specifications](https://www.bluetooth.com/specifications/)

seladb:
Thank you @dhruvkakadiya ! From briefly going over some of these links, it looks like HCI is a suite of protocols, not just one. Would you be interested in implementing them (or some of them)?

Issue #917, opened 2022-07-18
GTPv2 Support


ArmanTulegenov:
Would be great to add GTPv2 Support. Do you have some plans around it?

ArmanTulegenov:
Yes, I will try to add GTPv2 layer.

seladb:
@ArmanTulegenov do you know when you'll work on this?

mmanoj:
@ArmanTulegenov 

Let ne know any help required,I also like to contribute to this patch. Pls share PCAP file with GTPv2

seladb:
@mmanoj please feel free to implement GTPv2, let me know if you need my help!

Here are a few pcap files with GTPv2 I found online: 
- [`3_reattach.zip`](https://github.com/seladb/PcapPlusPlus/files/10299222/3_reattach.zip), found it here: https://blog.colte.network/pcap-library/
- [`test.0530152301.pcap.zip`](https://github.com/seladb/PcapPlusPlus/files/10299228/test.0530152301.pcap.zip), found it here: https://gitlab.com/wireshark/wireshark/-/issues/16594

mmanoj:
@seladb 

Thanks for the quick reply and support,sure I will looking in to it and update you.Currently I'm working on Radius 3GPP dictionary attributes decoding.I will check is it possible to open source with pcpp.

mmanoj:
Hi @seladb 

I'm working on initial work for GTPv2 message decoding, I will update you the progress,currently checking the core spec and protocol details.

seladb:
Sounds good @mmanoj , thanks for the update!

I am looking forward to seeing your PR.

cc @egecetin

mmanoj:
@seladb 
Can you explain the below code, are you looking the bit pattern, I'm trying to relate with wireshark output for GTPv1 to plan the GTP2 code

bool GtpV1Layer::isGTPv1(const uint8_t* data, size_t dataSize)
{
	if(data != NULL && dataSize >= sizeof(gtpv1_header) && (data[0] & 0xE0) == 0x20)
	{
		return true;
	}

	return false;
}

mmanoj:
![GTPv1-01](https://user-images.githubusercontent.com/1105057/215421727-20d7e92c-fd95-4740-9991-1b0e1e619c8b.png)

mmanoj:
above from [3_reattach.zip](https://github.com/seladb/PcapPlusPlus/files/10299222/3_reattach.zip)

[...]

Issue #906, opened 2022-06-30
How to access file header info?


GilbertoTheMighty:
Hello. I'm using pcpp::PcapFileReaderDevice reader(filename);
I would like to read [Magic Number](https://tools.ietf.org/id/draft-gharris-opsawg-pcap-00.html#section-4-5.2.1) from file header, which represents time format. I could not find anything that allows to access header struct. How to do this?

seladb:
Do you know if `libpcap` provides an API to read this magic number? PcapPlusPlus relies on libpcap/WinPcap/Npcap to read `.pcap` files so if there's no such API there's no way for PcapPlusPlus to fetch this information.

We did add the [`PCAP_TSTAMP_PRECISION_NANO`](https://github.com/seladb/PcapPlusPlus/blob/5dd6cec932e5dee2d040206c4b17c63994940ed8/Pcap%2B%2B/src/PcapFileDevice.cpp#L253) compilation flag to open files in nanosec precision, but that assumes you know the file's precision...

GilbertoTheMighty:
I've compiled pcap++ with PCAP_TSTAMP_PRECISION_NANO. 
Read pcap file, update timespec, write to another file => produces rounded timestamp.
1601234567.000051234 (written time) becomes 1601234567.000051000 i output file acconding to wireshark

GilbertoTheMighty:
[Seems like writePacket does not use this flag](https://github.com/seladb/PcapPlusPlus/blob/5dd6cec932e5dee2d040206c4b17c63994940ed8/Pcap%2B%2B/src/PcapFileDevice.cpp#L555)

GilbertoTheMighty:
I've also commented *1000 and /1000 in Common++\header\TimespecTimeval.h

After these change I was able to produce pcap file with nanosecond, which are seen in Wireshark

seladb:
Thanks for debugging the issue! Can you please open a PR with the fix?

seladb:
@GilbertoTheMighty will you consider opening a PR with the fix?

seladb:
> @GilbertoTheMighty will you consider opening a PR with the fix?

@GilbertoTheMighty I'm not sure if you saw my previous comment. Please let me know if you can open a PR with the fix

PR #1325, opened 2024-02-26
Added code to make pointer vector be thread safe

muthusaravanan3186:
Added code to make pointer vector be thread safe. Used mutex to lock every operation done on the pointer vector.


Dimi1010:
Kinda curious, is there a particular issue that requires the vector being thread safe?

Also, if it isnt required all the time, the feature can be made opt-in by injecting the `std::mutex` as a template parameter that can be provided a struct implementing no-op `lock`/`unlock`/`try_unlock` functions, instead when thread safety is not a requirement, saving the mutex overhead.

muthusaravanan3186:
We had a requirement when we need to read the packet simultaneously for some kind of edge processing. Yes mutex may be overhead in non-thread environment.

seladb (commented):
@muthusaravanan3186 I believe we can use `std::lock_guard` in this case which is simpler, no?

seladb:
> Kinda curious, is there a particular issue that requires the vector being thread safe?
> 
> Also, if it isn't required all the time, the feature can be made opt-in by injecting the `std::mutex` type as a template parameter that can be provided a struct implementing no-op `lock`/`unlock`/`try_unlock` methods instead when thread safety is not a requirement, saving the mutex overhead.

I think @Dimi1010 has a good point. Maybe we can add an optional parameter in the constructor saying whether or not to use the mutex

tigercosmos:
> > Kinda curious, is there a particular issue that requires the vector being thread safe?
> > Also, if it isn't required all the time, the feature can be made opt-in by injecting the `std::mutex` type as a template parameter that can be provided a struct implementing no-op `lock`/`unlock`/`try_unlock` methods instead when thread safety is not a requirement, saving the mutex overhead.
> 
> I think @Dimi1010 has a good point. Maybe we can add an optional parameter in the constructor saying whether or not to use the mutex

I think if we want to control using lock or not, maybe consider  using `ifdef USE_LOCK_POINTER_VECTOR` preprocessor directive, to prevent run-time branch overhead.

seladb:
> > > Kinda curious, is there a particular issue that requires the vector being thread safe?
> > > Also, if it isn't required all the time, the feature can be made opt-in by injecting the `std::mutex` type as a template parameter that can be provided a struct implementing no-op `lock`/`unlock`/`try_unlock` methods instead when thread safety is not a requirement, saving the mutex overhead.
> > 
> > 
> > I think @Dimi1010 has a good point. Maybe we can add an optional parameter in the constructor saying whether or not to use the mutex
> 
> I think if we want to control using lock or not, maybe consider using `ifdef USE_LOCK_POINTER_VECTOR` preprocessor directive, to prevent run-time branch overhead.

adding an additional `if` before locking the mutex shouldn't have a big performance penalty I think 🤔

Dimi1010:
@seladb @tigercosmos I was thinking more along the lines of how spdlog solved it with their logging sinks. Ends up requiring a template but you don't really have any runtime overhead for the non-mutex version, and it is not a global switch of all or nothing like a preprocessor directive would be.

Example from spdlog:
https://github.com/Dimi1010/spdlog/blob/v1.x/include%2Fspdlog%2Fsinks%2Fostream_sink.h#L46-L47

seladb:
> @seladb @tigercosmos I was thinking more along the lines of how spdlog solved it with their logging sinks. Ends up requiring a template but you don't really have any runtime overhead for the non-mutex version, and it is not a global switch of all or nothing like a preprocessor directive would be.
> 
> Example from spdlog: https://github.com/gabime/spdlog/blob/v1.x/include%2Fspdlog%2Fsinks%2Fostream_sink.h#L39-L40 https://github.com/gabime/spdlog/blob/v1.x/include%2Fspdlog%2Fsinks%2Fbase_sink-inl.h#L27

I'm not sure implementing a `null_mutex` has less performance penalty than adding a simple `if`... 🤔

Dimi1010:
> > @seladb @tigercosmos I was thinking more along the lines of how spdlog solved it with their logging sinks. Ends up requiring a template but you don't really have any runtime overhead for the non-mutex version, and it is not a global switch of all or nothing like a preprocessor directive would be.
> > Example from spdlog: https://github.com/gabime/spdlog/blob/v1.x/include%2Fspdlog%2Fsinks%2Fostream_sink.h#L39-L40 https://github.com/gabime/spdlog/blob/v1.x/include%2Fspdlog%2Fsinks%2Fbase_sink-inl.h#L27
> 
> I'm not sure implementing a `null_mutex` has less performance penalty than adding a simple `if`... 🤔

Might be, might not. They have `null_mutex` implemented like this:
```
struct null_mutex {
    void lock() const {}
    void unlock() const {}
};
```
whose `lock`/`unlock` compile down to just a return instruction back to caller,

The benefit over the if approach is that when you need a non-thread safe version, you don't need to construct and store a mutex anyway in the struct and it removes the need for ifs before every lock/unlock operation. It also allows you to actually use the `lock_guard` as you can't really just use it cleanly if you have an `if` block for wanting to lock and then the guard has to keep existing after the `if` block exists.

I think the potential minor performance cost of a function call over an `if` statement is worth it, from the point of code readability and maintainability perspective.

tigercosmos:
@Dimi1010 's proposal makes sense to me, especially readability and maintainability.

muthusaravanan3186:
are we fine with above implementation of using null mutex. If so i can go and make the necessary changes

seladb:
> are we fine with above implementation of using null mutex. If so i can go and make the necessary changes

Yes, I think we can go with this approach!

Dimi1010 (commented):
Left some comments on some cases where the mutex use does not actually keep the structure thread safe.

Honestly, addressing those changes to keep the vector thread safe internally in all cases seems to me being more of a hassle then just having multithreaded code that access the vector simultaneously keep their own synchronization between them then relying on the vector implementation to do the sync. A vector class has too many ways to access/modify the underlying data.

Common++/header/PointerVector.h
```diff
@@ -78,31 +92,51 @@ namespace pcpp
...
+		}
 
 		/**
 		 * Get the first element of the vector
 		 * @return An iterator object pointing to the first element of the vector
 		 */
-		VectorIterator begin() { return m_Vector.begin(); }
+		VectorIterator begin()
+		{
+			std::lock_guard<Mutex> lk(m_Mutex);
```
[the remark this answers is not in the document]
muthusaravanan3186: This is the case on normal scenario even without multi threading environment. we get an iterator and after some time vector reconstructs itself then the above iterator will be undefined.

Common++/header/PointerVector.h
```diff
@@ -55,6 +67,7 @@ namespace pcpp
 		 */
 		PointerVector(const PointerVector& other)
 		{
+			std::lock_guard<Mutex> lk(m_Mutex);
```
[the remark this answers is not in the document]
muthusaravanan3186: yes.. will modify the same

Common++/header/PointerVector.h
```diff
@@ -161,11 +208,13 @@ namespace pcpp
 		 */
 		const T* at(int index) const
 		{
+			std::lock_guard<Mutex> lk(m_Mutex);
 			return m_Vector.at(index);
```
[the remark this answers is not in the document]
muthusaravanan3186: Thanks for the comments. since we are mostly returning an iterator which is not a good idea for a multi-thread env.

Common++/header/PointerVector.h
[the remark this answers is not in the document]
Dimi1010: True, tho in a non-multithreaded scenario you don't have the option of another thread invalidating your iterator whenever. Iterators are invalidated in specific documented calls and the sequence of those calls in relation to iterator usage is deterministic.

Common++/header/PointerVector.h
```diff
@@ -111,13 +145,21 @@ namespace pcpp
...
 
 		/**
 		 * Returns a pointer of the first element in the vector
 		 * @return A pointer of the first element in the vector
 		 */
-		T* front() { return m_Vector.front(); }
+		T* front()
+		{
+			std::lock_guard<Mutex> lk(m_Mutex);
+			return m_Vector.front();
```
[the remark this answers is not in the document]
muthusaravanan3186: In that case vector is not suited well. As you said the iterator can be exploited in so many ways.

Common++/header/PointerVector.h
```diff
@@ -13,13 +14,23 @@
 namespace pcpp
 {
 
+    /**
+	 * @struct nullMutex
+     * A struct containing standard mutex operation but does not perform any action.
+	 * This mutex is used when the user decides PointerVector not be thread safe which be default
+	*/
+    struct nullMutex {
```
seladb, line 22: please use tabs instead of spaces

Common++/header/PointerVector.h
```diff
@@ -139,7 +182,10 @@ namespace pcpp
 		{
 			T* result = (*position);
 			VectorIterator tempPos = position;
-			tempPos = m_Vector.erase(tempPos);
+			{
+				std::lock_guard<Mutex> lk(m_Mutex);
+			    tempPos = m_Vector.erase(tempPos);
```
seladb, line 187: ditto: use tabs instead of spaces

tigercosmos:
@muthusaravanan3186  would you like to finish this PR?

muthusaravanan3186:
> @muthusaravanan3186 would you like to finish this PR?

@tigercosmos  Our requirement was single producer and single consumer. But  based on review comments it seems we need to address multiple consumer too. I just want to know what is the expectation here. If we need to change the PointerVector to hold shared ptr references; then it will be big change.. Please let me know what is expected here.

seladb:
> > @muthusaravanan3186 would you like to finish this PR?
> 
> @tigercosmos Our requirement was single producer and single consumer. But based on review comments it seems we need to address multiple consumer too. I just want to know what is the expectation here. If we need to change the PointerVector to hold shared ptr references; then it will be big change.. Please let me know what is expected here.

@muthusaravanan3186 we can do it in phases. Make some changes discussed above without using shared pointers, and then in the future open another PR to move to shared pointers. The improvements suggested in this PR will make `PointerVector` mostly thread safe (even if not 100%). We can mention the edge cases in the doxygen documentation

Common++/header/PointerVector.h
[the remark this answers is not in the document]
muthusaravanan3186: Done

Common++/header/PointerVector.h
```diff
@@ -78,31 +92,51 @@ namespace pcpp
...
+		}
 
 		/**
 		 * Get the first element of a constant vector
 		 * @return A const iterator object pointing to the first element of the vector
 		 */
-		ConstVectorIterator begin() const { return m_Vector.begin(); }
+		ConstVectorIterator begin() const
+		{
+			std::lock_guard<Mutex> lk(m_Mutex);
```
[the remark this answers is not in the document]
muthusaravanan3186: will make the changes in another PR

Common++/header/PointerVector.h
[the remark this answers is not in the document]
muthusaravanan3186: On next PR we are planning to use shared_ptr to hold  the vector elements

Common++/header/PointerVector.h
```diff
@@ -151,6 +197,7 @@ namespace pcpp
 		 */
 		T* at(int index)
 		{
+			std::lock_guard<Mutex> lk(m_Mutex);
 			return m_Vector.at(index);
```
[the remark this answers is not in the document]
muthusaravanan3186: On next PR we are planning to use shared_ptr to hold  the vector elements

Common++/header/PointerVector.h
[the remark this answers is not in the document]
muthusaravanan3186: On next PR we are planning to use shared_ptr to hold  the vector elements

Common++/header/PointerVector.h
muthusaravanan3186: done

Common++/header/PointerVector.h
muthusaravanan3186: done

Common++/header/PointerVector.h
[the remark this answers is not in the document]
muthusaravanan3186: we are planned to use shared_ptr in another PR

muthusaravanan3186:
@seladb  can you review now

tigercosmos:
@muthusaravanan3186  could you resolve the conflicts?

muthusaravanan3186:
@seladb i resolved all the conflicts. But on one of the conflict i am getting the below message
Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

tigercosmos:
> @seladb i resolved all the conflicts. But on one of the conflict i am getting the below message Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

Not sure what did you do with the merge, but you can resolve the conflicts at your forked repo.
Basically you can do the followings:
1. set https://github.com/seladb/PcapPlusPlus/ as upstream: `git remove upstream https://github.com/seladb/PcapPlusPlus/`
2. fetch upstream: `git fetch upstream master`
3. `git merge upstream master`
4. there will be some conflicts showing by git, and you need to resolve them.

seladb:
@muthusaravanan3186 you can look at this StackOverflow question: https://stackoverflow.com/questions/38949951/how-to-solve-merge-conflicts-across-forks

Please notice that you should resolve conflicts between `origin/dev` and `muthusaravanan3186/dev` (not between the `master` branches of origin and your fork)

muthusaravanan3186:
@seladb  Resolved all the conflicts.

Common++/header/PointerVector.h
```diff
@@ -55,22 +66,23 @@ namespace pcpp
...
+			try 
 			{
+				std::lock_guard<Mutex> lk(other.m_Mutex);
 				for (const auto iter : other)
-				{
-					T* objCopy = new T(*iter);
-					try
 					{
-						m_Vector.push_back(objCopy);
+						T* objCopy = new T(*iter);
```
tigercosmos, line 74: please revert the format.

Common++/header/PointerVector.h
```diff
@@ -55,22 +66,23 @@ namespace pcpp
...
-					catch (const std::exception&)
-					{
-						delete objCopy;
-						throw;
-					}
-				}
-			}
+			} 
 			catch (const std::exception&)
 			{
```
tigercosmos, line 87: we may need a lock here?

Common++/header/PointerVector.h
```diff
@@ -12,14 +13,23 @@
...
+	};
 
 	/**
 	 * @class PointerVector
 	 * A template class for representing a std::vector of pointers. Once (a pointer to) an element is added to this vector,
 	 * the element responsibility moves to the vector, meaning the PointerVector will free the object once it's removed from the vector
 	 * This class wraps std::vector and adds the capability of freeing objects once they're removed from it
 	 */
-	template<typename T>
+	template<typename T, typename Mutex=pcpp::nullMutex>
```
tigercosmos, line 32: please update the document.

Common++/header/PointerVector.h
Dimi1010: @tigercosmos in the catch block?

Dimi1010: That should not be an issue as this is a copy constructor. So nothing should have a reference to the object under construction yet. The main concern is locking the object that is being copied from as that could have other references.

tigercosmos: I see, make sense. About the lock in the try block, do you think it's more efficient to put at the outer one or the inner one; in other orders, locking the whole list or one by one is better?

Dimi1010: Locking and unlocking once is generally better, as it would introduce less performance overhead. Generally we are going though the entire list anyway, IMO, locking the whole list once is the better option.

tigercosmos: I also have a concern locking one by one will increase the overhead instead. Let's keep the current way.

Dimi1010: Another thing is that locking/unlocking for each element induces the potential of the list being modified in the middle of the copy operation. Thus making both approaches inequivalent from data consistency perspective.

Common++/header/PointerVector.h
muthusaravanan3186: changed to NullMutex. But detail namespace is only present on Pcap++. Since this on Common folder hoping will place it here which is more reasonable.

Common++/header/PointerVector.h
muthusaravanan3186: done

Common++/header/PointerVector.h
tigercosmos: You can update the comment, mentioning about the template argument. We use Doxygen to generate the document.

muthusaravanan3186: done

Common++/header/PointerVector.h
```diff
@@ -70,7 +83,7 @@ namespace pcpp
 						throw;
 					}
 				}
-			}
+			}
```
tigercosmos, line 86: nit: space

Common++/header/PointerVector.h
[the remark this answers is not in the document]
tigercosmos: Perhaps we want to add this line (and all other similar ones) later when you raise another PR.

tigercosmos (changes requested):
- [ ] Fix format, please ensure that you run `pre-commit` before submitting the code. You can also check if all CI works on your forked repo (Allow CI on the forked repo)
- [ ] Remove all changes that you plan to use on the next PR, e.g. `begin()`, `front()`. It doesn't make sense to have a WIP code here, please remove the unnecessary changes in this PR, and if you have time to raise another PR, we can have those changes back on that PR.

muthusaravanan3186:
@tigercosmos  There is no WIP PR yet; as discussed plan is to merge this PR and start to work on  use of shared_ptr.

tigercosmos:
> @tigercosmos There is no WIP PR yet; as discussed plan is to merge this PR and start to work on use of shared_ptr.

I am okay  If you want to add the support of shared_ptr, why not add it in this PR?
```cpp
template<typename T, typename Mutex=pcpp::NullMutex, typename PointerType=std::add_pointer_t<T> >
class PointerVector {

		void pushBack(PointerType element)
		{
			std::lock_guard<Mutex> lk(m_Mutex);
			m_Vector.push_back(element);
		}

		PointerType front()
		{
			std::lock_guard<Mutex> lk(m_Mutex);
			return m_Vector.front();
		}
}
```

muthusaravanan3186:
As discussed this was the plan

@muthusaravanan3186 we can do it in phases. Make some changes discussed above without using shared pointers, and then in the future open another PR to move to shared pointers. The improvements suggested in this PR will make PointerVector mostly thread safe (even if not 100%). We can mention the edge cases in the doxygen documentation

tigercosmos:
> As discussed this was the plan
> 
> @muthusaravanan3186 we can do it in phases. Make some changes discussed above without using shared pointers, and then in the future open another PR to move to shared pointers. The improvements suggested in this PR will make PointerVector mostly thread safe (even if not 100%). We can mention the edge cases in the doxygen documentation

Then I oppose adding unnecessary locks for the future support of the shared pointer, because (1) those locks are useless for `T*` (2) using the shared pointer is not on the plan.  Let's focus on handling the case that uses `T*`, and forget about the share pointers at this moment.

muthusaravanan3186:
@tigercosmos  Back to square one. we have null mutex for normal case and mutex for multi-threading case. This was discussed. Since it is vector of pointers and we are accessing the pointers which itself is not thread safe. Since changing Pointervector to handle shared Pointers is a huge change; we planned to implement in phases. If we need everything at single shot it may take time. Please let me know what is the expectation. If not will drop this PR and work on something else. We already modified PCAPPLUSPLUS according to our need.

seladb:
@muthusaravanan3186 @tigercosmos @Dimi1010 I'm trying to catch up here... I think we need to decide on 2 issues, please correct me if I'm wrong:
1. The iterator is not thread-safe because after the user receives it the vector can change and invalidate the iterator
2. The vector holds pointers that are not thread-safe by nature, meaning one thread can hold the pointer and another thread can delete it from the vector which will free the pointer

Which issue(s) are you concerned about and is blocking this PR?

Regarding (1) - do you have any (simple) suggestion on how to solve it?
Regarding (2) - we need to use a `std::shared_ptr` which is indeed a bigger and a breaking change

Dimi1010:
@seladb @tigercosmos @muthusaravanan3186 

Tbh, I don't really see an easy solution to (1) while keeping the current API of the class. The two ways I could see iteration being done both have problems:
- You either need custom iterator objects that hold a reference to the vector mutex or some other kind of black magic, and even then mutating though the iterator would require custom calls.
- You remove the iteration capability from the 'thread-safe' version and add a thread-safe method that makes a 'non-thread-safe' snapshot of the objects in the vector at time of call and return that to the calling thread to do as it sees fit.

Regarding point 2. Using `std::unique_ptr` and `std::shared_ptr` would remove the need for manual allocation tracking, so that would be useful, and `std::shared_ptr` would work in removing the problem of dangling pointers after removal from the vector. 


With all that said, I am personally of the opinion that the burden of thread safety should not be placed on the vector object itself, but on the program that requires the vector data to be accessed by multiple threads simultaneously. My reasoning is that the vector and raw arrays is general provide way too many ways to access or mutate the data to be easily secured, while the program attempting to access the data from different threads would have a far better knowledge of how the data is being accessed to provide thread-safe data access.

A simpler solution would be having the different threads make a local copy of the vector contents while under synchronization and then proceeding independently using their local copies. This type of solution minimizes the points of contention where sync lock needs to be acquired between the threads, as the alternative would require a sync lock on every element access or mutation. It also removes the need of the underlying objects that are being held by the array being thread-safe too.

seladb:
After reading @Dimi1010 's response I started wondering what is the use case of a thread-safe pointer vector if both (1) and (2) are not solved...

In general, a thread-safe container is only needed when one or more threads are changing the vector, usually while other threads are reading data from it:
- If (1) is not solved then the threads can't use an iterator (in a `for` loop for example), so the only way to access the vector data is by using `at()` or `front()`. While that's possible, it's not very convenient
- If (2) is not solved then no thread can remove elements from the vector, otherwise other threads may hold invalid pointers. That's a pretty major limitation that narrows down many of the use-cases for using the vector in a multithread environment

@muthusaravanan3186 can you share your use-case for thread-safe pointer vector? It might help us decide if we actually want to implement it or not

seladb:
@muthusaravanan3186 @Dimi1010 @tigercosmos let's decide if we want to keep this PR or close it?

muthusaravanan3186:
@seladb sorry for late reply.. let's close this PR. Let me if you have anything to work.. will start to work on that

tigercosmos:
IMO, the thread-safe issue is interesting. Perhaps put this to issue and keep tracking?

seladb:
> IMO, the thread-safe issue is interesting. Perhaps put this to issue and keep tracking?

@tigercosmos there was an issue for it: https://github.com/seladb/PcapPlusPlus/issues/234 but we closed it. We can reopen it if needed, but the real question is whether we want to address it, and if so - how?

seladb:
> @seladb sorry for late reply.. let's close this PR. Let me if you have anything to work.. will start to work on that

@muthusaravanan3186 thank you for being willing to help! You can choose any issue from the [issues list](https://github.com/seladb/PcapPlusPlus/issues) or suggest an improvement of your own. If you're looking for "quick wins" you can work on these issues:
- https://github.com/seladb/PcapPlusPlus/issues/906
- https://github.com/seladb/PcapPlusPlus/issues/1347

Or if you want to implement a new protocol:
- https://github.com/seladb/PcapPlusPlus/issues/917
- https://github.com/seladb/PcapPlusPlus/issues/1179

Dimi1010:
I vote close it. I don't believe the vector itself is the correct location for thread safety to be located as I explained in my earlier reasoning.

tigercosmos:
sure, let's close it.

PR #1331, opened 2024-03-01
Fixed a potential memory leak during PointerVector deep copy.

### LINKED ISSUE #1326: Refactor PointerVector to use std::unique_ptr internally?
It would remove the need to manually track 'new'/'delete' calls of objects and will minimize the risk of memory leaks.

`PonterVector`'s copy ctor has a potential memory leak:
```
PointerVector(const PointerVector& other)
{
	for (const auto iter : other)
	{
		T* objCopy = new T(*iter);
		m_Vector.push_back(objCopy);  // <-- If this throws objCopy is leaked.
	}
}
```

Dimi1010:
Closes #1326 for the most part. Full change to unique_ptr has issues with the iterators and keeping backwards compat.

### DIFF
--- a/Common++/header/PointerVector.h
+++ b/Common++/header/PointerVector.h
@@ -67,6 +67,7 @@ namespace pcpp
 					catch (const std::exception&)
 					{
 						delete objCopy;
+						throw;
 					}
 				}
 			}
@@ -76,6 +77,7 @@ namespace pcpp
 				{
 					delete obj;
 				}
+				throw;
 			}
 		}
 
--- a/Common++/header/PointerVector.h
+++ b/Common++/header/PointerVector.h
@@ -58,7 +58,14 @@ namespace pcpp
 			for (const auto iter : other)
 			{
 				T* objCopy = new T(*iter);
-				m_Vector.push_back(objCopy);
+				try
+				{
+					m_Vector.push_back(objCopy);
+				}
+				catch (const std::exception&)
+				{
+					delete objCopy;
+				}
 			}
 		}
 
--- a/Common++/header/PointerVector.h
+++ b/Common++/header/PointerVector.h
@@ -55,16 +55,26 @@ namespace pcpp
 		 */
 		PointerVector(const PointerVector& other)
 		{
-			for (const auto iter : other)
+			try
 			{
-				T* objCopy = new T(*iter);
-				try
+				for (const auto iter : other)
 				{
-					m_Vector.push_back(objCopy);
+					T* objCopy = new T(*iter);
+					try
+					{
+						m_Vector.push_back(objCopy);
+					}
+					catch (const std::exception&)
+					{
+						delete objCopy;
+					}
 				}
-				catch (const std::exception&)
+			}
+			catch (const std::exception&) 
+			{
+				for (auto obj : m_Vector)
 				{
-					delete objCopy;
+					delete obj;
 				}
 			}
 		}
--- a/Common++/header/PointerVector.h
+++ b/Common++/header/PointerVector.h
@@ -71,7 +71,7 @@ namespace pcpp
 					}
 				}
 			}
-			catch (const std::exception&) 
+			catch (const std::exception&)
 			{
 				for (auto obj : m_Vector)
 				{


Common++/header/PointerVector.h
```diff
@@ -58,7 +58,14 @@ namespace pcpp
 			for (const auto iter : other)
 			{
 				T* objCopy = new T(*iter);
-				m_Vector.push_back(objCopy);
+				try
```
tigercosmos, line 61: why do we need try-catch here? if we get an exception during the construction, the program should throw directly, Also, in `~PointerVector()`, the function already deletes the pointer.

Dimi1010: Because if the ctor throws you lose the pointer `objCopy` when the stack is unwinded and the memory allocated by `new T(*iter)` is leaked as you don't have any pointers referencing them after that.

Also did some tests and if an object ctor throws, the object's destructor is not called. All the initialized member fields do get their destructors called, but the object being constructed does not. So in actuality all the new deep copies of T are leaked if something fails during copy ctor.

Look at this code:
image
image

See how in the first version where it does not throw, `a` and `a2` are destructed, but in the second version where `A(const A& o)` throws, only `a` is destructed because `a2`'s copy ctor threw. This is essentially what will happen if PointerVector's copy ctor threw for any reason. The `m_Vector` is deconstructed properly, but you leak all the memory that the pointers being held within the vector point to.

Dimi1010: Also, yeah, mb, exception should be rethrown after delete.

Dimi1010: > Also, I am wondering if it ever has a chance to throw during `push()`?

`pushBack()` does not really have an issue with memory leaks internally, since it does no allocations directly. It is the caller's responsibility to have allocated the element T that is being pushed, and to safely do error handling if the push fails. and the `m_Vector.push_back()` will do its own recovery when throwing, so no need for interaction there.

tigercosmos: I understand your concern, but I am not sure if using the current two-level try-catch blocks is the best practice or not...
I agree that this patch ensures memory safety and `try` doesn't increase overhead. However, from the maintenance and conciseness perspective, I think we could find a better solution instead of using the two-level try-catch, what do you think?

Dimi1010: Technically a better way would be utilizing `unique_ptr` as a holding type for the internal vector. It would nicely handle most deallocations automatically. Tho the issue there ends up requiring making custom iterators if `PointerVector` is to keep its current API of returning pointers from the iterators.

seladb: I agree it's not very easy to read, but I think it's fine because it's quite short

seladb (approved):

Common++/header/PointerVector.h
```diff
@@ -55,10 +55,29 @@ namespace pcpp
...
+					catch (const std::exception&)
+					{
+						delete objCopy;
+						throw;
+					}
+				}
+			}
+			catch (const std::exception&)
+			{
+				for (auto obj : m_Vector)
```
tigercosmos, line 76: this part of logic needs to be fixed, you just delete the pointer of the element in the vector, but you didn't remove the element.

tigercosmos (changes requested):

Common++/header/PointerVector.h
tigercosmos: correct me if i am wrong,  I think the correct logic here should be
 ```cpp
 for (const auto iter : other)
{
    try 
    {
        T* objCopy = new T(*iter);
    } catch (std::exception& e) {
        delete objCopy;
        throw e;
    }
    try{
        m_Vector.push_back(objCopy);
    } catch (std::exception& e) {
        throw e;
    }
}
```

Dimi1010: Uh, not really. This has the exact same issue from where we started, but with extra steps.

If the new statement or the ctor for T in it throws, the memry allocated by new is automatically freed. Also objectCopy isnt really assigned so you don't have anything to delete.

`push_back` has a strong exception guarantee in the standard. Which means if the push failed, the vector is rolled back to the last valid state (pre-push), so the pointer is not added. That is at the core of the whole memory leak issue. If the push fails, the ptr is not added and you lose the `objCpy`, so you have to `delete` the memory as the exception is being propagated outwards before you leave the function scope.

tigercosmos: I understand now, thank you for clarifying. As per the current code logic, if an element cannot be successfully pushed into `m_Vector`, it triggers an exception within the inner try block. The outer try then catches this exception and enters the corresponding catch block, leading to the deletion of all pointers in `m_Vector`. Nevertheless, following these steps, the elements within `m_Vector` persist, but they now reference invalid memory, posing a potential issue?

Dimi1010: Since this is a copy constructor the rules for throwing exceptions out of ctors apply. All initialized member fields have their destructors called as the exception propagates out of the ctor and the object destructor is not called as it did not finish construction. 

Therefor `m_Vector` is deallocated wholesale as the exception leaves the ctor, so it is pointless to clear the vector if we don't have any complex logic using it. As it is, the exception is immediately rethrown so we skip to member field deallocation after the outer try block.

tigercosmos: True, thanks for correcting me.

tigercosmos (approved):

tigercosmos:
@Dimi1010  thanks

PR #1336, opened 2024-03-06
upgrade LightPcapNg to latest master

tigercosmos:
upgrade LightPcapNg to the latest master.
There are a few changes on LightPcapNg.
Since we also have our modifications, so I manually cherry-pick the difference between the current version and the latest master version of LightPcapNg.


3rdParty/LightPcapNg/LightPcapNg/src/light_alloc.c
```diff
@@ -54,8 +54,12 @@ light_pcapng light_alloc_block(uint32_t block_type, const uint32_t *block_body,
 
 	PADD32(block_body_length, &actual_size);
 
-	pcapng_block->block_total_length = actual_size; // This value MUST be a multiple of 4.
-	block_body_size = actual_size - 2 * sizeof(pcapng_block->block_total_length) - sizeof(pcapng_block->block_type);
+	pcapng_block->block_total_length = actual_size + 2 * sizeof(pcapng_block->block_total_length) + sizeof(pcapng_block->block_type); // This value MUST be a multiple of 4.
+	block_body_size = actual_size;
+	
+	// Test passes with the old code???
```
tigercosmos, line 60: I highlight here, by using the old version code, the test can pass, however, the latest code will fail with the following error.

```
ERROR at /home/mujin/PcapPlusPlus/3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c::__parse_options::67: 152 != 0
ERROR at /home/mujin/PcapPlusPlus/3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c::__parse_options::67: 152 != 0
TestPcapNgFileReadWriteAdv         : FAILED (/home/mujin/PcapPlusPlus/Tests/Pcap++Test/Tests/FileTests.cpp:521). Assert EQUAL failed:
   Actual:   
   Expected: Linux 3.18.1-1-ARCH
```

tigercosmos: It's related to this PR: https://github.com/rvelea/LightPcapNg/pull/6/files

seladb: I have no idea why they made this change, but the pcapng file in Wireshark shows the expected OS information. Maybe they have a bug? 🤔 

image

tigercosmos: @seladb  I decided to keep the old implementation.

3rdParty/LightPcapNg/LightPcapNg/src/light_alloc.c
```diff
@@ -54,8 +54,16 @@ light_pcapng light_alloc_block(uint32_t block_type, const uint32_t *block_body,
 
 	PADD32(block_body_length, &actual_size);
 
+	// ---> PCPP patch
+	// keep the original implementation, since it is not clear why the upstream code was changed, and the original implementation is working fine
```
tigercosmos, line 58: leave a comment here for future migration.

3rdParty/LightPcapNg/LightPcapNg/src/light_alloc.c
tigercosmos: please check the comment: https://github.com/seladb/PcapPlusPlus/pull/1336#discussion_r1513971881

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
```diff
@@ -442,6 +442,11 @@ char *light_pcapng_to_string(light_pcapng pcapng)
 
 uint32_t *light_pcapng_to_memory(const light_pcapng pcapng, size_t *size)
 {
+   if (pcapng == NULL) {
+		*size = 0;
```
egecetin, line 446: Possible null pointer dereference if `size = NULL`. Maybe we should add another `if` to check this

```
if (pcapng == NULL) {
    if (size != NULL) {
       *size = 0;
    }
   ....
}
```

tigercosmos: nice catch, fixed.

egecetin: I checked the code again looks like my second suggestion is better for remove the possibility completely because there are some other `*size =` calls in function. At first glance I misread the code sorry, even after this fix there is still possibility of null pointer dereference. So can you update the code with,

```
if (size == NULL) { // PCPP fix
   return NULL;
}
if (pcapng == NULL) {
  ...
}
```

tigercosmos: sure, seems better. just fixed.

seladb (approved):

Issue #1340, opened 2024-03-14
Power of 2 restriction for number of queues in DpdkDevice


lucashc:
Hi there,

First of all, thank you for this amazing project. I am currently using the DpdkDevice class to set-up a PCIe ethernet card.
I noticed that when using the `DpdkDevice::openMultiQueues` a check gets executed whether the number of queues is a power of 2. I believe that this is necessary for VF devices, but this is not necessary for raw PCIe devices like the devices supported by the MLX5 driver. 

I have a pure C program that calls DPDK directly, and it operators correctly on a MLX5 device with 15 RX/TX queues. The reason for having that precise number of queues is that on a 16-core system, one has 1 core as main core and then 15 worker queues, each of which is hooked up to one RX/TX queue pair. Using 16 queues in this scenario with 15 worker cores, would leave some packets dangling.

The specific code is: [`Pcap++/src/DpdkDevice.cpp:292`](https://github.com/seladb/PcapPlusPlus/blob/1ff377c5de22caf9907fcc31f3f71ea138346155/Pcap%2B%2B/src/DpdkDevice.cpp#L292).

Could this check be expanded such that it only runs on VF devices, but not on other devices?
Or better, is the check really necessary? Doesn't DPDk throw a descriptive error in that case?

seladb:
Thanks @lucashc for flagging it! I wasn't aware that PCIe cards don't have this restriction. There is a very easy fix for it: we already have an [`isVirtual()`](https://github.com/seladb/PcapPlusPlus/blob/1ff377c5de22caf9907fcc31f3f71ea138346155/Pcap%2B%2B/src/DpdkDevice.cpp#L546) method that we can use to determine whether or not to do this check. We only need to add it.

Would you consider opening a PR with the fix?

seladb:
Fixed in #1343

PR #1343, opened 2024-03-15
Fix power of 2 check #1340

lucashc:
Hi there,

This pull requests fixes issue #1340 by adding a check whether `DpdkDevice` is virtual and then checking if the the number of RX queues is a power of 2. In the case of PCIe-devices this is not necessary.


Pcap++/src/DpdkDevice.cpp
```diff
@@ -289,12 +289,15 @@ bool DpdkDevice::configurePort(uint8_t numOfRxQueues, uint8_t numOfTxQueues)
...
 
-	// verify num of RX queues is power of 2
-	bool isRxQueuePowerOfTwo = !(numOfRxQueues == 0) && !(numOfRxQueues & (numOfRxQueues - 1));
-	if (!isRxQueuePowerOfTwo)
+	// verify num of RX queues is power of 2 for virtual devices
+	if (isVirtual())
 	{
-		PCPP_LOG_ERROR("Num of RX queues must be power of 2 (because of DPDK limitation). Attempted to open device with " << numOfRxQueues << " RX queues");
-		return false;
+		bool isRxQueuePowerOfTwo = !(numOfRxQueues == 0) && !(numOfRxQueues & (numOfRxQueues - 1));
```
tigercosmos, line 295: let's refactor this: `!(numOfRxQueues == 0)` -> (numOfRxQueues != 0)

seladb: We can also do:
```cpp
bool isRxQueuePowerOfTwo = !(numOfRxQueues == 0 || (numOfRxQueues & (numOfRxQueues - 1)));
```

I'm not sure what is nicer though...

lucashc: Perhaps the check for power of 2 and the number of queues being zero should be separated? We should throw an error when the requested number of queues is zero, right? Even in the non-virtual case

seladb: @lucashc you're right, I totally agree! 👍

lucashc: Actually, do we support only sending packets? In that case it might make sense to allow zero rx queues. 
In any case, I will add the patch for separating the checks.

seladb: we could support only sending packets, but I think this is a very uncommon use-case. I think we can keep the restriction to disallow zero rx queues

seladb (approved):

seladb:
@lucashc thank you for this contribution to the project, it's very much appreciated! 🙏 ❤️

PR #1346, opened 2024-03-21
Check size of tcpFragmentList before delete fragment

MCredbear:
Fragment might be delete after calling m_OnMessageReadyCallback (eg. call TcpReassembly::closeConnection in m_OnMessageReadyCallback), so we need to check if it's already deleted .


Packet++/src/TcpReassembly.cpp
```diff
@@ -516,9 +516,12 @@ void TcpReassembly::checkOutOfOrderFragments(TcpReassemblyData* tcpReassemblyDat
 						}
 					}
 
-
-					// remove fragment from list
-					tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.erase(tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.begin() + index);
+					// fragment might be delete after calling m_OnMessageReadyCallback (eg. call TcpReassembly::closeConnection in m_OnMessageReadyCallback), so we need to check if it's already deleted 
+					if (tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.size() != 0)
```
tigercosmos, line 520: I think we should check if `tcpReassemblyData->twoSides[sideIndex].tcpFragmentList[tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.begin() + index]` exists, not checking the size?

MCredbear: Do you mean `if (tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.begin() + index != tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.end()` ?

MCredbear: I thought another problem: if we call reassemblePackage in m_OnMessageReadyCallback...... So maybe we should create a temporary fragment, pop the matehed fragment to the temporary one, send the temporary one to the m_OnMessageReadyCallback ?

tigercosmos: actually should be `if (tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.size() >  index)`, e.g. index=0, size=1; index=3, size=4

tigercosmos: > I thought another problem ...

could you try to write some tests for the cases that you mentioned? It will help us understand the issue easily.

MCredbear: I think the `index` has been outdated due to the operation in `m_OnMessageReadyCallback`, the temporary fragment might be the best solution.

tigercosmos: I am not very familiar with this part of code, but it sounds correct.
But still, we will need some tests to verify our assumption.

MCredbear: > > I thought another problem ...
> 
> could you try to write some tests for the cases that you mentioned? It will help us understand the issue easily.

This is the case which I face
```
void CaptureWorkerThread::tcpReassemblyMsgReadyCallback(const int8_t sideIndex, const pcpp::TcpStreamData& tcpData,
    void* userCookie)
{
    // extract the connection manager from the user cookie
    auto captureWorkerThread = static_cast<CaptureWorkerThread*>(userCookie);
    auto& connectionManager = captureWorkerThread->m_TcpConnectionManager;

    // check if this flow already appears in the connection manager. If not, igonre it
    auto flow = connectionManager.find(tcpData.getConnectionData().flowKey);
    if (flow == connectionManager.end()) {
        return;
        // connectionManager.insert(std::make_pair(tcpData.getConnectionData().flowKey, TcpConnectionData()));
        // flow = connectionManager.find(tcpData.getConnectionData().flowKey);
    }

    auto& connection = flow->second;

    if (connection.numOfDataPackets < 50) {
        connection.tcpData[connection.numOfDataPackets].side = sideIndex;
        connection.tcpData[connection.numOfDataPackets].data = tcpData.getData();
        connection.tcpData[connection.numOfDataPackets].length = tcpData.getDataLength();

        // count number of packets
        connection.numOfDataPackets += 1;

        // insert packets into database and close the connection after getting 50 packets in a flow
        if (connection.numOfDataPackets >= 50) {
            std::cout << "Inserting to Redis" << std::endl;
            captureWorkerThread->m_DataBase.insert(connection);
            connectionManager.erase(flow);
            captureWorkerThread->m_TcpReassembly.closeConnection(tcpData.getConnectionData().flowKey);
        }
    }
}
```

In this case, when `checkOutOfOrderFragments` calles `m_OnMessageReadyCallback` in the code above, the `m_OnMessageReadyCallback` calls `closeConnection`, and `closeConnection` calls `checkOutOfOrderFragments` again, the last `checkOutOfOrderFragments` clears the `tcpReassemblyData->twoSides[sideIndex].tcpFragmentList`, and causes error when the first `checkOutOfOrderFragments`  call `tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.erase(tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.begin() + index);`

tigercosmos: Referring to your example, maybe the appropriate approach might be verifying whether the connection is closed subsequent to the callback invocation.

you can refer to some similar cases here, we have crazy manipulations in the callback in other places:
https://github.com/seladb/PcapPlusPlus/blob/5d217fff8a24d11a90d3cf088f85bd58d3b4125a/Tests/Pcap%2B%2BTest/Tests/LiveDeviceTests.cpp#L442-L450

MCredbear: > Referring to your example, maybe the appropriate approach might be verifying whether the connection is closed subsequent to the callback invocation.
> 
> you can refer to some similar cases here, we have crazy manipulations in the callback in other places:
> 
> https://github.com/seladb/PcapPlusPlus/blob/5d217fff8a24d11a90d3cf088f85bd58d3b4125a/Tests/Pcap%2B%2BTest/Tests/LiveDeviceTests.cpp#L442-L450

I think verifying whether the connection is closed doesn't help when calling `reassemblePackage` in `m_OnMessageReadyCallback`, I will make a new commit to do with the temporary fragment soon.

tigercosmos: I believe that copying is not an optimal solution as it can reduce performance. The issue arises because `checkOutOfOrderFragments` is being called again while it is already in progress. To address this, consider introducing a flag that signals to the original `checkOutOfOrderFragments` whether it should continue execution after `m_OnMessageReadyCallback` returns. Alternatively, you can simply implement a check based on the size of the data as you currently did.

MCredbear: > I believe that copying is not an optimal solution as it can reduce performance. The issue arises because `checkOutOfOrderFragments` is being called again while it is already in progress. To address this, consider introducing a flag that signals to the original `checkOutOfOrderFragments` whether it should continue execution after `m_OnMessageReadyCallback` returns. Alternatively, you can simply implement a check based on the size of the data as you currently did.

No copy, just wait a minute

MCredbear: OK now, please check the new commit

tigercosmos: could you add a test in `Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp`?

we currently only have `tcpReassemblyMsgReadyCallback`, please have a test case that the connection is closed inside the callback.

MCredbear: > could you add a test in `Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp`?
> 
> we currently only have `tcpReassemblyMsgReadyCallback`, please have a test case that the connection is closed inside the callback.

I replace the `int index` with `iterator`, and I remove the `fragment` from `fragmentList` before calling `m_OnMessageReadyCallback`  and freeing it, there is no need to check whether the connection is closed I think.

tigercosmos: We cannot ensure the logic correctness unless we have a test case to verify it. In this case, I would like to know if your patch can work for the aforementioned case.

MCredbear: Sorry, I don't know write the test, but I can paste my code here

CaptureWorkerThread.h:
```
//
// Created by redbear on 24-3-7.
//

#ifndef CAPTURE_WORKER_H
#define CAPTURE_WORKER_H

#include <DnsLayer.h>
#include <DpdkDevice.h>
#include <DpdkDeviceList.h>
#include <IPLayer.h>
#include <TcpReassembly.h>

#include "DataBaes.h"
#include "TcpConnectionData.h"

#define MAX_RECEIVE_BURST 64

// typedef representing the connection manager
typedef std::unordered_map<uint32_t, TcpConnectionData> TcpConnectionManager;

class CaptureWorkerThread final : public pcpp::DpdkWorkerThread {
public:
    CaptureWorkerThread(uint32_t coreId, pcpp::DpdkDevice* device, int32_t rxQueueId);

    ~CaptureWorkerThread();

    bool run(uint32_t coreId) override;

    void stop() override;

    uint32_t getCoreId() const override;

private:
    static void handleDnsPacket(const pcpp::Packet& dnsPacket);

    /**
     * The callback being called by the TCP reassembly module whenever new data arrives on a certain connection
     */
    static void tcpReassemblyMsgReadyCallback(const int8_t sideIndex, const pcpp::TcpStreamData& tcpData,
        void* userCookie);

    /**
     * The callback being called by the TCP reassembly module whenever a new connection is found. This method adds the connection to the connection manager
     */
    static void tcpReassemblyConnectionStartCallback(const pcpp::ConnectionData& connectionData, void* userCookie);

    bool m_Stop = false;
    uint32_t m_CoreId;
    pcpp::DpdkDevice* p_Device;
    uint32_t m_RxQueueId;
    TcpConnectionManager m_TcpConnectionManager;
    pcpp::TcpReassemblyConfiguration m_TcpReassemblyConfiguration = pcpp::TcpReassemblyConfiguration(
        true, 5, 30, 50, true);
    pcpp::TcpReassembly m_TcpReassembly = pcpp::TcpReassembly(tcpReassemblyMsgReadyCallback, this,
        tcpReassemblyConnectionStartCallback,
        nullptr,
        // tcpReassemblyConnectionEndCallback,
        m_TcpReassemblyConfiguration);

    DataBase m_DataBase;
};

#endif // CAPTURE_WORKER_H

```

CaptureWorkerThread.cpp:
```
//
// Created by redbear on 24-3-7.
//

#include "CaptureWorkerThread.h"

#include "DataBaes.h"
#include "DnsBackTraceTable.h"

CaptureWorkerThread::CaptureWorkerThread(uint32_t coreId, pcpp::DpdkDevice* device, int32_t rxQueueId)
    : m_CoreId(coreId)
    , p_Device(device)
    , m_RxQueueId(rxQueueId)
{
}

CaptureWorkerThread::~CaptureWorkerThread()
{
}

bool CaptureWorkerThread::run(uint32_t coreId)
{
    pcpp::MBufRawPacket* mBufRawPackets[MAX_RECEIVE_BURST] = {};
    while (!m_Stop) {
        auto recievedPacketsCount = p_Device->receivePackets(mBufRawPackets, MAX_RECEIVE_BURST, m_RxQueueId);
        for (uint16_t i = 0; i < recievedPacketsCount; i += 1) {
            auto parsedPacket = pcpp::Packet(mBufRawPackets[i]);
            if (parsedPacket.isPacketOfType(pcpp::DNS))
                handleDnsPacket(parsedPacket);
            else {
                try {
                    m_TcpReassembly.reassemblePacket(parsedPacket);
                } catch (std::exception& e) {
                }
            }
        }
    }
    return true;
}

void CaptureWorkerThread::stop()
{
    m_Stop = true;
}

uint32_t CaptureWorkerThread::getCoreId() const
{
    return m_CoreId;
}

void CaptureWorkerThread::handleDnsPacket(const pcpp::Packet& dnsPacket)
{
    const auto dnsLayer = dnsPacket.getLayerOfType<pcpp::DnsLayer>();
    const auto ipLayer = dnsPacket.getLayerOfType<pcpp::IPLayer>();
    if (dnsLayer && ipLayer) {
        if (dnsLayer->getDnsHeader()->queryOrResponse == 1) {
            const auto timeStamp = dnsPacket.getRawPacketReadOnly()->getPacketTimeStamp().tv_sec;
            const auto query = dnsLayer->getFirstQuery();
            if (query) {
                const auto domain = query->getName();
                std::vector<std::string> resolvedIps;
                for (auto answer = dnsLayer->getFirstAnswer(); answer != nullptr;
                     answer = dnsLayer->getNextAnswer(answer)) {
                    if (answer->getDnsType() == pcpp::DNS_TYPE_A) {
                        auto resolvedIp = answer->getData().castAs<pcpp::IPv4DnsResourceData>()->getIpAddress().toString();
                        resolvedIps.push_back(resolvedIp);
                    } else if (answer->getDnsType() == pcpp::DNS_TYPE_AAAA) {
                        auto resolvedIp = answer->getData().castAs<pcpp::IPv6DnsResourceData>()->getIpAddress().toString();
                        resolvedIps.push_back(resolvedIp);
                    }
                }
                const auto queryIp = ipLayer->getDstIPAddress().toString();
                DnsBackTraceTable::updateTable(queryIp, resolvedIps, timeStamp, domain);
            }
        }
    }
}

void CaptureWorkerThread::tcpReassemblyMsgReadyCallback(const int8_t sideIndex, const pcpp::TcpStreamData& tcpData,
    void* userCookie)
{
    // extract the connection manager from the user cookie
    auto captureWorkerThread = static_cast<CaptureWorkerThread*>(userCookie);
    auto& connectionManager = captureWorkerThread->m_TcpConnectionManager;

    // check if this flow already appears in the connection manager. If not, igonre it
    auto flow = connectionManager.find(tcpData.getConnectionData().flowKey);
    if (flow == connectionManager.end()) {
        return;
        // connectionManager.insert(std::make_pair(tcpData.getConnectionData().flowKey, TcpConnectionData()));
        // flow = connectionManager.find(tcpData.getConnectionData().flowKey);
    }

    auto& connection = flow->second;

    if (connection.numOfDataPackets < 50) {
        connection.tcpData[connection.numOfDataPackets].side = sideIndex;
        connection.tcpData[connection.numOfDataPackets].data = tcpData.getData();
        connection.tcpData[connection.numOfDataPackets].length = tcpData.getDataLength();

        // count number of packets
        connection.numOfDataPackets += 1;

        if (connection.numOfDataPackets >= 50) {
            std::cout << "Inserting to Redis" << std::endl;
            captureWorkerThread->m_DataBase.insert(connection);
            connectionManager.erase(flow);
            captureWorkerThread->m_TcpReassembly.closeConnection(tcpData.getConnectionData().flowKey);
        }
    }
}

void CaptureWorkerThread::tcpReassemblyConnectionStartCallback(const pcpp::ConnectionData& connectionData,
    void* userCookie)
{
    // get a pointer to the connection manager
    auto captureWorkerThread = static_cast<CaptureWorkerThread*>(userCookie);
    auto& connectionManager = captureWorkerThread->m_TcpConnectionManager;
    // look for the connection in the connection manager
    auto connection = connectionManager.find(connectionData.flowKey);

    // assuming it's a new connection
    if (connection == connectionManager.end()) {
        // add it to the connection manager
        connectionManager.insert(std::make_pair(connectionData.flowKey, TcpConnectionData()));
    }
}

// void CaptureWorkerThread::tcpReassemblyConnectionEndCallback(const pcpp::ConnectionData& connectionData,
//     pcpp::TcpReassembly::ConnectionEndReason reason, void* userCookie)
// {
//     // get a pointer to the connection manager
//     auto captureWorkerThread = static_cast<CaptureWorkerThread*>(userCookie);
//     auto &connectionManager = captureWorkerThread->m_TcpConnectionManager;

//     // find the connection in the connection manager by the flow key
//     auto connection = connectionManager.find(connectionData.flowKey);

//     // connection wasn't found - shouldn't get here
//     if (connection == connectionManager.end())
//         return;

//     // remove the connection from the connection manager
//     connectionManager.erase(connection);
// }

```
At least in my case, this patch works fine

MCredbear: > As for an additional test - if I understand correctly we want a test where we close a connection manually during an `OnTcpMessageReady` callback. It should be very complex to write. Let me know if you have an issue writing it, I might be able to help.

Yes, but I think maybe just modify current `TcpReassemblyTests.cpp` is enough.

seladb: @MCredbear yes, we can add a test in `TcpReassemblyTests.cpp`. Do you think you can work on it?

MCredbear: OK, I'll have a try.

tigercosmos: > we currently only have `tcpReassemblyMsgReadyCallback`, please have a test case that the connection is closed inside the callback.

As I mentioned above, you can see how we do for `tcpReassemblyMsgReadyCallback` in `TcpReassemblyTests.cpp`, and maybe have a callback `tcpReassemblyMsgCallbackConnectionClose`. You may also want to refer to `PcapPlusPlus/Tests/Pcap++Test/Tests/LiveDeviceTests.cpp` for more examples of testing callbacks.

seladb: We currently have [`tcpReassemblyTest`](https://github.com/seladb/PcapPlusPlus/blob/5d217fff8a24d11a90d3cf088f85bd58d3b4125a/Tests/Pcap%2B%2BTest/Tests/TcpReassemblyTests.cpp#L177) that runs TCP reassembly and registers to the 3 callbacks (message ready, connections start, connection end). However, the [`tcpReassemblyMsgReadyCallback`](https://github.com/seladb/PcapPlusPlus/blob/5d217fff8a24d11a90d3cf088f85bd58d3b4125a/Tests/Pcap%2B%2BTest/Tests/TcpReassemblyTests.cpp#L98C13-L98C42) doesn't close the connection in the middle.

I think that modifying [`tcpReassemblyTest`](https://github.com/seladb/PcapPlusPlus/blob/5d217fff8a24d11a90d3cf088f85bd58d3b4125a/Tests/Pcap%2B%2BTest/Tests/TcpReassemblyTests.cpp#L177) to support this use case won't be easy. Hence, I'd recommend writing a new function, equivalent to [`tcpReassemblyTest`](https://github.com/seladb/PcapPlusPlus/blob/5d217fff8a24d11a90d3cf088f85bd58d3b4125a/Tests/Pcap%2B%2BTest/Tests/TcpReassemblyTests.cpp#L177) and a new "message ready callback" that closes the connection after it got X packets (while there are still out-of-ordered packets in the buffer).

@MCredbear if you need an existing scenario that has out-of-order packets you can use this test: https://github.com/seladb/PcapPlusPlus/blob/5d217fff8a24d11a90d3cf088f85bd58d3b4125a/Tests/Pcap%2B%2BTest/Tests/TcpReassemblyTests.cpp#L416

MCredbear:
I add a new function to specially test that case, appologize for the ugly code

Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp
seladb, line 1: In order for the test to run you need to add `PTF_RUN_TEST` here: https://github.com/seladb/PcapPlusPlus/blob/e48be80c8671ca2018db24947c7d71df9c41b1fd/Tests/Pcap%2B%2BTest/main.cpp#L302

Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp
```diff
@@ -211,6 +268,47 @@ static bool tcpReassemblyTest(const std::vector<pcpp::RawPacket>& packetStream,
 }
 
 
+// ~~~~~~~~~~~~~~~~~~~
+// tcpReassemblyTestManuallyCloseConnOnMesgReady()
+// ~~~~~~~~~~~~~~~~~~~
+
+static bool tcpReassemblyTestManuallyCloseConnOnMesgReady(const std::vector<pcpp::RawPacket>& packetStream, TcpReassemblyMultipleConnStatsWithReassembly& results, bool monitorOpenCloseConns, bool closeConnsManually)
```
seladb, line 275: I think we can make this method simpler - I don't think that `monitorOpenCloseConns` and `closeConnsManually` are needed...

Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp
```diff
@@ -55,6 +55,28 @@ struct TcpReassemblyMultipleConnStats
...
+	FlowKeysList flowKeysList;
+
+	std::vector<timeval>timestamps;
+	void clear()
+	{
+		stats.clear();
+		flowKeysList.clear();
+	}
+
+	pcpp::TcpReassembly *tcpReassmbly = nullptr;
```
seladb, line 77: Maybe instead of duplicating the entire struct we can just inherit from `TcpReassemblyMultipleConnStats` and add `pcpp::TcpReassembly *tcpReassmbly`?
Or another option is to add `pcpp::TcpReassembly *tcpReassmbly` to `TcpReassemblyMultipleConnStats`

MCredbear:
图片
图片
Sorry, it occurs error again after I deploy my project in a 20Gb/s network..... Currently I'm still figuring what causes this.

tigercosmos:
> 图片 图片 Sorry, it occurs error again after I deploy my project in a 20Gb/s network..... Currently I'm still figuring what causes this.

are you able to reproduce it? adding the root cause in the test is helpful for the edge cases.

MCredbear:
> > 图片 图片 Sorry, it occurs error again after I deploy my project in a 20Gb/s network..... Currently I'm still figuring what causes this.
> 
> are you able to reproduce it? adding the root cause in the test is helpful for the edge cases.

Yes, and my new commit fixes it, thought I think this error shouldn't be caused...

davidbolvansky:
I am using TcpReassembly and it crashes with double free / invalid pointer in checkOutOfOrderFragments, but only on ARM, interestingly.

Looking at this, it seems this PR should fix crashes we see. I have a pcap with recorded traffic, but unfortunely I cannot share it. But I could help your with testing.

tigercosmos:
> I am using TcpReassembly and it crashes with double free / invalid pointer in checkOutOfOrderFragments, but only on ARM, interestingly.
> 
> Looking at this, it seems this PR should fix crashes I see. I have a pcap with recorded traffic, but unfortunelly I cannot share it. But I could help you with testing.

That would be great, would you like to help test the PR with your pcap file on the ARM machine?

davidbolvansky:
Yes, the bug happens when you call closeConnection from OnMessageReadyCallback.

@seladb

tigercosmos:
@davidbolvansky could you please furnish a concise reproducible code sample illustrating the issue you're encountering?

davidbolvansky:
As I said the bug happens when you call tcpReassembly.closeConnection(connData.flowKey) from OnMessageReadyCallback - OnMessageEndCallback is called for first time.

Then OnMessageEndCallback is called again, due normal fin. This second invocation triggers double free.

Tests/Pcap++Test/TestDefinition.h
```diff
@@ -64,6 +64,7 @@ PTF_TEST_CASE(TestTcpReassemblySanity);
 PTF_TEST_CASE(TestTcpReassemblyRetran);
 PTF_TEST_CASE(TestTcpReassemblyMissingData);
 PTF_TEST_CASE(TestTcpReassemblyOutOfOrder);
+PTF_TEST_CASE(TestTcpReassemblyOutOfOrderManuallyCloseConnOnMesgReady);
```
seladb, line 67: Can we make the test name shorter?
Maybe: `TestTcpReassemblyOOOCloseConn`

seladb:
@MCredbear can you please address the remaining comments so we can get the PR merged?

Tests/Pcap++Test/TestDefinition.h
MCredbear: Is `TestTcpReassemblyOutOfOrderWithManualClose` fine?

seladb: We should probably change that, but we currently have a hard-coded limit of 35 characters for test names, here is an example: https://github.com/seladb/PcapPlusPlus/blob/0c75603db71a9bef02803e9b0893a5074fcbd5da/Tests/PcppTestFramework/PcppTestFrameworkRun.h#L99

Tests that have a longer name will result in a weird UI print, something like:
```
...
...
TestTcpReassemblyMalformedPkts     : PASSED
TestTcpReassemblyMultipleConns     : PASSED
TestTcpReassemblyIPv6              : PASSED
TestTcpReassemblyIPv6MultConns     : PASSED
TestTcpReassemblyIPv6_OOO          : PASSED
TestTcpReassemblyCleanup           : PASSED
TestTcpReassemblyMaxOOOFrags       : PASSED
TestTcpReassemblyMaxSeq            : PASSED
TestTcpReassemblyDisableOOOCleanup : PASSED
TestTcpReassemblyOutOfOrderWithManualClose : PASSED
...
```

seladb:
@MCredbear please also address the previous comment:
https://github.com/seladb/PcapPlusPlus/pull/1346#discussion_r1545563549
https://github.com/seladb/PcapPlusPlus/pull/1346#discussion_r1545563703
https://github.com/seladb/PcapPlusPlus/pull/1346#discussion_r1545564233
https://github.com/seladb/PcapPlusPlus/pull/1346#discussion_r1571927077

And fix the pre-commit issue: https://github.com/seladb/PcapPlusPlus/actions/runs/8772352536/job/24071180251?pr=1346

seladb:
@MCredbear the new test fails in CI 😕

MCredbear:
Emm, I don't know what causes the new CI test error

egecetin:
@MCredbear In output side there are three different HTTP payloads with `Content-Length` 2248, 5261 and 6634. But for expected side there are four different HTTP payloads with `Content-Length` 2428, 5261, 6634 and 6649. So, it looks like the last request/reply pattern does not exist. If it is expected behavior, you should change the expected data.

Tests/Pcap++Test/TestDefinition.h
```diff
@@ -64,6 +64,7 @@ PTF_TEST_CASE(TestTcpReassemblySanity);
 PTF_TEST_CASE(TestTcpReassemblyRetran);
 PTF_TEST_CASE(TestTcpReassemblyMissingData);
 PTF_TEST_CASE(TestTcpReassemblyOutOfOrder);
+PTF_TEST_CASE(TestTcpReassemblyOOOWithManualClose);
```
tigercosmos, line 67: what is "OOO"?

Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp
```diff
@@ -211,6 +246,43 @@ static bool tcpReassemblyTest(const std::vector<pcpp::RawPacket>& packetStream,
...
+
+	tcpReassembly = new pcpp::TcpReassembly(tcpReassemblyManuallyCloseConnMsgReadyCallback, &results, tcpReassemblyConnectionStartCallback, tcpReassemblyConnectionEndCallback);
+
+	for (auto iter : packetStream)
+	{
+		pcpp::Packet packet(&iter);
+		tcpReassembly->reassemblePacket(packet);
+	}
+
+	//for(TcpReassemblyMultipleConnStats::Stats::iterator iter = results.stats.begin(); iter != results.stats.end(); iter++)
```
tigercosmos, line 265: forget to remove?

Tests/Pcap++Test/TestDefinition.h
seladb: OOO == Out Of Order

tigercosmos: I don't like the abbreviation, not straightforward, I would recommand `TestTcpReassemblyOutOfOrderWithManualClose`

seladb: The problem is that we currently have a limit of 35 chars per test (you can see it everywhere in `PcppTestFramework.h` so longer test names will look misaligned when running them. Perhaps we can remove this limitation and automatically detect the longest test name, but I'm not sure it'd be straightforward to do 🤔 

We can also extend the 35 limit to a larger number...

tigercosmos: @seladb  maybe we should increase the number? `TestTcpReassemblyOutOfOrderWithManualClose` length is 42, so I guess changing it to 50 is enough.

tigercosmos:
@MCredbear  would you like to finish the PR?

MCredbear:
> @MCredbear would you like to finish the PR?

Yes, sorry for the laggy reply, my computer's SSD broken last month, I'll commit later

MCredbear:
@seladb could you help me with the Tests? I'm confusing now....

seladb:
> @seladb could you help me with the Tests? I'm confusing now....

@MCredbear I'm looking...

seladb:
@MCredbear I looked into why the tests don't pass and I'm working on a fix.

The main problem with calling `closeConnection()` inside `onMessageReadyCallback` is that it may cause a recursive call in `checkOutOfOrderFragments()`, which leads to issues. Here is a scenario describing how it happens:
1. Assume we are in the middle of processing a TCP flow and there are some out-of-order fragments (packets) in the buffer
2. `checkOutOfOrderFragments()` is called and manages to find a packet in the out-of-order buffer that matches the current sequence
3. It calls the `onMessageReadyCallback` with that fragment (packet)
4. The callback calls `closeConnection()`
5. `closeConnection()` contains logic to call [`checkOutOfOrderFragments()`](https://github.com/seladb/PcapPlusPlus/blob/09261f05d513183df12a184f0c01f3fd2044f208/Packet%2B%2B/src/TcpReassembly.cpp#L672)
6.  `checkOutOfOrderFragments()` manages to find a packet in the out-of-order buffer that matches the current sequence
7. It calls the `onMessageReadyCallback`
8. The callback calls `closeConnection()` (again)
9.  `closeConnection()`  calls `checkOutOfOrderFragments()` (again)
10. And so on...

It's not an endless loop because eventually the packets in the out-of-order buffer will run out, but it creates issues when going back in the recursion. Also, `checkOutOfOrderFragments()` wasn't designed to run in a recursive manner.

The fix I'm working on is to prevent this recursion by adding a flag called `m_ProcessingOutOfOrder` that will be checked inside `checkOutOfOrderFragments()` and will return immediately if it's turned on.

seladb:
@MCredbear please take a look at the changes I made in this commit: https://github.com/seladb/PcapPlusPlus/pull/1346/commits/bdf16164177dfdff2f33c2350752dcaeba144b9e

All tests pass now.

seladb:
@MCredbear did you see my previous comments?
Let me know if we can merge the PR

Packet++/src/TcpReassembly.cpp
```diff
@@ -473,34 +473,40 @@ void TcpReassembly::handleFinOrRst(TcpReassemblyData* tcpReassemblyData, int8_t
...
 		do
 		{
-			index = 0;
+			it = tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.begin();
 			foundSomething = false;
 
 			// first fragment list iteration - go over the whole fragment list and see if can find fragments that match the current sequence
 			// or have smaller sequence but have big enough payload to get new data
-			while (index < (int)tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.size())
+			while (it != tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.end() && (tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.size() > 0))
```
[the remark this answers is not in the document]
seladb: You're right! Fixed in this commit: https://github.com/seladb/PcapPlusPlus/pull/1346/commits/8f0a546b4dd8e1e8a19433015355c71ef7a36ff4

Packet++/src/TcpReassembly.cpp
```diff
@@ -473,34 +473,40 @@ void TcpReassembly::handleFinOrRst(TcpReassemblyData* tcpReassemblyData, int8_t
...
+
+	m_ProcessingOutOfOrder = true;
 	bool foundSomething = false;
 
 	do
 	{
 		PCPP_LOG_DEBUG("Starting first iteration of checkOutOfOrderFragments - looking for fragments that match the current sequence or have smaller sequence");
 
-		int index = 0;
+		PointerVector<TcpFragment>::VectorIterator it;
```
Dimi1010, line 488: I don't really like how the iterator is declared here and then used all throughout the function. Sure perhaps it is more efficient (not really sure how much tho, since its stack allocated), but IMO it makes it harder to read the function code as I have to scroll up to figure out what that iterator is and why it is not declared at the loop usage, but somewhere else.

So far, it is only used twice. both times to do full loops through `tcpFragmentList`.
- Once on L498 through L568
- Again on L591 through L607, where the previous value is overwritten by a `begin()` call.

IMO, the iterator declaration should be pulled next to the L498 loop and the L591 loop be converted to a `for-loop`

Packet++/src/TcpReassembly.cpp
```diff
@@ -473,34 +473,40 @@ void TcpReassembly::handleFinOrRst(TcpReassemblyData* tcpReassemblyData, int8_t
...
 		foundSomething = false;
 
 		do
 		{
-			index = 0;
+			it = tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.begin();
```
Dimi1010, line 493: We could probably cache the `tcpReassemblyData->twoSizes[sideIndex]` result to a reference variable? It has been accessed 18 times in the function.

Packet++/src/TcpReassembly.cpp
```diff
@@ -581,30 +587,30 @@ void TcpReassembly::checkOutOfOrderFragments(TcpReassemblyData* tcpReassemblyDat
...
+			it++;
 		}
 
 		// this means fragment list is not empty at this stage
-		if (closestSequenceFragIndex > -1)
+		if (closestSequenceFragIt != tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.end())
 		{
 			// get the fragment with the closest sequence
-			TcpFragment* curTcpFrag = tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.at(closestSequenceFragIndex);
+			TcpFragment* curTcpFrag = tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.getAndRemoveFromVector(closestSequenceFragIt);
```
Dimi1010, line 613: ditto: `unique_ptr`

Packet++/src/TcpReassembly.cpp
```diff
@@ -473,34 +473,40 @@ void TcpReassembly::handleFinOrRst(TcpReassemblyData* tcpReassemblyData, int8_t
 
 void TcpReassembly::checkOutOfOrderFragments(TcpReassemblyData* tcpReassemblyData, int8_t sideIndex, bool cleanWholeFragList)
 {
+	if (m_ProcessingOutOfOrder)
+	{
+		return;
+	}
+
+	m_ProcessingOutOfOrder = true;
```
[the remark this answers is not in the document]
seladb: I implemented something in this commit: https://github.com/seladb/PcapPlusPlus/pull/1346/commits/6fc76c6902396c65108fe5842c91ca02ebede4f0

Do you think it's good enough, or do you have a suggestion for a better implementation?

Packet++/src/TcpReassembly.cpp
```diff
@@ -473,34 +473,40 @@ void TcpReassembly::handleFinOrRst(TcpReassemblyData* tcpReassemblyData, int8_t
...
+			while (it != tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.end())
 			{
-				TcpFragment* curTcpFrag = tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.at(index);
-
 				// if fragment sequence matches the current sequence
-				if (curTcpFrag->sequence == tcpReassemblyData->twoSides[sideIndex].sequence)
+				if ((*it)->sequence == tcpReassemblyData->twoSides[sideIndex].sequence)
 				{
+					// pop the fragment from fragment list
+					auto curTcpFrag = tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.getAndRemoveFromVector(it);
```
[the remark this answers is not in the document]
seladb: Good call! Updated in commit https://github.com/seladb/PcapPlusPlus/pull/1346/commits/baac906fd2d26b5abaa2f807af5c8d287db4712c

Packet++/src/TcpReassembly.cpp
```diff
@@ -511,18 +517,18 @@ void TcpReassembly::checkOutOfOrderFragments(TcpReassemblyData* tcpReassemblyDat
...
 
 					continue;
 				}
 
 				// if fragment sequence has lower sequence than the current sequence
-				if (SEQ_LT(curTcpFrag->sequence, tcpReassemblyData->twoSides[sideIndex].sequence))
+				if (SEQ_LT((*it)->sequence, tcpReassemblyData->twoSides[sideIndex].sequence))
 				{
+					// pop the fragment from fragment list
+					auto curTcpFrag = tcpReassemblyData->twoSides[sideIndex].tcpFragmentList.getAndRemoveFromVector(it);
```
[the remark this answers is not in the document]
seladb: Updated in commit https://github.com/seladb/PcapPlusPlus/pull/1346/commits/baac906fd2d26b5abaa2f807af5c8d287db4712c

Packet++/src/TcpReassembly.cpp
seladb: Updated in commit https://github.com/seladb/PcapPlusPlus/pull/1346/commits/baac906fd2d26b5abaa2f807af5c8d287db4712c

Packet++/src/TcpReassembly.cpp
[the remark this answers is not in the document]
Dimi1010: Its good enough. Doesn't really need much more.

Packet++/src/TcpReassembly.cpp
seladb: Update in this commit: https://github.com/seladb/PcapPlusPlus/pull/1346/commits/d1b610d27ec78704539fb6f7f7119b404c8b3e89

Packet++/src/TcpReassembly.cpp
seladb: Updated in this commit: https://github.com/seladb/PcapPlusPlus/pull/1346/commits/64562380a1b754123e8f3e461c5a6c1771bb8a11

Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp
```diff
@@ -120,6 +121,40 @@ static void tcpReassemblyMsgReadyCallback(int8_t sideIndex, const pcpp::TcpStrea
 }
 
 
+// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+// tcpReassemblyManuallyCloseConnMsgReadyCallback()
+// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+static void tcpReassemblyManuallyCloseConnMsgReadyCallback(int8_t sideIndex, const pcpp::TcpStreamData& tcpData, void* userCookie)
+{
+	TcpReassemblyMultipleConnStats::Stats &stats = ((TcpReassemblyMultipleConnStats*)userCookie)->stats;
```
Dimi1010, line 130: Nit: c-style cast

Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp
```diff
@@ -120,6 +121,40 @@ static void tcpReassemblyMsgReadyCallback(int8_t sideIndex, const pcpp::TcpStrea
...
+
+	if (sideIndex != iter->second.curSide)
+	{
+		iter->second.numOfMessagesFromSide[sideIndex]++;
+		iter->second.curSide = sideIndex;
+	}
+
+	((TcpReassemblyMultipleConnStats *)userCookie)->timestamps.push_back(tcpData.getTimeStamp());
+	iter->second.numOfDataPackets++;
+	iter->second.reassembledData += std::string((char*)tcpData.getData(), tcpData.getDataLength());
```
Dimi1010, line 149: Nit, ditto: C style cast

Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp
```diff
@@ -120,6 +121,40 @@ static void tcpReassemblyMsgReadyCallback(int8_t sideIndex, const pcpp::TcpStrea
...
+
+	// if numOfDataPackets hits 10, close the connection manually
+	if (iter->second.numOfDataPackets >= 10)
+	{
+		((TcpReassemblyMultipleConnStats*)userCookie)->tcpReassmbly->closeConnection(tcpData.getConnectionData().flowKey);
```
Dimi1010, line 154: Nit, ditto: C style cast

Dimi1010 (approved):
Some minor nits about C-style casts. Other that that LGTM.

Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp
seladb: Updated in this commit: https://github.com/seladb/PcapPlusPlus/pull/1346/commits/1aa0df801960052b3d2acc9b6e4450d16fddb728

Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp
seladb: Updated in this commit: https://github.com/seladb/PcapPlusPlus/pull/1346/commits/1aa0df801960052b3d2acc9b6e4450d16fddb728

Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp
seladb: Updated in this commit: https://github.com/seladb/PcapPlusPlus/pull/1346/commits/1aa0df801960052b3d2acc9b6e4450d16fddb728

seladb (approved):

seladb:
Thank you @MCredbear for working on this, much appreciated! 🙏 ❤️

MCredbear:
> Thank you @MCredbear for working on this, much appreciated! 🙏 ❤️

You are welcome, and sorry for the late reply, my school life is too busy this days that I forget to look at the PR

davidbolvansky:
Perfect, thanks.

Any plans for new release with this bugfix?

Dimi1010:
@davidbolvansky Next release is aimed to be in August.

If you need the bugfix urgently, the bugfix is also part of the master branch now, you so could potentially build from that.

PR #1353, opened 2024-04-03
Introduce ASN.1 BER decoder and encoder

seladb:
> Abstract Syntax Notation One (ASN.1) is a standard interface description language (IDL) for defining data structures that can be serialized and deserialized in a cross-platform way

Taken from Wikipedia: https://en.wikipedia.org/wiki/ASN.1

ASN.1 is used in a variety of protocols such as LDAP, SNMP, TLS X.509 Certificates, and more.

This PR introduces ASN.1 decoder and encoder. In later PR I'll add protocols (layers) that use this decoder/encoder to parse their data.


Packet++/header/Asn1Codec.h
```diff
@@ -0,0 +1,462 @@
...
+		 * @tparam Asn1RecordType The type to cast to
+		 * @return A pointer to the type after casting
+		 */
+		template <class Asn1RecordType>
+		Asn1RecordType* castAs()
+		{
+			auto result = dynamic_cast<Asn1RecordType*>(this);
+			if (result == nullptr)
+			{
+				throw std::runtime_error("Cast failed, instance isn't of the requested type");
```
tigercosmos, line 180: we should use `std::bad_cast` here

Packet++/src/Asn1Codec.cpp
```diff
@@ -0,0 +1,569 @@
...
+	void Asn1Record::decodeValueIfNeeded()
+	{
+		if (m_EncodedValue != nullptr)
+		{
+			decodeValue(m_EncodedValue, true);
+			m_EncodedValue = nullptr;
+		}
+	}
+
+	Asn1GenericRecord::Asn1GenericRecord(Asn1TagClass tagClass, bool isConstructed, uint8_t tagType, const uint8_t* value, size_t valueLen)
```
tigercosmos, line 325: any reason not to use the member initializer lists?

Packet++/src/Asn1Codec.cpp
```diff
@@ -0,0 +1,569 @@
...
+		m_Value = data[0] != 0;
+	}
+
+	std::vector<uint8_t> Asn1BooleanRecord::encodeValue() const
+	{
+		uint8_t byte = (m_Value ? 0xff : 0x00);
+		return { byte };
+	}
+
+	Asn1NullRecord::Asn1NullRecord() : Asn1PrimitiveRecord(static_cast<uint8_t>(Asn1UniversalTagType::Null))
```
tigercosmos, line 564: why not have a overloading for `Asn1PrimitiveRecord` to use `Asn1UniversalTagType` as the argument?

seladb: Do you mean create another c'tor for `Asn1PrimitiveRecord` that takes an argument of type `Asn1UniversalTagType`?

tigercosmos: yes `Asn1PrimitiveRecord(Asn1UniversalTagType type)`, I think it is clearer.

Packet++/src/Asn1Codec.cpp
```diff
@@ -0,0 +1,569 @@
...
+			throw std::invalid_argument("Cannot decode ASN.1 record, data doesn't contain the entire record");
+		}
+
+		decodedRecord->m_TotalLength = tagLen + lengthLen + decodedRecord->m_ValueLength;
+
+		if (!lazy)
+		{
+			try
+			{
+				decodedRecord->decodeValue((uint8_t*)data + tagLen + lengthLen, lazy);
```
tigercosmos, line 121: btw, since it's new code, can we use static_cast for all of these?

Packet++/src/Asn1Codec.cpp
```diff
@@ -0,0 +1,569 @@
...
+		m_TagType = tagType;
+		m_TagClass = Asn1TagClass::Universal;
+		m_IsConstructed = false;
+	}
+
+	Asn1IntegerRecord::Asn1IntegerRecord(uint32_t value) : Asn1PrimitiveRecord(static_cast<uint8_t>(Asn1UniversalTagType::Integer))
+	{
+		m_Value = value;
+
+		if (m_Value <= std::pow(2, sizeof(uint8_t) * 8))
```
tigercosmos, line 426: we can use `std::numeric_limits<uint8_t>::max();`

Packet++/src/Asn1Codec.cpp
```diff
@@ -0,0 +1,569 @@
...
+			}
+			case 3:
+			{
+				uint8_t tempArr[sizeof(uint32_t)];
+				auto hostValue = htobe32(static_cast<uint32_t>(m_Value));
+				memcpy(tempArr, &hostValue, sizeof(uint32_t));
+				std::copy(tempArr + 1, tempArr + m_ValueLength + 1, std::back_inserter(result));
+				break;
+			}
+			case 4:
```
tigercosmos, line 507: maybe use `sizeof(uint32_t)` here is clearer?

Packet++/src/Asn1Codec.cpp
```diff
@@ -0,0 +1,569 @@
...
+			{
+				uint8_t tempArr[sizeof(uint32_t)];
+				auto hostValue = htobe32(static_cast<uint32_t>(m_Value));
+				memcpy(tempArr, &hostValue, m_ValueLength);
```
tigercosmos, line 511: maybe we can have consistency, since you use `memcpy(tempArr, &hostValue, sizeof(uint32_t));` for case 3?

Packet++/src/Asn1Codec.cpp
```diff
@@ -0,0 +1,569 @@
...
+		auto decodedRecord = decodeTagAndCreateRecord(data, dataLen, tagLen);
+
+		int lengthLen;
+		try
+		{
+			lengthLen = decodedRecord->decodeLength(data + tagLen, dataLen - tagLen);
+		}
+		catch (...)
+		{
+			delete decodedRecord;
```
egecetin, line 105: Since I haven't checked the entire code I might miss a point but maybe `unique_ptr` can be used for 
`decodedRecord` and for return type of `decodeTagAndCreateRecord`?

Packet++/src/Asn1Codec.cpp
seladb: All members are part of the base class `Asn1Record` so we can't initialize them in a member initializer list: https://stackoverflow.com/questions/7405740/how-can-i-initialize-base-class-member-variables-in-derived-class-constructor

Packet++/src/Asn1Codec.cpp
seladb: I'm not sure I understand your suggestion...

Packet++/src/Asn1Codec.cpp
seladb: `decodeInternal()` returns a pointer that we can push into a `PointerVector` in case of a `Asn1ConstructedRecord`: https://github.com/seladb/PcapPlusPlus/blob/3dea318a164da4f3f723c917df70192ba73d7bd4/Packet%2B%2B/src/Asn1Codec.cpp#L390

Packet++/src/Asn1Codec.cpp
seladb: Fixed in this commit: https://github.com/seladb/PcapPlusPlus/pull/1353/commits/4b4865f2aa01e3c5ac619f91d7a6d0ff0227934c

Packet++/src/Asn1Codec.cpp
seladb: Fixed in this commit: https://github.com/seladb/PcapPlusPlus/pull/1353/commits/4b4865f2aa01e3c5ac619f91d7a6d0ff0227934c

Packet++/src/Asn1Codec.cpp
seladb: Fixed in this commit: https://github.com/seladb/PcapPlusPlus/pull/1353/commits/4b4865f2aa01e3c5ac619f91d7a6d0ff0227934c

Packet++/header/Asn1Codec.h
seladb: Fixed in these commits:
https://github.com/seladb/PcapPlusPlus/pull/1353/commits/4b4865f2aa01e3c5ac619f91d7a6d0ff0227934c
https://github.com/seladb/PcapPlusPlus/pull/1353/commits/cb0f1c02a0fef9c012cfc83da7c36df7683f90f7

Packet++/src/Asn1Codec.cpp
seladb: Oh got it. Fixed in https://github.com/seladb/PcapPlusPlus/pull/1353/commits/155788fa5b55d836931a8df30cedc602d77cd807

Packet++/src/Asn1Codec.cpp
```diff
@@ -0,0 +1,574 @@
...
+			case 2:
+			{
+				m_Value = be16toh(*(uint16_t*)data);
+				break;
+			}
+			case 3:
+			{
+				uint8_t tempData[4] = {0};
+				memcpy(tempData + 1, data, 3);
+				m_Value = be32toh(*(uint32_t*)tempData);
```
tigercosmos, line 469: nitpick: reinterpret_cast?

tigercosmos (approved):
LGTM. IMO, including the specification reference (URLs) in the comments would be advantageous

Packet++/src/Asn1Codec.cpp
seladb: Good catch! fixed here: https://github.com/seladb/PcapPlusPlus/pull/1353/commits/aab23e30a37b85a79d4aaee1322fda9c57d5fe0a

PR #1357, opened 2024-04-12
GVCP support

### LINKED ISSUE #1321: Introduce GigE Vision Control Protocol (GVCP) protocol
GigE Vision[[1]](https://en.wikipedia.org/wiki/GigE_Vision#cite_note-1) is an [interface standard](https://en.wikipedia.org/wiki/Interface_standard) introduced in 2006 for high-performance industrial cameras.

To support it, we will add a new layer beyond the UDP layer.

There are plenty of commands in GVCP, the first step would be:
- recognized UDP port 3956 as GVCP
- parse GVCP command header
- parse GVCP acknowledge header

Fully support of all commands and their response could be the next step.

see more references:
- Wikipedia: https://en.wikipedia.org/wiki/GigE_Vision
- Wireshark: https://www.wireshark.org/docs/dfref/g/gvcp.html
- Related repos: 
  - https://github.com/Touseefelahi/GigeVision
  - https://github.com/AravisProject/aravis
- Spec file: https://www.automate.org/vision/vision-standards/download-the-gige-vision-standard

tigercosmos:
Per #1321

This PR adds a layer for the GigE Vision Control Protocol (GVCP). It supports parsing, creation and editing.

I rebuilt the branch as one commit on the current `dev`. The commit addresses all open review comments.

## Design

- `GvcpRequestLayer` and `GvcpAcknowledgeLayer` represent all messages.
- Discovery and Force IP messages have their own layer classes.
- The header structs are plain data in `pcpp::internal`. The public API uses getters and setters.
- `GvcpLayer::parseGvcpLayer()` validates the data. `UdpLayer` falls back to `PayloadLayer` if the data is not valid.
- `getHeaderLen()` returns the full message length, because GVCP is always the last layer.
- `getPayloadData()` returns the message data after the header.
- `computeCalculateFields()` updates the data size field.
- Unexpected command and status values map to `Unknown`.
- The protocol type is `65`, because GTPv2 uses `57`.

## Fixes

- The acknowledge layers now keep the previous layer and the packet.
- `getVersion()` now converts the byte order.
- The string getters no longer read past the end of a field.

## Tests

The tests cover parsing, creation, editing and malformed data. All 267 tests of `Packet++Test` pass on macOS.


Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,632 @@
...
+ * @author An-Chi Liu (phy.tiger@gmail.com)
+ */
+
+/**
+ * @namespace pcpp
+ * @brief The main namespace for the PcapPlusPlus lib
+ */
+namespace pcpp
+{
+	namespace detail
```
Dimi1010, line 20: Maybe use `pcpp::internal`?

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,632 @@
...
+	{
+		static constexpr size_t kGvcpMagicNumber = 0x42;
+		static constexpr size_t kGvcpRequestHeaderLength = 8;
+		static constexpr size_t kGvcpAckHeaderLength = 8;
+		static constexpr size_t kGvcpDiscoveryBodyLength = 248;
+		static constexpr size_t kGvcpForceIpBodyLength = 56;
+	}  // namespace detail
+
+	typedef uint8_t GvcpFlag;  // flag bits are specified by each command
```
Dimi1010, line 29: nit: maybe use cpp11 `using` for alias declaration?

```suggestion
	using GvcpFlag = uint8_t;  // flag bits are specified by each command
```

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,632 @@
...
+		/// @brief get the GVCP command
+		GvcpCommand getCommand() const
+		{
+			return getGvcpHeader()->getCommand();
+		}
+
+		// implement Layer's abstract methods
+		std::string toString() const override
+		{
+			return "";
```
Dimi1010, line 533: Should this be empty?

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,632 @@
...
+		 */
+		GvcpCommand getCommand() const
+		{
+			return getGvcpHeader()->getCommand();
+		}
+
+		// implement Layer's abstract methods
+		std::string toString() const override
+		{
+			return "";
```
Dimi1010, line 607: ditto, `GvcpRequestLayer::toString`

Packet++/src/GvcpLayer.cpp
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	/*---------------------- Class GvcpLayer ----------------------------*/
+
+	GvcpLayer::GvcpLayer(uint8_t* data, size_t dataSize, Layer* prevLayer, Packet* packet)
+	    : Layer(data, dataSize, prevLayer, packet)
+	{
+		m_Protocol = Gvcp;
+		m_DataLen = dataSize;
+		m_Data = data;
```
Dimi1010, line 33: Isn't this already set by `Layer::Layer(...)`?

Packet++/src/GvcpLayer.cpp
```diff
@@ -0,0 +1,101 @@
...
+		os << "0x" << std::hex << static_cast<uint16_t>(status) << std::dec;
+		return os;
+	}
+
+	/*---------------------- Class GvcpLayer ----------------------------*/
+
+	GvcpLayer::GvcpLayer(uint8_t* data, size_t dataSize, Layer* prevLayer, Packet* packet)
+	    : Layer(data, dataSize, prevLayer, packet)
+	{
+		m_Protocol = Gvcp;
```
Dimi1010, line 31: Hmm... we should probably add an overload to `Layer::Layer(...)` that accepts a protocol parameter too, instead of needing to be set manually afterwards.

Packet++/header/GvcpLayer.h
tigercosmos: sure

Packet++/header/GvcpLayer.h
tigercosmos: Fixed.

Packet++/header/GvcpLayer.h
tigercosmos: I missed it. Added.

Packet++/header/GvcpLayer.h
tigercosmos: I missed it. Added.

Packet++/src/GvcpLayer.cpp
tigercosmos: yes, this can be removed.

Packet++/src/GvcpLayer.cpp
tigercosmos: True, we should do the refactoring for this issue later.

tigercosmos:
@seladb  The PR is ready.  Please take a look.

Packet++/src/GvcpLayer.cpp
Dimi1010: Opened a separate issue for this.

tigercosmos:
Don't know why the fuzz tests failed?

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,626 @@
...
+		{}
+
+		GvcpFlag getFlag() const
+		{
+			return flag;
+		}
+
+		GvcpCommand getCommand() const
+		{
+			return static_cast<GvcpCommand>(netToHost16(command));
```
seladb, line 135: I think we can move all implementations that use `netToHost16` to `GvcpLayer.cpp` so we don't expose `SystemUtils.h` in the header file

Packet++/src/UdpLayer.cpp
```diff
@@ -134,6 +135,17 @@ namespace pcpp
...
+		{
+			if (GvcpLayer::verifyRequest(udpData))
+			{
+				m_NextLayer = new GvcpRequestLayer(udpData, udpDataLen, this, m_Packet);
+			}
+			else
+			{
+				m_NextLayer = new GvcpAcknowledgeLayer(udpData, udpDataLen, this, m_Packet);
+			}
+		}
```
seladb, line 148: It would be nicer to add a static method to `GvcpLayer` to create a GCVP layer if valid, or `PayloadLayer` otherwise. Then we can do:
```cpp
else if (GvcpLayer::isGvcpPort(portSrc) || GvcpLayer::isGvcpPort(portDst))
{
	m_NextLayer = GvcpLayer::parseGvcpLayer(udpData, udpDataLen, this, m_Packet);
}
```

README.md
```diff
@@ -265,6 +265,8 @@ PcapPlusPlus currently supports parsing, editing and creation of packets of the
 47. SSH - parsing only (no editing capabilities)
 48. Telnet - parsing only (no editing capabilities)
 49. Generic payload
+50. GVCP
```
seladb, line 268: Let's add GVCP before `Generic payload`?

Tests/Packet++Test/PacketExamples/gvcp_discovery_ack.pcap
seladb, line 1: to reduce the number of files, maybe we should consolidate all packets under one pcap file?

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,626 @@
...
+		Error = 0x8FFF,
+		Unknown = 0xFFFF
+	};
+
+	std::ostream& operator<<(std::ostream& os, GvcpResponseStatus status);
+
+#pragma pack(push, 1)
+	/// @brief Gvcp request header
+	/// @note refer to the spec "15.1 Request Header". The data is stored as big-endian.
+	struct GvcpRequestHeader
```
seladb, line 108: I have a general design comment: I think it'd be nicer to add the property getters / setters to the layer class and not to the struct. That's the way we usually design layers. We can do something like this (below is pseudo code):
```cpp
class GvcpAcknowledgeLayer : public GvcpLayer
{
    uint16_t getAckId() const { return getGvcpHeader()->ackId; }
}
```

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,626 @@
...
+		bool hasAllowBroadcastFlag() const
+		{
+			constexpr GvcpFlag kAllowBroadcastFlag = 0b0001000;
+			return (flag & kAllowBroadcastFlag) == kAllowBroadcastFlag;
+		}
+	};
+
+	/// @brief Gvcp acknowledge header
+	/// @note refer to the spec "15.2 Acknowledge Header". The data is stored as big-endian.
+	struct GvcpAckHeader
```
seladb, line 190: I'm still learning the GVCP protocol, but shouldn't this struct inherit from `GvcpRequestHeader`? 
If not, isn't there a common header for all GVCP packets?

seladb (commented):
I just started reviewing this PR because I still need to learn this protocol. But here are a few general comments

Packet++/header/GvcpLayer.h
tigercosmos: The request header is different from the acknowledge header, so I separated them.

Tests/Packet++Test/PacketExamples/gvcp_discovery_ack.pcap
tigercosmos: Then how do we find the correct packet for testing? I intended to separate them. How about putting them into a directory?

Packet++/header/GvcpLayer.h
seladb: Is there a common header for all GVCP packets?

Packet++/header/GvcpLayer.h
tigercosmos: How about having `getAckId` for both the layer and the header so that a user can get the data either way?

Tests/Packet++Test/PacketExamples/gvcp_discovery_ack.pcap
seladb: I'm only referring to the pacp files, we should still separate the `.dat` files

Packet++/header/GvcpLayer.h
seladb: I think that we can "hide" the struct, and only expose the getters / setters. We can make `getGvcpHeader()` private / protected

Packet++/header/GvcpLayer.h
tigercosmos: Unfortunately, no.

The request (or called command):
image

The response (or called acknowledge):
image

README.md
tigercosmos: sure.

seladb (commented):
@tigercosmos here is my suggestion of the structure we can use, which follows ideas implemented in other PcapPlusPlus layers. This code snippet is just the high level interface and doesn't include all the methods, implementation and lower level details, but it's mostly to demonstrate my suggestion.

Please let me know what you think and we can discuss it further.

```cpp
struct gvcp_command_header
{
	uint8_t gcvpCommand = internal::kGvcpMagicNumber;  // always fixed
	uint8_t flag = 0;  // 0-3 bits are specified by each command, 4-6 bits are reserved, 7 bit is acknowledge
	uint16_t command = 0;
	uint16_t dataSize = 0;
	uint16_t requestId = 0;
};

struct gvcp_ack_header
{
	uint16_t status = 0;
	uint16_t command = 0;
	uint16_t dataSize = 0;
	uint16_t ackId = 0;
};

struct gvcp_discovery_message : public gvcp_command_header
{
};

struct gvcp_discovery_ack_message : public gvcp_ack_header
{
	uint16_t versionMajor = 0;
	uint16_t versionMinor = 0;
	uint32_t deviceMode = 0;
	uint16_t reserved = 0;
	uint8_t macAddress[6] = { 0 };
	uint32_t supportedIpConfigOptions = 0;
	uint32_t ipConfigCurrent = 0;
	uint8_t reserved2[12] = { 0 };
	uint32_t ipAddress = 0;
	uint8_t reserved3[12];
	uint32_t subnetMask = 0;
	uint8_t reserved4[12] = { 0 };
	uint32_t defaultGateway = 0;
	char manufacturerName[32] = { 0 };
	char modelName[32] = { 0 };
	char deviceVersion[32] = { 0 };
	char manufacturerSpecificInformation[48] = { 0 };
	char serialNumber[16] = { 0 };
	char userDefinedName[16] = { 0 };
};

class GvcpLayer : public Layer
{
public:
	GvcpLayer() = delete;

	static bool isGvcpPort(uint16_t port);
	static GvcpLayer* parseGvcpLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet);

	// implement Layer's abstract methods
	void parseNextLayer() override;
	void computeCalculateFields() override;
	OsiModelLayer getOsiModelLayer() const override;
};

class GvcpRequestLayer : public GvcpLayer
{
public:
	GvcpRequestLayer() = delete;

	static bool isDataValid(const uint8_t* data, size_t dataLen);

	GvcpFlag getFlag() const;
	void setFlag(const GvcpFlag flag);

	GvcpCommand getCommand() const;

	uint16_t getLength() const;

	uint16_t getRequestId() const;
	void setRequestId(uint16_t requestId);
};

class GvcpAckLayer : public GvcpLayer
{
public:
	GvcpAckLayer() = delete;

	static bool isDataValid(const uint8_t* data, size_t dataLen);

	GvcpResponseStatus getStatus() const;
	void setStatus(GvcpResponseStatus status);

	GvcpCommand getCommand() const;

	uint16_t getLength() const;

	uint16_t getAckId() const;
	void setAckId(uint16_t ackId);
};

class GvcpDiscoveryRequestLayer : public GvcpRequestLayer
{
public:
	// constructors

	GvcpDiscoveryRequestLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet);
	GvcpDiscoveryRequestLayer();

	// implement Layer's abstract methods

	std::string toString() const override;
	size_t getHeaderLen() const override;
};

class GvcpDiscoveryAckLayer : public GvcpAckLayer
{
	// constructors

	GvcpDiscoveryAckLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet);
	GvcpDiscoveryAckLayer(/** a list of all params: uint16_t majorVersion, uint16_t minorVersion, uint32_t deviceMode, ...*/);

	// getters and setters

	std::pair<uint16_t, uint16_t> getVersion() const;

	pcpp::MacAddress getMacAddress() const;
	void setMacAddress(const pcpp::MacAddress macAddress);

	// ...

	// implement Layer's abstract methods

	std::string toString() const override;
	size_t getHeaderLen() const override;
};

class GvcpForceIpRequestLayer : public GvcpRequestLayer
{
	// ...
};

class GvcpForceIpAcktLayer : public GvcpAckLayer
{
	// ...
};
```

tigercosmos:
@seladb  Reading your sample code, basically I got your intention. So "Discovery" and "ForceIP" have their own layer, and for other commands, they will fall back to general `GVCPRequestLayer` and `GvcpAcknowledgeLayer`. I don't see a problem for now, and I will try to refactor the PR based on this and see if I encounter any issues.

seladb:
> @seladb Reading your sample code, basically I got your intention. So "Discovery" and "ForceIP" have their own layer, and for other commands, they will fall back to general `GVCPRequestLayer` and `GvcpAcknowledgeLayer`. I don't see a problem for now, and I will try to refactor the PR based on this and see if I encounter any issues.

Yes, other messages should fall under `GvcpRequestLayer` or `GvcpAckLayer`. In that way we can create specific layers for more messages in the future

tigercosmos:
@seladb  The code has been updated, could you take a look?

tigercosmos:
@seladb  ping.

seladb:
> @seladb ping.

Sorry for the delay, I'll review it soon

Common++/header/Logger.h
```diff
@@ -1,10 +1,10 @@
 #pragma once
 
-#include <stdio.h>
+#include <iomanip>
 #include <iostream>
 #include <sstream>
-#include <iomanip>
 #include <stdint.h>
+#include <stdio.h>
```
seladb, line 7: Are these changes relevant for this PR?

Packet++/header/ProtocolType.h
```diff
@@ -352,6 +352,11 @@ namespace pcpp
 	 */
 	const ProtocolType WireGuard = 56;
 
+	/*
+	 * GVCP protocol
+	 */
+	const ProtocolType Gvcp = 57;
```
seladb, line 358: Maybe change to capital letters `GVCP`?

Tests/Packet++Test/PacketExamples/gvcp_discovery_ack.pcap
seladb, line 1: nit: maybe use Wireshark's `mergecap` tool to merge all pcap files into 1 pcap file?

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,776 @@
...
+
+#include "IpAddress.h"
+#include "Layer.h"
+#include "MacAddress.h"
+#include "SystemUtils.h"
+#include <iostream>
+
+/**
+ * @file GvcpLayer.h
+ * @author An-Chi Liu (phy.tiger@gmail.com)
```
seladb, line 11: nit: I'm not strictly against it, but we usually don't put the author inside the code.

ditto in the other new files in this PR

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,776 @@
...
+		static constexpr size_t kGvcpRequestHeaderLength = 8;
+		static constexpr size_t kGvcpAckHeaderLength = 8;
+		static constexpr size_t kGvcpDiscoveryBodyLength = 248;
+		static constexpr size_t kGvcpForceIpBodyLength = 56;
+	}  // namespace internal
+
+	using GvcpFlag = uint8_t;  // flag bits are specified by each command
+
+	/// @brief Gvcp command
+	/// See spec "18 Command and Acknowledge Values"
```
seladb, line 32: Does this assume the user has the spec?

ditto in the other enums, structs and classes

Packet++/header/GvcpLayer.h
seladb: @tigercosmos this was not changed. Please see other protocol implementations like the recent `WireGuardLayer` or `BgpLayer`

seladb:
@tigercosmos I didn't review the entire PR because [this comment](https://github.com/seladb/PcapPlusPlus/pull/1357#discussion_r1783983577) wasn't addressed yet. Can you please address this and I'll review the rest of the PR?

tigercosmos:
> @tigercosmos I didn't review the entire PR because [this comment](https://github.com/seladb/PcapPlusPlus/pull/1357#discussion_r1783983577) wasn't addressed yet. Can you please address this and I'll review the rest of the PR?

@seladb  What do you mean not changed? I have changed all layers. Or do you mean putting the header definition inside the layer? I don't do that because the layer will be too big.

Packet++/header/GvcpLayer.h
tigercosmos: Yes, the user can get a copy of the spec by themselves.

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,776 @@
...
+		 */
+		GvcpRequestHeader* getGvcpHeader() const
+		{
+			return reinterpret_cast<GvcpRequestHeader*>(m_Data);  // the header is at the beginning of the data
+		}
+
+		/// @brief get the GVCP command
+		GvcpCommand getCommand() const
+		{
+			return getGvcpHeader()->getCommand();
```
tigercosmos, line 384: @seladb  I think you didn't read the code here.

Packet++/header/GvcpLayer.h
tigercosmos: @seladb  please read the implementation in the layer class. I intend to keep the function for the header. The layer will use the functions from the header. In this way, user can get the field in either the header or the layer.

seladb: I think it makes it harder to use and differs from other implementations we have in our code base.

When a user has a layer instance, for example `GvcpRequestHeader`, they expect to have direct access to the getters and setters:
```cpp
auto gvcpLayer = ...;
gvcpLayer->getFlag();
```

Instead what they need to do with the current implementation is:
```cpp
auto gvcpLayer = ...;
gvcpLayer->getGvcpHeader->getFlag();
```

I think the first implementation is more clear, and aligns better to other protocols

Packet++/header/GvcpLayer.h
seladb: Maybe instead of (or in addition to) adding a reference to the spec, we can add some documentation?

Packet++/header/GvcpLayer.h
tigercosmos: @seladb  I think you misunderstand my statement. I meant there are two ways to get the flag value in your example.

```cpp
auto gvcpLayer = .... ;
gvcpLayer->getFlag() == gvcpLayer->getGvcpHeader->getFlag();
```

seladb: I can't see the first option `gvcpLayer->getFlag()` in the code base, am I missing something? 🤔 

Also - why do we need to have both? Isn't it enough to keep the getters and setters in the layer?

Packet++/header/GvcpLayer.h
seladb: I agree we don't want to explain the entire protocol in the doxygen documentation, but maybe give just one sentence for each enum / enum value / struct / class?

I don't think we have only a reference to the spec in any other protocol 🤔

Packet++/header/GvcpLayer.h
tigercosmos: An example is here: https://github.com/seladb/PcapPlusPlus/pull/1357#discussion_r1784026612

I thought it's more convenient if the header and the layer both are able to get the field, but I guess it's fine to keep the layer's one.

Packet++/header/GvcpLayer.h
tigercosmos: Sure, for some classes maybe they are worth having a sentence comment. But for most of the enum values, I don't see a good reason to add "a sentence" to explain them as the spec doesn't explain that much as well.

seladb: We document enum values as well... but if you think it's too much then I'm ok with not documenting enum values. Worst case we can add it later if needed. So let's at least add one sentence for structs/enums/classes?

Packet++/header/GvcpLayer.h
tigercosmos: @seladb  I found that there is a reason that at least `GvcpRequestHeader`  and `GvcpAckHeader` need to have `getCommand()` method. When we parse the layer, I first check the header and decide which layer to create:

```cpp
GvcpLayer* GvcpLayer::parseGvcpLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
{
	if (GvcpLayer::verifyRequest(data))
	{	
		auto* header = reinterpret_cast<GvcpRequestHeader*>(data);
		if (header->getCommand() == GvcpCommand::DiscoveredCmd)
		{
			return new GvcpDiscoveryRequestLayer(data, dataLen, prevLayer, packet);
		}
		else if (header->getCommand() == GvcpCommand::ForceIpCmd)
		{
			return new GvcpForceIpRequestLayer(data, dataLen, prevLayer, packet);
		}
		else
		{
			return new GvcpRequestLayer(data, dataLen, prevLayer, packet);
		}
	}
	else
	{
		auto* header = reinterpret_cast<GvcpAckHeader*>(data);
		if (header->getCommand() == GvcpCommand::DiscoveredAck)
		{
			return new GvcpDiscoveryAcknowledgeLayer(data, dataLen);
		}
		else if (header->getCommand() == GvcpCommand::ForceIpAck)
		{
			return new GvcpForceIpAcknowledgeLayer(data, dataLen);
		}
		else
		{
			return new GvcpAcknowledgeLayer(data, dataLen, prevLayer, packet);
		}
	}
}
```

Of course we can do something like the following code, but it's inefficient:
```cpp
auto layer = new GvcpRequestLayer(data, dataLen, prevLayer, packet);
if (layer->getCommand() == GvcpCommand::DiscoveredCmd)
{
	delete layer;
	return  new GvcpDiscoveryRequestLayer(data, dataLen, prevLayer, packet);
}
```

What do you think?

Packet++/header/GvcpLayer.h
tigercosmos: Since it's one line function, I think it's clearer to be in the header. I forward declare the function, so we don't need to include the whole header file.

Packet++/src/UdpLayer.cpp
tigercosmos: Please check the implementation in `parseGvcpLayer`.

Common++/header/Logger.h
tigercosmos: No, reverted.

Packet++/header/GvcpLayer.h
tigercosmos: Okay.

Tests/Packet++Test/PacketExamples/gvcp_discovery_ack.pcap
tigercosmos: Done.

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+		uint16_t requestId = 0;
+
+	public:
+		// ------------- methods --------------
+		GvcpRequestHeader() = default;
+
+		GvcpRequestHeader(GvcpFlag flag, GvcpCommand command, uint16_t dataSize, uint16_t requestId)
+		    : flag(flag), command(hostToNet16(static_cast<uint16_t>(command))), dataSize(hostToNet16(dataSize)),
+		      requestId(hostToNet16(requestId))
+		{}
```
seladb, line 123: Why do we need c'tors for a struct? 🤔 

What we usually do is declare all properties as public and then we don't need c'tors.
These structs are internal to the GVCP layers classes and shouldn't be exposed outside, so it doesn't matter if their properties are public.

If you want you can move these structs as `protected` inside of the GVCP layer classes

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+		Error = 0x8FFF,
+		Unknown = 0xFFFF
+	};
+
+	std::ostream& operator<<(std::ostream& os, GvcpResponseStatus status);
+
+#pragma pack(push, 1)
+	/// @brief Gvcp request header
+	/// @note refer to the spec "15.1 Request Header". The data is stored as big-endian.
+	struct GvcpRequestHeader
```
seladb, line 105: nit: we usually define structs with snake_case, for example: `gvcp_request_header`. Same for the other structs

Packet++/src/GvcpLayer.cpp
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	/*---------------------- Class GvcpLayer ----------------------------*/
+
+	GvcpLayer::GvcpLayer(uint8_t* data, size_t dataSize, Layer* prevLayer, Packet* packet)
+	    : Layer(data, dataSize, prevLayer, packet, GVCP)
+	{}
+
+	GvcpLayer* GvcpLayer::parseGvcpLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
+	{
```
seladb, line 28: I think this method should also check if the data is not null and in the minimum size, for example `sizeof(GvcpRequestHeader)` if it's a request or `sizeof(GvcpAckHeader)` if it's a response, otherwise it should return a `PayloadLayer`

Packet++/src/GvcpLayer.cpp
```diff
@@ -0,0 +1,143 @@
...
+				return new GvcpDiscoveryRequestLayer(data, dataLen, prevLayer, packet);
+			}
+			else if (header->getCommand() == GvcpCommand::ForceIpCmd)
+			{
+				return new GvcpForceIpRequestLayer(data, dataLen, prevLayer, packet);
+			}
+			else
+			{
+				return new GvcpRequestLayer(data, dataLen, prevLayer, packet);
+			}
```
seladb, line 43: We can use `switch` here

Packet++/src/GvcpLayer.cpp
```diff
@@ -0,0 +1,143 @@
...
+				return new GvcpDiscoveryAcknowledgeLayer(data, dataLen);
+			}
+			else if (header->getCommand() == GvcpCommand::ForceIpAck)
+			{
+				return new GvcpForceIpAcknowledgeLayer(data, dataLen);
+			}
+			else
+			{
+				return new GvcpAcknowledgeLayer(data, dataLen, prevLayer, packet);
+			}
```
seladb, line 59: ditto: We can use `switch` here

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+		      requestId(hostToNet16(requestId))
+		{}
+
+		GvcpCommand getCommand() const
+		{
+			return static_cast<GvcpCommand>(netToHost16(command));
+		}
+	};
+	static_assert(sizeof(GvcpRequestHeader) == internal::kGvcpRequestHeaderLength,
+	              "Gvcp request header size should be 8 bytes");
```
seladb, line 131: Why do we need this `static_assert`? 🤔 

Same for the other `static_assert` in this file...

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+
+		/**
+		 * @brief Check if the broadcast is allowed
+		 * @return true The broadcast acknowledge is allowed
+		 */
+		bool hasAllowBroadcastFlag() const
+		{
+			constexpr GvcpFlag kAllowBroadcastFlag = 0b0001000;
+			return (flag & kAllowBroadcastFlag) == kAllowBroadcastFlag;
+		}
```
seladb, line 148: Why do we need this method in the struct?

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+	};
+
+	std::ostream& operator<<(std::ostream& os, GvcpResponseStatus status);
+
+#pragma pack(push, 1)
+	/// @brief Gvcp request header
+	/// @note refer to the spec "15.1 Request Header". The data is stored as big-endian.
+	struct GvcpRequestHeader
+	{
+		friend class GvcpRequestLayer;
```
seladb, line 107: Once we make the properties public we don't need this `friend class..`

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+
+		// implement Layer's abstract methods
+		void computeCalculateFields() override
+		{}
+
+		// implement Layer's abstract methods
+		OsiModelLayer getOsiModelLayer() const override
+		{
+			return OsiModelLayer::OsiModelApplicationLayer;
+		}
```
seladb, line 285: Why are these methods `protected`? They should be `public`

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+		 * @param[in] packet A pointer to the Packet instance where layer will be stored in
+		 */
+		GvcpRequestLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet);
+
+		/**
+		 * @brief Construct a new GvcpRequestLayer object
+		 * @param[in] data A pointer to the data including the header and the payload
+		 * @param[in] dataSize The size of the data in bytes
+		 */
+		GvcpRequestLayer(const uint8_t* data, size_t dataSize);
```
seladb, line 306: Why do we need this c'tor if we have the other c'tor below?

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+		}
+
+		// implement Layer's abstract methods
+		std::string toString() const override;
+
+		// implement Layer's abstract methods
+		size_t getHeaderLen() const override
+		{
+			return sizeof(GvcpRequestHeader);
+		}
```
seladb, line 387: The header length should include the entire layer length - in this case since GVCP layer is always last, it should return `m_DataLen` for all type of GVCP messages

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+		                          uint16_t payloadDataSize = 0, GvcpFlag flag = 0, uint16_t requestId = 1);
+
+		/**
+		 * @brief Get the header object
+		 * @return GvcpRequestHeader* A pointer to the header object
+		 */
+		GvcpRequestHeader* getGvcpHeader() const
+		{
+			return reinterpret_cast<GvcpRequestHeader*>(m_Data);  // the header is at the beginning of the data
+		}
```
seladb, line 327: this method can probably be private/protected

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+			return getGvcpHeader()->flag;
+		}
+
+		/**
+		 * @brief Get the data size from the header
+		 */
+		uint16_t getDataSize() const
+		{
+			return netToHost16(getGvcpHeader()->dataSize);
+		}
```
seladb, line 343: ditto: let's move all implementations that use `netToHost...` to the cpp file

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+		GvcpRequestHeader() = default;
+
+		GvcpRequestHeader(GvcpFlag flag, GvcpCommand command, uint16_t dataSize, uint16_t requestId)
+		    : flag(flag), command(hostToNet16(static_cast<uint16_t>(command))), dataSize(hostToNet16(dataSize)),
+		      requestId(hostToNet16(requestId))
+		{}
+
+		GvcpCommand getCommand() const
+		{
+			return static_cast<GvcpCommand>(netToHost16(command));
```
seladb, line 127: This method should also handle unexpected values and return `GvcpCommand::Unknown`

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+		{
+			return netToHost16(getGvcpHeader()->requestId);
+		}
+
+		/**
+		 * @brief Get the command from the header
+		 */
+		GvcpCommand getCommand() const
+		{
+			return static_cast<GvcpCommand>(netToHost16(getGvcpHeader()->command));
```
seladb, line 358: We can do:
```cpp
return getGvcpHeader()->getCommand();
```

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+		 * @param[in] packet A pointer to the Packet instance where layer will be stored in
+		 */
+		GvcpAcknowledgeLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet);
+
+		/**
+		 * @brief Construct a new GvcpRequestLayer object
+		 * @param[in] data A pointer to the data including the header and the payload
+		 * @param[in] dataSize The size of the data in bytes
+		 */
+		GvcpAcknowledgeLayer(const uint8_t* data, size_t dataSize);
```
seladb, line 408: ditto: why do we need this c'tor if we have the c'tor below?

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+		                     uint16_t payloadDataSize = 0, uint16_t ackId = 0);
+
+		/**
+		 * @brief Get the header object
+		 * @return GvcpAckHeader* A pointer to the header object
+		 */
+		GvcpAckHeader* getGvcpHeader() const
+		{
+			return reinterpret_cast<GvcpAckHeader*>(m_Data);  // the header is at the beginning of the data
+		}
```
seladb, line 429: ditto: this method can be private/protected

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+
+		/**
+		 * @return the response status from the header
+		 */
+		GvcpResponseStatus getStatus() const
+		{
+			return static_cast<GvcpResponseStatus>((netToHost16(getGvcpHeader()->status)));
```
seladb, line 436: This method should handle cases where `getGvcpHeader()->status` is an unexpected value and return `GvcpResponseStatus::Unknown`

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
...
+		}
+
+		/**
+		 * @return the response command type from the header
+		 */
+		GvcpCommand getCommand() const
+		{
+			return static_cast<GvcpCommand>(netToHost16(getGvcpHeader()->command));
```
seladb, line 444: ditto: we can do:
```
return getGvcpHeader()->getCommand();
```

Packet++/header/GvcpLayer.h
seladb: @tigercosmos please address this comment also?

seladb:
@tigercosmos a gentle reminder to address the PR comments ☝️

tigercosmos:
> @tigercosmos a gentle reminder to address the PR comments ☝️

Thanks. I will continue on this after October. Quite busy these days.

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,772 @@
+#pragma once
+
+#include "IpAddress.h"
+#include "Layer.h"
+#include "MacAddress.h"
+#include <iostream>
+
```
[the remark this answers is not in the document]
tigercosmos: Added.

Packet++/header/GvcpLayer.h
tigercosmos: Sure.

Packet++/header/GvcpLayer.h
tigercosmos: I added a constructor for it because the byte order will change, and it's better to handle it by the constructor. In addition, `GvcpRequestHeader` is public in purpose here. Users are free to get the raw header if they want, so there is also a public function `getGvcpHeader()` in the layers. I don't think we should stop users from getting the header.

Packet++/header/GvcpLayer.h
tigercosmos: Fixed.

Packet++/header/GvcpLayer.h
tigercosmos: `static_assert` ensures the correct definition in the struct, as we know the header's actual byte size.

Packet++/header/GvcpLayer.h
tigercosmos: That's true. Removed.

Packet++/header/GvcpLayer.h
tigercosmos: Sure, fixed.

Packet++/header/GvcpLayer.h
tigercosmos: What if the user wants to build the layer from the bytes of the whole layer? The following constructor is for building the layer with the header information and the payload data.

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,758 @@
...
+		static constexpr size_t kGvcpMagicNumber = 0x42;
+		static constexpr size_t kGvcpRequestHeaderLength = 8;
+		static constexpr size_t kGvcpAckHeaderLength = 8;
+		static constexpr size_t kGvcpDiscoveryBodyLength = 248;
+		static constexpr size_t kGvcpForceIpBodyLength = 56;
+	}  // namespace internal
+
+	using GvcpFlag = uint8_t;  // flag bits are specified by each command
+
+	/// @brief Gvcp command defines the command values and the corresponding acknowledge values
```
seladb, line 32: 1. nit: let's use GVCP (capital letters) instead of Gvcp in the doxygen documentation (not in class names etc).
Please search and replace other instances of Gvcp in this file and replace them too
2. In order to keep the convention we have everywhere else, maybe use this format for doxygen comments?
    ```cpp
    /**
     * comment...
     */
    ```

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,758 @@
...
+
+		// ------------- methods --------------
+		gvcp_request_header() = default;
+
+		gvcp_request_header(GvcpFlag flag, GvcpCommand command, uint16_t dataSize, uint16_t requestId)
+		    : flag(flag), command(hostToNet16(static_cast<uint16_t>(command))), dataSize(hostToNet16(dataSize)),
+		      requestId(hostToNet16(requestId))
+		{}
+
+		GvcpCommand getCommand() const
```
seladb, line 133: If we decide to keep this struct and the other structs public, we should add doxygen documentation for each public member and public method

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,758 @@
...
+		ActionAck = 0x0101,
+		Unknown = 0xFFFF
+	};
+
+	/// output operator for GvcpCommand
+	std::ostream& operator<<(std::ostream& os, GvcpCommand command);
+
+	/// @brief GVCP response status can be returned in an acknowledge message or a GVSP header.
+	/// See more in the spec "Table 19-1: List of Standard Status Codes"
+	enum class GvcpResponseStatus : uint16_t
```
seladb, line 74: This enum has doxygen documentation for its values but `GvcpCommand` doesn't. I think we should be consistent - either we add doxygen to all values in all enums, or we don't...

Packet++/header/GvcpLayer.h
```diff
@@ -0,0 +1,758 @@
...
+		WriteProtect = 0x8004,    ///< The addressed register cannot be written to
+		BadAlignment = 0x8005,    ///< A badly aligned address offset or data size was specified.
+		AccessDenied =
+		    0x8006,  ///< An attempt was made to access an address location which is currently/momentary not accessible
+		Busy = 0x8007,  ///< A required resource to service the request is not currently available. The request may be
+		                ///< retried at a later time
+		LocalProblem = 0x8008,       // deprecated
+		MsgMismatch = 0x8009,        // deprecated
+		InvalidProtocol = 0x800A,    // deprecated
+		NoMsg = 0x800B,              // deprecated
```
seladb, line 91: These `// deprecated` comments are not in the doxygen format so these values won't appear in the API documentation. We should change them to `///< deprecated`.

Same for other `// deprecated` comments below

Packet++/header/GvcpLayer.h
seladb: I think I mentioned it in some comments, but I'm not sure.... In my opinion there is no need to expose both `getGvcpHeader()` and getters/setters for each property, that will create a somewhat confusing API.

In older protocols we used to only expose the "raw header" (like `getGvcpHeader()`), but in newer protocols we usually expose getters and setters which I think provides a cleaner API. We also got feedback from users that they prefer the getters/setters over the "raw header"

Packet++/header/GvcpLayer.h
seladb: I'm not sure I understand... we define the struct, why would it change? 🤔

Packet++/header/GvcpLayer.h
seladb: That's usually not required. If users have raw data they create a `pcpp::Packet` out of it and use this c'tor: `GvcpRequestLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)`

Packet++/header/GvcpLayer.h
seladb: Please see this comment: https://github.com/seladb/PcapPlusPlus/pull/1357#discussion_r1821925821

Packet++/header/GvcpLayer.h
tigercosmos: Some enums in the spec have explanations but some don't. I will add the comments for the enum values as long as the spec has the sentences for the values.

Packet++/header/GvcpLayer.h
tigercosmos: Sure, let me not expose it.

Packet++/header/GvcpLayer.h
tigercosmos: sure fixed.

Packet++/header/GvcpLayer.h
tigercosmos: the struct is now under `internal`.

Packet++/header/GvcpLayer.h
tigercosmos: fixed.

tigercosmos:
Still WIP, not ready yet.

seladb:
> Still WIP, not ready yet.

@tigercosmos maybe you can move the PR to draft until it's ready?

tigercosmos:
Ops, I didn't notice this. Changed.

Packet++/src/GvcpLayer.cpp
tigercosmos: changed.

Packet++/header/GvcpLayer.h
tigercosmos: People may change the struct, but we know the exact packet size. This is a double check just in case.

Packet++/header/GvcpLayer.h
tigercosmos: same

Packet++/header/GvcpLayer.h
tigercosmos: I don't get it. The header length should not cover the payload. GVCP layer includes GVCP header and GVCP payload. In this case, why is `getHeaderLen()`  not the GVCP header length?

Packet++/header/GvcpLayer.h
tigercosmos: fixed.

Packet++/header/GvcpLayer.h
tigercosmos: fixed.

Packet++/header/GvcpLayer.h
tigercosmos: fixed.

seladb:
@tigercosmos this PR is open for a long time, do you plan to continue working on it or should we close it?

tigercosmos:
Let's close it.

tigercosmos (commented):
@seladb  @Dimi1010  I have reworked this PR and addressed all issues above. Please help review, thanks!

Dimi1010 (approved):
LGTM code wise. Tho, I haven't gone in depth into the actual spec.

PR #1362, opened 2024-04-16
close the socket

### LINKED ISSUE #1361: PcapLiveDevice socket is not close
in file Pcap++/src/PcapLiveDevice.cpp, line 853 and 909 create socket, but never close

tigercosmos:
fix #1361

### DIFF
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -850,16 +850,28 @@ void PcapLiveDevice::setDeviceMtu()
 	memset(&ifr, 0, sizeof(ifr));
 	strncpy(ifr.ifr_name, m_Name.c_str(), sizeof(ifr.ifr_name) - 1);
 
-	int socketfd = socket(AF_INET, SOCK_DGRAM, IPPROTO_IP);
-	if (ioctl(socketfd, SIOCGIFMTU, &ifr) == -1)
+	int socketfd = -1;
+	try
 	{
-		PCPP_LOG_DEBUG("Error in retrieving MTU: ioctl() returned -1");
+		socketfd = socket(AF_INET, SOCK_DGRAM, IPPROTO_IP);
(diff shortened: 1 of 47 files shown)


tigercosmos:
@seladb  this is ready for review.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -859,6 +859,7 @@ void PcapLiveDevice::setDeviceMtu()
 	}
 
 	m_DeviceMtu = ifr.ifr_mtu;
+	::close(socketfd);
```
seladb, line 862: Maybe we should refactor this code to make sure the socket always closes?
I mean something like this:
```cpp
int socketfd = -1;
try
{
	int socketfd = socket(AF_INET, SOCK_DGRAM, IPPROTO_IP);
	...
}
catch (const std::exception& e)
{
 	//log exception using e.what()
}
if (socketfd != -1)
{
	::close(socketfd);
}
```

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -914,6 +915,7 @@ void PcapLiveDevice::setDeviceMacAddress()
 	}
 
 	m_MacAddress = MacAddress(ifr.ifr_hwaddr.sa_data[0], ifr.ifr_hwaddr.sa_data[1], ifr.ifr_hwaddr.sa_data[2], ifr.ifr_hwaddr.sa_data[3], ifr.ifr_hwaddr.sa_data[4], ifr.ifr_hwaddr.sa_data[5]);
+	::close(socketfd);
```
seladb, line 918: ditto

Pcap++/src/PcapLiveDevice.cpp
tigercosmos: fixed.

seladb (approved):

PR #1368, opened 2024-04-18
Improve nanosecond precision support for pcap

### LINKED ISSUE #906: How to access file header info?

egecetin:
~Instead of compile time constant definitions~ use `pcap_get_tstamp_precision()` to detect timestamp precision which introduced with libpcap 1.5.1

This should also fix #906 

- [x] Update pcap open functions to control precision
- [x] Add arguments to specify desired precision
- [x] Keep backward compatibility since original Winpcap does not support nanosecond precision
- [x] Manually (or with CI) test different precisions for pcap/pcapng files
    - Read Microseconds -> Write Microseconds 
    - Read Microseconds -> Write Nanoseconds
    - Read Nanoseconds -> Write Microseconds 
    - Read Nanoseconds -> Write Nanoseconds


.gitignore
```diff
@@ -61,6 +61,7 @@ Tests/Pcap++Test/cUrl/curl_output.txt
 *.VC.db
 *.vcxproj.user
 **/.vs
+CMakeSettings.json
```
egecetin, line 64: This is Visual Studio generated file to hold configuration when remote compilation (ssh/wsl etc) requested

egecetin:
Manually tested, it looks like it works for `pcap` files. Can you also test before merging for different precisions?

seladb:
@egecetin why can't we write tests for it? 🤔

egecetin:
@seladb If i can figure out a test flow yes, it is possible 😀. One way I'm thinking.

- Craft two `RawPacket`s with one microsecond and one nanosecond precision
- Write both of them to two `pcap` files with one opened with microsecond and the other nanosecond
- Read these files
- Compare the timestamp headers of read packets with crafted ones. Microsecond precision should truncate, nanosecond should persist.

If I'm right only missed part with this flow, if somehow, I write mathematically inverse operations in write and read only the pcap file itself will be wrong but all other cases should be covered right? 🤔

seladb:
> @seladb If i can figure out a test flow yes, it is possible 😀. One way I'm thinking.
> 
>     * Craft two `RawPacket`s with one microsecond and one nanosecond precision
> 
>     * Write both of them to two `pcap` files with one opened with microsecond and the other nanosecond
> 
>     * Read these files
> 
>     * Compare the timestamp headers of read packets with crafted ones. Microsecond precision should truncate, nanosecond should persist.
> 
> 
> If I'm right only missed part with this flow, if somehow, I write mathematically inverse operations in write and read only the pcap file itself will be wrong but all other cases should be covered right? 🤔

Yes, I think it's a valid test that mostly covers what is mentioned in the ticket: https://github.com/seladb/PcapPlusPlus/issues/906#issuecomment-1176744903
You can even extend the test a bit more:
- Craft two `RawPackets` with one microsecond and one nanosecond precision
- Write both of them to two pcap files with one opened with microsecond and the other nanosecond
- Read both files in 2 ways:
  1. Read microsecond file with microsecond precision
  2. Read microsecond file with nanosecond precision
  3. Read nanosecond file with microsecond precision
  4. Read nanosecond file with nanosecond precision
- Make sure you get the expected result in each case:
  1. crafted packet timestamp == read packet timestamp
  2. crafted packet timestamp * 1000 == read packet timestamp
  3. crafted packet timestamp / 1000 == read packet timestamp
  4. crafted packet timestamp == read packet timestamp

egecetin:
Actually, for the reading side there is no option implemented. It always uses as much high as precision always. So only WinPcap will use microsecond, and all other systems/tests will use nanosecond for reading mode. So, I added a function `isTimestampPrecisionNanoSupported` to check compiled binary supports the nano precision.

egecetin:
@seladb It is ready to review again

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -55,6 +55,18 @@ std::string IFileDevice::getFileName() const
...
+{
+#if defined(PCAP_TSTAMP_PRECISION_NANO)
+	return true;
+#else
+	PCPP_LOG_DEBUG(
+		"PcapPlusPlus was compiled without nano precision support which requires libpcap > 1.5.1. Please "
+		"recompile PcapPlusPlus with nano precision support to use this feature. Using default microsecond precision");
+	return false;
+#endif
+}
```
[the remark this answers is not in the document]
egecetin: If there are no compilation errors, it can be great. I'll check it but not sure because if i'm not mistaken some functions introduced after 1.5.1. So, we will probably see some "undefined ..." errors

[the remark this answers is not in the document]
egecetin: https://github.com/seladb/PcapPlusPlus/blob/6d17ca81fbc80dfd4fa3529b18ed81a753f192f3/Pcap%2B%2B/src/PcapFileDevice.cpp#L255 

This function is introduced with nanosecond precision support in `libpcap`. So, unfortunately all checks must be preprocessor directives

egecetin:
@seladb Probably I addressed all your comments. Can you review it again when you have time?

Pcap++/header/PcapFileDevice.h
```diff
@@ -16,6 +16,20 @@ typedef struct pcap_dumper pcap_dumper_t;
...
+	 */
+    enum class FileTimestampPrecision : int8_t
+    {
+		/// Precision is unknown or not set/determined
+        Unknown = -1,
+		/// Precision is in microseconds
+        Micro,
+		/// Precision is in nanoseconds
+        Nano
+    };
```
seladb, line 32: There is a mix of tabs and spaces, can you move everything to tabs?

Pcap++/header/PcapFileDevice.h
```diff
@@ -127,6 +141,16 @@ namespace pcpp
 		*/
 		LinkLayerType getLinkLayerType() const { return m_PcapLinkLayerType; }
 
+		/**
+		 * @return The precision of the timestamps in the file.
+		*/
+		FileTimestampPrecision getTimestampPrecision() const { return m_Precision; }
```
seladb, line 147: For better test coverage, maybe we can add tests for this method?

It'll actually be interesting to test it before opening the file and after opening it

seladb: You only added tests for `PcapFileWriterDevice::getTimestampPrecision()` but not for `PcapFileReaderDevice::getTimestampPrecision()`...

seladb (approved):
Please see this last comment: https://github.com/seladb/PcapPlusPlus/pull/1368#discussion_r1612839104

Can you fix it before merging?

Pcap++/header/PcapFileDevice.h
egecetin: @seladb `libpcap` scales precision because of `pcap_open_offline_with_tstamp_precision` behaviour. So, I added a test and updated comments of reader function. Can you take a look again before merging for anything i missed?

https://www.tcpdump.org/manpages/pcap_open_offline.3pcap.html

> pcap_open_offline_with_tstamp_precision() takes an additional precision argument specifying the time stamp precision desired; if PCAP_TSTAMP_PRECISION_MICRO is specified, packet time stamps will be supplied in seconds and microseconds, and if PCAP_TSTAMP_PRECISION_NANO is specified, packet time stamps will be supplied in seconds and nanoseconds. If the time stamps in the file do not have the same precision as the requested precision, they will be scaled up or down as necessary before being supplied.

seladb (approved):

PR #1375, opened 2024-04-26
Replaced IPNetwork's raw pointer usage with smart pointers.

Dimi1010:
Main points of this PR are the replacement of raw pointers, `new` and `delete` calls in IPNetwork with usages of `std::unique_ptr<T>` and a backported implementation of `std::make_unique`.


Dimi1010:
Also, just asking, is it desirable for internal `detail` namespaces to be documented in the public API?

tigercosmos:
> Also, just asking, is it desirable for internal `detail` namespaces to be documented in the public API documentation?

yes, we should document everything.

Common++/header/IpAddress.h
```diff
@@ -678,15 +680,15 @@ namespace pcpp
 		 */
 		IPNetwork(const IPAddress& address, uint8_t prefixLen)
 		{
+			using namespace pcpp::detail;
+
 			if (address.isIPv4())
 			{
-				m_IPv4Network = new IPv4Network(address.getIPv4(), prefixLen);
-				m_IPv6Network = nullptr;
+				m_IPv4Network = stdbackports::make_unique<IPv4Network>(address.getIPv4(), prefixLen);
```
tigercosmos, line 687: I don't like implementation, make it too complicated. Can we just declare the `std::unique_ptr` without `make_shared`?

Dimi1010: It can be done, but then you still have the `new` sprinkled here and there every time you construct a `unique_ptr`, essentially expanding every call to `unique_ptr<T>(new T(args...))`

Decided to backport `make_unique` as that is generally the accepted standard for unique pointer creation post-Cpp14. There are some issues with using the `unique_ptr` with `new`-ed pointers where unspecified execution order can cause a memory leak, if you are constructing temporaries.

Finally it makes a potential transition to Cpp14 easier, as you would just have to cut the `stdbackports` namespace at the call site to `std`.

tigercosmos: @Dimi1010  `std::make_shared` is supported by c++11 for non-array types. In your PR, there is no `make_shared` for array type, so actually we can just use `std::make_shared` here.

Dimi1010: @tigercosmos Eh, `std::make_shared` has the overhead of allocating a bigger block to keep a control block. Also is `shared_ptr` is heavier than `unique_ptr` as it includes reference counting instead of being a thin wrapper over a `T*`.

Personally, if I had to choose between going `make_shared` and `unique_ptr<T>(new T())`, I would choose the latter, unless I needed to share the pointer.

tigercosmos: Personally, I prefer utilizing the std function over the backported alternative. I'll leave a comment to provide other team members with options.

seladb: I tend to agree with @tigercosmos , the `stdbackports` implementation is quite complex and probably not needed in a lot of places. I'd prefer to use `unique_ptr<T>(new T())` which is easier to understand. Once we move to C++14 we can refactor this and use `make_unique`

Dimi1010: Reverted to simple `std::unique_ptr<T>(new T())` calls in 35625ee and d74f9f1

tigercosmos (approved):
LGTM, thanks

seladb (approved):

Issue #657, opened 2021-05-26
Packet::setRawPacket parses layers it should not when parseUntilLayer is set


weyrick:
Hello, I found that `Packet::setRawPacket` is parsing too many layers when `parseUntilLayer` is set. If `parseUntilLayer` is set to `OsiModelTransportLayer` and the packet contains DNS traffic, it will still parse into the DNS layer even though it should stop at the UDP layer. 

This is the offending loop: https://github.com/seladb/PcapPlusPlus/blob/34987748e000b33f69a72770f70aa26e257ac0a4/Packet%2B%2B/src/Packet.cpp#L60

There seems to be a hack in the code to clean up the extra layer if it went too far:
https://github.com/seladb/PcapPlusPlus/blob/34987748e000b33f69a72770f70aa26e257ac0a4/Packet%2B%2B/src/Packet.cpp#L76

But this is extra work, and the result is that the DNS payload is parsed twice (because I do my own DNS parsing later).

seladb:
@weyrick thanks for reaching out!
The issue is that there might be several layers of the same OSI layer. The most trivial example I can think of is multiple VLAN layers. If we change `<=` to `<` only the first VLAN layer will be parsed. 
The only way to know what the next layer would be is to parse it. Maybe we can introduce "partial parsing" or "detect layer" functionality that does the minimum work needed to detect what is the next layer 🤔

weyrick:
Hi @seladb, ok interesting. Right so how about add a pure virtual `Layer::identifyNextLayer` and give each layer the ability to peek into the payload to see what is next without parsing it. This is essentially what `parseNextLayer` does already, except that it creates the layer immediately when it finds it, for example:

```
	if (HttpMessage::isHttpPort(portDst) && HttpRequestFirstLine::parseMethod((char*)payload, payloadLen) != HttpRequestLayer::HttpMethodUnknown)
		m_NextLayer = new HttpRequestLayer(payload, payloadLen, this, m_Packet);
```

So `identifyNextLayer` could do:
```
	if (HttpMessage::isHttpPort(portDst) && HttpRequestFirstLine::parseMethod((char*)payload, payloadLen) != HttpRequestLayer::HttpMethodUnknown)
        return std::make_pair(pcpp::HTTP, pcpp::OsiModelApplicationLayer);
```

or similar. Then this information could be used by `setRawPacket` and `parseNextLayer`. Make sense?

seladb:
[...]

PR #1376, opened 2024-04-26
Refactor parseNextLayer to include parseUntilLayer and parseUntil

rtmeng:
This pull request implements the fix to #657 by refactoring the function `parseNextLayer()` to include parameters `parseUntil` and `parseUntilLayer`, which fixes the issue of `Packet::setRawPacket` parsing layers it should not when `parseUntilLayer` is set. Like mentioned in the Issue, this solution doesn't break the API and relies on the layers to not parse too far if it's not requested. 

Main Changes:
- Changed each layer's `parseNextLayer()` to check if the next layer should be parsed
- Fixed inconsistencies of using the keyword override


Packet++/src/BgpLayer.cpp
```diff
@@ -76,12 +76,15 @@ std::string BgpLayer::getMessageTypeAsString() const
...
 
-void BgpLayer::parseNextLayer()
+void BgpLayer::parseNextLayer(ProtocolType parseUntil, OsiModelLayer parseUntilLayer)
 {
 	size_t headerLen = getHeaderLen();
 	if (m_DataLen <= headerLen || headerLen == 0)
 		return;
 
+	if (getProtocol() == parseUntil || getOsiModelLayer() == parseUntilLayer)
+        return;
```
egecetin, line 86: It will be a pain to change a bit but all `return` statements in cpp files unfortunately requires a tab indentation

rtmeng: Got it, somehow didn’t notice that locally, I’ll go through and add them

tigercosmos:
Could you also add tests for these changes?

seladb:
@rtmeng do you plan to continue working on this PR?

rtmeng:
Sorry just saw this @seladb, I won’t be able to in the next month so I can close this PR and reopen it later if still applicable.

PR #1381, opened 2024-05-04
ensure data is not nullptr

### LINKED ISSUE #1380: SNI Extension with zero length
We use PcapPlusPlus to get the server hostname from SSLServerNameIndicationExtension. However we found a crash once in a machine where the extension has length 0.

As a result when fetching the hostname we get a null pointer exception and our program crashes here https://github.com/seladb/PcapPlusPlus/blob/4d48615c9bfb2ef0c61d50e1e1dd3e3d63a82be0/Packet%2B%2B/src/SSLHandshake.cpp#L1121. 

We have seen this issue once. What scenario can cause this? Should the caller always check if getData() is not null before calling getHostName()?

OS Version:            macOS 14.4.1 (23E224)

Crash at:
0x10246dcec pcpp::SSLServerNameIndicationExtension::getHostName() const + 44

tigercosmos:
fixed #1380

### DIFF
--- a/Packet++/src/SSLHandshake.cpp
+++ b/Packet++/src/SSLHandshake.cpp
@@ -1117,6 +1117,10 @@ uint8_t* SSLExtension::getData() const
 
 std::string SSLServerNameIndicationExtension::getHostName() const
 {
+	if (!getData())
+	{
+		return "";
+	}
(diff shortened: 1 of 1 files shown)


tigercosmos:
ops, it's already in https://github.com/seladb/PcapPlusPlus/pull/1377

PR #1382, opened 2024-05-04
Changed BpfFilterWrapper to use smart pointers internally.

Dimi1010:
Added custom deleter type `BpfProgramDeleter` to encapsulate cleanup logic for a `bpf_program` pointer.

Changed `BpfFilterWrapper` to keep a `unique_ptr` with custom deleter instead of raw pointer to offload the cleanup logic to the smart pointer instead of handling it manually.


Pcap++/src/PcapFilter.cpp
```diff
@@ -63,8 +67,9 @@ bool BpfFilterWrapper::setFilter(const std::string& filter, LinkLayerType linkTy
 			return false;
 		}
 
+		// TODO: Do we technically need 'free_program' here? Both m_program and m_filter string are overwritten directly after that.
```
tigercosmos, line 70: I agree that we should remove this.

Pcap++/header/PcapFilter.h
```diff
@@ -78,7 +90,7 @@ namespace pcpp
 	private:
 		std::string m_FilterStr;
 		LinkLayerType m_LinkType;
-		bpf_program* m_Program;
+		std::unique_ptr<bpf_program, detail::BpfProgramDeleter> m_Program;
```
tigercosmos, line 93: love this.

tigercosmos (approved):
Thanks, LGTM.

Pcap++/header/PcapFilter.h
```diff
@@ -69,6 +69,18 @@ namespace pcpp
 		LESS_OR_EQUAL
 	} FilterOperator;
 
+	namespace detail
```
seladb, line 72: maybe rename this to `internal` as we have in [other places](https://github.com/seladb/PcapPlusPlus/blob/4d48615c9bfb2ef0c61d50e1e1dd3e3d63a82be0/Common%2B%2B/header/IpUtils.h#L59)?

Pcap++/src/PcapFilter.cpp
seladb: Yes I agree also

Pcap++/header/PcapFilter.h
tigercosmos: We also have `detail` here: https://github.com/seladb/PcapPlusPlus/blob/4d48615c9bfb2ef0c61d50e1e1dd3e3d63a82be0/Pcap%2B%2B/header/PcapFilter.h#L542

it's better to unify them.

seladb: yes I agree! I personally like `internal` better than `detail`

Dimi1010: Implemented 539201b

Pcap++/src/PcapFilter.cpp
Dimi1010: Removed the call in df9f37e. Would still like an opinion on the previous comment tho.

Pcap++/src/PcapFilter.cpp
```diff
@@ -28,15 +28,32 @@ bool GeneralFilter::matchPacketWithFilter(RawPacket* rawPacket)
...
+	}
+
+	/**
+	 * @class PcapTDeleter
+	 * A deleter that cleans up a pcap_t structure by calling pcap_close.
+	 */
+	struct PcapTDeleter
+	{
+		void operator()(pcap_t* ptr) const { pcap_close(ptr); }
+	};
```
seladb, line 46: This is nice! I think we can use it in more places in the code, especially in `PcapLiveDevice` and `PcapFileDevice.h`. But for now we can leave this code here

seladb (approved):

seladb:
@tigercosmos can you also re-review before we merge?

Pcap++/src/PcapFilter.cpp
tigercosmos: current way is fine.

tigercosmos:
thanks @Dimi1010

PR #1385, opened 2024-05-05
Replaced raw arrays with std::array in IPv4 and IPv6 address classes.

Dimi1010:
Overview of changes:
- Replaced raw array usages with `std::array` in internal implementation of `IPv4Address` and `IPv6Address`.

Overview of new additions:
- Added new constructors to `IPv4Address` and `IPv6Address` that take `std::array<uint8_t, 4>` and `std::array<uint8_t, 16>` respectively.
- Added new getter method `toByteArray()` to to `IPv4Address` and `IPv6Address` that return `std::array<uint8_t, 4>` and `std::array<uint8_t, 16>` respectively.

### DIFF
--- a/Common++/header/IpAddress.h
+++ b/Common++/header/IpAddress.h
@@ -64,32 +64,27 @@ namespace pcpp
 		IPv4Address(const std::string& addrAsString);
 
 		/**
-		 * Converts the IPv4 address into a 4B integer
-		 * @return a 4B integer in network byte order representing the IPv4 address
+		 * @return A 4-byte integer in network byte order representing the IPv4 address
 		 */
 		inline uint32_t toInt() const;
 
 		/**
-		 * Returns a view of the IPv4 address as a 4-byte raw C-style array
-		 * @return A non-owning pointer to 4-byte array representing the IPv4 address
+		 * @return A non-owning pointer to 4-byte C-style array representing the IPv4 address
 		 */
 		const uint8_t* toBytes() const { return m_Bytes.data(); }
 
 		/**
-		 * Returns a view of the IPv4 address as a std::array of bytes
 		 * @return A reference to a 4-byte standard array representing the IPv4 address
 		 */
 		const std::array<uint8_t, 4>& toByteArray() const { return m_Bytes; }
 
 		/**
-		 * Returns a std::string representation of the address
 		 * @return A string representation of the address
 		 */
 		std::string toString() const;
 
 		/**
-		 * Determine whether the address is a multicast address
-		 * @return True if an address is multicast
+		 * @return True if an address is multicast, false otherwise.
 		 */
 		bool isMulticast() const;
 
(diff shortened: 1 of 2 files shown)


Common++/header/IpAddress.h
```diff
@@ -160,7 +172,7 @@ namespace pcpp
 	uint32_t IPv4Address::toInt() const
 	{
 		uint32_t addr;
-		memcpy(&addr, m_Bytes, sizeof(m_Bytes));
+		memcpy(&addr, m_Bytes.data(), m_Bytes.size() * sizeof(uint8_t));
```
tigercosmos, line 175: why not just use `sizeof(m_Bytes)`?

tigercosmos (approved):

Common++/header/IpAddress.h
```diff
@@ -65,7 +72,12 @@ namespace pcpp
 		/**
 		 * Returns a pointer to 4-byte array representing the IPv4 address
 		 */
-		const uint8_t* toBytes() const { return m_Bytes; }
+		const uint8_t* toBytes() const { return m_Bytes.data(); }
+
+		/**
+		 * Returns a reference to a 4-byte standard array representing the IPv4 address
+		 */
```
seladb, line 79: Maybe we can do:
```cpp
/**
 * @return A reference to a 4-byte standard array representing the IPv4 address
 */
```

We can also update the documentation of `toBytes()`

Common++/header/IpAddress.h
```diff
@@ -193,7 +211,12 @@ namespace pcpp
 		/**
 		 * Returns a pointer to 16-byte array representing the IPv6 address
 		 */
-		const uint8_t* toBytes() const { return m_Bytes; }
+		const uint8_t* toBytes() const { return m_Bytes.data(); }
+
+		/**
+		 * Returns a reference to a 16-byte standard array representing the IPv6 address
+		 */
```
seladb, line 218: ditto

Common++/header/IpAddress.h
Dimi1010: Updated the documentation b3228c1, please check it and see if its good.

seladb: In getters we usually tend to avoid the description and only have a `@return` clause. For example;
```cpp
/**
 * @return A reference to a 4-byte standard array representing the IPv4 address
 */
```

Instead of:
```cpp
/**
 * Returns a reference to a 4-byte standard array representing the IPv4 address
 * @return A reference to a 4-byte standard array representing the IPv4 address
 */
```

Dimi1010: I guess it makes sense for the getters, tho I was mostly following the same pattern as the `toString` method directly below the `toBytes`, as I consider the methods more similar to the `toString` converter method then to a plain getter method.

seladb: Yes, we changed it at some point because it doesn't make a lot of sense to repeat the same sentence twice... feel free to change `toString()` as well if you want 🙂

Dimi1010: Fixed 1d145b7

seladb (approved):

PR #1386, opened 2024-05-06
Pcap device list refactoring to use smart pointers.

Dimi1010:
This PR mainly aims to update `PcapLiveDeviceList` and `PcapRemoteDeviceList` to utilize cpp11's `unique_ptr` for dynamic memory management.

Overview of changes:
- [x] Refactor `PcapLiveDeviceList` to use smart pointers internally for memory management.
- [x] Refactor `PcapRemoteDeviceList` to use smart pointers internally for memory management.
- [x] Refactor `PcapRemoteDevice` to keep ~~its own `RemoteAuthentication` copy or~~ a shared reference of `RemoteAuthentication` with `PcapRemoteDeviceList` via `std::shared_ptr` as it currently uses a raw pointer to share authentication with the device list.
- [x] Add `shared_ptr` overloads to `PcapLiveDeviceList` methods. Deprecate raw pointer methods.
- [x] Add `shared_ptr` overloads to `PcapRemoteDeviceList` methods. Deprecate raw pointer methods.
- [x] Add unit tests for smart pointer API methods for `PcapLiveDeviceList`.
- [x] Add unit tests for smart pointer API methods for `PcapRemoteDeviceList`.
- [ ] Find a better way to keep the old iterators than keeping a duplicate list?
- [x] Add `unique_ptr` overloads for `clone` to `PcapLiveDevice` ~~and `PcapRemoteDevice`~~.
- [x] Add unit tests for smart pointer API method for `PcapLiveDevice`.

Additional minor changes:
- Marked `PcapLiveDeviceList` and `PCapRemoteDeviceList`'s copy ctors as explicitly `deleted` instead of keeping them `private`.
- Moved internal helper deleters to `MemoryUtils.h`.
- Centralized code for fetching device list from Libpcap/Winpcap/Npcap in `DeviceUtils.h/cpp`. (It was duplicated a couple of times)


tigercosmos:
> RemoteAuthentication

IMO, the unique copy makes sense, since once the authentication is set, it should never be possible to change by others.

Dimi1010:
> > RemoteAuthentication
> 
> IMO, the unique copy makes sense, since once the authentication is set, it should never be possible to change by others.

Atm, made it so the user provides a `unique_ptr` to the factory method that is then internally changed to `shared_ptr` by the device list and shared with the list's devices. So all devices use the same auth object but it should not be exposed externally, outside the device list <-> device relationship.

tigercosmos:
> > > RemoteAuthentication
> > 
> > 
> > IMO, the unique copy makes sense, since once the authentication is set, it should never be possible to change by others.
> 
> Atm, made it so the user provides a `unique_ptr` to the factory method that is then internally changed to `shared_ptr` by the device list and shared with the list's devices. So all devices use the same auth object but it should not be exposed externally, outside the device list <-> device relationship.

True, as `PcapRemoteDeviceList` doesn't expose `PcapRemoteAuthentication`, `PcapRemoteAuthentication` can directly be a shared object among the list and device.

Pcap++/header/PcapLiveDeviceList.h
```diff
@@ -49,50 +57,105 @@ namespace pcpp
 
 		/**
 		 * @return A vector containing pointers to all live devices currently installed on the machine
+		 * @deprecated This method is deprecated in favor of the SmartPtrAPI overload.
+		 */
+		PCPP_DEPRECATED_RAW_PTR_API const std::vector<PcapLiveDevice*>& getPcapLiveDevicesList() const;
```
Dimi1010, line 62: `PcapLiveDeviceList` provides a getter API to get all registered devices, while `PcapRemoteDeviceList` provides a container like API to get all registered devices. 

In my opinion, it would be best if both lists provided a standardized API. Which one would be better?
Personally, I prefer the iterator approach, although it will be a bit trickier to implement as I dislike returning reference getters, as they make the underlying implementation difficult to change in the future.

Pcap++/header/PcapRemoteDeviceList.h
```diff
@@ -97,53 +174,86 @@ namespace pcpp
...
+		 * @param[in] apiTag Disambiguating tag for SmartPtrAPI.
+		 * @return A shared pointer to the PcapRemoteDevice if found, nullptr otherwise
+		 */
+		std::shared_ptr<PcapRemoteDevice> getRemoteDeviceByIP(const std::string& ipAddrAsString, SmartPtrApiTag apiTag) const;
 
 		/**
 		 * @return An iterator object pointing to the first PcapRemoteDevice in list
 		 */
-		RemoteDeviceListIterator begin() { return m_RemoteDeviceList.begin(); }
+		RemoteDeviceListIterator begin() { return m_RemoteDeviceListView.begin(); }
```
Dimi1010, line 237: See comment in `PcapLiveDeviceList` on `getPcapLiveDevicesList`

Pcap++/header/PcapLiveDeviceList.h
```diff
@@ -1,12 +1,16 @@
...
+#include "DeprecationUtils.h"
 #include "IpAddress.h"
 #include "PcapLiveDevice.h"
 #include <vector>
+#include <memory>
 
 
 /// @file
 
+#define PCPP_DEPRECATED_RAW_PTR_API PCPP_DEPRECATED("This method is deprecated in favor of the SmartPtrAPI overload.")
```
Dimi1010, line 12: Doxygen generation really dislikes this macro for some reason, when used on a member function. No idea why... tried several things to fix them including mirroring the way the `define` is written in `TcpLayer.h`, but none of them worked. Would like a second opinion on why the generation fails on the macro.

Dimi1010: Doxygen test fails with the following:
```
Preprocessing /tmp/cirrus-ci-build/Pcap++/head/tmp/cirrus-ci-build/Pcap++/header/PcapLiveDeviceList.h:62: error: Found ';' while parsing initializer list! (doxygen could be confused by a macro call without semicolon) (warning treated as error, aborting now)
er/MBufRawPacket.h...
```

Dimi1010 (commented):
@seladb @tigercosmos could you please look at the comments and give an opinion?

Pcap++/header/PcapLiveDeviceList.h
tigercosmos: I think it makes sense to have both a getter and an iterator for a container-like object. 

Regarding the iterator type, since `std::shared_ptr<T>` and `T*` are pointer types, the users' existing code (such as `->`) should not have much impact.

However, if we really don't want to break the API, as you said, it's still possible to have some conversion, but it's just troublesome.

Pcap++/header/DeviceUtils.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
```
tigercosmos, line 1: suggest to put all util functions in one file, as there is not much code.

Pcap++/header/PcapLiveDevice.h
```diff
@@ -24,6 +25,15 @@ typedef struct pcap_addr pcap_addr_t;
 */
 namespace pcpp
 {
+	/*
+	 * @class SmartPtrApiTag
+	 * Helper tag to disambiguate smart pointer API.
+	 */
+	struct SmartPtrApiTag{};
+	/**
+	 * Helper tag constant for disambuguating smart pointer API.
```
tigercosmos, line 34: typo: disambiguating, btw I rarely see this word, I wonder if it is precise here.

Pcap++/header/PcapLiveDevice.h
```diff
@@ -130,6 +137,8 @@ namespace pcpp
 		static void onPacketArrivesNoCallback(uint8_t* user, const struct pcap_pkthdr* pkthdr, const uint8_t* packet);
 		static void onPacketArrivesBlockingMode(uint8_t* user, const struct pcap_pkthdr* pkthdr, const uint8_t* packet);
 	public:
+		PcapLiveDevice(const PcapLiveDevice& other) = delete;
```
tigercosmos, line 140: please also implement moving construction/assignment.

Pcap++/header/PcapLiveDeviceList.h
```diff
@@ -49,50 +57,105 @@ namespace pcpp
...
 
 		/**
 		 * Get a pointer to the live device by its IP address. IP address can be both IPv4 or IPv6
 		 * @param[in] ipAddr The IP address defined for the device
-		 * @return A pointer to the live device if this IP address exists. NULL otherwise
+		 * @return A pointer to the live device if this IP address exists, nullptr otherwise
+		 * @deprecated This method is deprecated in favor of the SmartPtrAPI overload.
+		 */
+		PCPP_DEPRECATED_RAW_PTR_API PcapLiveDevice* getPcapLiveDeviceByIp(const IPAddress& ipAddr) const;
+		/**
```
tigercosmos, line 76: need a space.

Pcap++/header/PcapLiveDevice.h
tigercosmos: btw, if the underlying are shared pointers, why do we disallow coping?

Pcap++/src/PcapFilter.cpp
```diff
@@ -66,7 +58,7 @@ bool BpfFilterWrapper::setFilter(const std::string& filter, LinkLayerType linkTy
 
 	if (filter != m_FilterStr || linkType != m_LinkType)
 	{
-		std::unique_ptr<pcap_t, internal::PcapTDeleter> pcap = std::unique_ptr<pcap_t, internal::PcapTDeleter>(pcap_open_dead(linkType, DEFAULT_SNAPLEN));
+		std::unique_ptr<pcap_t, internal::PcapCloseDeleter> pcap = std::unique_ptr<pcap_t, internal::PcapCloseDeleter>(pcap_open_dead(linkType, DEFAULT_SNAPLEN));
```
tigercosmos, line 61: I would write `auto pcap = std::unique_ptr<pcap_t, internal::PcapCloseDeleter>(pcap_open_dead(linkType, DEFAULT_SNAPLEN));`

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -402,39 +404,35 @@ void PcapLiveDevice::close()
...
 	}
 
-	pcap_if_t* currInterface = interfaceList;
-	while (currInterface != nullptr)
+	for (pcap_if_t* currInterface = interfaceList.get(); currInterface != nullptr; currInterface = currInterface->next)
 	{
-		if(!strcmp(currInterface->name, getName().c_str()))
-			break;
-		currInterface = currInterface->next;
+		if (!strcmp(currInterface->name, getName().c_str()))
```
tigercosmos, line 424: std::strcmp

Pcap++/src/PcapLiveDeviceList.cpp
```diff
@@ -253,34 +247,71 @@ void PcapLiveDeviceList::setDnsServers()
...
+
+const std::vector<PcapLiveDevice*>& PcapLiveDeviceList::getPcapLiveDevicesList() const
+{
+	updateLiveDeviceListView();
+	return m_LiveDeviceListView;
+}
+
 PcapLiveDevice* PcapLiveDeviceList::getPcapLiveDeviceByIp(const IPAddress& ipAddr) const
+{
+	// Technically this creates and deconstructs an extra shared ptr leading to some inneficiencies but its shorter.
```
tigercosmos, line 273: please check the typo and grammar.

Pcap++/header/PcapLiveDevice.h
Dimi1010: > btw, if the underlying are shared pointers, why do we disallow coping?

IDK. I think the reason a `LiveDevice` itself is not copiable is something with the underlying Pcap descriptors. Although there is still the `clone` method so, not really sure. The current change is mostly a stylistic from the pre-Cpp11 style of private ctors to post-Cpp11 deleted ctors.

Pcap++/header/PcapLiveDeviceList.h
Dimi1010: > I think it makes sense to have both a getter and an iterator for a container-like object.

Both is an option. 
 
> Regarding the iterator type, since `std::shared_ptr<T>` and `T*` are pointer types, the users' existing code (such as `->`) should not have much impact.

Yeah, they shouldn't really. unless they decided to save the value from the iterator somewhere. That is a breaking change, but it would only need a `get()` call from the smart pointer to fix, as old code should still keep the list alive to keep the devices alive.

> However, if we really don't want to break the API, as you said, it's still possible to have some conversion, but it's just troublesome.

I tried my hand at converting iterator a bit earlier. It is technically possible but it quickly becomes templated black magic.

Pcap++/header/DeviceUtils.h
tigercosmos: I see. originally I thought we could merge `MemoryUtils` and `DeviceUtils`, but it did make sense to be separated.

Pcap++/header/PcapLiveDeviceList.h
tigercosmos: I prefer to remove the old API, and have both getter and new iterator. Then we can also remove the hack of `SmartPtrApiTag` and `m_LiveDeviceListView`

@seladb  what do you say?

Pcap++/header/PcapLiveDevice.h
tigercosmos: we can keep the old `clone`, but I think we can now embrace the copy constructor/assignment?

Dimi1010: Ah, figured out at least partly why its private / deleted. `LiveDevice` is a virtual class and usually handled by pointers to base class (i.e. `WinPcapLive` device is used through `PcapLiveDevice` pointer). A copy ctor used from a base pointer will only use the base copy ctor, not the actual copy ctor of the class the base ptr points to.

Pcap++/src/PcapFilter.cpp
Dimi1010: Added 5d9dcfd

Pcap++/header/PcapLiveDeviceList.h
Dimi1010: Fixed 0528e92. Thanks

Pcap++/header/PcapLiveDevice.h
Dimi1010: Fixed a95c946. I think it should be better now?

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: Fixed 4ae2edb

Pcap++/header/PcapLiveDevice.h
Dimi1010: > please also implement moving construction/assignment.

Added 4bcb50e

Dimi1010:
This PR will be split into multiple smaller PRs as part of #1431.

PR #1392, opened 2024-05-11
Add a benchmark which uses Google benchmark library

egecetin:
~~Automatically generate performance benchmark information for website using Google Benchmark~~

Add a google benchmark based benchmark application


tigercosmos:
Where can I see the benchmark result?

egecetin:
@tigercosmos Check "Run benchmark" step https://github.com/seladb/PcapPlusPlus/actions/runs/10644856460/job/29510043289

tigercosmos:
Will it show a warning if the benchmark performance decreases too much? Otherwise, people will not click into the CI and check the result.

egecetin:
@tigercosmos Actually my aim is to generate and update the benchmark results automatically. I enabled PR runs for just testing. I'll disable it when the work is done, and the runs will be only from master or dev. Because benchmark results can be very flaky without a dedicated machine/runner unfortunately. This workflow and integration with Google benchmark are just for easier benchmarking in local PCs and to give an idea about to performance in PcapPlusPlus website (since current benchmark really old)

Benchmarks/benchmark.cpp
```diff
@@ -0,0 +1,281 @@
...
+			}
+			++layerTypes[curLayer->getProtocol()];
+
+			// Count OSI layers. If the OSI layer is not in the map, add it
+			if (osiLayers.find(curLayer->getOsiModelLayer()) == osiLayers.end())
+			{
+				osiLayers[curLayer->getOsiModelLayer()] = 0;
+			}
+			++osiLayers[curLayer->getOsiModelLayer()];
+		}
```
seladb, line 153: If we want to measure the packet parsing, why do we need this? It's enough to just parse the packet, no? 🤔

seladb:
@egecetin I'm wondering if there's much value in publishing our benchmark without comparing PcapPlusPlus to other libraries. What do you think?

egecetin (commented):
@seladb Actually it is just for give an idea about the performance in terms of packet per second and for easier local benchmarking (both disk and cpu sides). Github's hardware is good enough to get good results. This is why i added long sentence to README and a workflow to CI. I agree it alone does not have true meaning unfortunately so why I left the old benchmark also. I can  (at least might try 🙂) also update to comparing with other libraries periodically. Maybe once a month or at least manually run after every release in a different repo.

Benchmarks/benchmark.cpp
egecetin: My concern is if some compiler decides the parsedPacket is not used it might try to remove the parsing maybe. But I'm not sure

seladb: maybe we can do something simple like asserting the first layer is not null? 🤔

Examples/PcapPlusPlus-benchmark/README.md
seladb, line 1: I think we should re-structure this README file to make it more clear that there are actually 2 different benchmark applications included here

Examples/PcapPlusPlus-benchmark/README.md
```diff
@@ -6,3 +6,12 @@ This is a benchmark application used for measuring PcapPlusPlus performance. It
 See this page for more details: https://pcapplusplus.github.io/docs/benchmark
 
 This application currently compiles on Linux only (where benchmark was running on)
+
+Additionally, there is an integration of Google Benchmark library for benchmarking. For enabling, configure the library with `-DPCAPPP_USE_GOOGLEBENCHMARK=ON`. Each benchmark can be affected from different sources. You can check the table below for more information. If you have performance critical applications which will use PcapPlusPlus, please benchmark it in your environment to get better and more precise results. Also using different pcaps, both in terms of size (larger is better to prevent open/close overhead) and protocol/session variety, can also help to give idea about performance of PcapPlusPlus in your environment.
```
seladb, line 10: I used ChatGPT to find better wording for "Each benchmark can be affected from different sources" and it suggested: "Each benchmark can be influenced by various factors"

Examples/PcapPlusPlus-benchmark/README.md
```diff
@@ -6,3 +6,12 @@ This is a benchmark application used for measuring PcapPlusPlus performance. It
...
+
+|     Benchmark     |   Operation   |  Can be affected  |
```
seladb, line 12: ditto: "Influencing factors"

Examples/PcapPlusPlus-benchmark/benchmark-google.cpp
```diff
@@ -0,0 +1,263 @@
...
+	pcpp::RawPacket rawPacket;
+	for (auto _ : state)
+	{
+		if (!reader.getNextPacket(rawPacket))
+		{
+			// If the rawPacket is empty there should be an error
+			if (totalBytes == 0)
+			{
+				state.SkipWithError("Cannot read packet");
+				break;
```
seladb, line 38: shouldn't this use `return`?

If yes, we can simplify the code:
```cpp
if (!reader.getNextPacket(rawPacket))
{
	// If the rawPacket is empty there should be an error
	if (totalBytes == 0)
	{
		state.SkipWithError("Cannot read packet");
		break;
	}
	
	state.PauseTiming();
	reader.close();
	reader.open();
	state.ResumeTiming();
}
```

Examples/PcapPlusPlus-benchmark/benchmark-google.cpp
```diff
@@ -0,0 +1,263 @@
...
+			}
+			else  // If the rawPacket is not empty, it means the file is over
+			{
+				state.PauseTiming();
+				reader.close();
+				reader.open();
```
seladb, line 44: Why do we need to close and re-open the file? 🤔

Examples/PcapPlusPlus-benchmark/benchmark-google.cpp
```diff
@@ -0,0 +1,263 @@
...
+				state.ResumeTiming();
+				continue;
+			}
+		}
```
seladb, line 127: ditto, same comments as above

Examples/PcapPlusPlus-benchmark/benchmark-google.cpp
```diff
@@ -0,0 +1,263 @@
...
+		pcpp::MacAddress dstMac(randNum, randNum, randNum, randNum, randNum, randNum);
+
+		// Craft packet
+		std::unique_ptr<pcpp::EthLayer> ethLayer(new pcpp::EthLayer(srcMac, dstMac));
+
+		std::unique_ptr<pcpp::IPv4Layer> ipv4Layer(nullptr);
+		std::unique_ptr<pcpp::IPv6Layer> ipv6Layer(nullptr);
+
+		std::unique_ptr<pcpp::TcpLayer> tcpLayer(nullptr);
+		std::unique_ptr<pcpp::UdpLayer> udpLayer(nullptr);
```
seladb, line 166: A slightly nicer code could be:

```cpp
pcpp::Packet packet;

pcpp::MacAddress srcMac(randNum, randNum, randNum, randNum, randNum, randNum);
pcpp::MacAddress dstMac(randNum, randNum, randNum, randNum, randNum, randNum);
packet.addLayer(new pcpp::EthLayer(srcMac, dstMac), true);

// Randomly choose between IPv4 and IPv6
if (randNum % 2)
{
	packet.addLayer(new pcpp::IPv4Layer(randNum, randNum), true);
}
else
{
	std::array<uint8_t, 16> srcIP = { randNum, randNum, randNum, randNum, randNum, randNum, randNum, randNum,
		                              randNum, randNum, randNum, randNum, randNum, randNum, randNum, randNum };
	std::array<uint8_t, 16> dstIP = { randNum, randNum, randNum, randNum, randNum, randNum, randNum, randNum,
		                              randNum, randNum, randNum, randNum, randNum, randNum, randNum, randNum };

	packet.addLayer(new pcpp::IPv6Layer(srcIP, dstIP), true);
}

// Randomly choose between TCP and UDP
if (randNum % 2)
{
	packet.addLayer(new pcpp::TcpLayer(randNum % 65536, randNum % 65536), true);
}
else
{
	packet.addLayer(new pcpp::UdpLayer(randNum % 65536, randNum % 65536), true);
}

packet.computeCalculateFields();
```

The second boolean argument in `addLayer()` is `ownInPacket` which means the packet will free the layer memory when the packet is destroyed

Examples/PcapPlusPlus-benchmark/benchmark-google.cpp
egecetin: Google benchmark decides runtime length so the pcap file can (or mostly) is shorter than the actual runtime. So, for rewinding I closed and opened the same file again to get more packets until benchmark finishes.

Examples/PcapPlusPlus-benchmark/CMakeLists.txt
```diff
@@ -1,12 +1,39 @@
-add_executable(benchmark benchmark.cpp)
+add_executable(BenchmarkExample benchmark.cpp)
 
-target_link_libraries(benchmark PUBLIC PcapPlusPlus::Pcap++)
+target_link_libraries(BenchmarkExample PUBLIC PcapPlusPlus::Pcap++)
 
-set_target_properties(benchmark PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PCAPPP_BINARY_EXAMPLES_DIR}")
+set_target_properties(BenchmarkExample PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PCAPPP_BINARY_EXAMPLES_DIR}")
+
+if(PCAPPP_USE_GOOGLEBENCHMARK)
```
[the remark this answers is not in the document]
egecetin: Two reasons actually,

- Are want Google Benchmark as a dependency always? Google benchmarks require a compiler to support C++14
- This is not really important I can fix it probably but for MacOS (cross-compiling cases x86 -> ARM or vice versa I can't remember now) there was an error, and I can't figure out why. But then noticed the README specifies the benchmarks only run in linux. If all platforms supported, we should also remove this mention from README

Examples/PcapPlusPlus-benchmark/README.md
egecetin: Check 557839a

Examples/PcapPlusPlus-benchmark/README.md
egecetin: Check 557839a

Examples/PcapPlusPlus-benchmark/README.md
egecetin: Fixed with 557839a

Examples/PcapPlusPlus-benchmark/benchmark-google.cpp
egecetin: Fixed with 557839a

Examples/PcapPlusPlus-benchmark/CMakeLists.txt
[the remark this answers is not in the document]
seladb: 1. I don't mind having Google Benchmark as a dependency of this specific example, especially because it downloads it automatically
2. Is it possible to avoid compiling this example if the complier doesn't support C++14? I'm not a CMake expert but I assume this is possible?
3. > Then noticed the README specifies the benchmarks only run in linux. If all platforms supported, we should also remove this mention from README

    I didn't see where it's mentioned... I managed to compile and run it on Windows with MSVC 🤔

.github/workflows/build_and_test.yml
```diff
@@ -299,6 +299,7 @@ jobs:
             host-arch: x86_64
             arch: arm64
             config-zstd: OFF
+            additional-flags: -DCMAKE_CROSSCOMPILING=1 -DRUN_HAVE_STD_REGEX=0
```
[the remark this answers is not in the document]
egecetin: Otherwise cross-compile fails because of Google Benchmark. It can't detect regex implementation somehow. Check the build log https://github.com/seladb/PcapPlusPlus/actions/runs/11327526535/job/31498794905. The workaround mentioned here: https://github.com/google/benchmark/issues/351

Examples/PcapPlusPlus-benchmark/CMakeLists.txt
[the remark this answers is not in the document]
egecetin: 1. Agreed
2. It should be I'll take a look
3. Probably a forgotten mention https://github.com/seladb/PcapPlusPlus/blob/db905ad02e3112b134106d89c604a51ace595f76/Examples/PcapPlusPlus-benchmark/README.md?plain=1#L8

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
seladb: Is there a way to avoid compiling Google Benchmark if regex implementation isn't found or if `CMAKE_CROSSCOMPILING` is turned on? 🤔

Examples/PcapPlusPlus-benchmark/CMakeLists.txt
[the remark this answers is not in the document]
seladb: > 3. Probably a forgotten mention 

I think it can be compiled on all platforms so we can drop this comment

Examples/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.14)
```
egecetin, line 1: `fetchcontent_makeavailable` added in 3.14

egecetin:
@seladb It is ready to review if you have time

Examples/PcapPlusPlus-benchmark/benchmark-google.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+int main(int argc, char** argv)
+{
+	// Initialize the benchmark
+	benchmark::Initialize(&argc, argv);
+
+	// Parse command line arguments to find the pcap file name
+	for (int idx = 1; idx < argc; ++idx)
+	{
+		if (strcmp(argv[idx], "--pcap-file") == 0)
```
seladb, line 208: Since `--pcap-file` is a mandatory parameter (and the only param for now), maybe we don't need the param itself and can run the benchmark like this:
```
.\BenchmarkExampleGoogle.exe D:\my_packets.pcap
```

egecetin: @seladb The problem is arguments are also used by google benchmark. It supports some flags as default, for example `--benchmark_filter` or `--benchmark_out_format` etc.

seladb: Oh ok, so we can keep it 👍

tigercosmos (approved):
LGTM, I guess this is just for the tutorial; otherwise, I think the test cases of the benchmarks could be more comprehensive.

PR #1399, opened 2024-05-23
Use token for codecov uploads

egecetin:
Recently most uploads failed because of rate limiting. Using token should fix the issue.

https://docs.codecov.com/docs/adding-the-codecov-token

image

image


tigercosmos:
I am wondering if the token is under the organization or person?

egecetin:
@tigercosmos It should be under @seladb's account since the repo is under person account

seladb:
Maybe that's the reason coverage went down from 80+% to 74%? 🤔

egecetin:
@seladb Did you add token to repo with following instructions?

https://docs.codecov.com/docs/adding-the-codecov-token

It still says,
`[2024-05-24T06:40:41.530Z] ['verbose'] Passed token was 0 characters long`

egecetin:
Maybe it can be caused because changes still in dev branch also if you add a token

seladb:
> Maybe it can be caused because changes still in dev branch, if you add a token

I added the token just now:
image

PR #1400, opened 2024-05-23
Fixup TcpLayer deprecation macro cleanup.

Dimi1010:
The macro `PCPP_DEPRECATED_TCP_OPTION_TYPE` was always being undefined at the end of `TcpLayey.h` even if it was externally defined.

### DIFF
--- a/Packet++/header/TcpLayer.h
+++ b/Packet++/header/TcpLayer.h
@@ -5,11 +5,8 @@
 #include "TLVData.h"
 #include <string.h>
 
-#ifndef PCPP_DEPRECATED_TCP_OPTION_TYPE
-#define PCPP_DEPRECATED_TCP_OPTION_TYPE__LOCAL_DEFINE
 #define PCPP_DEPRECATED_TCP_OPTION_TYPE PCPP_DEPRECATED("enum TcpOptionType is deprecated; " \
                                                         "Use enum class TcpOptionEnumType instead")
-#endif
 
 /// @file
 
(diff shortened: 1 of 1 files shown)


Packet++/header/TcpLayer.h
```diff
@@ -6,6 +6,7 @@
 #include <string.h>
 
 #ifndef PCPP_DEPRECATED_TCP_OPTION_TYPE
```
tigercosmos, line 8: we already have `#pragma once`, so maybe we don't need `PCPP_DEPRECATED_TCP_OPTION_TYPE`?

Dimi1010: Wdym? They are different things?

seladb, line 8: Maybe we don't need `#ifndef PCPP_DEPRECATED_TCP_OPTION_TYPE`. We can assume no one will define it externally, same as we have with other macros in this project 🤔

Dimi1010: Well, if we are assuming that then yes, the ifndefs are superfluous.

tigercosmos (approved):

Packet++/header/TcpLayer.h
seladb, line 1: Maybe we should also remove `#undef PCPP_DEPRECATED_TCP_OPTION_TYPE` in the end of the file? 🤔

tigercosmos: only the current file uses this macro, so we shouldn't expose it.

seladb: yeah, I guess you're right. We can keep the `#unded`

seladb (approved):

PR #1403, opened 2024-05-24
Export ASN.1 records to string

seladb:


Packet++/header/Asn1Codec.h
```diff
@@ -169,6 +169,11 @@ namespace pcpp
 		 */
 		size_t getTotalLength() const { return m_TotalLength; }
 
+		/**
+		 * @return A string representation of the record
+		 */
+		std::string toString();
```
tigercosmos, line 175: should be `const`?

Packet++/header/Asn1Codec.h
```diff
@@ -280,6 +289,8 @@ namespace pcpp
 		void decodeValue(uint8_t* data, bool lazy) override;
 		std::vector<uint8_t> encodeValue() const override;
 
+		std::vector<std::string> toStringList() override;
```
tigercosmos, line 292: should be `const`?

Packet++/src/Asn1Codec.cpp
```diff
@@ -12,7 +15,106 @@
...
+			default:
+				return "Private";
+		}
+	}
+
+	std::string toString(Asn1UniversalTagType tagType)
+	{
+		switch (tagType)
+		{
+			case Asn1UniversalTagType::EndOfContent:
```
tigercosmos, line 39: recommend to use a hash table here.

Packet++/src/Asn1Codec.cpp
```diff
@@ -409,6 +547,20 @@ namespace pcpp {
 		return result;
 	}
 
+	std::vector<std::string> Asn1ConstructedRecord::toStringList()
+	{
+		decodeValueIfNeeded();
+		std::vector<std::string> result = {Asn1Record::toStringList().front()};
+		for (auto subRecord : m_SubRecords)
```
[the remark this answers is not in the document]
seladb: `m_SubRecords` is a `PointerVector<Asn1Record>` so each element in the iterator is of type `Asn1Record*`

Packet++/header/Asn1Codec.h
```diff
@@ -211,6 +216,10 @@ namespace pcpp
 
 		uint8_t encodeTag();
 		std::vector<uint8_t> encodeLength() const;
+
+		virtual std::vector<std::string> toStringList();
```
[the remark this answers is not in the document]
seladb: This is a protected method that is not exposed as part of our API, so documentation is not required

Packet++/header/Asn1Codec.h
seladb: Here is the problem:
- `toString()` calls `toStringList()` internally, which means  `toStringList()` should also be `const`
- Some of the overrides of ` `toStringList()` use `getValue()` or `decodeValueIfNeeded()` internally, which means these methods should also be `const`
- `getValue()` also uses `decodeValueIfNeeded()` internally
- `decodeValueIfNeeded()` cannot be `const` because it decodes the record, hence changes the class members

Do you see any way around it? 🤔

Packet++/header/Asn1Codec.h
seladb: Please see this comment: https://github.com/seladb/PcapPlusPlus/pull/1403#discussion_r1614032304

Packet++/src/Asn1Codec.cpp
```diff
@@ -536,16 +693,52 @@ namespace pcpp {
...
 
 	std::vector<uint8_t> Asn1OctetStringRecord::encodeValue() const
 	{
-		return {m_Value.begin(), m_Value.end()};
+		if (m_IsPrintable)
+		{
+			return {m_Value.begin(), m_Value.end()};
+		}
+
+		auto rawValueSize = static_cast<size_t>(m_Value.size() / 2);
```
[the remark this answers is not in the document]
seladb: `m_Value` in this case contains a string of hex stream, for example: `"46f722c636e3"`. When we encode it we need to convert the string back to a byte array, so the size of the byte array will be the size of the string divided by 2:
```cpp
std::string hexStream = "46f722c636e3" // size is 12

// equivalent to:

std::vector<uint8_t> rawValue = {0x46, 0xf7, 0x22, 0xc6, 0x36, 0xe3}; // size is 6
```

Packet++/src/Asn1Codec.cpp
```diff
@@ -566,6 +759,11 @@ namespace pcpp {
 		return { byte };
 	}
 
+	std::vector<std::string> Asn1BooleanRecord::toStringList()
```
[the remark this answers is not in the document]
seladb: We need a vector if the string is over one line, which happens in constructed records which are nested:
https://github.com/seladb/PcapPlusPlus/blob/f6da2566442653db568afdf6dc32e7ea74392978/Packet++/src/Asn1Codec.cpp#L550

It'll look something like this:
```
Application (0) (constructed), Length: 2+80
   Integer, Length: 2+1, Value: 3
  OctetString, Length: 2+59, Value: cn=Administrator,cn=Users,dc=cloudshark-a,dc=example,dc=com
  ContextSpecific (0), Length: 2+14
```

Packet++/src/Asn1Codec.cpp
```diff
@@ -327,6 +429,42 @@ namespace pcpp {
...
+	std::string Asn1Record::toString()
+	{
+		auto lines = toStringList();
+
+		auto commaSeparated = [](std::string str1, std::string str2)
+		{
+			return std::move(str1) + '\n' + std::move(str2);
+		};
+
+		return std::accumulate(std::next(lines.begin()), lines.end(),lines[0], commaSeparated);
```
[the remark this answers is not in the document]
seladb: Fixed in this commit: https://github.com/seladb/PcapPlusPlus/pull/1403/commits/c2fffbeed339314526d9eb4b7d3b4bdd084657a6

Packet++/src/Asn1Codec.cpp
seladb: Fixed in this commit: https://github.com/seladb/PcapPlusPlus/pull/1403/commits/c2fffbeed339314526d9eb4b7d3b4bdd084657a6

Packet++/header/Asn1Codec.h
tigercosmos: I think you can use `mutable` for `decodeValueIfNeeded`? Then we can still have `toString` as `const`.

Packet++/header/Asn1Codec.h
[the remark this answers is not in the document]
tigercosmos: got it.

Packet++/src/Asn1Codec.cpp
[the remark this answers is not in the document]
tigercosmos: I see, it's clear now. Perhaps add the comment to explain it in the code?

Packet++/header/Asn1Codec.h
seladb: As far as I know `mutable` only applies to class members (not methods). `decodeValueIfNeeded()` calls `decodeValue()` internally, which is implemented differently in every record type. If we decide to go with the `mutable` approach we probably need to change a lot of class members to `mutable` which isn't very nice I think...

Another approach is to do that:
```cpp
std::string Asn1Record::toString() const
{
	auto lines = const_cast<Asn1Record*>(this)->toStringList();
	...
}
```

But it's not nice either 😕

Packet++/src/Asn1Codec.cpp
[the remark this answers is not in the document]
seladb: Sure, added in this commit: https://github.com/seladb/PcapPlusPlus/pull/1403/commits/ba4d48fe91a09dccd32c468512c9528e14c38059

Packet++/header/Asn1Codec.h
tigercosmos: @seladb I think most people don't expect calling `toString` to change the class members, and modifying members to make a string doesn't seem to be a good practice.

Is it possible to refactor the code to separate the mutation logic? Or maybe we can change the function name? or at least `const_cast` is acceptable?

btw, I notice that `bool lazy` appears in many places, but most of the `lazy`s are not used, maybe we can also improve this...

seladb: The idea in `lazy` decoding/parsing is to avoid parsing the entire record and its nested records until needed. The reason is that ASN.1 records can be quite extensive ([here](https://lapo.it/asn1js/#MG8CAQJjagQACgECCgEAAgEAAgEAAQEAqTyBHDIuMTYuODQwLjEuMTEzNzMwLjMuMy4yLjQ2LjGCEGRlcGFydG1lbnROdW1iZXKDBz49TjQ3MDmEAf8wGQQCY24EAnNuBA90ZWxlcGhvbmVOdW1iZXI) is an example), so decoding all records ahead of time might significantly increase the memory required to store a parsed packet. 

The "price" we pay in lazy parsing is that calling so-called "innocent" methods like `toString()` might affect the class members because it triggers the parsing/decoding process. But I think it's a reasonable price to avoid bloating the memory. Using `const_cast` or even refactoring class members to be `mutable` is a bit of "cheating" because it hides the fact that calling these methods actually change the class members.

tigercosmos: 1. I wonder if it is possible to call the decoding function before calling `toString()` because currently you put the decoding function inside `toString`. The behavior should be the same, but `toString` can be `const`. What do you think about this?
2. for `lazy`, I understand the purpose of the lazy initialization, but in some classes, they don't use the `lazy` argument. It is an open question if it is necessary to override the function with `lazy`, or maybe we can just declare a new function that doesn't have `lazy`. For example, `void Asn1IntegerRecord::decodeValue(uint8_t* data, bool lazy)` implementation doesn't use `lazy`, so it can be `void Asn1IntegerRecord::decodeValue(uint8_t* data)`.

seladb: > 1. I wonder if it is possible to call the decoding function before calling `toString()` because currently you put the decoding function inside `toString`. The behavior should be the same, but `toString` can be `const`. What do you think about this?

I'm unsure how we can guarantee decoding happened before the user called `toString()`?

> 2\. for `lazy`, I understand the purpose of the lazy initialization, but in some classes, they don't use the `lazy` argument. It is an open question if it is necessary to override the function with `lazy`, or maybe we can just declare a new function that doesn't have `lazy`. For example, `void Asn1IntegerRecord::decodeValue(uint8_t* data, bool lazy)` implementation doesn't use `lazy`, so it can be `void Asn1IntegerRecord::decodeValue(uint8_t* data)`.

If we do that, how can we generically allow lazy decoding for all types of records? Indeed, some use lazy decoding and some don't, but when PcapPlusPlus decodes a packet it doesn't know what's in it, so it automatically asks for lazy decoding. `toString()` is a generic method in the parent class `Asn1Record` and it doesn't know if the specific type uses lazy decoding or not.

I'm feeling I'm missing something in your proposal, can you be more specific on the changes you propose? 🤔

tigercosmos: @seladb  I made a PR for the explanation. Please check https://github.com/seladb/PcapPlusPlus/pull/1409/files

seladb: Thanks @tigercosmos for making the PR! The main issue I see with this approach is cascading record decoding, here is an example:

Consider we have the following record structure:
```
Constructed record
  Integer record (value == 1)
  OctetString record (value == "my string")
  Constructed record
    Boolean record (value == true)
    Integer record (value == 1)
    Constructed record
      OctetString record (value == "my string 2")
```

- :white_check_mark: When we call `pcpp::Asn1Record::decode(data, dataLen, false)`, it decodes the entire structure as exepcted
-  :white_check_mark: When we call `pcpp::Asn1Record::decode(data, dataLen, true)`, it doesn't decode anything so we get the following:
   ```
   Constructed record (value not decoded)
   ```
- :x: However, if we try to get the sub-records of the constructed value it decodes the entire structure:
   ```cpp
   auto constructedRecord = pcpp::Asn1Record::decode(data, dataLen, true);
   auto subRecords = constructedRecord.getSubRecords();
   ```
   `getSubRecords()` calls `decodeValueIfNeeded()` which calls `decodeValue()` which decodes the entire structure. I'd expect the following:
   ```
   Constructed record
     Integer record (value not decoded)
     OctetString record (value not decoded)
     Constructed record (value not decoded)
   ```
   But instead I get:
   ```
   Constructed record
     Integer record (value == 1)
     OctetString record (value == "my string")
     Constructed record
       Boolean record (value == true)
       Integer record (value == 1)
       Constructed record
         OctetString record (value == "my string 2")
```

egecetin (approved):

Packet++/header/Asn1Codec.h
tigercosmos: @seladb it's clear now, thanks for the explanation. 
About the `const` stuff, I cannot provide a better solution for now, so I won't be against the current design.

PR #1406, opened 2024-05-24
Fixed live device clone method returning wrong device type on win32 machines.

Dimi1010:
Fixes `clone` behavior of `PcapLiveDevice` and `WinPcapLiveDevice` and adds unit tests checking device classes when cloning.

Originally as the `clone` method was not virtual when a pointer referencing a `WinPcapLiveDevice` was cloned a device of the base class `PcapLiveDevice` was returned, instead of the derived class.

### DIFF
--- a/Pcap++/header/WinPcapLiveDevice.h
+++ b/Pcap++/header/WinPcapLiveDevice.h
@@ -56,11 +56,8 @@ namespace pcpp
 		 */
 		int getMinAmountOfDataToCopyFromKernelToApplication() const { return m_MinAmountOfDataToCopyFromKernelToApplication; }
 
-		/**
-		 * Clones the current device class
-		 * @return Pointer to the copied class
-		 */
-		PcapLiveDevice* clone() const override;
+	protected:
+		PcapLiveDevice* cloneInternal(pcap_if_t& devInterface) const override;
 	};
 
 } // namespace pcpp
--- a/Pcap++/header/WinPcapLiveDevice.h
+++ b/Pcap++/header/WinPcapLiveDevice.h
@@ -55,6 +55,12 @@ namespace pcpp
 		 * setMinAmountOfDataToCopyFromKernelToApplication())
 		 */
 		int getMinAmountOfDataToCopyFromKernelToApplication() const { return m_MinAmountOfDataToCopyFromKernelToApplication; }
+
+		/**
+		 * Clones the current device class
+		 * @return Pointer to the copied class
+		 */
+		PcapLiveDevice* clone() const override;
 	};
 
 } // namespace pcpp
--- a/Pcap++/header/WinPcapLiveDevice.h
+++ b/Pcap++/header/WinPcapLiveDevice.h
@@ -57,7 +57,7 @@ namespace pcpp
 		int getMinAmountOfDataToCopyFromKernelToApplication() const { return m_MinAmountOfDataToCopyFromKernelToApplication; }
 
 	protected:
-		PcapLiveDevice* cloneInternal(pcap_if_t& devInterface) const override;
+		WinPcapLiveDevice* cloneInternal(pcap_if_t& devInterface) const override;
 	};
 
 } // namespace pcpp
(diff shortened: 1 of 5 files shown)


Pcap++/src/WinPcapLiveDevice.cpp
```diff
@@ -126,6 +126,36 @@ bool WinPcapLiveDevice::setMinAmountOfDataToCopyFromKernelToApplication(int size
...
+
+	if (currInterface)
+		retval = new WinPcapLiveDevice(currInterface, true, true, true);
+	else
+		PCPP_LOG_ERROR("Can't find interface " << getName().c_str());
+
+	pcap_freealldevs(interfaceList);
+	return retval;
+}
+
```
seladb, line 158: Both `clone()` implementations have mostly all common code. Maybe we can do something like this (it's all pseudo-code that I didn't actually run):
```cpp
class PcapLiveDevice
{
public:
    PcapLiveDevice* clone() const
    {
        // call pcap_findalldevs, iterate over the interfaces and find the right one
        pcap_if_t* currInterface = ...
        if(currInterface)
	    result = cloneInternal(currInterface);
        else
	    PCPP_LOG_ERROR("Can't find interface " << getName().c_str());
        
        // call pcap_freealldevs and do cleanup

        return result;
    }
protected:
    virtual PcapLiveDevice* cloneInternal(void* iface)
    {
        // need to case from `void*` to `pcap_if_t*` to avoid exposing `pcap_if_t*` in the .h file
        return new PcapLiveDevice(iface, true, true, true)
    }
}

class WinPcapLiveDevice : public PcapLiveDevice
{
protected:
    WinPcapLiveDevice* cloneInternal(void* iface) override
    {
        // need to case from `void*` to `pcap_if_t*` to avoid exposing `pcap_if_t*` in the .h file
        return new WinPcapLiveDevice(iface, true, true, true)
    }
}

```

Dimi1010: Sure, can do that. Btw, `pcap_if_t` is already forward declared in PcapLiveDevice.h anyway, so no need for the void ptr shenanigans.

seladb: oh, you're right! 👍

Dimi1010: Implemented 0a037af

Pcap++/header/WinPcapLiveDevice.h
```diff
@@ -55,6 +55,9 @@ namespace pcpp
 		 * setMinAmountOfDataToCopyFromKernelToApplication())
 		 */
 		int getMinAmountOfDataToCopyFromKernelToApplication() const { return m_MinAmountOfDataToCopyFromKernelToApplication; }
+
+	protected:
+		PcapLiveDevice* cloneInternal(pcap_if_t& devInterface) const override;
```
seladb, line 60: nit: I think the signature can change a bit to return`WinPcapLiveDevice*`:
```cpp
WinPcapLiveDevice* cloneInternal(pcap_if_t& devInterface) const override;
```

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -266,8 +266,26 @@ PTF_TEST_CASE(TestPcapLiveDeviceClone)
...
+
+		liveDev = originalDev->clone();
+	}
+
 	PTF_ASSERT_NOT_NULL(liveDev);
+
+#ifdef _WIN32
+	// Tests if the clone is correctly returns a Windows Live Device on Windows systems.
+	PTF_ASSERT_NOT_NULL(dynamic_cast<pcpp::WinPcapLiveDevice*>(liveDev));
+#endif // _WIN32
```
seladb, line 287: Maybe move this to `TestWinPcapLiveDevice`? It's cleaner that way and we don't need the `#ifdef _WIN32`

Pcap++/header/WinPcapLiveDevice.h
Dimi1010: Implemented 62b8b12

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
seladb: I'm sorry, I didn't notice we have a dedicate test for live device clone: `TestPcapLiveDeviceClone`

seladb (approved):

PR #1408, opened 2024-05-25
Remove Ubuntu 18.04 and upgrade Fedora to 39 and RHEL to 9.4

clementperon:
Ubuntu 18.04 is EOL since April 2023.

Also it doesn't support recent NodeJS this imply we can't properly bump Github CI Actions.

Let's drop it and add support of Ubuntu 24.04 instead


tigercosmos (approved):

tigercosmos (changes requested):
Seems some CIs need to be fixed.

For Ubuntu 24, it's about PEP668: https://peps.python.org/pep-0668/#marking-an-interpreter-as-using-an-external-package-manager.

We can simply apply `--break-system-packages` for the pip.

.github/workflows/build_and_test.yml
```diff
@@ -99,14 +99,14 @@ jobs:
       - name: Test PcapPlusPlus
         run: |
           ${{ matrix.python }} -m pip install -U pip
```
egecetin, line 101: Error from this line. `pip` itself also should break system packages

.github/workflows/build_and_test.yml
```diff
@@ -99,14 +99,14 @@ jobs:
...
-          ${{ matrix.python }} -m pip install -r ci/run_tests/requirements.txt
+          ${{ matrix.python }} -m pip install --break-system-packages -r ci/run_tests/requirements.txt
           ${{ matrix.python }} ci/run_tests/run_tests.py --interface eth0 ${{ matrix.test-flags }}
 
       - name: Test Examples
         run: |
           cd Tests/ExamplesTest
           ${{ matrix.python }} -m pip install -U pip
```
egecetin, line 108: same

.github/workflows/build_and_test.yml
```diff
@@ -98,15 +98,15 @@ jobs:
 
       - name: Test PcapPlusPlus
         run: |
-          ${{ matrix.python }} -m pip install -U pip
-          ${{ matrix.python }} -m pip install -r ci/run_tests/requirements.txt
+          ${{ matrix.python }} -m pip install --break-system-packages -U pip
```
egecetin, line 101: What about using a new environment variable like `additional-flags`  to set empty or `--break-system-packages`. Changing these lines should fix maybe, what do you think?

```
${{ matrix.python }} -m pip install ${{ matrix.additional-python-flags }} -U pip
${{ matrix.python }} -m pip install ${{ matrix.additional-python-flags }} -r ci/run_tests/requirements.txt
```

tigercosmos: can't we just have an if-else statement?

egecetin: It is another option but adding an if else statement for step could lead more complex workflow file, no?

tigercosmos: Both work, so I am fine with either one.

clementperon: I have also bump Fedora so no more issue

egecetin: Now RedHat with same error

seladb: Sorry I'm late for the party... 🙂 
The most robust solution is to use `venv` (or `virtualenv`) as described here: https://stackoverflow.com/a/75696359
This is the standard way to install packages in python. I thought we could avoid it in our CI, but apparently, we can't. But it shouldn't be a big effort to add it (I can help if needed)

clementperon: Does it makes sense to use a venv in a container? We all the system will be clean and restart from scratch at next iteration ?

seladb: there shouldn't be any problem using `venv` in a docker container. `venv` is a local python environment that contains all the packages (and versions) the application needs. It doesn't override any OS settings or anything like that.

`venv` is often used in production python applications (which usually run inside docker / kubernetes) to ensure the application has the exact package versions it needs to run properly. This is very important in python because there is no compilation process

clementperon: I know what venv is, my point is do we mind to take effort to support venv in a CI.
Where we are already in an isolated environement.

we are trying to isolate python packages, in a isolated container environement

seladb: That's the reason we didn't use `venv` until now 🙂 
But if there are conflicts with the system Python packages, using `venv` should solve it. 

Or we can use the `--break-system-packages` flag in all envs, I wouldn't use it selectively on specific envs

tigercosmos: I think using `--break-system-packages` is fine enough unless someone want to put effort to change to venv, which can do it in another PR.

seladb: I just noticed `pip` on RHEL doesn't have the `--break-system-packages` flag, probably because the `pip` version is too old. So I guess we can either add this flag selectively or move to `venv`...

seladb: @clementperon do you think switching to [RHEL licensing manager](https://github.com/seladb/PcapPlusPlus-DockerImages/pull/40) should solve it? 🤔

clementperon: I have bump Pip in the same MR

seladb:
@clementperon now that we updated RHEL, maybe we can fix this PR and merge it?

.github/workflows/package.yml
```diff
@@ -26,18 +26,18 @@ jobs:
     strategy:
       matrix:
         include:
+          - image: ubuntu2404
+            config-zstd: OFF
```
clementperon, line 30: @seladb you should also set it here

```suggestion
            config-zstd: OFF
            additional-pip-flags: --break-system-packages
```

seladb (approved):

PR #1409, opened 2024-05-25
Asn1 fixes demo0525

tigercosmos:


Packet++/src/Asn1Codec.cpp
```diff
@@ -203,16 +203,16 @@ namespace pcpp
 
 		if (!lazy)
 		{
+			 // not lazy going here
 			try
 			{
-				decodedRecord->decodeValue(const_cast<uint8_t*>(data) + tagLen + lengthLen, lazy);
```
tigercosmos, line 208: the `lazy` here is always `false` anyway

Packet++/src/Asn1Codec.cpp
```diff
@@ -404,7 +404,7 @@ namespace pcpp
 	{
 		if (m_EncodedValue != nullptr)
 		{
-			decodeValue(m_EncodedValue, true);
```
tigercosmos, line 407: the `true` here doesn't make any effect, since all `decodeValue`s don't use `lazy` argument.

PR #1410, opened 2024-05-26
Refactor of RpcapdServerInitializer

Dimi1010:
This PR refactors the `RpcapdServerInitializer` object and contains the following changes:
- Fixed leak of sub-process thread handle by the initializer.
- Fixed sub-process not terminating if the main process is killed without calling object destructors.
- `getHandle` method marked as `const`.
- Explicitly deleted copy constructor and copy assignment operator for `RpcapdServerInitializer`.
NB: Move constructor can be added at a later date if needed.
- Removed the usages of `ostringstream` in favor of straight `string` allocation with `reserve` + `append`.
- Added assert to remote device test that the device pointer is not null.

### DIFF
--- a/Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
+++ b/Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
@@ -10,7 +10,6 @@
 #include "../Common/GlobalTestArgs.h"
 #include "../Common/TestUtils.h"
 #include "../Common/PcapFileNamesDef.h"
-#include <sstream>
 #if defined(_WIN32)
 #include "PcapRemoteDevice.h"
 #include "PcapRemoteDeviceList.h"
--- a/Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
+++ b/Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
@@ -163,8 +163,8 @@ public:
 		ZeroMemory( &pi, sizeof(pi) );
 		if (!CreateProcess (
 				TEXT(cmd.c_str()),
-				// CreateProcessW (Unicode version) modifies the argument string inplace during internal processing. 
-				// This can potentially cause access violation, if used with a compile time constant string, 
+				// CreateProcessW (Unicode version) modifies the argument string inplace during internal processing.
+				// This can potentially cause access violation, if used with a compile time constant string,
 				// but since the current usage uses a heap allocated string, it should be fine.
 				const_cast<char*>(TEXT(args.c_str())),
 				nullptr, nullptr, false,
--- a/Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
+++ b/Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
@@ -187,7 +187,22 @@ public:
 	}
 
 	RpcapdServerInitializer(const RpcapdServerInitializer&) = delete;
+	RpcapdServerInitializer(RpcapdServerInitializer&& other) noexcept 
+		: m_ProcessHandle(other.m_ProcessHandle), m_JobHandle(other.m_JobHandle)
+	{
+		other.m_ProcessHandle = nullptr;
+		other.m_JobHandle = nullptr;
+	}
 	RpcapdServerInitializer& operator=(const RpcapdServerInitializer&) = delete;
+	RpcapdServerInitializer& operator=(RpcapdServerInitializer&& other) noexcept
+	{
+		killProcessAndCloseHandles();
+		m_ProcessHandle = other.m_ProcessHandle;
+		m_JobHandle = other.m_JobHandle;
+		other.m_ProcessHandle = nullptr;
+		other.m_JobHandle = nullptr;
+		return *this;
+	}
 
 	~RpcapdServerInitializer() { killProcessAndCloseHandles(); }
 
--- a/Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
+++ b/Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
@@ -97,17 +97,64 @@ class RpcapdServerInitializer
 {
 private:
 	HANDLE m_ProcessHandle;
+	HANDLE m_JobHandle;
 
+	void killProcessAndCloseHandles()
+	{
+		if (m_ProcessHandle != nullptr)
+		{
+			TerminateProcess(m_ProcessHandle, 0);
+			CloseHandle(m_ProcessHandle);
+			m_ProcessHandle = nullptr;
+		}
+
+		if (m_JobHandle != nullptr)
+	{
+			CloseHandle(m_JobHandle);
+			m_JobHandle = nullptr;
+		}
+	}
 public:
-
-	RpcapdServerInitializer(bool activateRemoteDevice, const std::string &ip, uint16_t port) : m_ProcessHandle(nullptr)
+	RpcapdServerInitializer(bool activateRemoteDevice, const std::string& ip, uint16_t port)
+		: m_ProcessHandle(nullptr), m_JobHandle(nullptr)
 	{
 		if (!activateRemoteDevice)
 			return;
 
 		std::string cmd = "rpcapd\\rpcapd.exe";
-		std::ostringstream args;
-		args << "rpcapd\\rpcapd.exe -b " << ip << " -p " << port << " -n";
+		std::string args;
+		// Reserves the size of the arguments string in 1 allocation.
+		// Sizeof C-string includes the NULL terminator in the size.
+		// For the purposes of this calculation the NULL terminator's inclusion in the size is used to simulate the space delimiter after the argument.
+		args.reserve(
+			sizeof("-b") + ip.size() + 
+			sizeof(" -p") + 5 /* The maximum digits a uint16_t can have is 5 */ + 
+			sizeof(" -n") - 1 /* Subtracts one as the last NULL terminator is not needed */
+		);
+		args += "-b ";
+		args += ip;
+		args += " -p";
+		args += std::to_string(port);
+		args += " -n";
+
+		m_JobHandle = CreateJobObject(nullptr, nullptr);
+		if (m_JobHandle == nullptr)
+		{
+			throw std::runtime_error("Failed to create a job object with error code: " + std::to_string(GetLastError()));
+		}
+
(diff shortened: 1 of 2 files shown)


Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -97,54 +96,106 @@ class RpcapdServerInitializer
...
+	RpcapdServerInitializer(bool activateRemoteDevice, const std::string& ip, uint16_t port)
+		: m_ProcessHandle(nullptr), m_JobHandle(nullptr)
 	{
 		if (!activateRemoteDevice)
 			return;
 
 		std::string cmd = "rpcapd\\rpcapd.exe";
-		std::ostringstream args;
-		args << "rpcapd\\rpcapd.exe -b " << ip << " -p " << port << " -n";
+		std::string args;
```
tigercosmos, line 124: As you already count the size of usage. recommend to use `std::snprintf`

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -97,54 +96,106 @@ class RpcapdServerInitializer
...
-			TerminateProcess(m_ProcessHandle, 0);
-			CloseHandle(m_ProcessHandle);
+			DWORD errCode = GetLastError();
+			killProcessAndCloseHandles();
+			throw std::runtime_error("Failed assigning process to job object with code: " + std::to_string(errCode));
 		}
 	}
 
-	HANDLE getHandle() { return m_ProcessHandle; }
+	RpcapdServerInitializer(const RpcapdServerInitializer&) = delete;
```
tigercosmos, line 193: also `RpcapdServerInitializer(RpcapdServerInitializer&&) = delete;`?

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -97,54 +96,106 @@ class RpcapdServerInitializer
...
+		m_JobHandle = CreateJobObject(nullptr, nullptr);
+		if (m_JobHandle == nullptr)
+		{
+			throw std::runtime_error("Failed to create a job object with error code: " + std::to_string(GetLastError()));
+		}
+
+		// Sets up the job limits so closing the job will automatically kill all processes assigned to the job.
+		// This will prevent the subprocess continuing to live if the current process is killed without unwinding the stack (i.e. std::terminate is called),
+		// as the OS itself will kill the subprocesses when the last job handle is closed.
+		JOBOBJECT_EXTENDED_LIMIT_INFORMATION jobLimitInfo;
```
tigercosmos, line 148: I am sure the following Unix works, but not sure for Windows:
`JOBOBJECT_EXTENDED_LIMIT_INFORMATION jobLimitInfo = {}`?

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -97,54 +96,106 @@ class RpcapdServerInitializer
...
 
 		std::string cmd = "rpcapd\\rpcapd.exe";
-		std::ostringstream args;
-		args << "rpcapd\\rpcapd.exe -b " << ip << " -p " << port << " -n";
+		std::string args;
+		// Reserves the size of the arguments string in 1 allocation.
+		// Sizeof C-string includes the NULL terminator in the size.
+		// For the purposes of this calculation the NULL terminator's inclusion in the size is used to simulate the space delimiter after the argument.
+		args.reserve(
+			sizeof("rpcapd\\rpcapd.exe -b") + ip.size() +
```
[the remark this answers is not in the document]
Dimi1010: `sizeof` is evaluated at compile time, so it is compiled down to a bunch of integer arithmetic.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
Dimi1010: What is the benefit of `std::snprintf`?

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
Dimi1010: Move ctor and assignment operator are not implicitly declared if there are user defined copy ctor, copy assignment operator or destructor, so there is no need to declare them as deleted explicitly.

Dimi1010: Also, the object can be safely moveable, just not copiable.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
Dimi1010: Implemented 334a4d0

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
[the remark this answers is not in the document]
tigercosmos: we can simplify the code as:
```
    // Define a buffer large enough to hold the formatted string
    char buffer[256];

    // Use snprintf to format the string
    std::snprintf(buffer, sizeof(buffer), "rpcapd\\rpcapd.exe -b %s -p %d -n", ip.c_str(), port);
```

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: I didn't see any place that uses the move constructor, though.

In addition, it's recommended to follow the rule of 5: if we already spell out the copy constructor, it would be better to spell out the move constructor.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
tigercosmos: we can simplify the code as:
```
    // Define a buffer large enough to hold the formatted string
    char buffer[256];

    // Use snprintf to format the string
    std::snprintf(buffer, sizeof(buffer), "rpcapd\\rpcapd.exe -b %s -p %d -n", ip.c_str(), port);
```

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
Dimi1010: Sure, will add the move ctor. Should be simple enough.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
Dimi1010: well, writing it out, it ended up closer to this.
```
std::array<char, 256> args;
int res = std::snprintf(args.data(), args.size(), "rpcapd\\rpcapd.exe -b %s -p %d -n", ip.c_str(), port);
if (res < 0)
{
	throw std::runtime_error("Error during string formatting");
}
else if (res >= args.size())
{
	// We should never get here with reasonable input but you never know.
	throw std::runtime_error("Buffer overflow. Are you sure the IP is a valid string?");
}
```

Dimi1010: Implemented a8316a9

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
Dimi1010: Fixed 60233e2

tigercosmos (approved):

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -97,54 +98,115 @@ class RpcapdServerInitializer
...
+		}
+	}
+public:
+	RpcapdServerInitializer(bool activateRemoteDevice, const std::string& ip, uint16_t port)
+		: m_ProcessHandle(nullptr), m_JobHandle(nullptr)
 	{
 		if (!activateRemoteDevice)
 			return;
 
 		std::string cmd = "rpcapd\\rpcapd.exe";
```
tigercosmos, line 125: we can move this line to around L163 (new)

Dimi1010: True, tho IMO I prefer the variables being initialized in the sequence they are used in the `CreateProcess` call.

tigercosmos:
@seladb I think this is ready for review.

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
```diff
@@ -97,54 +98,115 @@ class RpcapdServerInitializer
...
+	RpcapdServerInitializer& operator=(const RpcapdServerInitializer&) = delete;
+	RpcapdServerInitializer& operator=(RpcapdServerInitializer&& other) noexcept
+	{
+		killProcessAndCloseHandles();
+		m_ProcessHandle = other.m_ProcessHandle;
+		m_JobHandle = other.m_JobHandle;
+		other.m_ProcessHandle = nullptr;
+		other.m_JobHandle = nullptr;
+		return *this;
+	}
```
seladb, line 205: Do we really need all of these methods? We don't the copy c'tor or the assignment operator anywhere

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
seladb, line 125: nit: I don't remember why I defined this string twice (here and in line 127). Maybe we can use `cmd` when we write `args`?

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
Dimi1010: Well, the class is not safe to copy, but since it only holds POD members, a copy ctor will be implicitly declared by the compiler if not explicitly deleted. As for the move ctor, @tigercosmos wanted them added as deleted too, but since the class is technically movable, I decided to write them out instead of declaring them deleted.

seladb: ok, I have no specific objection. It's a helper class in our test suite so it doesn't have to be perfect 🙂

Tests/Pcap++Test/Tests/LiveDeviceTests.cpp
Dimi1010: Added 519b35a

seladb (approved):

PR #1412, opened 2024-05-28
Add LDAP layer

seladb:
This PR adds a layer to parse/craft the LDAP protocol (Lightweight Directory Access Protocol).
You can read about it here:
- https://ldap.com/
- https://en.wikipedia.org/wiki/Lightweight_Directory_Access_Protocol

This protocol uses ASN.1 BER to encode the data, so it heavily uses the `Asn1Codec`.

This is just the first PR out of a series of PRs I plan to open. This PR adds an `LdapLayer` with parsing and crafting of generic LDAP messages. Later on, I'll add PRs to parse specific LDAP messages.


.github/workflows/build_and_test.yml
```diff
@@ -490,7 +495,7 @@ jobs:
           - os: windows-2022
             platform: "Visual Studio 17 2022"
             arch: "x64"
-            pcap_lib: "npcap"
+            pcap_lib: "winpcap"
```
egecetin, line 498: This is duplicate, maybe one of them is `Win32`

https://github.com/seladb/PcapPlusPlus/blob/1fd0df5510d237fdb7db639e2559ac2cdb9b87e6/.github/workflows/build_and_test.yml#L491-L498

seladb: it's all just temporary because the Npcap web-site and its resources are all down for at least a day or two 😕 
Is https://npcap.com/ working for you? 🤔 
I wanted to test that my code works, and it's not related to WinPcap/Npcap in any way, so I figured I'll use WinPcap until Npcap is back...

egecetin: I understand that this is a temporary change but original matrix include 1 Win32 job and 3 x64 job. I mean since we have two options (npcap and winpcap), two versions (2019 and 2022) and two archs (win32 and x64) there should be more jobs 😁

egecetin: More clearly, did we forget to add or we tried to decrease testing time/workflow count?

seladb: we can have more options, but I think the most common options today are the combination of x64, Npcap and vs2019/2022. Win32 and WinPcap are mostly deprecated I believe

seladb:
The CI is currently failing because Npcap is down for some reason 😕 
But the code itself doesn't rely on it, so it's ready for review.

As a temp solution, I updated VS build to use WinPcap (couldn't do it for MinGW for some reason, but not sure it's worth the effort). I'll revert it once Npcap is back

Packet++/header/LdapLayer.h
```diff
@@ -0,0 +1,267 @@
...
+			try
+			{
+				result = (thisPtr->*method)();
+				return true;
+			}
+			catch (...)
+			{
+				return false;
+			}
+		}
```
Dimi1010, line 258: This function can be made `static`. Also why can't this use the implicit `this` pointer?

Packet++/src/LdapLayer.cpp
```diff
@@ -0,0 +1,210 @@
...
+		{static_cast<uint8_t>(LdapOperationType::ExtendedResponse), LdapOperationType::ExtendedResponse},
+		{static_cast<uint8_t>(LdapOperationType::IntermediateResponse), LdapOperationType::IntermediateResponse}
+	};
+
+	std::string LdapOperationType::toString() const
+	{
+		return LdapOperationTypeToString.at(m_Value);
+	}
+
+	LdapOperationType LdapOperationType::fromIntValue(uint8_t value)
```
tigercosmos, line 62: should call `fromUintValue`? 

also `IntToLdapOperationType` should be `UintToLdapOperationType`?

Packet++/src/LdapLayer.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+	uint16_t LdapLayer::getMessageID() const
+	{
+		return getRootAsn1Record()->getSubRecords().at(0)->castAs<Asn1IntegerRecord>()->getValue();
+	}
+
+	std::vector<LdapControl> LdapLayer::getControls() const
+	{
+		std::vector<LdapControl> controls;
+		if (getRootAsn1Record()->getSubRecords().size() < 3)
```
tigercosmos, line 172: is 3 a magic number? can we have a comment on it?

tigercosmos: also for the following number `2`, and `1`.

Tests/Packet++Test/PacketExamples/ldap.pcapng
egecetin, line 1: Not critical but pcap file has 21 packets but the tests only use four of them (Four `dat` files). Maybe remove unused packets from the pcap file 🤔 or am i missing something 🤔

seladb: Yes, as I mentioned in the [PR description](https://github.com/seladb/PcapPlusPlus/pull/1412#issue-2319983172) this is just the first PR out of a series of PRs I plan to open with different LDAP messages

Common++/header/GeneralUtils.h
```diff
@@ -64,4 +64,17 @@ namespace pcpp
 		int mask = alignment - 1;
 		return (number + mask) & ~mask;
 	}
+
+	/**
+	 * A template class to calculate enum class hash
+	 * @tparam EnumClass
+	 */
+	template<typename EnumClass>
```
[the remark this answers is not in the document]
seladb: Nice!! Modified here: https://github.com/seladb/PcapPlusPlus/pull/1412/commits/f111859e07461404aaaf75bf4c47dcbdde952f02

Common++/header/GeneralUtils.h
```diff
@@ -64,4 +64,17 @@ namespace pcpp
...
+	struct EnumClassHash
+	{
+		size_t operator()(const EnumClass& value) const
+		{
+			return static_cast<int>(value);
```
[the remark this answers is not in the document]
seladb: Nice!! Modified here: https://github.com/seladb/PcapPlusPlus/pull/1412/commits/f111859e07461404aaaf75bf4c47dcbdde952f02

Packet++/src/LdapLayer.cpp
seladb: Updated in this commit: https://github.com/seladb/PcapPlusPlus/pull/1412/commits/ff0bcc1a6f6fc74d75a3e99a72392e3934c768de

Packet++/src/LdapLayer.cpp
seladb: This is the LDAP specification that you can find here: https://github.com/vlm/ldap-server-example/blob/master/Lightweight-Directory-Access-Protocol-V3.asn1

```
        LDAPMessage ::= SEQUENCE {
             messageID       MessageID,
             protocolOp      CHOICE {
                  bindRequest           BindRequest,
                  bindResponse          BindResponse,
                  unbindRequest         UnbindRequest,
                  searchRequest         SearchRequest,
                  searchResEntry        SearchResultEntry,
                  searchResDone         SearchResultDone,
                  searchResRef          SearchResultReference,
                  modifyRequest         ModifyRequest,
                  modifyResponse        ModifyResponse,
                  addRequest            AddRequest,
                  addResponse           AddResponse,
                  delRequest            DelRequest,
                  delResponse           DelResponse,
                  modDNRequest          ModifyDNRequest,
                  modDNResponse         ModifyDNResponse,
                  compareRequest        CompareRequest,
                  compareResponse       CompareResponse,
                  abandonRequest        AbandonRequest,
                  extendedReq           ExtendedRequest,
                  extendedResp          ExtendedResponse,
                  ...,
                  intermediateResponse  IntermediateResponse },
             controls       [0] Controls OPTIONAL }
```

- Message ID is a mandatory first parameter
- LDAP operation type is a mandatory second parameter
- Controls is an optional third parameter

Packet++/header/LdapLayer.h
seladb: I'd like to implement `tryGet` for `LdapLayer` and future layer classes that will derive from `LdapLayer`. I tried different options and this is the only way I could make it work. Can you suggest a better way?

You can find my draft branch with implementation of more LDAP message types here: https://github.com/seladb/PcapPlusPlus/tree/ldap

Packet++/src/LdapLayer.cpp
tigercosmos: I would suggest using a meaningful `constexpr` variable for better understanding the code.
e.g. `constexpr int messageIDindex = 1`

seladb: That's a good idea! Added in this commit: https://github.com/seladb/PcapPlusPlus/pull/1412/commits/7a7480f4f49dedb4e3361d45015e6dd71368d2b1

Packet++/header/LdapLayer.h
Dimi1010: Can't really think of any other ways off the top of my head, so I guess it is fine.

One suggestion for better readability to replace 
`result = (thisPtr->*method)();` with `result = std::mem_fn(method)(thisPtr);`,
as I was mildly confused what was happening on that line when I read it the first time.
Requires `<functional>`.

seladb: sure, added in this commit: https://github.com/seladb/PcapPlusPlus/pull/1412/commits/4c075713ca22909d62b0cd54a88a239016d49139

Packet++/src/LdapLayer.cpp
```diff
@@ -0,0 +1,210 @@
...
+		{LdapOperationType::CompareResponse,       "CompareResponse"},
+		{LdapOperationType::AbandonRequest,        "AbandonRequest"},
+		{LdapOperationType::SearchResultReference, "SearchResultReference"},
+		{LdapOperationType::ExtendedRequest,       "ExtendedRequest"},
+		{LdapOperationType::ExtendedResponse,      "ExtendedResponse"},
+		{LdapOperationType::IntermediateResponse,  "IntermediateResponse"},
+		{LdapOperationType::Unknown,               "Unknown"}
+	};
+
+	const std::unordered_map<uint8_t, LdapOperationType> UintToLdapOperationType{
```
Dimi1010, line 34: nit: Maybe add a sanity check `static_assert` that `LdapOperationType::Value`'s backing field is not larger than `uint8_t`, since we are initialzing via `static_cast`s?

Packet++/src/LdapLayer.cpp
```diff
@@ -0,0 +1,210 @@
...
+	{
+		return LdapOperationTypeToString.at(m_Value);
+	}
+
+	LdapOperationType LdapOperationType::fromUintValue(uint8_t value)
+	{
+		if (UintToLdapOperationType.find(value) != UintToLdapOperationType.end())
+		{
+			return UintToLdapOperationType.at(value);
+		}
```
Dimi1010, line 67: `find` already returns an iterator to the found element. We can use that instead of searching the map twice.

Packet++/src/LdapLayer.cpp
seladb: `LdapOperationType::Value` is an enum of type `uint8_t`: https://github.com/seladb/PcapPlusPlus/blob/394de3ebac90d6a91674e7ecbe7f9e888129b3c7/Packet%2B%2B/header/LdapLayer.h#L27

So I think it's safe to assume all values are of type `uint8_t`. Besides, this is an internal map that doesn't rely on any user input so I'm not sure this check is necessary

Packet++/src/LdapLayer.cpp
seladb: Good catch! Fixed in commit https://github.com/seladb/PcapPlusPlus/pull/1412/commits/33c9dcbc0d84234e69baa23a8db370179a4ce602

Common++/header/GeneralUtils.h
```diff
@@ -64,4 +65,17 @@ namespace pcpp
...
 	}
+
+	/**
+	 * A template class to calculate enum class hash
+	 * @tparam EnumClass
+	 */
+	template<typename EnumClass, typename std::enable_if<std::is_enum<EnumClass>::value , bool>::type = false>
+	struct EnumClassHash
+	{
+		size_t operator()(const EnumClass& value) const
```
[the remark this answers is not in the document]
seladb: Yes, I agree. Fixed in commit https://github.com/seladb/PcapPlusPlus/pull/1412/commits/33c9dcbc0d84234e69baa23a8db370179a4ce602

Packet++/src/LdapLayer.cpp
tigercosmos: if it's only internal, probably the current implemetation is fine.

tigercosmos (approved):

Dimi1010 (approved):

PR #1424, opened 2024-06-02
Add LDAP search request and result entry layers

seladb:
This PR adds 2 additional LDAP layers:
- Search request operation: https://datatracker.ietf.org/doc/html/rfc4511#section-4.5.1
- Search result entry message: https://datatracker.ietf.org/doc/html/rfc4511#section-4.5.2


README.md
```diff
@@ -256,14 +256,15 @@ PcapPlusPlus currently supports parsing, editing and creation of packets of the
...
 40. HTTP headers (request & response)
-41. NTP (v3, v4)
-42. Radius
-43. S7 Communication (S7comm)
-44. SMTP
-45. SOME/IP
-46. SSH - parsing only (no editing capabilities)
-47. Telnet - parsing only (no editing capabilities)
-48. Generic payload
+41. LDAP
```
seladb, line 259: In the previous PR I forgot to add LDAP to the README file. Adding it now...

Packet++/header/LdapLayer.h
```diff
@@ -266,10 +288,295 @@ namespace pcpp
...
+		static constexpr int baseObjectIndex = 0;
+		static constexpr int scopeIndex = 1;
+		static constexpr int derefAliasIndex = 2;
+		static constexpr int sizeLimitIndex = 3;
+		static constexpr int timeLimitIndex = 4;
+		static constexpr int typesOnlyIndex = 5;
+		static constexpr int filterIndex = 6;
+		static constexpr int attributesIndex = 7;
+
+		LdapSearchRequestLayer(std::unique_ptr<Asn1Record>& asn1Record, uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
```
Dimi1010, line 510: Did not notice that `LdapLayer` had a similar constructor on the last PR, so this comment is regarding that class's ctor too.

Generally do **NOT** pass `unique_ptr` by non-`const` reference, pass it by value (and use moves if necessary). 

Passing it as a non-`const` reference introduces ambiguity, as it symbolizes that the smart pointer *might* be taken ownership of, which is not the case for most functions. 
**NB:** If the function actually does *conditionally* take ownership, then that is a valid use case for non-`const` reference.

If ownership is guaranteed to be transferred to the function, have the `unique_ptr` as a value parameter. It might require a `move` from the caller into the value, but it shows the caller that the object is not his concern anymore and the previous variable that held the pointer is guaranteed to be invalid after that. Both `unique_ptr` and `shared_ptr` are cheap to move.

If ownership transfer is not required, pass it as a `const` reference to the `unique_ptr`, if access to the smart pointer internals are required for some reason, or better yet, a raw pointer or reference to the object itself, as that will work irrespective of how the memory is being managed by the caller (smart pointers, new/delete, stack, etc...).

Packet++/header/LdapLayer.h
```diff
@@ -266,10 +288,295 @@ namespace pcpp
...
+		}
+	protected:
+		friend LdapLayer* LdapLayer::parseLdapMessage(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet);
+
+		static constexpr int objectNameIndex = 0;
+		static constexpr int attributesIndex = 1;
+		static constexpr int attributeTypeIndex = 0;
+		static constexpr int attributeValueIndex = 1;
+
+		LdapSearchResultEntryLayer(std::unique_ptr<Asn1Record>& asn1Record, uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
```
Dimi1010, line 559: Ditto, `unique ptr` reference.

Dimi1010 (changes requested):

Packet++/header/LdapLayer.h
seladb: That's a good point, thanks for noticing it! Fixed here: https://github.com/seladb/PcapPlusPlus/pull/1424/commits/17330d797dd5b398c82657b267b04fb032415f55

Packet++/header/LdapLayer.h
seladb: Good catch! Fixed here: https://github.com/seladb/PcapPlusPlus/pull/1424/commits/17330d797dd5b398c82657b267b04fb032415f55

Packet++/src/LdapLayer.cpp
```diff
@@ -139,9 +139,16 @@ namespace pcpp {
...
+					return new LdapSearchRequestLayer(asn1Record, data, dataLen, prevLayer, packet);
+				case LdapOperationType::SearchResultEntry:
+					return new LdapSearchResultEntryLayer(asn1Record, data, dataLen, prevLayer, packet);
+				case LdapOperationType::Unknown:
+					return nullptr;
+				default:
+					return new LdapLayer(asn1Record, data, dataLen, prevLayer, packet);
 			}
 
 			return nullptr;
```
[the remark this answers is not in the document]
seladb: Good catch! Fixed here: https://github.com/seladb/PcapPlusPlus/pull/1424/commits/17330d797dd5b398c82657b267b04fb032415f55

Dimi1010 (approved):
LGTM.

Issue #1347, opened 2024-03-22
Fix building warning due to LightPcapNg


tigercosmos:
build log
```
[  3%] Building C object 3rdParty/LightPcapNg/CMakeFiles/light_pcapng.dir/LightPcapNg/src/light_io.c.o
/tmp/cirrus-ci-build/3rdParty/LightPcapNg/LightPcapNg/src/light_io.c:91:20: warning: incompatible pointer types assigning to 'FILE *' (aka 'struct __sFILE *') from 'light_file' (aka 'struct light_file_t *') [-Wincompatible-pointer-types]
        pcapng->stream.fd = light_open(file_name, LIGHT_OREAD);
                          ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/cirrus-ci-build/3rdParty/LightPcapNg/LightPcapNg/src/light_io.c:117:17: warning: incompatible pointer types passing 'FILE *' (aka 'struct __sFILE *') to parameter of type 'light_file' (aka 'struct light_file_t *') [-Wincompatible-pointer-types]
        if (light_read(pcapng->stream.fd, &block_type, sizeof(block_type)) == -1 ||
                       ^~~~~~~~~~~~~~~~~
/tmp/cirrus-ci-build/3rdParty/LightPcapNg/LightPcapNg/include/light_platform.h:54:30: note: passing argument to parameter 'fd' here
size_t light_read(light_file fd, void *buf, size_t count);
                             ^
/tmp/cirrus-ci-build/3rdParty/LightPcapNg/LightPcapNg/src/light_io.c:118:15: warning: incompatible pointer types passing 'FILE *' (aka 'struct __sFILE *') to parameter of type 'light_file' (aka 'struct light_file_t *') [-Wincompatible-pointer-types]
                        light_read(pcapng->stream.fd, &block_total_length, sizeof(block_total_length)) == -1) {
                                   ^~~~~~~~~~~~~~~~~
/tmp/cirrus-ci-build/3rdParty/LightPcapNg/LightPcapNg/include/light_platform.h:54:30: note: passing argument to parameter 'fd' here
size_t light_read(light_file fd, void *buf, size_t count);
                             ^
[...]

PR #1425, opened 2024-06-03
github: ci: fix pipeline due to bump of GCC in mingw

clementperon:


tigercosmos (approved):

clementperon:
@tigercosmos I have changed, the issue is that we can't enforce GCC version with MSYS2, so instead I have disable the warning

clementperon:
Temporary fix for https://github.com/seladb/PcapPlusPlus/issues/1347

tigercosmos (approved):

seladb (approved):

PR #1426, opened 2024-06-03
Refactor of IP utils to use exceptions on invalid input.

Dimi1010:
- Refactors `sockaddr2in*` functions to throw exceptions when invalid `sockaddr` is provided.
- Adds `try_sockaddr2in*` functions that mirror the old functionality to return `nullptr` on failed extractions.
- Adds overload for `sockaddr2string` that takes only `sockaddr` as parameter and returns a `string`.

### DIFF
--- a/Common++/header/IpUtils.h
+++ b/Common++/header/IpUtils.h
@@ -67,12 +67,26 @@ namespace pcpp
 		in_addr* sockaddr2in_addr(struct sockaddr *sa);
 
 		/**
+		 * Attempt to extract IPv4 address from sockaddr.
+		 * @param[in] sa - input sockaddr
+		 * @return Pointer to address in in_addr format or nullptr if extraction fails.
+		 */
+		in_addr* try_sockaddr2in_addr(struct sockaddr* sa);
+
+		/**
 		 * Extract IPv6 address from sockaddr
 		 * @param[in] sa - input sockaddr
 		 * @return Address in in6_addr format
 		 * @throws std::invalid_argument Sockaddr family is not AF_INET6 or sockaddr is nullptr. 
 		 */
 		in6_addr* sockaddr2in6_addr(struct sockaddr *sa);
+
+		/**
+		 * Attempt to extract IPv6 address from sockaddr.
+		 * @param[in] sa - input sockaddr
+		 * @return Pointer to address in in6_addr format or nullptr if extraction fails.
+		 */
+		in6_addr* try_sockaddr2in6_addr(struct sockaddr* sa);
 
 		/**
 		 * Converts a sockaddr format address to its string representation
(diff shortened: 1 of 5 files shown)


tigercosmos (approved):

Common++/header/IpUtils.h
```diff
@@ -62,22 +63,48 @@ namespace pcpp
...
+		void sockaddr2string(sockaddr const* sa, char* resultString, size_t resultBufLen);
+
+		/**
+		 * Converts a spckaddr format address to its string representation
+		 * @param[in] sa Address in sockaddr format
+		 * @return String representation of the address
+		 * @throws std::invalid_argument Sockaddr family is not AF_INET or AF_INET6, or sockaddr is nullptr.
 		 */
-		void sockaddr2string(struct sockaddr *sa, char* resultString);
+		std::string sockaddr2string(sockaddr const* sa);
```
seladb, line 107: do we use this overload of `sockaddr2string` anywhere? 🤔

Common++/src/IpUtils.cpp
```diff
@@ -18,37 +19,117 @@ namespace pcpp
...
 			if (sa == nullptr)
+				throw std::invalid_argument("sockaddr is nullptr");
+
+			switch (sa->sa_family)
+			{
+			case AF_INET:
+				return &(reinterpret_cast<sockaddr_in*>(sa)->sin_addr);
+			default:
+				throw std::invalid_argument("sockaddr family is not AF_INET.");
+			}
```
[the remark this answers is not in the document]
Dimi1010: Eh, not really needed. I just thought it looked nicer when I wrote it. They both compile down to the same thing, so it could be either one.

Common++/header/IpUtils.h
Dimi1010: Atm, not really, but I thought it was a good overload to have, in case of just needing a `string`.

seladb: These are internal APIs that are not used a lot. I'd suggest reducing it to only what we need

Common++/src/IpUtils.cpp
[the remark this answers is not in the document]
seladb: I think using an `if` instead of `switch` is nicer in this case 🤔

Common++/header/IpUtils.h
Dimi1010: Removed 202fa55

Common++/src/IpUtils.cpp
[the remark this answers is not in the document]
Dimi1010: Sure. Fixed 40a8360

Common++/header/IpUtils.h
Dimi1010: Fixed 202fa55 and 518b837.

Accidentally removed the wrong function from the header in the first commit, whops.

seladb (approved):

Common++/header/IpUtils.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <stdint.h>
+#include <string>
```
Dimi1010, line 4: Unused include.
```suggestion
```

PR #1428, opened 2024-06-03
Cpp casts and const changes.

Dimi1010:
Part of #977 

- C cast -> Cpp casts
- Some `const` qualifiers added.
- Fixed uninitialized member variable.

### DIFF
--- a/Packet++/src/IPv6Layer.cpp
+++ b/Packet++/src/IPv6Layer.cpp
@@ -187,7 +187,7 @@ void IPv6Layer::removeAllExtensions()
 
 bool IPv6Layer::isFragment() const
 {
-	IPv6FragmentationHeader* fragHdr = getExtensionOfType<IPv6FragmentationHeader>();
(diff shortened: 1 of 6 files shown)


Packet++/src/IPv6Layer.cpp
```diff
@@ -187,7 +187,7 @@ void IPv6Layer::removeAllExtensions()
 
 bool IPv6Layer::isFragment() const
 {
-	IPv6FragmentationHeader* fragHdr = getExtensionOfType<IPv6FragmentationHeader>();
+	IPv6FragmentationHeader const* fragHdr = getExtensionOfType<IPv6FragmentationHeader>();
```
tigercosmos, line 190: why not `return getExtensionOfType<IPv6FragmentationHeader>() != nullptr`

tigercosmos (approved):

Packet++/src/IPv6Layer.cpp
Dimi1010: Good idea. Added 94f6a17

seladb (approved):

Issue #1431, opened 2024-06-05
General refactor of PcapLiveDeviceList and PcapRemoteDeviceList.


Dimi1010:
This is an issue to track progress that will be done over several PRs.

Task overview:
1. - [x] Refactor `PcapLiveDeviceList` to use smart pointers internally for memory management.
2. - [ ] Refactor `PcapRemoteDeviceList` to use smart pointers internally for memory management.
(#1488)
4. - [x] Refactor `PcapRemoteDevice` to keep a shared reference of `RemoteAuthentication` with `PcapRemoteDeviceList` via `std::shared_ptr` as it currently uses a raw pointer to share authentication with the device list.
5. - [ ] ~~Unify the public "getDevice" API of both lists using smart pointers.~~
6. - [ ] Unify the public API for iteration and fetching all devices. 
(#1488)
7. - [x] Refactor `PcapRemoteDeviceList`'s factories to return `unique_ptr` instead of raw pointers.
8. - [x] Centralize code calling Libpcap/Winpcap/Npcap to fetch devices into a utility header. 

### 1. Use smart pointers for internal memory management in `PcapLiveDeviceList`
The current implementation uses a vector of raw pointers and manually handles `new`/`delete` calls. The move towards smart pointers will automate the need for manual memory management.

### 2. Use smart pointers for internal memory management in `PcapRemoteDeviceList`
The current implementation uses a vector of raw pointers and manually handles `new`/`delete` calls. The move towards smart pointers will automate the need for manual memory management.

### 3. Use `shared_ptr` to store `RemoteAuthentication` in `PcapRemoteDeviceList` and `PcapRemoteDevice`.
The current implementation uses a raw pointer to `RemoteAuthentication` shared between `PcapRemoteDeviceList` and its `PcapRemoteDevice`s. The move towards smart pointers will relieve the need for `PcapRemoteDeviceList` to need to manage the lifetime of the authentication object manually, and will prevent potentially deallocating it while another `PcapRemoteDevice` is using it.

### 4. ~~Unifying the public API for fetching single devices.~~
Currently the public APIs of `PcapLiveDeviceList` and `PcapRemoteDeviceList` use similar but not entirely the same syntax for getter functions.
- `PcapLiveDeviceList` uses `getPcapLiveDeviceBy*` returning a raw pointer to `PcapLiveDevice`
[...]

PR #1434, opened 2024-06-05
Centralize code for fetching pcap devices.

### LINKED ISSUE #1433: Centralize code calling Libpcap/Winpcap/Npcap to fetch devices into a utility header.

Dimi1010:
Part of #1431. Implements point 7.


Pcap++/src/PcapLiveDeviceList.cpp
```diff
@@ -39,32 +41,30 @@ PcapLiveDeviceList::~PcapLiveDeviceList()
...
-	if (err < 0)
+	std::unique_ptr<pcap_if_t, internal::PcapFreeAllDevsDeleter> interfaceList;
+	try
+	{
+		interfaceList = internal::getAllLocalPcapDevices();
+	}
+	catch (const std::exception& e)
 	{
-		PCPP_LOG_ERROR("Error searching for devices: " << errbuf);
+		PCPP_LOG_ERROR(e.what());
```
tigercosmos, line 51: recommend to have a meaningful error message here.

Pcap++/src/PcapLiveDeviceList.cpp
```diff
@@ -39,32 +41,30 @@ PcapLiveDeviceList::~PcapLiveDeviceList()
...
 	}
 
 	PCPP_LOG_DEBUG("Pcap lib version info: " << IPcapDevice::getPcapLibVersionInfo());
 
-	pcap_if_t* currInterface = interfaceList;
-	while (currInterface != nullptr)
+
+	for (pcap_if_t* currInterface = interfaceList.get(); currInterface != nullptr; currInterface = currInterface->next)
```
tigercosmos, line 57: nice catch

Pcap++/src/PcapRemoteDeviceList.cpp
```diff
@@ -20,51 +22,39 @@ PcapRemoteDeviceList* PcapRemoteDeviceList::getRemoteDeviceList(const IPAddress&
...
+	std::unique_ptr<pcap_if_t, internal::PcapFreeAllDevsDeleter> interfaceList;
+	try
 	{
-		PCPP_LOG_ERROR("Error retrieving device on remote machine. Error string is: " << errorBuf);
-		return NULL;
+		interfaceList = internal::getAllRemotePcapDevices(ipAddress, port, pRmAuth);
+	}
+	catch (const std::exception& e)
+	{
+		PCPP_LOG_ERROR(e.what());
```
tigercosmos, line 41: same here

tigercosmos (approved):

Pcap++/src/PcapLiveDeviceList.cpp
Dimi1010: The error messages are actually the same as before. It is just that the DeviceUtil functions have the error messages thrown in a `runtime_error` as the exception reason and this code catches that and logs the message in the exception.

Pcap++/CMakeLists.txt
```diff
@@ -24,6 +26,8 @@ add_library(
 
 set(public_headers
     header/Device.h
+    header/DeviceUtils.h
+    header/MemoryUtils.h
```
seladb, line 30: These 2 new files contain internal utils, so they shouldn't be exposed as part of the API...

Pcap++/header/DeviceUtils.h
```diff
@@ -0,0 +1,34 @@
...
+		/**
+		 * Fetches a list of all network devices on a remote machine that WinPcap/NPcap can find.
+		 * @param[in] ipAddress IP address of the remote machine.
+		 * @param[in] port Port to use when connecting to the remote machine.
+		 * @param[in] pRmAuth Pointer to an authentication structure to use when connecting to the remote machine. Nullptr if no authentication is required.
+		 * @return A smart pointer to an interface list structure.
+		 * @throws std::runtime_error The system encountered an error fetching the devices.
+		 */
+		std::unique_ptr<pcap_if_t, PcapFreeAllDevsDeleter> getAllRemotePcapDevices(const IPAddress& ipAddress, uint16_t port, pcap_rmtauth* pRmAuth = nullptr);
+#endif // _WIN32
```
seladb, line 32: This is very specific to remote devices, why do we need to extract it to a different header file? 🤔

Pcap++/header/MemoryUtils.h
seladb, line 1: Maybe call this file `PcapUtils.h`? It only contains structs for freeing pcap devices.

By the way, the `PcapCloseDeleter` can also be used in `PcapFileDevice.cpp` and `PcapFilter.cpp` (maybe in subsequent PRs?)

Pcap++/header/DeviceUtils.h
Dimi1010: Mostly so all the pcap fetch devices calls are in the same place.

Pcap++/header/MemoryUtils.h
Dimi1010: It is supposed to contain utilities for smart pointer, not pcap deleters specifically.
See this: https://github.com/seladb/PcapPlusPlus/pull/1386#discussion_r1623892654

seladb: I'm curious about which other utilities you plan to have in this file? 🤔

Pcap++/CMakeLists.txt
Dimi1010: Where do you want me to put them then?

Pcap++/header/DeviceUtils.h
seladb: to be honest, remote pcap devices are quite a small niche and I don't think it's used a lot. I'd recommend keeping the code inside `PcapRemoteDevice`, and keep these util files mostly focused on common code that is used in several places

Pcap++/CMakeLists.txt
seladb: I'm not a CMake expert, but I think there is a way to decide which header files are "public" (meaning exposed as part of PcapPlusPlus API) and which are "private". This may also affect our doxygen API documentation: https://pcapplusplus.github.io/api-docs/next/ (which currently looks at all the header files in `Pcap++/header` 😕 )

Maybe @clementperon can help here?

Pcap++/header/MemoryUtils.h
Dimi1010: Hmm, I was planning it to be a general deleters header, but I guess if we want to separate the deleters by use case we can rename it to `PcapUtils.h`. 

What is your opinion on merging `DeviceUitls.h` into the renamed header as @tigercosmos originally proposed?

Pcap++/header/DeviceUtils.h
Dimi1010: IDK, I find it better organized to have all the code that "gets pcap to fetch a list of X devices" in one place, even if this specific function that is not used in a lot of places.

@tigercosmos Any thoughts on the matter?

Pcap++/header/MemoryUtils.h
Dimi1010: Renamed 0d91ddf

Pcap++/header/DeviceUtils.h
tigercosmos: I agree with @seladb , unless this function is used somewhere else except for `PcapRemoteDevice` in our codebase.

Dimi1010: Ok, moved the function to an anonymous namespace in `PcapRemoteDeviceList` as it is only used there.

Although I think I spotted an issue with `PcapRemoteDevice`'s `clone` method. It reuses the the `PcapLiveDevice`'s clone which fetches `getAllLocalDevices`. Btw, why is clone reusing re-fetching the devices instead of just copying the device data?

Pcap++/header/MemoryUtils.h
Dimi1010: > By the way, the `PcapCloseDeleter` can also be used in `PcapFileDevice.cpp` and `PcapFilter.cpp` (maybe in subsequent PRs?)

Replaced the locally defined `PcapTDeleter` with `PcapUtils.h`'s `PcapCloseDeleter` in `PcapFilter.h` as the `TDeleter` was the initial prototype class that got renamed into the `CloseDeleter`.

Pcap++/CMakeLists.txt
Dimi1010: Well, removed the headers from the public header list. (fb685bf)

Things appear to be building fine, and the only difference is that the headers are missing from the include folder of the install command output.

Don't know if the docs generation has changed anything. The CI action still shows that the headers are being parsed.

seladb: I think that's ok. For doxygen we probably need to exclude these files in [`Doxyfile`](https://github.com/PcapPlusPlus/pcapplusplus.github.io/blob/master/static/api-docs/next/doxygen/Doxyfile) and [`Doxyfile-ci`](https://github.com/PcapPlusPlus/pcapplusplus.github.io/blob/master/static/api-docs/next/doxygen/Doxyfile-ci) which are in the [web-site repo](https://github.com/PcapPlusPlus/pcapplusplus.github.io/). Here is how to exclude files:
https://stackoverflow.com/questions/34776315/doxygen-special-command-to-ignore-a-file

Do you want to open a PR for it or prefer I do it?

Pcap++/header/DeviceUtils.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+/// @file
+
+#include <memory>
+#include "IpAddress.h"
+#include "PcapUtils.h"
+
+// Forward declaration
+struct pcap_rmtauth;
```
seladb, line 10: This is probably not needed anymore?

seladb (approved):
See one comment, otherwise LGTM

Pcap++/CMakeLists.txt
Dimi1010: Hmm, maybe we can mark the documentation as internal for doxygen in the code?

Something like this: https://stackoverflow.com/a/18443973/11922936
That way the docs won't be generated for the public documentation but it will retain the option for generating internal documentation if we want one in the future.

tigercosmos (approved):

Pcap++/header/DeviceUtils.h
Dimi1010: Removed. f343771

Pcap++/CMakeLists.txt
Dimi1010: Enclosed the internal docs in a conditional section (617273e). I think this might remove them from the public documentation.

seladb: Nice!! You can test it locally or merge this PR and I'll test it later. It's not a blocker for merging because the documentation isn't final for a non-releases and we can fix it later if needed

tigercosmos: let's make it in another PR if necessary.

PR #1441, opened 2024-06-06
Use smart pointers for RemoteAuthentication in PcapRemoteDeviceList.

Dimi1010:
Part of #1431. Implements point 3.


Pcap++/src/PcapRemoteDevice.cpp
```diff
@@ -19,14 +19,14 @@ pcap_rmtauth PcapRemoteAuthentication::getPcapRmAuth() const
 	return result;
 }
 
-PcapRemoteDevice::PcapRemoteDevice(pcap_if_t* iface, PcapRemoteAuthentication* remoteAuthentication, const IPAddress& remoteMachineIP, uint16_t remoteMachinePort)
+PcapRemoteDevice::PcapRemoteDevice(pcap_if_t* iface, std::shared_ptr<PcapRemoteAuthentication> remoteAuthentication, const IPAddress& remoteMachineIP, uint16_t remoteMachinePort)
 	: PcapLiveDevice(iface, false, false, false)
+	, m_RemoteMachineIpAddress(remoteMachineIP)
```
tigercosmos, line 24: recommend the reorder of the construction list by the order of the arguments.

tigercosmos (approved):

Pcap++/src/PcapRemoteDevice.cpp
Dimi1010: It is currently ordered the way the members are declared in the class, which is the actual way they will be initialized.

tigercosmos: I see

Pcap++/src/PcapRemoteDevice.cpp
seladb, line 1: There are a few more occurrences of `NULL`, maybe we should replace them with `nullptr`?

seladb (approved):

Pcap++/src/PcapRemoteDevice.cpp
Dimi1010: Added 501bda6. There are some C-style casts left over, but those are handled in #1443.

PR #1445, opened 2024-06-11
Refactor of PfRingDeviceList internals.

Dimi1010:
This PR brings internal implementation changes from `PcapLiveDeviceList` internals to `PfRingDeviceList`.

Changes:
- Explicitly marked copy/move ctors as deleted.
- Changed device fetching to be handled by `getAllLocalPcapDevices`.
- PfRingDevice pointers are now managed by unique pointers.
- Replaced `calcPfRingVersion` with a local function to the translation unit to remove the need to use `void*`.
- Added custom deleter for `pfring_close` to manage PF_RING handles with smart pointers.


tigercosmos (approved):
LGTM

Pcap++/src/PfRingDeviceList.cpp
```diff
@@ -4,14 +4,53 @@
...
+
+			char versionAsString[25];
+			std::snprintf(versionAsString, 25, "PF_RING v.%u.%u.%u\n",
+				(version & 0xFFFF0000) >> 16,
+				(version & 0x0000FF00) >> 8,
+				 version & 0x000000FF);
+
+			return std::string(versionAsString);
+		}
+	}
```
seladb, line 51: I'm not sure this needs to be in its own namespace. Why not just put it in the cpp file?

Pcap++/src/PfRingDeviceList.cpp
```diff
@@ -4,14 +4,53 @@
...
+		 * Reads the ring version of a PF_RING handle.
+		 * @param[in] ring A PF_RING handle.
+		 * @return A string representation of the ring version.
+		 */
+		std::string readPfRingVersion(pfring* ring)
+		{
+			uint32_t version;
+			if (pfring_version(ring, &version) < 0)
+			{
+				throw std::runtime_error("Couldn't retrieve PF_RING version, pfring_version returned an error");
```
seladb, line 40: I'm not sure it should throw an exception, it can just return an empty string

Pcap++/src/PfRingDeviceList.cpp
Dimi1010: I guess, it can be. I wasn't exactly sure on, if `Foo::(unnamed namespace)` also had internal linkage, but reading more about it, they do.

seladb: does it really matter? It's anyway not exposed to the outside 🤔

Pcap++/src/PfRingDeviceList.cpp
tigercosmos: true, probably "unknown"

seladb: It can be "unknown" also, I don't have a strong opinion

Pcap++/src/PfRingDeviceList.cpp
Dimi1010: I find it more clear having `internal to this file only` functions wrapped in an unnamed namespace, so they can't be accidentally exported. Other than that and the internal linkage, not really sure.

Pcap++/src/PfRingDeviceList.cpp
Dimi1010: Eh, it can return an empty or "Unknown" string, since its internal function.

I preferred throwing an exception because if `pfring_version` fails, the version isn't really "read" is it?
In addition to that, how often do we expect the reading of the version to fail?

If we look at the only current call site at line 94, the "happy" path currently is:
1. Read string.
2. Save string to variable.
3. Carry on...

If we are returning an empty string, the path will always need to check if the string is empty or equal to a magic string ("Unknown") after step 2. With the exception, we aren't paying overhead if the read succeeds, only if it fails.

So the question is, are we expecting the read to fail like half the time, or once in a blue moon?
The usage of exceptions, essentially removes overhead from the happy path and adds **more** overhead to the exceptional path.
- If we are expecting the read to fail a lot, an error code approach might be better.
- If the read is expected to succeed most of the time an exception approach is better.

seladb: I don't expect `pfring_version()` to fail after `pfring_open()` succeeded. Still, I think throwing an exception makes the code a bit more complex than it needs to be. The caller of `readPfRingVersion()` doesn't care what the version is - it could be any string, including an empty string or "Unknown". `readPfRingVersion()` can decide which string to return, and if `pfring_version()` fails it can write a message to the log

Pcap++/src/PfRingDeviceList.cpp
seladb: ok sure 👍

Pcap++/src/PfRingDeviceList.cpp
Dimi1010: The caller does care tho. Both before and after the refactor.

See the old lines 56-57. `calcPfRingVersion` is called for every device until the version string is populated, and on error it does not populate the version string. The refactor mirrors that behaviour by not writing the result of `readPfRingVersion`, if an exception occurs.

I guess, we can make the result be an empty string and write that regardless, as the `if` condition will still retry for the next device, but a populated error string will require another condition to ignore it.

seladb: ah you're right, I missed that. So I think the method can return an empty string if it fails?

Dimi1010: sure, if we start using it at more places, then we might discuss exceptions again.

Dimi1010: Changed to empty string. dfa993d

seladb (approved):

PR #1446, opened 2024-06-11
Refactor executeShellCommand to throw exceptions.

Dimi1010:
This PR refactors the internal `executeShellCommand` function to throw a `runtime_error` on failure to execute, instead of returning the magic string ("ERROR").

Additionally refactored `PfRingDeviceList` to utilize the helper function during construction.

### DIFF
--- a/Common++/src/SystemUtils.cpp
+++ b/Common++/src/SystemUtils.cpp
@@ -51,6 +51,22 @@ int gettimeofday(struct timeval * tp, struct timezone * tzp)
 }
 #endif
 
+/// @cond PCPP_INTERNAL
+
+namespace
+{
+	/**
+	 * @class PcloseDeleter
+	 * A deleter that cleans up a FILE handle using pclose.
+	 */
+	struct PcloseDeleter
+	{
+		void operator()(FILE* ptr) const { PCLOSE(ptr); }
+	};
+} // namespace
+
+/// @endcond
+
 namespace pcpp
 {
 
@@ -184,14 +200,6 @@ void createCoreVectorFromCoreMask(CoreMask coreMask, std::vector<SystemCore>& re
 		coreMask = coreMask >> 1;
 		i++;
 	}
-}
-
-namespace
-{
-	struct PcloseDeleter
-	{
-		void operator()(FILE* ptr) const { PCLOSE(ptr); }
-	};
 }
 
 std::string executeShellCommand(const std::string& command)
--- a/Common++/src/SystemUtils.cpp
+++ b/Common++/src/SystemUtils.cpp
@@ -6,6 +6,7 @@
 #endif
 #include <stdexcept>
 #include <memory>
+#include <array>
 #include <iostream>
 #include <mutex>
 #include <signal.h>
@@ -201,12 +202,12 @@ std::string executeShellCommand(const std::string& command)
 		throw std::runtime_error("Error executing command: " + command);
 	}
 
-	char buffer[128];
+	std::array<char, 128> buffer;
 	std::string result;
 	while(!feof(pipe.get()))
 	{
-		if(fgets(buffer, 128, pipe.get()) != nullptr)
-			result += buffer;
+		if(fgets(buffer.data(), buffer.size(), pipe.get()) != nullptr)
+			result += buffer.data(); // Using the C-string overload of string append.
 	}
 	return result;
 }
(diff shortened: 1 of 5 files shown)


Common++/src/SystemUtils.cpp
```diff
@@ -183,18 +185,29 @@ void createCoreVectorFromCoreMask(CoreMask coreMask, std::vector<SystemCore>& re
 	}
 }
 
-std::string executeShellCommand(const std::string &command)
+namespace
```
tigercosmos, line 188: can't we just put `PcloseDeleter` outside?

Common++/src/SystemUtils.cpp
```diff
@@ -183,18 +185,29 @@ void createCoreVectorFromCoreMask(CoreMask coreMask, std::vector<SystemCore>& re
...
 {
-	FILE* pipe = POPEN(command.c_str(), "r");
-	if (!pipe) return "ERROR";
+	struct PcloseDeleter
+	{
+		void operator()(FILE* ptr) const { PCLOSE(ptr); }
```
tigercosmos, line 192: better check the `ptr`
```       
 if (ptr) {
            PCLOSE(ptr);
        }
```

Common++/src/SystemUtils.cpp
```diff
@@ -183,18 +185,29 @@ void createCoreVectorFromCoreMask(CoreMask coreMask, std::vector<SystemCore>& re
...
+
+std::string executeShellCommand(const std::string& command)
+{
+	std::unique_ptr<FILE, PcloseDeleter> pipe = std::unique_ptr<FILE, PcloseDeleter>(POPEN(command.c_str(), "r"));
+	if (!pipe)
+	{
+		throw std::runtime_error("Error executing command: " + command);
+	}
+
 	char buffer[128];
```
tigercosmos, line 204: recommend `std::array<char, 128> buffer;`

Pcap++/src/DpdkDeviceList.cpp
```diff
@@ -258,24 +258,25 @@ bool DpdkDeviceList::verifyHugePagesAndDpdkDriver()
...
+		{
+			try
 			{
-				PCPP_LOG_ERROR("None of igb_uio, uio_pci_generic, vfio-pci kernel modules are loaded so DPDK cannot be initialized. Please run <PcapPlusPlus_Root>/setup_dpdk.sh");
-				return false;
+				execResult = executeShellCommand("modinfo -d vfio-pci");
+				PCPP_LOG_DEBUG("vfio-pci module is loaded");
 			}
-			else
+			catch (const std::runtime_error&)
```
tigercosmos, line 273: why not print the error?

Common++/src/SystemUtils.cpp
Dimi1010: The deleter is made to be used only by the smart pointer and those already skip calling the deleter on nullptr.

Common++/src/SystemUtils.cpp
Dimi1010: Define outside? Outside the anonymous namespace?

Pcap++/src/DpdkDeviceList.cpp
Dimi1010: Imo, the original error message is sufficiently descriptive if the entire try-catch chain fails.

Also printing in the last catch will only ever print the last (vfio-pci) command as error.

We can print it as a debug message, but we already have debug messages on success so idk.

Common++/src/SystemUtils.cpp
tigercosmos: yes, any reason that you need the anonymous namespace here?

Dimi1010: Well, mostly to have the helper functions and structs local to the translation unit. Although I guess it would be better to pull the unnamed namespace up to the global level.

Common++/src/SystemUtils.cpp
Dimi1010: Added bcceb79

Common++/src/SystemUtils.cpp
Dimi1010: Pulled the structure into a top level unnamed namespace (c285b6f).

seladb (approved):

PR #1451, opened 2024-06-17
Clang format for `Common++`

### LINKED ISSUE #1248: introducing clang-format

tigercosmos:
apply clang-format for `Common++`


.clang-format
```diff
@@ -1,7 +1,8 @@
 ---
 BasedOnStyle: Microsoft
 UseTab: Always
+BreakBeforeBraces: Allman
 AllowShortFunctionsOnASingleLine: All
 NamespaceIndentation: All
-
+SortIncludes: false
```
tigercosmos, line 7: prevent some possible includes' error due to the ordering.

.clang-format
```diff
@@ -1,7 +1,8 @@
 ---
 BasedOnStyle: Microsoft
 UseTab: Always
+BreakBeforeBraces: Allman
```
tigercosmos, line 4: the style like this:
```
void Foo ()
{
// ...
}
```

.github/workflows/build_and_test.yml
```diff
@@ -30,6 +30,7 @@ jobs:
         run: |
           apk update && apk add cppcheck python3-dev
           python3 -m pip install cmake-format
+          pip install clang-format==18.1.6
```
tigercosmos, line 33: specify the clang-formant here.

.pre-commit-config.yaml
```diff
@@ -16,9 +16,11 @@ repos:
   - repo: https://github.com/pocc/pre-commit-hooks
     rev: v1.3.5
     hooks:
+      - id: clang-format
+        args: ["--style=file"]  # Use the .clang-format file for configuration
+        files: ^Common\+\+/.*\.(cpp|h)$
```
tigercosmos, line 21: apply on `Common++` for now.

Common++/header/DeprecationUtils.h
```diff
@@ -3,35 +3,37 @@
...
+#define DISABLE_WARNING_PUSH DO_PRAGMA(GCC diagnostic push)
+#define DISABLE_WARNING_POP DO_PRAGMA(GCC diagnostic pop)
+#define DISABLE_WARNING(warningName) DO_PRAGMA(GCC diagnostic ignored #warningName)
 
-        #define DISABLE_WARNING_DEPRECATED     DISABLE_WARNING(-Wdeprecated-declarations)
-    #else
-        #pragma message("WARNING: Disabling of warnings is not implemented for this compiler")
-        #define DISABLE_WARNING_PUSH
-        #define DISABLE_WARNING_POP
+// clang-format off
```
tigercosmos, line 29: here is the manual fix, since the clang-formant cannot do the formating well here. it tries to separate `-Wdeprecated-declarations` to `-Wdeprecated - declarations`

tigercosmos:
don't squash the commit.

seladb:
@tigercosmos the PR title should be:
"Clang format for `Common++`", not `Commit++` 🙂

tigercosmos:
sorry, fixed the title.

btw, seems that the CI already runs the check for clang-format

tigercosmos:
Sorry, I think it's okay to squash the commits, and we can add the ignore hash in the next PR.

Common++/header/DeprecationUtils.h
```diff
@@ -3,35 +3,37 @@
...
+// clang-format on
+#else
+#pragma message("WARNING: Disabling of warnings is not implemented for this compiler")
+#define DISABLE_WARNING_PUSH
+#define DISABLE_WARNING_POP
 
-        #define DISABLE_WARNING_DEPRECATED
-    #endif
+#define DISABLE_WARNING_DEPRECATED
+#endif
```
Dimi1010, line 38: Can we add comments to the `endif`s if clang is going to format away the spaces (or disable that formatting?). It was fine without when they were indented, but it gets really hard to follow nested chains when they are flat.

tigercosmos: @Dimi1010  sure, apply indents for the macros

.clang-format
```diff
@@ -1,7 +1,10 @@
...
 AllowShortFunctionsOnASingleLine: All
 NamespaceIndentation: All
-
+SortIncludes: false
+IndentPPDirectives: AfterHash
```
tigercosmos, line 8: let the macro has indent

.clang-format
```diff
@@ -1,7 +1,10 @@
...
+IndentWidth: 4
```
tigercosmos, line 9: macro indent width

Common++/header/Logger.h
```diff
@@ -52,64 +55,64 @@ namespace pcpp
...
-		NetworkUtils, ///< NetworkUtils module (Pcap++)
+		PcapLogModuleRemoteDevice,		///< WinPcapRemoteDevice module (Pcap++)
+		PcapLogModuleLiveDevice,		///< PcapLiveDevice module (Pcap++)
+		PcapLogModuleFileDevice,		///< FileDevice module (Pcap++)
+		PcapLogModulePfRingDevice,		///< PfRingDevice module (Pcap++)
+		PcapLogModuleMBufRawPacket,		///< MBufRawPacket module (Pcap++)
+		PcapLogModuleDpdkDevice,		///< DpdkDevice module (Pcap++)
+		PcapLogModuleKniDevice,			///< KniDevice module (Pcap++)
+		PcapLogModuleXdpDevice,			///< XdpDevice module (Pcap++)
+		NetworkUtils,					///< NetworkUtils module (Pcap++)
```
seladb, line 115: Also here - I'm not sure why clang-format decided to use tabs instead of spaces. It doesn't look as it should:

image

Common++/header/Logger.h
```diff
@@ -153,7 +158,8 @@ namespace pcpp
 		 * @param[in] method The method in PcapPlusPlus code the log message is coming from
 		 * @param[in] line The line in PcapPlusPlus code the log message is coming from
 		 */
-		typedef void (*LogPrinter)(LogLevel logLevel, const std::string& logMessage, const std::string& file, const std::string& method, const int line);
+		typedef void (*LogPrinter)(LogLevel logLevel, const std::string &logMessage, const std::string &file,
+								   const std::string &method, const int line);
```
seladb, line 162: In this case clang-format is using a combination of spaces and tabs, it's even more weird...

Common++/header/Logger.h
```diff
@@ -187,7 +193,11 @@ namespace pcpp
 		 * Set all PcapPlusPlus modules to a certain log level
 		 * @param[in] level The log level to set all modules to
 		 */
-		void setAllModulesToLogLevel(LogLevel level) { for (int i=1; i<NumOfLogModules; i++) m_LogModulesArray[i] = level; }
+		void setAllModulesToLogLevel(LogLevel level)
+		{
+			for (int i = 1; i < NumOfLogModules; i++)
+				m_LogModulesArray[i] = level;
```
seladb, line 199: Is there a way to always enforce curly braces even if there's only one statement?

Common++/src/Logger.cpp
```diff
@@ -4,55 +4,50 @@
...
-	delete logStream;
-	if (logLevel == Logger::Error)
+	void Logger::defaultLogPrinter(LogLevel logLevel, const std::string &logMessage, const std::string &file,
+								   const std::string &method, const int line)
 	{
-		m_LastError = logMessage;
+		std::ostringstream sstream;
+		sstream << file << ": " << method << ":" << line;
+		std::cout << std::left << "[" << std::setw(5) << Logger::logLevelAsString(logLevel) << ": " << std::setw(45)
+				  << sstream.str() << "] " << logMessage << std::endl;
```
seladb, line 33: I actually like the previous version better, but not sure if there is a way to keep it...

Common++/src/SystemUtils.cpp
```diff
@@ -70,341 +70,300 @@ namespace
...
+	const SystemCore SystemCores::Core31 = {0x80000000, 31};
+
+	const SystemCore SystemCores::IdToSystemCore[MAX_NUM_OF_CORES] = {
+		SystemCores::Core0,	 SystemCores::Core1,  SystemCores::Core2,  SystemCores::Core3,	SystemCores::Core4,
+		SystemCores::Core5,	 SystemCores::Core6,  SystemCores::Core7,  SystemCores::Core8,	SystemCores::Core9,
+		SystemCores::Core10, SystemCores::Core11, SystemCores::Core12, SystemCores::Core13, SystemCores::Core14,
+		SystemCores::Core15, SystemCores::Core16, SystemCores::Core17, SystemCores::Core18, SystemCores::Core19,
+		SystemCores::Core20, SystemCores::Core21, SystemCores::Core22, SystemCores::Core23, SystemCores::Core24,
+		SystemCores::Core25, SystemCores::Core26, SystemCores::Core27, SystemCores::Core28, SystemCores::Core29,
+		SystemCores::Core30, SystemCores::Core31};
```
seladb, line 113: is there a way to ask clang-format to keep it the way it was, meaning one value in each line?

Common++/header/IpAddressUtils.h
```diff
@@ -21,83 +21,83 @@ namespace pcpp
 	 * Overload of the equal-to operator
 	 * @return True if the addresses are equal, false otherwise
 	 */
-	bool operator==(const IPv4Address& lhs, const in_addr& rhs);
+	bool operator==(const IPv4Address &lhs, const in_addr &rhs);
```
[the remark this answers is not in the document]
tigercosmos: I just follow the default choice. let me change it.

Common++/header/GeneralUtils.h
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
-#include <string>
 #include <stdint.h>
+#include <string>
```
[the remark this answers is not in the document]
tigercosmos: I added that option later, but I think it's better to reorder as long as there is no build error?
I made the reordering to off because it may break the rest of the codebase.

Common++/header/IpAddress.h
```diff
@@ -31,7 +30,7 @@ namespace pcpp
 	 */
 	class IPv4Address
 	{
-	public:
+	  public:
```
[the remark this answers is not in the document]
tigercosmos: let me find if it can be addressed.

Common++/header/IpAddress.h
```diff
@@ -93,20 +92,21 @@ namespace pcpp
...
+		bool operator<(const IPv4Address &rhs) const
 		{
 			uint32_t intVal = toInt();
-			std::reverse(reinterpret_cast<uint8_t*>(&intVal), reinterpret_cast<uint8_t*>(&intVal) + sizeof(intVal));
+			std::reverse(reinterpret_cast<uint8_t *>(&intVal), reinterpret_cast<uint8_t *>(&intVal) + sizeof(intVal));
 
 			uint32_t rhsIntVal = rhs.toInt();
-			std::reverse(reinterpret_cast<uint8_t*>(&rhsIntVal), reinterpret_cast<uint8_t*>(&rhsIntVal) + sizeof(rhsIntVal));
+			std::reverse(reinterpret_cast<uint8_t *>(&rhsIntVal),
+						 reinterpret_cast<uint8_t *>(&rhsIntVal) + sizeof(rhsIntVal));
```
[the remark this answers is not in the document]
tigercosmos: unless we change to spaces, otherwise I don' t think we can address this.

Common++/src/SystemUtils.cpp
tigercosmos: I agree, this doesn't look good.

Common++/header/GeneralUtils.h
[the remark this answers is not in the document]
seladb: Once we agree on all the formatting rules, I suggest we re-run the existing Common++ codebase with `clang-format` to make sure the changes match the formatting rules

Common++/src/SystemUtils.cpp
tigercosmos: there is `BreakArrays`, but only for JSON https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakarrays

Common++/header/IpAddress.h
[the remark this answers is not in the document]
seladb: Maybe try to use the other options of `UseTab`? I'm not sure what each one does exactly, but maybe one of them addresses this issue:
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#usetab

Common++/src/SystemUtils.cpp
seladb: Maybe `Cpp11BracedListStyle: false`? 🤔 
https://www.reddit.com/r/C_Programming/comments/izosbi/how_to_clangformat_struct_initializer_list/?rdt=38280

Common++/header/IpAddressUtils.h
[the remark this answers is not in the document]
tigercosmos: fixed.

Common++/header/GeneralUtils.h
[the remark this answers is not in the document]
tigercosmos: I reset it.

Common++/header/IpAddress.h
```diff
@@ -1,16 +1,15 @@
 #pragma once
 
-#include <stdint.h>
-#include <string.h>
-#include <string>
 #include <algorithm>
-#include <ostream>
 #include <array>
 #include <memory>
+#include <ostream>
```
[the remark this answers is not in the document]
tigercosmos: fixed.

Common++/header/IpAddress.h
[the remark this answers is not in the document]
tigercosmos: fixed.

Common++/header/Logger.h
tigercosmos: fixed.

Common++/src/Logger.cpp
tigercosmos: I cannot find the solution for this.

Common++/src/SystemUtils.cpp
tigercosmos: it helps a little bit.
```
	const SystemCore SystemCores::IdToSystemCore[MAX_NUM_OF_CORES] = {
		SystemCores::Core0,  SystemCores::Core1,  SystemCores::Core2,  SystemCores::Core3,  SystemCores::Core4,
		SystemCores::Core5,  SystemCores::Core6,  SystemCores::Core7,  SystemCores::Core8,  SystemCores::Core9,
		SystemCores::Core10, SystemCores::Core11, SystemCores::Core12, SystemCores::Core13, SystemCores::Core14,
		SystemCores::Core15, SystemCores::Core16, SystemCores::Core17, SystemCores::Core18, SystemCores::Core19,
		SystemCores::Core20, SystemCores::Core21, SystemCores::Core22, SystemCores::Core23, SystemCores::Core24,
		SystemCores::Core25, SystemCores::Core26, SystemCores::Core27, SystemCores::Core28, SystemCores::Core29,
		SystemCores::Core30, SystemCores::Core31
	};
```

Common++/header/IpAddress.h
```diff
@@ -41,19 +40,27 @@ namespace pcpp
...
+		}
 
 		/**
 		 * A constructor that creates an instance of the class out of a 4-byte standard array.
 		 * @param[in] bytes The address as 4-byte standard array in network byte order
 		 */
-		IPv4Address(const std::array<uint8_t, 4>& bytes) : m_Bytes(bytes) {}
+		IPv4Address(const std::array<uint8_t, 4>& bytes) : m_Bytes(bytes)
+		{
+		}
```
seladb, line 63: This formatting change doesn't make a lot of sense, is there an option to avoid it?

Common++/header/LRUList.h
```diff
@@ -1,10 +1,10 @@
 #pragma once
 
-#include <unordered_map>
 #include <list>
+#include <unordered_map>
```
seladb, line 4: For some reason it still changes the include order, any idea why?

Common++/header/MacAddress.h
```diff
@@ -50,9 +53,8 @@ namespace pcpp
 		 * If the string doesn't represent a valid MAC address, the constructor throws an exception.
 		 * @param[in] addr the string representing the MAC address in format "00:00:00:00:00:00"
 		 */
-		template<typename T, typename = typename std::enable_if<std::is_convertible<T, std::string>::value>::type>
-		MacAddress(const T& addr) : MacAddress(static_cast<std::string>(addr)) {};
-
+		template <typename T, typename = typename std::enable_if<std::is_convertible<T, std::string>::value>::type>
+		MacAddress(const T& addr) : MacAddress(static_cast<std::string>(addr)){};
```
seladb, line 57: Here it keeps the empty curly braces `{}` in the same line, as opposed to `IpAddress.h` where it didn't. Any idea why?

Common++/src/OUILookup.cpp
```diff
@@ -8,103 +8,104 @@
...
+								++ctrRead;
+							}
 						}
 					}
 				}
 			}
+
+			vendorMap.insert({
+			    std::stoull(line.key()), { val["vendor"], vLocalMaskedFilter }
+            });
```
seladb, line 57: There is a weird mix of tabs and spaces in these lines, but not sure if it can be fixed 🤔

Common++/src/SystemUtils.cpp
seladb: yeah, still doesn't look very nice...

. git-blame-ignore-revs
seladb, line 1: Should we add this as empty file in this commit, or add it later after we merge?

tigercosmos: I think we can have it now, we will add the file anyway.

Common++/src/OUILookup.cpp
tigercosmos: it is in purpose. the indent is tab, the alignment is space.
for example:
```
<tab> void  foo(int a,
<tab><spaces>int b) {}
```

Common++/src/SystemUtils.cpp
tigercosmos: I think it's fine though, actually shorter and cleaner.

Common++/header/LRUList.h
tigercosmos: yeah, weird.

Common++/src/OUILookup.cpp
seladb: yeah, but why does line 57 has spaces instead of tabs? Maybe we should just move `});` to line 56?

. git-blame-ignore-revs
seladb: I don't have a strong opinion, but I think it's better to add the file after we have the commit hash. But if you feel strongly about keeping it, I'm ok with it

Common++/header/MacAddress.h
tigercosmos: maybe because it's a template?

Common++/src/SystemUtils.cpp
seladb: ok 👍

Common++/src/OUILookup.cpp
tigercosmos: yeah... you are right.. it's weird here as well.

Common++/header/IpAddress.h
tigercosmos: this one is due to `BreakBeforeBraces: Allman`

Common++/header/MacAddress.h
tigercosmos: I think I know why, probably due to `;`

tigercosmos: yeah, after I removed `;`, it becomes normal.
But for now I can do manually fixing, but after when we apply a lot of files, it's hard to pick these syntax issues one by one.

Common++/header/IpAddress.h
seladb: If we change to `Custom` we can set `SplitEmptyFunction: false` which I think should solve it:
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#bracewrapping

Common++/header/MacAddress.h
seladb: oh, maybe we should just remove the `;`

Common++/header/LRUList.h
tigercosmos: I don't know why, but I cannot resolve this one. the only way is to insert empty line to ensure the order not to change, but in this case, it's okay to let it change.

Common++/header/MacAddress.h
seladb: > yeah, after I removed `;`, it becomes normal. But for now I can do manually fixing, but after when we apply a lot of files, it's hard to pick these syntax issues one by one.

we should fix things whenever we see them, of course we might miss some...

Common++/header/IpAddress.h
tigercosmos: actually, I would stand for splitting the braces in this case. shall we just use the default `Allman` settings to prevent too much customization from using `Custom`?

Common++/src/OUILookup.cpp
seladb: did you try moving `});` from line 57 to line 56 before running `clang-format`?

Common++/header/LRUList.h
seladb: ok 👍

Common++/header/IpAddress.h
seladb: I don't know the default values of the different flags under `BraceWrapping`. If all the defaults are what we need, I think it's ok to set:
```
BreakBeforeBraces: Custom
BraceWrapping:
  SplitEmptyFunction: false
```

Common++/src/OUILookup.cpp
tigercosmos: I tried, it doesn't help 
no matter this;
```
vendorMap.insert({std::stoull(line.key()), { val["vendor"], vLocalMaskedFilter }});
```
or
```
			vendorMap.insert(
				{std::stoull(line.key()), { val["vendor"], vLocalMaskedFilter }});
```

they will be formatted as:
```
			vendorMap.insert({
				std::stoull(line.key()), { val["vendor"], vLocalMaskedFilter }
            });
```

one way is to use `// clang-format off` here, but I would suggest just let it be this if it doesn't break the build.
according to my experience, when there is too many braces, clang-format becomes stupid.

Common++/header/IpAddress.h
tigercosmos: I tried your config, and the effect is like this:

```
void matchNetwork(const std::string& network) {}
```
become this:
```
void matchNetwork(const std::string& network)
{}
```

Common++/src/OUILookup.cpp
seladb: yes, we can just leave it the way it is... 😕

Common++/header/IpAddress.h
seladb: this is slightly better I think. The rest of the curly braces formatting stays the same?

tigercosmos: if there is something in the braces, yes, it will be the same.
```
void matchNetwork(const std::string& network)
{
   // ...
}
```

tigercosmos: sure, applied your config.

seladb (approved):

PR #1452, opened 2024-06-17
Fix build failures on MinGW during package CI

seladb:
This PR addresses this issue: https://github.com/seladb/PcapPlusPlus/issues/1347

I'm not entirely sure how the MinGW build we have in the `package` workflow is different from the one we have in the `build_and_test` workflow (both have the same versions and are running on the same type of VM), but anyway I addressed the build failure and they all seem to pass now: 
`package`: https://github.com/seladb/PcapPlusPlus/actions/runs/9577752848
`build_and_test`: https://github.com/seladb/PcapPlusPlus/actions/runs/9577752845

### DIFF
--- a/Packet++/src/Asn1Codec.cpp
+++ b/Packet++/src/Asn1Codec.cpp
@@ -622,7 +622,7 @@ namespace pcpp
 	{
 		std::vector<uint8_t> result;
 
-		#if !(defined(__MIGW32__) || defined(__MIGW64__))
+		#if !(defined(__MINGW64_VERSION_MAJOR) || defined(__MINGW32_MAJOR_VERSION))
 		result.reserve(m_ValueLength);
 		#endif
 
--- a/Packet++/src/Asn1Codec.cpp
+++ b/Packet++/src/Asn1Codec.cpp
@@ -621,7 +621,7 @@ namespace pcpp
 	std::vector<uint8_t> Asn1IntegerRecord::encodeValue() const
 	{
 		std::vector<uint8_t> result;
-		result.reserve(m_ValueLength);
+//		result.reserve(m_ValueLength);
 
 		switch (m_ValueLength)
 		{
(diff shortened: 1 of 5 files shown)


Packet++/src/Asn1Codec.cpp
```diff
@@ -621,7 +621,6 @@ namespace pcpp
 	std::vector<uint8_t> Asn1IntegerRecord::encodeValue() const
 	{
 		std::vector<uint8_t> result;
-		result.reserve(m_ValueLength);
```
seladb, line 624: I have no idea why the compiler complains here, but I think it's fine to remove it:
https://github.com/seladb/PcapPlusPlus/actions/runs/9562329238/job/26358481698

tigercosmos: can we ignore this line only for `MinGW`?

seladb: It's a compilation error so not sure we can ignore it, but I think removing it is ok, I don't foresee a big performance impact because of it

tigercosmos: I mean `ifndef MinGW` or some similar syntax. 
I think it's better to keep the `reserve` as it still works for other platforms.

tigercosmos (approved):

tigercosmos:
@seladb  last thing before merging, could you add a note for every place you made changes?
such as `// PCPP patch`
since it's a forked code, it will be easier to trace in the future.

seladb:
> @seladb last thing before merging, could you add a note for every place you made changes? such as `// PCPP patch` since it's a forked code, it will be easier to trace in the future.

Done in commit https://github.com/seladb/PcapPlusPlus/pull/1452/commits/dbfee2a4f86fd443366e4773e86ebff719c82a26

PR #1458, opened 2024-06-23
Move doxygen to GitHub Actions

seladb:
- Move doxygen to GitHub Actions and remove it from Cirrus CI
- Temporary comment-out all Cirrus CI tasks so we can merge PRs


.github/workflows/build_and_test.yml
```diff
@@ -15,684 +15,725 @@ permissions:
   contents: read
 
 jobs:
-  pre-commit:
+  doxygen:
     runs-on: ubuntu-latest
-    container: seladb/alpine317
+    container: seladb/ubuntu1804:latest
```
tigercosmos, line 20: @seladb why ubuntu18? we should use the lastest LTS.

seladb: We need doxygen to be installed. We installed it in all of our Ubuntu images: https://github.com/seladb/PcapPlusPlus-DockerImages/blob/master/Dockerfile-ubuntu2204

I updated the image to Ubuntu 22.04

clementperon:
@seladb should It be a different workflow?

seladb:
> @seladb should It be a different workflow?

I'm not sure it's necessary, because a major part is to test doxygen which is part of `build_and_test`

tigercosmos:
> > @seladb should It be a different workflow?
> 
> I'm not sure it's necessary, because a major part is to test doxygen which is part of `build_and_test`

@seladb  I think it is worth doing that, since we can easily track the CI category at the panel:
image

clementperon:
> > > @seladb should It be a different workflow?
> > 
> > 
> > I'm not sure it's necessary, because a major part is to test doxygen which is part of `build_and_test`
> 
> @seladb I think it is worth doing that, since we can easily track the CI category at the panel:

@seladb @tigercosmos to me the other reason is that the triggering mode could differs from the build_package but maybe i'm wrong, So it don't have a strong opinion.

.cirrus.yml
```diff
@@ -2,97 +2,68 @@ env:
   BUILD_DIR: "build"
   GCOVR_FLAGS: "--gcov-ignore-parse-errors --exclude-throw-branches --filter Common --filter Pcap --filter Packet --xml"
```
clementperon, line 3: maybe we could remove cirrus file no ?

clementperon (approved):

.cirrus.yml
seladb: I think we can remove it after we migrate all jobs to GitHub Actions. In each PR I'll remove the jobs that are migrated

PR #1459, opened 2024-06-23
Move Zstd tests from CirrusCI to GitHub Actions

seladb:
- Move Zstd tests from CirrusCI to GitHub Actions
- Comment out the rest of Cirrus CI jobs so we can merge the PR

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -55,9 +55,6 @@ jobs:
           - image: ubuntu2204
             python: python3
             config-zstd: OFF
-          - image: ubuntu2204
-            python: python3
-            config-zstd: ON
           - image: ubuntu2204-icpx
             python: python3
             config-zstd: OFF
@@ -66,6 +63,9 @@ jobs:
           - image: ubuntu2004
             python: python3
             config-zstd: OFF
+          - image: ubuntu2004-zstd
+            python: python3
+            config-zstd: ON
           - image: ubuntu1804
             python: python3.8
             config-zstd: OFF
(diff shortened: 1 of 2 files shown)


clementperon (approved):

.github/workflows/build_and_test.yml
```diff
@@ -233,15 +236,22 @@ jobs:
...
-        os-version: [macos-12, macos-13]
-        arch: [x86_64, arm64]
-        # Handle ZSTD build by Cirrus CI
-        # config-zstd: [ON, OFF]
-        config-zstd: [OFF]
-        exclude:
-          # excludes ZSTD on Arm64
-          - arch: arm64
+        include:
+          - os-version: macos-12
```
tigercosmos, line 240: no macos-14? maybe we should just maintain 13 and 14?

seladb: macOS 14 is M1 only: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

macOS 12 and 13 are also x86_64

tigercosmos: @seladb  I see, then don't we test zstd on macos 14 arm?

seladb: I don't think it's necessary to test Zstd on all archs. 

I tried on arm64 and it [didn't work](https://github.com/seladb/PcapPlusPlus/actions/runs/9631691995/job/26563813940), probably because the Zstd version installed on this machine isn't compiled with arm64, so x86_64 is currently the only option

Dimi1010 (approved):

PR #1460, opened 2024-06-23
Freebsd CI

tigercosmos:


.github/workflows/build_and_test.yml
```diff
@@ -561,6 +561,44 @@ jobs:
...
+        prepare: |
+          pkg install -y python3 bash cmake git gmake gsed libpcap tcpreplay
+
+        run: |
+          echo "Building PcapPlusPlus"
+          chmod a+rw /dev/bpf*
+          cmake -S . -B Dist
+          cmake --build Dist -j$(sysctl -n hw.ncpu)
+
+          echo "Setting up the network interface for the tests"
```
tigercosmos, line 586: since we cannot split the commands when we already in a `run`, just echoing the names.

note that we cannot have multiple runs, because the each run will init a new VM.

tigercosmos:
@seladb  please check.

.github/workflows/build_and_test.yml
seladb: > note that we cannot have multiple runs, because the each run will init a new VM.

Why can't we split it into multiple runs? As far as I know, all `run` commands share the same environment.
One thing they might not share is the `interface_name` variable, but I think you can calculate it once and store it in `$GITHUB_ENV`: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files

I'd suggest splitting into multiple steps according to the `echo` commands you used

tigercosmos: yes, you're right. I think in a wrong way

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
tigercosmos: fixed.

.github/workflows/build_and_test.yml
tigercosmos: since a "run" here actually go into the VM environment, so we must to stay in the same "run"

tigercosmos: got the error if we split to many runs: https://github.com/seladb/PcapPlusPlus/actions/runs/9638713557/job/26579832666?pr=1460

seladb: > since a "run" here actually go into the VM environment, so we must to stay in the same "run"

Yes you're right! this step sets up a  FreeBSD VM so if we use it in multiple steps, multiple VMs will be created

seladb (approved):

PR #1468, opened 2024-06-24
`PointerVector` smart pointer intergration.

Dimi1010:
Core changes to `PointerVector`:
- Added overload to `pushBack` that accepts `unique_ptr<T>`. 
- Added new methods `getAndDetach` that work analogous to `getAndRemoveFromVector` but return `unique_ptr<T>`.
- Added move constructor that that transfers the ownership of the elements moved to the vector.
- Added copy/move assignment operators.
- Added `const` overloads to `front()`/`back()` accessors.
- Added exceptions when `nullptr` is being pushed into the vector.


Common++/header/PointerVector.h
```diff
@@ -26,84 +28,130 @@ namespace pcpp
...
 		 */
 		PointerVector()
 		{}
 
+		/**
+		 * Copies the vector along with all elements inside it.
+		 * All elements inside the copied vector are duplicates and the originals remain unchanged.
+		 * @param[in] other The vector to copy from.
+		 * @remarks As the vector is copied via deep copy, all pointers obtained from the copied vector
+		 *	reference the duplicates and not the original.
```
tigercosmos, line 49: is the tab at the beginning made by clang-format?

Dimi1010: I don't think so. I used a tab there for indentation, tho in VS it shows as a 2-space tab.
image

Common++/header/PointerVector.h
```diff
@@ -195,6 +257,29 @@ namespace pcpp
...
+			return getAndDetach(m_Vector.begin() + index);
+		}
+
+		/**
+		 * Removes an element from the vector and transfers ownership to the returned unique pointer.
+		 * @param[in, out] position An iterator pointing to the element to detach.
+		 *	The iterator is shifted to the following element after the detach completes.
+		 * @return An unique pointer that holds ownership of the detached element.
+		 */
+		std::unique_ptr<T> getAndDetach(VectorIterator& position)
```
tigercosmos, line 276: I wonder if it is necessary to return the new `position`, I think the user will only care about the return value, so maybe `const ectorIterator&`?

Common++/header/PointerVector.h
```diff
@@ -216,6 +301,57 @@ namespace pcpp
...
+			std::vector<T*> copyVec;
+
+			try
+			{
+				for (const auto iter : origin)
+				{
+					T* objCopy = new T(*iter);
+					try
+					{
+						copyVec.push_back(objCopy);
```
tigercosmos, line 320: how about `emplace_back(std::move(objCopy))`?

Common++/header/PointerVector.h
```diff
@@ -216,6 +301,57 @@ namespace pcpp
...
+		 * Frees all elements inside the vector.
+		 * Calling this function with non-heap allocated pointers is UB.
+		 * @param[in] origin The vector of elements to free.
+		 * @remarks The vector's contents are not cleared and will point to invalid locations in memory.
+		 */
+		static void freeVectorUnsafe(std::vector<T*> const& origin)
+		{
+			for (auto& obj : origin)
+			{
+				delete obj;
```
tigercosmos, line 351: do we need to check nulltpr?

Common++/header/PointerVector.h
tigercosmos: sorry, I meant between " *" and "reference".

Common++/header/PointerVector.h
tigercosmos: sure.

Common++/header/PointerVector.h
Dimi1010: The checks happen on `pushBack`. The only other way to somehow introduce `nullptr` to the vector is via assignment to the non-`const` iterators, and if you get to that point, you are kinda deliberately shooting yourself in the foot, IMO.

tigercosmos: yeah.. you're right, I think my foot is blooding now XD

Common++/header/PointerVector.h
Dimi1010: Yeah, I meant that one. I used a tab there, so its not from the formatter.

tigercosmos: I see, seems that clang-format doesn't catch it.
I think our codebase doesn't use tab in comments, could you change to spaces?

Dimi1010 (commented):
Just adding some reminders of things I noticed, that I should probably fix.

Common++/header/PointerVector.h
```diff
@@ -26,84 +28,130 @@ namespace pcpp
...
 		 */
 		PointerVector()
 		{}
 
+		/**
+		 * Copies the vector along with all elements inside it.
+		 * All elements inside the copied vector are duplicates and the originals remain unchanged.
+		 * @param[in] other The vector to copy from.
+		 * @remarks As the vector is copied via deep copy, all pointers obtained from the copied vector
+		 * reference the duplicates and not the original.
```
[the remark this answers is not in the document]
Dimi1010: Fixed 3295ddb

Common++/header/PointerVector.h
```diff
@@ -26,84 +28,130 @@ namespace pcpp
...
+		 */
+		PointerVector(const PointerVector& other) : m_Vector(deepCopyUnsafe(other.m_Vector))
+		{}
+
+		/**
+		 * Move constructor. All elements along with their ownership is transferred to the new vector.
+		 * @param[in] other The vector to move from.
+		 */
+		PointerVector(PointerVector&& other) noexcept : m_Vector(std::move(other.m_Vector))
```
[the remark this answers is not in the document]
Dimi1010: Fixed ef64030

Common++/header/PointerVector.h
```diff
@@ -26,84 +28,130 @@ namespace pcpp
...
+				m_Vector = deepCopyUnsafe(other.m_Vector);
 			}
+			// If an exception is thrown during the copy operation, restore old values and rethrow.
 			catch (const std::exception&)
 			{
-				for (auto obj : m_Vector)
-				{
-					delete obj;
-				}
+				m_Vector = oldValues;
```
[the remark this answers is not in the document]
Dimi1010: Fixed 3c8c44c

seladb (approved):

Common++/header/PointerVector.h
```diff
@@ -56,7 +56,9 @@ namespace pcpp
 		 * @param[in] other The vector to move from.
 		 */
 		PointerVector(PointerVector&& other) noexcept : m_Vector(std::move(other.m_Vector))
-		{}
+		{
+			other.m_Vector.clear();
```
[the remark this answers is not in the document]
Dimi1010: The vector move ctor technically does not have a well defined state after move. The standard just says it should be in a valid state, but it does not guarantee it will always be empty state. It ysually is, but that is implementation defined.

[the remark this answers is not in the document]
tigercosmos: I see. thanks for the explanation.

PR #1471, opened 2024-06-24
Refactored clang version check into a python script.

### LINKED ISSUE #1470: check clang-version on Windows
pre-commit calls `ci/cmake-format-all.sh` to check the version. However, it doesn't work on Windows.

Dimi1010:
Fixes #1470 
The PR changes the clang version check script from a bash script to a python script. The idea is that since pre-commit is a python package, the system running it will (hopefully) have a system-wide python interpreter installed and can run the checker script regardless of the OS.


ci/check-clang-format-version.py
```diff
@@ -0,0 +1,22 @@
+import subprocess
+
+EXPECTED_CLANG_VERSION = "18.1.6"
+
+
+def main():
+    result = subprocess.run("clang-format --version", capture_output=True)
```
tigercosmos, line 7: `subprocess.run(["clang-format", "--version"]...`

.pre-commit-config.yaml
```diff
@@ -5,8 +5,8 @@ repos:
     hooks:
       - id: check-clang-format-version
         name: Check clang-format version
-        entry: ./ci/check-clang-format-version.sh
-        language: script
+        entry: python ./ci/check-clang-format-version.py
```
tigercosmos, line 8: explicitly use Python3?

Dimi1010: Sure. `os.run` is not supported pre python 3.6 anyway.

ci/check-clang-format-version.py
```diff
@@ -0,0 +1,22 @@
...
+    result.check_returncode()
+
+    version_str = result.stdout.decode("utf-8").split(" ")[2].strip()
+    if version_str != EXPECTED_CLANG_VERSION:
+        print(
+            f"Error: Found clang-format version {version_str}, but {EXPECTED_CLANG_VERSION} is required."
+        )
+        exit(1)
```
seladb, line 15: In python the convention is to throw an exception. It's exit value is `1`:
```python
raise ValueError(f"Error: Found clang-format version {version_str}, but {EXPECTED_CLANG_VERSION} is required.")
```

ci/check-clang-format-version.py
```diff
@@ -0,0 +1,22 @@
...
+
+    print("Clang format version satisfied.")
+    exit(0)
```
seladb, line 18: This line is not needed. The default exit value is `0`

Dimi1010: Fixed c0b4713

ci/check-clang-format-version.py
Dimi1010: Fixed c0b4713

ci/check-clang-format-version.py
```diff
@@ -0,0 +1,18 @@
+import subprocess
+
+EXPECTED_CLANG_VERSION = "18.1.6"
+
+
+def main():
+    result = subprocess.run(("clang-format", "--version"), capture_output=True)
+    result.check_returncode()
+
+    version_str = result.stdout.decode("utf-8").split(" ")[2].strip()
```
seladb, line 10: nit: the default split character is space, so you can just do: `.split()`

ci/check-clang-format-version.py
Dimi1010: Fixed c6efe07

seladb (approved):
Small nit, otherwise LGTM

ci/check-clang-format-version.py
Dimi1010: Thanks. Updated in de4d3d8

PR #1457, opened 2024-06-21
check clang-format version

tigercosmos:
add a script to check  the clang-format version.

PR #1472, opened 2024-06-26
fix pre-commit clang-tidy

tigercosmos:
seems that I didn't do it correctly in https://github.com/seladb/PcapPlusPlus/pull/1457

according to the experiment, if we put clang-format in "repo: local", it doesn't function.
No matter how I add incorrect formats in the file, it doesn't throw errors.

Therefore revert the previous change.


tigercosmos:
@seladb  this PR is ready to merge.

seladb (approved):

PR #1476, opened 2024-06-30
Refactor of `PcapRemoteDeviceList`'s factory functions.

Dimi1010:
Part of #1431. implements point 6.

- Added a new factory function `createRemoteDeviceList` with overloads with/without authentication returning `unique_ptr`.
- Deprecated the old `getRemoteDeviceList` due to the functions returning raw pointers.
- Removed default constructor of `PcapRemoteDeviceList`
- Added parameterized constructor to `PcapRemoteDeviceList` for detailed initialization.
- Removed unused private methods:
  - `setRemoteMachineIpAddress`
  - `setRemoteMachinePort`
  - `setRemoteAuthentication`
- Changed typedefs to Cpp11 standard.
- Typo fixes

### DIFF
--- a/Pcap++/src/PcapRemoteDeviceList.cpp
+++ b/Pcap++/src/PcapRemoteDeviceList.cpp
@@ -74,14 +74,14 @@ PcapRemoteDeviceList* PcapRemoteDeviceList::getRemoteDeviceList(const IPAddress&
 
 std::unique_ptr<PcapRemoteDeviceList> PcapRemoteDeviceList::createRemoteDeviceList(const IPAddress& ipAddress, uint16_t port, PcapRemoteAuthentication const* remoteAuth)
 {
-	std::shared_ptr<PcapRemoteAuthentication> remoteAuthCopy;
+	std::shared_ptr<PcapRemoteAuthentication> pRemoteAuthCopy;
 	pcap_rmtauth* pRmAuth = nullptr;
 	pcap_rmtauth rmAuth;
 	if (remoteAuth != nullptr)
 	{
 		PCPP_LOG_DEBUG("Authentication requested. Username: " << remoteAuth->userName << ", Password: " << remoteAuth->password);
-		remoteAuthCopy = std::make_shared<PcapRemoteAuthentication>(*remoteAuth);
-		rmAuth = remoteAuthCopy->getPcapRmAuth();
+		pRemoteAuthCopy = std::make_shared<PcapRemoteAuthentication>(*remoteAuth);
+		rmAuth = pRemoteAuthCopy->getPcapRmAuth();
 		pRmAuth = &rmAuth;
 	}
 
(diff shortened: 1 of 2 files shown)


Pcap++/src/PcapRemoteDeviceList.cpp
```diff
@@ -48,19 +48,40 @@ namespace pcpp
...
 
 PcapRemoteDeviceList* PcapRemoteDeviceList::getRemoteDeviceList(const IPAddress& ipAddress, uint16_t port, PcapRemoteAuthentication* remoteAuth)
 {
+	auto result = PcapRemoteDeviceList::createRemoteDeviceList(ipAddress, port, remoteAuth);
+	return result.release();
+}
+
+std::unique_ptr<PcapRemoteDeviceList> PcapRemoteDeviceList::createRemoteDeviceList(const IPAddress& ipAddress, uint16_t port, PcapRemoteAuthentication const* remoteAuth)
+{
+	std::shared_ptr<PcapRemoteAuthentication> sRemoteAuth;
```
tigercosmos, line 77: I don't like the naming `sRemoteAuth`, it's not the convention in this project.

Pcap++/src/PcapRemoteDeviceList.cpp
```diff
@@ -75,20 +96,29 @@ PcapRemoteDeviceList* PcapRemoteDeviceList::getRemoteDeviceList(const IPAddress&
...
+			devices.push_back(pNewRemoteDevice.get());
+			pNewRemoteDevice.release();
+		}
+	}
+	catch (const std::exception& e)
 	{
-		PcapRemoteDevice* pNewRemoteDevice = new PcapRemoteDevice(currInterface, resultList->m_RemoteAuthentication,
-				resultList->getRemoteMachineIpAddress(), resultList->getRemoteMachinePort());
-		resultList->m_RemoteDeviceList.push_back(pNewRemoteDevice);
+		for (auto dev : devices)
```
tigercosmos, line 113: `dev` is confusing, but since the code is short here, so probably it is fine.

Pcap++/src/PcapRemoteDeviceList.cpp
tigercosmos: however, we don't have many `std::shared_ptr` currently.  So if other people agree with this, this can become the official coding style.

Dimi1010: Originally, I had a private overload where `remoteAuth` was a shared ptr to avoid that, but scrapped it as it was not strictly nessesary. 

Perhaps that would be a better solution?

Dimi1010: Tbh, I just named it that way coz i needed the variable to not conflict. If there are suggestions on a potential name, i'm happy to hear them.

Pcap++/src/PcapRemoteDeviceList.cpp
Dimi1010: Fixed edfa5de

Pcap++/src/PcapRemoteDeviceList.cpp
Dimi1010: Changed it to `remoteAuthCopy` to be more explicit that its a copy in 4f66704. Not sure if it should be `remoteAuthCopy` or `pRemoteAuthCopy` tho.

tigercosmos: I am fine with either, but seems `pXXX` is more common in the codebase?

Dimi1010: Updated cf3d1c2

seladb (approved):

PR #1483, opened 2024-07-05
cmake: use Ccache if it's available

clementperon:
If ccache tool is available use it for both compiling and linking

I will introduce in another MR the save/restore of the ccache database for github ci runners


tigercosmos:
Just triggered the CI. Let's see if CI shows there is a CCache or not.

tigercosmos:
Checked the CI, seems that we need to set up CCache ourselves

seladb (approved):

Issue #1384, opened 2024-05-05
Find alternative package for netifaces 


zhengfeihe:
I found `netifaces` package is no longer under maintained since python3.9 , detailed info here: https://github.com/al45tair/netifaces/issues/78. Some other projects have moved to using other packages, eg `psutil`: https://github.com/ros2/ros2cli/pull/875

Maybe we can also move to using other under maintained, reliable packages as well.

zhengfeihe:
I'd like to work on this after I finish my current pr.

seladb:
@zhengfeihe are you still planning to work on this?

seladb:
By the way, there is this project: https://github.com/SamuelYvon/netifaces-2

But I'm not sure it'll work well on FreeBSD: https://github.com/SamuelYvon/netifaces-2/issues/36

tigercosmos:
FreeBSD can be treated as a special case, I guess.

seladb:
> FreeBSD can be treated as a special case, I guess.

Maybe... or we can try to fix this package to work with FreeBSD, or we can choose another library 🙂

zhengfeihe:
> @zhengfeihe are you still planning to work on this?

Yes.. Just got delayed onmy current pr. I will start working on this soon.

tigercosmos:
@zhengfeihe  I think you might want to work on this, going to assign you.

zhengfeihe:
> @zhengfeihe I think you might want to work on this, going to assign you.

Thanks! I already started.

tigercosmos:
As we need to support Ubuntu24.04, this issue seems to become critical at the moment.
Some error happens in the CI of the PR #1486: [Log](https://github.com/seladb/PcapPlusPlus/actions/runs/9825638625/job/27126039912)

tigercosmos:
@zhengfeihe  please confirm if you can submit the PR for this by today; otherwise, I am sorry that I have to unassign you and take over this issue.

zhengfeihe:
> @zhengfeihe please confirm if you can submit the PR for this by today; otherwise, I am sorry that I have to unassign you and take over this issue.

I will submit the pr by the end of the day.

tigercosmos:
addressed in https://github.com/seladb/PcapPlusPlus/pull/1489

PR #1486, opened 2024-07-06
support ubuntu 24 in CI

tigercosmos:
- use `venv` for pip installation
- add Ubuntu 24.04 in the CI


seladb:
@tigercosmos we need to install venv on all Linux containers to make it work. In Ubuntu the package is `python3-venv` I think, we need to figure out what the package name in the other distros

tigercosmos:
> @tigercosmos we need to install venv on all Linux containers to make it work. In Ubuntu the package is `python3-venv` I think, we need to figure out what the package name in the other distros

I see, let me figure out.

tigercosmos:
blocking by #1384 due to:
```
  DEPRECATION: netifaces is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
```

.github/workflows/build_and_test.yml
```diff
@@ -80,6 +84,19 @@ jobs:
...
+          if [ "${{ matrix.image }}" = "alpine317" ]; then
+            apk add python3 py3-virtualenv py3-virtualenv python3-dev gcc musl-dev linux-headers
+          elif [ "${{ matrix.image }}" = "rhel94" ]; then
+            yum install -y python3 python3-virtualenv python3-devel
+          elif [ "${{ matrix.image }}" = "fedora39" ]; then
+            dnf install -y python3 python3-virtualenv python3-devel
+          else
+            apt-get update
+            apt-get install -y python3 python3-venv python3-dev
+          fi
```
clementperon, line 98: this should be done in the container

.github/workflows/build_and_test.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           apk update && apk add cppcheck python3-dev
           python3 -m pip install cmake-format clang-format==18.1.6
 
+      # TODO: investigate how to run pre-commit with `venv`
       - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
```
clementperon, line 35: Did you try 
```suggestion
      - name: Activate virtualenv
        run: |
          ${{ matrix.python }} -m venv ./venv
          . .venv/bin/activate
          echo PATH=$PATH >> $GITHUB_ENV

      - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
```

.github/workflows/build_and_test.yml
```diff
@@ -52,6 +53,9 @@ jobs:
     strategy:
       matrix:
         include:
+          - image: ubuntu2404
+            python: python3
```
clementperon, line 57: now that everybody use python: python3 maybe we could drop it no ?

.github/workflows/build_and_test.yml
```diff
@@ -97,15 +114,19 @@ jobs:
...
         run: |
-          ${{ matrix.python }} -m pip install -U pip
+          ${{ matrix.python }} -m venv ./venv
+          source ./venv/bin/activate
+          ${{ matrix.python }} -m pip install pip
           ${{ matrix.python }} -m pip install -r ci/run_tests/requirements.txt
           ${{ matrix.python }} ci/run_tests/run_tests.py --interface eth0 ${{ matrix.test-flags }}
 
       - name: Test Examples
+        shell: bash
```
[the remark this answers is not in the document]
tigercosmos: the default shell is `sh`, and to `source` the venv activator, we need bash.

.github/workflows/build_and_test.yml
```diff
@@ -97,15 +114,19 @@ jobs:
         run: cmake --build "$BUILD_DIR" -j
 
       - name: Test PcapPlusPlus
+        shell: bash
         run: |
-          ${{ matrix.python }} -m pip install -U pip
+          ${{ matrix.python }} -m venv ./venv
+          source ./venv/bin/activate
+          ${{ matrix.python }} -m pip install pip
```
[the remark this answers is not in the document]
tigercosmos: not sure, maybe yes, because we use venv now.

.github/workflows/build_and_test.yml
tigercosmos: ya, we can do that as well, @seladb can you update the containers?

clementperon: I can do it if you want, but you just need to open a MR here:
https://github.com/seladb/PcapPlusPlus-DockerImages

tigercosmos: cool, I didn't know this.

tigercosmos: @seladb  shouldn't [/PcapPlusPlus-DockerImages](https://github.com/seladb/PcapPlusPlus-DockerImages) under the organization?

seladb: I guess we can move it, but maybe do it after the release?

.github/workflows/build_and_test.yml
clementperon: @tigercosmos it shoul be "source ./venv/bin/activate"

tigercosmos: `source` is only for bash.  However, I cannot control to let pre-commit action to use bash. I will just convert the change, and I don't see any issue that pre-commit doesn't run in the venv.

.github/workflows/build_and_test.yml
tigercosmos: got it.

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
tigercosmos: fixed.

.github/workflows/build_and_test.yml
tigercosmos: removed all.

.github/workflows/build_and_test.yml
tigercosmos: fixed.

tigercosmos:
@seladb  ready for review.

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
seladb: In `sh`, instead of using `source` you can use `.`:
```
. ./venv/bin/activate
```

https://unix.stackexchange.com/questions/748962/how-to-activate-python-venv-in-a-non-bash-shell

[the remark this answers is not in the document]
tigercosmos: I tried, but it doesn't work.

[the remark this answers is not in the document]
tigercosmos: okay, let me try again.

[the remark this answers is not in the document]
tigercosmos: ah, well, seems it works :)

.github/workflows/build_and_test.yml
```diff
@@ -32,6 +32,7 @@ jobs:
           apk update && apk add cppcheck python3-dev
           python3 -m pip install cmake-format clang-format==18.1.6
 
+      # TODO: investigate how to run pre-commit with `venv`
```
egecetin, line 35: Not related with this PR directly but, it is possible to use [pre-commit.ci](https://https://pre-commit.ci/) but we should remove cppcheck and clang-format from pre-commit since they require pre-installed binaries.

clementperon (approved):

.github/workflows/build_and_test.yml
```diff
@@ -111,16 +107,17 @@ jobs:
 
       - name: Test PcapPlusPlus
         run: |
-          ${{ matrix.python }} -m pip install -U pip
-          ${{ matrix.python }} -m pip install -r ci/run_tests/requirements.txt
-          ${{ matrix.python }} ci/run_tests/run_tests.py --interface eth0 ${{ matrix.test-flags }}
+          python3 -m venv ./venv
```
Dimi1010, line 110: nit: Maybe move this line to it's own step? It looks weird to me that the venv is created inside the 'Test PCPP' step but the following steps also depend on it.

.github/workflows/build_and_test.yml
```diff
@@ -347,15 +345,15 @@ jobs:
...
           python -m pip install -r ci/run_tests/requirements.txt
           python ci/run_tests/run_tests.py --interface en0
 
       - name: Test Examples
         if: ${{ matrix.host-arch == matrix.arch }}
         run: |
           cd Tests/ExamplesTest
-          python -m pip install -U pip
           python -m pip install -r requirements.txt
           python -m pytest --interface en0 --use-sudo --root-path=../../Dist/examples_bin
```
Dimi1010, line 358: Why does the first step use 'venv' and the second doesn't?

seladb (approved):
Please fix @Dimi1010 's comments, otherwise LGTM

.github/workflows/build_and_test.yml
tigercosmos: Sure. Moved it to a step.

PR #1487, opened 2024-07-07
Fix memory leak in PointerVector pushBack

### LINKED ISSUE #1465: Potential memory leak in PointerVector pushBack, if internal vector reallocation fails.
In large parts of the code base, `PoiterVector::pushBack()` is being used as `pVector.pushBack(new T(...));`. With the current `pushBack` implementation, if the internal call to `m_vector.push_back()` requires a reallocation and fails, the pointed element is leaked in those cases. 

Opening an issue instead of directly addressing it in a PR, because there are a couple possible ways to address it:
- Assume ownership is transferred at the `pushBack` call and clean up the element on error.
- Provide a flag `cleanupOnError` to let the call site determine if the vector should handle the cleanup or the call site.
- Assume the call site will handle cleanup of the pointed element on exception and don't do anything in the vector itself.

Dimi1010:
Fixes #1465

- Added a new flag to raw pointe push_back that automatically frees the element if the `pushBack` operation fails.
- Added a new overload disallowing direct pushing of `nullptr_t` into the vector.


Common++/header/PointerVector.h
```diff
@@ -121,19 +122,36 @@ namespace pcpp
 			m_Vector.clear();
 		}
 
+		/**
+		 * Adding a nullptr to the vector is not allowed.
+		 */
+		void pushBack(std::nullptr_t element, bool freeElementOnError = false) = delete;
```
tigercosmos, line 128: it's pretty design.

Common++/header/PointerVector.h
```diff
@@ -121,19 +122,36 @@ namespace pcpp
...
+
 		/**
 		 * Add a new (pointer to an) element to the vector
 		 * @param[in] element A pointer to an element to assume ownership of.
+		 * @param[in] freeElementOnError If set to true, the element is freed if an exception is thrown during the push.
 		 * @throws std::invalid_argument The provided pointer is a nullptr.
 		 */
-		void pushBack(T* element)
+		void pushBack(T* element, bool freeElementOnError = false)
```
tigercosmos, line 136: I wonder if setting `freeElementOnError` default as `true` makes sense? Seems that most of the time we set it as `true`.

Dimi1010: Maybe, @seladb wanted it set to false [here](https://github.com/seladb/PcapPlusPlus/issues/1465#issuecomment-2211723443) tho

tigercosmos: I see, @seladb do you consider let it to be true?

seladb: I thought about it more and changed my mind: https://github.com/seladb/PcapPlusPlus/issues/1465#issuecomment-2252113356

tigercosmos: Cool, it makes sense to me as well.

seladb (approved):

PR #1456, opened 2024-06-21
 apply clang-format to `Tests`

tigercosmos:
part of https://github.com/seladb/PcapPlusPlus/issues/1248

PR #1370, opened 2024-04-20
cmake: add reproducible build option

clementperon:
In order to validate that the clang-format doesn't mess our code we could check if the build is reproductible or not.
https://github.com/seladb/PcapPlusPlus/pull/1257

This has been tested with MacOS / Clang.
LibPacket++.a is properly rebuilt from scratch and shasum are equal.

Needs more test under Linux and Windows

PR #1488, opened 2024-07-07
Consolidate shared DeviceList functionality under a common base class.

Dimi1010:
## Summary
This is a rework of #1431, parts 1 and 2.

The list classes now provide direct functions for iteration and fetching devices. The previous approach of retrieving another 'List' from a 'List' class was redundant and tied the internal implementation to it, making future changes more difficult.

### Changes:
- **PointerVector<T>**:
  - Introduced a deleter extension point by changing it to `PointerVector<T, Deleter>`.

- **DeviceListBase<DeviceType>**:
  - Added a new base class `DeviceListBase<DeviceType>` to implement common functionality shared among device lists.
  - Common functionality includes:
    - A protected `PointerVector<DeviceType>` member variable.
    - Element accessor methods: `at()`, `front()`, `back()`.
    - Iterator accessor methods: `begin()`, `end()`, `cbegin()`, `cend()`.
    - Capacity accessors: `size()`, `empty()`.

- **Refactoring**:
  - Refactored all device list classes to inherit from `DeviceListBase`.
  - Removed the iterator API implementation from `RemoteDeviceList`, which is now superseded by the inherited iterator API from `DeviceListBase`.
  - Deprecated the `getPcap*DevicesList` methods, which have been superseded by the direct accessor methods provided by `DeviceListBase`.

### Affected Lists:
- `PcapLiveDeviceList`
- `PfRingDeviceList`
- `DpdkDeviceList`

- **Deprecated Methods**:
  - Updated all usages of deprecated methods in the library to use the new, non-deprecated methods.


Pcap++/src/PcapRemoteDeviceList.cpp
```diff
@@ -83,9 +84,9 @@ PcapRemoteDeviceList* PcapRemoteDeviceList::getRemoteDeviceList(const IPAddress&
 
 	for (pcap_if_t* currInterface = interfaceList.get(); currInterface != nullptr; currInterface = currInterface->next)
 	{
-		PcapRemoteDevice* pNewRemoteDevice = new PcapRemoteDevice(currInterface, resultList->m_RemoteAuthentication,
-				resultList->getRemoteMachineIpAddress(), resultList->getRemoteMachinePort());
-		resultList->m_RemoteDeviceList.push_back(pNewRemoteDevice);
+		std::unique_ptr<PcapRemoteDevice> pNewRemoteDevice = std::unique_ptr<PcapRemoteDevice>(
```
tigercosmos, line 87: use `auto` to shorten the line.

Pcap++/src/PcapRemoteDeviceList.cpp
```diff
@@ -83,9 +84,9 @@ PcapRemoteDeviceList* PcapRemoteDeviceList::getRemoteDeviceList(const IPAddress&
...
+			new PcapRemoteDevice(currInterface, resultList->m_RemoteAuthentication, resultList->getRemoteMachineIpAddress(), resultList->getRemoteMachinePort()));
```
tigercosmos, line 88: a little too long, please consider splitting it.

Pcap++/src/PcapRemoteDeviceList.cpp
Dimi1010: Fixed b479b08

Pcap++/src/PcapRemoteDeviceList.cpp
Dimi1010: Fixed b479b08. Should be better now.

Pcap++/src/PcapRemoteDeviceList.cpp
```diff
@@ -84,8 +84,9 @@ PcapRemoteDeviceList* PcapRemoteDeviceList::getRemoteDeviceList(const IPAddress&
 
 	for (pcap_if_t* currInterface = interfaceList.get(); currInterface != nullptr; currInterface = currInterface->next)
 	{
-		std::unique_ptr<PcapRemoteDevice> pNewRemoteDevice = std::unique_ptr<PcapRemoteDevice>(
-			new PcapRemoteDevice(currInterface, resultList->m_RemoteAuthentication, resultList->getRemoteMachineIpAddress(), resultList->getRemoteMachinePort()));
+		auto pNewRemoteDevice = std::unique_ptr<PcapRemoteDevice>(
```
tigercosmos, line 87: I understand why you put the `new PcapRemoteDevice` inside `std::uique_ptr`, but I think it's safe to do the following here, which should be easier to read(?)
```
auto newRemoteDevice = new PcapRemoteDevice(currInterface, resultList->m_RemoteAuthentication, resultList->getRemoteMachineIpAddress(), resultList->getRemoteMachinePort()));
std::unique_ptr<PcapRemoteDevice> pNewRemoteDevic(newRemoteDevice);
```
maybe the first line can be split by the arguments as well.

Dimi1010: Hmm, It is possible but I kinda dislike doubling up on the pointer variables.

I suppose we can do this, if it is more clear?:
```
auto pNewRemoteDevice = std::unique_ptr<PcapRemoteDevice>(
	new PcapRemoteDevice(
		currInterface,
		resultList->m_RemoteAuthentication, 
		resultList->getRemoteMachineIpAddress(), 
		resultList->getRemoteMachinePort()
	)
);
```
But it is just going to get squashed to this:
```
auto pNewRemoteDevice = std::unique_ptr<PcapRemoteDevice>(
    new PcapRemoteDevice(currInterface, resultList->m_RemoteAuthentication,
                         resultList->getRemoteMachineIpAddress(), resultList->getRemoteMachinePort()));
```
when the formatting comes to it.

tigercosmos: I think it's clearer. don't hesitate to use `// clang-format off` if the formatted is no good.

Dimi1010: Fixed the formatting in 6a4c330

Dimi1010:
@tigercosmos added `PfRingDeviceList` to the refactor last minute. Sry for the re-review request.

tigercosmos (approved):

seladb:
@Dimi1010 CI fails for some reason in generating the codecov report, and idea why? 🤔

clementperon:
> Owner
> ###

I have the same issue in my MR 
https://github.com/seladb/PcapPlusPlus/pull/1370

I didn't touch to the code at all

seladb:
@clementperon `gcovr` complains it finds the same method `pcpp::PcapLiveDeviceList::getInstance()` in multiple lines: 51 ,63. 
Apparently line 51 is the location of the method before the change, and line 63 is the location after the change:
image

gcovr also complains about a new file version when it compares the file to the file in `.ccache`. Maybe the issue is related to ccache we just added?

```
Stderr of gcov was >>/__w/PcapPlusPlus/PcapPlusPlus/Examples/IcmpFileTransfer/Common.cpp:source file is newer than notes file '__w/PcapPlusPlus/PcapPlusPlus/.ccache/d/d/fac836f17d373ef9882523799d4b3d-1445515.gcno'
(the message is displayed only once per source file)
/__w/PcapPlusPlus/PcapPlusPlus/Packet++/header/Packet.h:source file is newer than notes file '__w/PcapPlusPlus/PcapPlusPlus/.ccache/d/d/fac836f17d373ef9882523799d4b3d-1445515.gcno'
/__w/PcapPlusPlus/PcapPlusPlus/Packet++/header/IcmpLayer.h:source file is newer than notes file '__w/PcapPlusPlus/PcapPlusPlus/.ccache/d/d/fac836f17d373ef9882523799d4b3d-1445515.gcno'
/__w/PcapPlusPlus/PcapPlusPlus/Packet++/header/IPv4Layer.h:source file is newer than notes file '__w/PcapPlusPlus/PcapPlusPlus/.ccache/d/d/fac836f17d373ef9882523799d4b3d-1445515.gcno'
/__w/PcapPlusPlus/PcapPlusPlus/Packet++/header/TLVData.h:source file is newer than notes file '__w/PcapPlusPlus/PcapPlusPlus/.ccache/d/d/fac836f17d373ef9882523799d4b3d-1445515.gcno'
/__w/PcapPlusPlus/PcapPlusPlus/Packet++/header/IPLayer.h:source file is newer than notes file '__w/PcapPlusPlus/PcapPlusPlus/.ccache/d/d/fac836f17d373ef9882523799d4b3d-1445515.gcno'
/__w/PcapPlusPlus/PcapPlusPlus/Packet++/header/EthLayer.h:source file is newer than notes file '__w/PcapPlusPlus/PcapPlusPlus/.ccache/d/d/fac836f17d373ef9882523799d4b3d-1445515.gcno'
/__w/PcapPlusPlus/PcapPlusPlus/Common++/header/PcapPlusPlusVersion.h:source file is newer than notes file '__w/PcapPlusPlus/PcapPlusPlus/.ccache/d/d/fac836f17d373ef9882523799d4b3d-1445515.gcno'
/__w/PcapPlusPlus/PcapPlusPlus/Pcap++/header/PcapLiveDeviceList.h:source file is newer than notes file '__w/PcapPlusPlus/PcapPlusPlus/.ccache/d/d/fac836f17d373ef9882523799d4b3d-1445515.gcno'
__w/PcapPlusPlus/PcapPlusPlus/.ccache/d/d/fac836f17d373ef9882523799d4b3d-1445515.gcda:cannot open data file, assuming not executed
<< End of stderr
Exception was >>Got function pcpp::PcapLiveDeviceList::getInstance() on multiple lines: 51, 63.
	You can run gcovr with --merge-mode-functions=MERGE_MODE.
	The available values for MERGE_MODE are described in the documentation.<< End of stderr
```

Dimi1010:
@seladb @clementperon It probably has something to do with `ccache`. Look at #1487. Commit [b13d203](https://github.com/seladb/PcapPlusPlus/pull/1487/commits/b13d203977b4705a10425d8fd6e16595c9151387) passes all checks and then after merge, commit [e418d41](https://github.com/seladb/PcapPlusPlus/pull/1487/commits/e418d41f35cdf80a120f740700e7cec53ed21383) suddenly fails its report with a similar issue. The only notable things merged to the dev branch between the 2 commits are #1476, #1492 and #1495.

egecetin:
@seladb @clementperon @Dimi1010 Maybe remove all codecov generated files in source tree with grep/wildcard from extension just before saving ccache?

seladb:
> @seladb @clementperon @Dimi1010 Maybe remove all codecov generated files in source tree with grep/wildcard from extension just before saving ccache?

@egecetin I reverted the codecov upgrade, please see my message on Slack. I'm not sure it'll fix this issue, but it'll probably fix the CI failure in this PR: https://github.com/seladb/PcapPlusPlus/pull/1456

Pcap++/header/PcapLiveDeviceList.h
```diff
@@ -56,7 +68,9 @@ namespace pcpp
 
 		/**
 		 * @return A vector containing pointers to all live devices currently installed on the machine
+		 * @deprecated This method has been deprecated in favor of begin/end iteration API.
 		 */
+		PCPP_DEPRECATED("Deprecated in favor of begin/end iteration API")
 		const std::vector<PcapLiveDevice*>& getPcapLiveDevicesList() const { return m_LiveDeviceListView; };
```
seladb, line 74: There are things you can do with a vector that you can't do with an iterator, such as getting the element at index X, for example: `PcapLiveDeviceList::getInstance().getPcapLiveDevicesList()[1]`, or get the first or last element. 

Of course we can implement all of those as part of `PcapLiveDeviceList`, but isn't it easier to just expose the vector? 🤔 

And if we expose the vector it's probably a bit confusing to also expose an iterator over the `PcapLiveDeviceList` class itself

Dimi1010: The argument about indexing is true, tho I don't really agree that exposing the vector is theright choice. It might be easier to do, but it will cause more headaches in the long run. 

One of which we already encountered when refactoring the pointer management to be automatic w.r.t the current class. That issue being the nessesity to have a vector alive (the "view" class) to return a reference to. This is the primary reason I have for this deprecation.

Also while I already mentioned it, conceptually, it does not look right to me to have a class named 'List' have to be queried for another container. If it was called 'Manager' or something like that, then fine, I guess. IMO 'List' implies a container and at the moment, what we have does not behave like a container.

seladb: > The argument about indexing is true, tho I don't really agree that exposing the vector is the right choice. It might be easier to do, but it will cause more headaches in the long run.

Which headaches do you think it'll cause in the long run?



> One of which we already encountered when refactoring the pointer management to be automatic w.r.t the current class. That issue being the nessesity to have a vector alive (the "view" class) to return a reference to. This is the primary reason I have for this deprecation.

I agree that keeping the view isn't amazing, but it's good enough I think, so we can return the vector to the user if needed



> Also while I already mentioned it, conceptually, it does not look right to me to have a class named 'List' have to be queried for another container. If it was called 'Manager' or something like that, then fine, I guess. IMO 'List' implies a container and at the moment, what we have does not behave like a container.

I agree the name for this class could be improved, maybe "Manager" or "Container" would be better. But since those classes exist for a very long time changing them would be a pain

Dimi1010: > Which headaches do you think it'll cause in the long run?

Well, primarily the one that it binds the internal implementation to the external API, which makes potential changes to the implementation difficult, which we already encountered. I especially don't like it in these classes as they are one of the main blocks of the library and the API should be relatively stable.

> I agree that keeping the view isn't amazing, but it's good enough I think, so we can return the vector to the user if needed

Eh, I guess, I mostly added it as a temporary band-aid while I run though the deprecations... which I guess I should have known better how all temporary solutions and up...

One thing I kinda want is to unify the iteration and access API for all the different lists. At the moment it is kinda all over the place. `LiveDevice`, `PfRing` and `Dpdk` all expose a reference to vector, `RemoteDevice` exposes iterators and `Kni` exposes nothing. One of my ideas is to have a base templated 'List' container which implements all the accessors and then all the lists inherit from that.

> I agree the name for this class could be improved, maybe "Manager" or "Container" would be better. But since those classes exist for a very long time changing them would be a pain.

Eh, `List` is fine as a name, it is just that it would be better if it behaved like a proper container since it is already named like one.

seladb: Assuming we go with your suggestion, we probably need to implement most of the vector getters like: count, first, last, getByIndex, iterator, etc. right?

Dimi1010: I think these ones would be sufficient:
- at / operator[]
- front / back
- (c)begin/(c)end
- maybe r(c)begin/r(c)end
- empty/size

These are the main ones that would most likely be used.

seladb: It's a bit weird, because all of these method will call `m_LiveDeviceList` with the same method, it'll look like:
```cpp
methodX()
{
    return m_LiveDeviceList.methodX();
}
```

The only thing we gain from not exposing the vector is:
- We make `PcapLiveDeviceList` the container
- We make it read-only, so no one can modify it

seladb: > n alternative could be to just return the result of `getAll*Devices()` by value instead of reference. That would also fix the issue of the internal implementation being bound to the API. At the expense of the list being generated every time it is called and not being updated after that (due to `reset()`, etc calls). Although I am hesitant to change a reference return to a value return directly as it might affect existing code.

If you think that's better then I'm ok with it. It will change the existing interface, but do you see any potential usage of this method that might break? 🤔

Dimi1010: > > n alternative could be to just return the result of `getAll*Devices()` by value instead of reference. That would also fix the issue of the internal implementation being bound to the API. At the expense of the list being generated every time it is called and not being updated after that (due to `reset()`, etc calls). Although I am hesitant to change a reference return to a value return directly as it might affect existing code.
> 
> If you think that's better then I'm ok with it. It will change the existing interface, but do you see any potential usage of this method that might break? 🤔

Well, I kinda threw it there as a potential alternative. Tho looking into it, the change from a reference to a value return looks like a minefield. 

See this code for example.
```
struct Object
{
	std::vector<int> const& getValueByRef() { return m_value; }
	std::vector<int> const getValue() { return m_value; }
	
	std::vector<int> m_value;
};

int main()
{ 
	Object obj;
        
        // This a fine and possibly common usage of the current implementation.
	std::vector<int> const& valueRef = obj.getValueByRef();
        // This is what the code might look like after a direct change.
        // It returns a temporary which's address is assigned to the reference, which compiles but is UB.
        // How long does the temporary live? Nobody knows, so the reference can become dead at any moment.
	std::vector<int> const& valueRef2 = obj.getValue(); 
	return 0;
}
```

So yeah... the return by value would require its own separate function to not break a lot of stuff. :/

On the other hand... tried to do the original idea of having public access methods similar to vector and I think it went pretty well, especially with the base class.

Pcap++/src/DpdkDeviceList.cpp
```diff
@@ -192,11 +183,15 @@ bool DpdkDeviceList::initDpdkDevices(uint32_t mBufPoolSizePerDevice, uint16_t mB
...
-		DpdkDevice* newDevice = new DpdkDevice(i, mBufPoolSizePerDevice, mBufDataSize);
+		auto newDevice = std::unique_ptr<DpdkDevice>(new DpdkDevice(i, mBufPoolSizePerDevice, mBufDataSize));
 		PCPP_LOG_DEBUG("DpdkDevice #" << i << ": Name='" << newDevice->getDeviceName() << "', PCI-slot='" << newDevice->getPciAddress() << "', PMD='" << newDevice->getPMDName() << "', MAC Addr='" << newDevice->getMacAddress() << "'");
-		m_DpdkDeviceList.push_back(newDevice);
+		m_DeviceList.pushBack(std::move(newDevice));
 	}
 
+	// Full update of all elements of the view vector to synchronize them with the main vector.
+	m_DpdkDeviceListView.resize(m_DeviceList.size());
+	std::copy(m_DeviceList.begin(), m_DeviceList.end(), m_DpdkDeviceListView.begin());
```
tigercosmos, line 193: This can be simplified to `m_DpdkDeviceListView = m_DeviceList`, and usually this is slightly faster.

Pcap++/src/DpdkDeviceList.cpp
```diff
@@ -206,15 +201,15 @@ DpdkDevice* DpdkDeviceList::getDeviceByPort(int portId) const
 	if (!isInitialized())
 	{
 		PCPP_LOG_ERROR("DpdkDeviceList not initialized");
-		return NULL;
+		return nullptr;
 	}
 
-	if ((uint32_t)portId >= m_DpdkDeviceList.size())
+	if (static_cast<uint32_t>(portId) >= m_DeviceList.size())
```
tigercosmos, line 207: isn't `size()` `size_t`?

Pcap++/src/KniDeviceList.cpp
```diff
@@ -86,71 +83,65 @@ KniDevice* KniDeviceList::createDevice(
 )
 {
 	if (!isInitialized())
-		return NULL;
+		return nullptr;
```
tigercosmos, line 85: Better to add curly braces for it.

Pcap++/src/DpdkDeviceList.cpp
```diff
@@ -192,11 +183,15 @@ bool DpdkDeviceList::initDpdkDevices(uint32_t mBufPoolSizePerDevice, uint16_t mB
 	// Initialize a DpdkDevice per port
 	for (int i = 0; i < numOfPorts; i++)
 	{
-		DpdkDevice* newDevice = new DpdkDevice(i, mBufPoolSizePerDevice, mBufDataSize);
+		auto newDevice = std::unique_ptr<DpdkDevice>(new DpdkDevice(i, mBufPoolSizePerDevice, mBufDataSize));
 		PCPP_LOG_DEBUG("DpdkDevice #" << i << ": Name='" << newDevice->getDeviceName() << "', PCI-slot='" << newDevice->getPciAddress() << "', PMD='" << newDevice->getPMDName() << "', MAC Addr='" << newDevice->getMacAddress() << "'");
-		m_DpdkDeviceList.push_back(newDevice);
+		m_DeviceList.pushBack(std::move(newDevice));
```
tigercosmos, line 188: Is `std::move` needed? If so, KniDeviceList.cpp:105 also needs it.

Dimi1010: Move is needed here because `newDevice` is a `unique ptr`. In KniDevice, I explicitly left it as a raw ptr, because the private delete on KniDevice was causing issues with the unique_ptr and it needed to be returned as a result afterwards.

tigercosmos: sure.

Pcap++/src/DpdkDeviceList.cpp
Dimi1010: I'm pretty sure that is not gonna work. `m_DeviceList` is a `PointerVector<T>` while `m_DpdkDeviceListView` is `std::vector<T>`.

I tested it with this code snippet:
```
pcpp::PointerVector<int> pVec;
std::vector<int*> vec;

pVec.pushBack(new int(1));
pVec.pushBack(new int(2));
pVec.pushBack(new int(3));

vec = pVec;
```

And it gives a no matching operator '=' compile time error.

Pcap++/src/KniDeviceList.cpp
Dimi1010: Fixed 2ae51d2

Pcap++/src/DpdkDeviceList.cpp
Dimi1010: Fixed 2ae51d2

Pcap++/src/DpdkDeviceList.cpp
tigercosmos: Ops, I didn't notice that the types are different. Sorry for the confusing.

Pcap++/header/PcapLiveDeviceList.h
```diff
@@ -21,10 +23,9 @@ namespace pcpp
 	 * devices are initialized on startup and wrap the network interfaces installed on the machine. This class enables access to them through
 	 * their IP addresses or get a vector of all of them so the user can search them in some other way
 	 */
-	class PcapLiveDeviceList
+	class PcapLiveDeviceList : public internal::DeviceListBase<PcapLiveDevice>
```
seladb, line 26: If we make this class a container, maybe we also want to rename the other methods like:
- `getPcapLiveDeviceByIp` -> `getByIp`
- `getPcapLiveDeviceByName` -> `getByName`
- `getPcapLiveDeviceByIpOrName` -> `getByIpOrName`

We can deprecate the old names and later remove them.

Also, `getDnsServers()` doesn't seems to really belong here, but I don't know what to do with it...

Pcap++/header/PfRingDeviceList.h
```diff
@@ -46,7 +48,9 @@ namespace pcpp
 		/**
 		 * Return a list of all available PF_RING devices
 		 * @return a list of all available PF_RING devices
+		 * @deprecated This method has been deprecated in favor of direct accessor API.
 		 */
+		PCPP_DEPRECATED("Deprecated in favor of direct accessor API")
 		const std::vector<PfRingDevice*>& getPfRingDevicesList() const { return m_PfRingDeviceListView; }
```
seladb, line 54: You missed updating it in the tests: https://github.com/Dimi1010/PcapPlusPlus/blob/2ae51d24cc81103afa70d28995011cfe965d829e/Tests/Pcap%2B%2BTest/Tests/PfRingTests.cpp#L203

Pcap++/src/DpdkDeviceList.cpp
seladb: Why should this be a `unique_ptr` and not a raw pointer? 🤔

Pcap++/header/DpdkDeviceList.h
seladb, line 1: did you have a chance to test it? Our tests don't cover DPDK unfortunately, mostly because DPDK is not easy to test in CI...

Pcap++/header/PcapRemoteDeviceList.h
```diff
@@ -26,37 +27,36 @@ namespace pcpp
...
 		 * Iterator object that can be used for iterating all PcapRemoteDevice in list
 		 */
-		using RemoteDeviceListIterator = typename std::vector<PcapRemoteDevice*>::iterator;
+		using RemoteDeviceListIterator = iterator;
 
 		/**
 		 * Const iterator object that can be used for iterating all PcapRemoteDevice in a constant list
 		 */
-		using ConstRemoteDeviceListIterator = typename std::vector<PcapRemoteDevice*>::const_iterator;
+		using ConstRemoteDeviceListIterator = const_iterator;
```
seladb, line 53: Maybe we can mark them as deprecated? User can use `iterator` and `const_iterator`

Pcap++/header/DeviceListBase.h
```diff
@@ -0,0 +1,167 @@
...
+namespace pcpp
+{
+	namespace internal
+	{
+		/**
+		 * @class DeviceListBase
+		 * A base class for all device lists in PcapPlusPlus. This class is used to store a list of devices and provide
+		 * access to them
+		 */
+		template <class DeviceType, class DeviceDeleter = std::default_delete<DeviceType>> class DeviceListBase
```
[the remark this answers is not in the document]
Dimi1010: I would prefer keeping it separate from `PointerVector` as an abstraction layer. 

Changing it to a generic `ReadOnlyPointerVector` would mean that it can't keep the protected destructor (as that may be used in other places) and will have to have a virtual public destructor to not cause issues with deletion from base pointer.

Also, it might allow centralising other shared functions between the different lists if there are any to be had.

Pcap++/header/PcapLiveDeviceList.h
Dimi1010: Sure. Tbh, it might even be possible to get some those into the base class with some template magic fuckery, as some are shared (LiveDevice and RemoteDevice basically share the same getter methods).

Pcap++/src/DpdkDeviceList.cpp
Dimi1010: I try to use unique ptr whereever possible for the RAII deletion and cleared ownership semantics.

Pcap++/header/DpdkDeviceList.h
Dimi1010: Not really, as I don't have dpdk installed. Does it work on windows?

Pcap++/header/PcapRemoteDeviceList.h
Dimi1010: Deprecated them in cee1e43

Pcap++/header/PfRingDeviceList.h
Dimi1010: Updated ab45201

Pcap++/header/PcapLiveDeviceList.h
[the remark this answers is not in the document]
Dimi1010: Updated 75c4e32

Pcap++/header/DeviceListBase.h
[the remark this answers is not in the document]
seladb: In order to support a protected d'tor we can create `ReadOnlyPointerVector` with a non-protected destructor, and then inherit from it and add a protected d'tor, or just add a protected d'tor to all list classes

Pcap++/header/DpdkDeviceList.h
seladb: DPDK doesn't work on Windows (as far as I know). But you can test it on a Linux VM. If you're not familiar with please let me know and I'll test it

Pcap++/header/PcapRemoteDeviceList.h
seladb: Maybe you should put `PCPP_DEPRECATED` above the `using` line instead of inline, and then MinGW will compile? 🤔

seladb: Thinking about it again - not sure putting the macro above the `using` statement will do what we want, but I'm not 100% sure... Worst case we can either drop `RemoteDeviceListIterator` and `ConstRemoteDeviceListIterator` completely, even though it's a breaking change (I don't believe many people use it anyway), or just keep it without deprecation

Pcap++/header/DeviceListBase.h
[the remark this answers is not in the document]
Dimi1010: > In order to support a protected d'tor we can create ReadOnlyPointerVector with a non-protected destructor, and then inherit from it and add a protected d'tor, or just add a protected d'tor to all list classes.

That does not solve the problem tho... the protected destructor is to prevent accidental deletion from a base class pointer, as `DeviceListBase`'s inheritance is public and avoiding the need for having `virtual` destructor which adds runtime overhead. Having a base class above `DeviceListBase` with a public destructor just pushes the problem up the chain, as now code like this is valid:
```
ReadOnlyPointerVector<T>* p = new DerivedFromDeviceBaseList();
delete p; // Calls ~ReadOnlyPointerVector(). Does not destruct anything down the inheritance chain.
```

This can be solved by inheriting privately from a potential `ReadOnlyPointerVector` but really do we even need a general purpose `ReadOnlyPointerVector` at the moment? What would be the difference between that and using `PointerVector<T> const&` where it is needed. I'm of the opinion to leave it be for now and add a `ReadOnlyPointerVector` if another place actually requires it.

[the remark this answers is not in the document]
Dimi1010: I mean, the underlying type does not need to be an actual device for `DeviceListBase` either. It can be a mock device for the unit tests.

Pcap++/header/PcapLiveDeviceList.h
Dimi1010: Changed the functions to syntax `getDeviceBy*`. It is shorter and `Dpdk` and `Kni` lists already used that syntax.

Related commits: a73d4b9, ffac432, 151f09f.

Pcap++/header/PcapRemoteDeviceList.h
Dimi1010: Ok, had to break out my voodoo doll but managed to get it working in bb695b8.

Pcap++/header/DpdkDeviceList.h
Dimi1010: @seladb might be quicker for you to test it, as I am slightly confused its installation process.

seladb:
@Dimi1010 although it's not a breaking change, it's a pretty big change that I'm not sure we should introduce just before the upcoming release. We already have a few deprecation and breaking changes...

What do you think about waiting with this PR until the release?

Dimi1010:
@seladb Sure, no problem.

tigercosmos:
> @Dimi1010 although it's not a breaking change, it's a pretty big change that I'm not sure we should introduce just before the upcoming release. We already have a few deprecation and breaking changes...
> 
> What do you think about waiting with this PR until the release?

I'm also uncertain about the statement. Even if we don't include this in the upcoming release, the next one is likely to introduce even more breaking changes.

seladb:
> > @Dimi1010 although it's not a breaking change, it's a pretty big change that I'm not sure we should introduce just before the upcoming release. We already have a few deprecation and breaking changes...
> > What do you think about waiting with this PR until the release?
> 
> I'm also uncertain about the statement. Even if we don't include this in the upcoming release, the next one is likely to introduce even more breaking changes.

I hope that we'll release the version more frequently than the existing one, so hopefully each version will include fewer API changes and breaking changes

Pcap++/src/DpdkDeviceList.cpp
```diff
@@ -34,6 +34,7 @@
 #include <rte_mbuf.h>
 #include <rte_version.h>
 
+#	include <memory>
```
Dimi1010, line 37:
```suggestion
#include <memory>
```

Pcap++/src/PfRingDeviceList.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include <cstdio>
 #include <array>
+#	include <memory>
```
Dimi1010, line 7:
```suggestion
#include <memory>
```

Pcap++/header/DeviceListBase.h
```diff
@@ -0,0 +1,167 @@
...
+			}
+
+			/**
+			 * @brief Returns a pointer to first device.
+			 * @return A pointer to the specified device.
+			 */
+			DeviceType* front()
+			{
+				return m_DeviceList.front();
+			}
```
tigercosmos, line 74: Miss an empty line.

Pcap++/header/PcapLiveDeviceList.h
```diff
@@ -65,42 +70,108 @@ namespace pcpp
...
+		}
 
 		/**
 		 * Get a pointer to the live device by its IP address represented as string. IP address can be both IPv4 or IPv6
 		 * @param[in] ipAddrAsString The IP address defined for the device as string
 		 * @return A pointer to the live device if this IP address is valid and exists. nullptr otherwise
 		 */
-		PcapLiveDevice* getPcapLiveDeviceByIp(const std::string& ipAddrAsString) const;
+		PcapLiveDevice* getDeviceByIp(const std::string& ipAddrAsString) const;
+		/**
```
tigercosmos, line 128: Also missing an empty line.

Pcap++/header/DeviceListBase.h
Dimi1010: Fixed 344a656

Pcap++/header/PcapLiveDeviceList.h
Dimi1010: Fixed 344a656

Pcap++/header/DeviceListBase.h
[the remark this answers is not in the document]
Dimi1010: Added tests fc73879

tigercosmos (approved):

Pcap++/header/KniDevice.h
```diff
@@ -129,7 +129,7 @@ namespace pcpp
 	class KniDevice : public IDevice
 	{
 		friend class KniDeviceList;
-		friend class internal::KniDeviceDeleter;
+		friend struct internal::KniDeviceDeleter;
```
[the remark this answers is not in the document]
Dimi1010: In this line, nothing except it silences a GCC warning about it being declared as a struct in one place and a class in another.

Dimi1010:
@seladb can we revisit this one for merge?

seladb:
@Dimi1010 this PR is pretty big, hard to review and affects many parts of the code. Maybe we can break it up to multiple PRs?
The first one can introduce the `DeviceListBase` class and update one of the simpler lists (the one with the least changes). Then subsequent PRs can modify more lists. Please let me know if that's ok with you

Dimi1010:
This should be fine since the lists themselves don't have any interconnected dependencies.

Dimi1010:
Closing as majority is implemented through PRs: #1787 #1790 #1791 #1794 #1795 #1796 #1797

PR #1492, opened 2024-07-08
ci: github: save and restore ccache database

clementperon:
In order to speedup compilation save and restore ccache database

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -665,7 +665,14 @@ jobs:
 
       - name: Install dependencies
         run: |
-          sudo apt update && sudo apt -y install libpcap-dev libbpf-dev tcpreplay
+          sudo apt update && sudo apt -y install ccache libpcap-dev libbpf-dev tcpreplay
+
+      - name: Restore Ccache
+        id: ccache-restore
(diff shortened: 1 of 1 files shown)


tigercosmos:
cool! do you feel it faster?

clementperon:
> cool! do you feel it faster?

You you can check it went from 3-4min to <1min

tigercosmos (approved):

seladb:
@clementperon how does the compiler know to use ccache? Should we configure `CMake` with the ccache directory? 🤔

clementperon:
> @clementperon how does the compiler know to use ccache? Should we configure `CMake` with the ccache directory? 🤔

Because we merged this :) 
https://github.com/seladb/PcapPlusPlus/pull/1492#issuecomment-2232113960

tigercosmos:
@clementperon  Seems that there are conflicts. You may want to fix it.

clementperon:
> > > @clementperon how does the compiler know to use ccache? Should we configure `CMake` with the ccache directory? 🤔
> > 
> > 
> > Because we merged this :) #1483
> 
> But how does `ccache` know to use the `${{ github.workspace }}/.ccache` directory to get the cached data? Is it the default directory?

Yes that's the default directory for Ccahe

clementperon:
From the documentation: https://ccache.dev/manual/4.2.html#_location_of_the_primary_configuration_file

seladb (approved):

PR #1497, opened 2024-07-12
Change PcapLiveDevice to store IP information as IPAddresses.

### LINKED ISSUE #1439: PcapRemoteDevice's clone method does not work.
`PcapRemoteDevice`'s clone method directly inherits from `PcapLiveDevice`'s clone method.
The method incorrectly attempts to fetch the device from local devices, failing and returning `nullptr`.

Solution would be to add an override to the clone method, correctly dispatching to fetch the remote devices, or refactoring the clone call to skip the fetching if possible.

Dimi1010:
Fixes #1439 and #1499.

This IP changes the internal structure of `PcapLiveDevice` to hold the IP information in `IPAddress` classes instead of `pcap_addr` structs. This fixes the issues where IP information might be randomly overwritten. Additionally the clean separation between the device information and the device proper allows cleaner clone procedure where the cached device information can be used to instantiate another `Device` object. The reworked procedure also fixes the clone issues with `PcapRemoteDevice`.

### `PcapLiveDevice` changes
A new protected structure `DeviceInformationDetails` has been added to `PcapLiveDevice` to hold the relatively static information about the device (name, description, addresses). The previously separate device members are replaced by a single member structure that is now used to hold that information.

**NB: The change removes the previously deprecated method `getAddresses()` and should NOT be merged before its deprecation cycle is complete**

In addition to the changes a new constructor overload has been added to `PcapLiveDevice` and derived classes taking the abovementioned structure instead of `pcap_if_t` pointer, allowing device construction from cached information.

### Optimizing the clone procedure
During the clone procedure, the `m_InterfaceDetails` structure is used to inject the current device's parameters to the clone's constructor without the need to re-fetch and search through the entire device list from the Pcap layer. The `clone` method is updated to a virtual method and performs the construction of a new device object directly without re-fetching the devices.

The method `cloneInternal` has been removed as superfluous, and derived implementations have been updated to override the `clone` method instead.

### Fix for clone of `PcapRemoteDevice`.
- Added an override of `clone` to `PcapRemoteDevice` to instantiate a new remote device from the current one's interface details.
- Added unit tests to check if the return object of a `PcapRemoteDevice::clone` method returns a non-null `PcapLiveDevice` pointer that can be successfully downcasted to a `PcapRemoteDevice` pointer.


Pcap++/header/PcapLiveDevice.h
```diff
@@ -80,16 +80,30 @@ namespace pcpp
...
+			explicit DeviceInterfaceDetails(pcap_if_t* pInterface);
+			/// @brief Name of the device.
+			std::string name;
+			/// @brief Description of the device.
+			std::string description;
+			/// @brief IP addresses associated with the device.
+			std::vector<IPAddress> addresses;
+			/// @brief Flag to indicate if the device is a loopback device.
+			bool isLoopback;
+		};
```
seladb, line 96: very nit: no need for doxygen documentation for private/protected structs

Pcap++/header/PcapRemoteDevice.h
```diff
@@ -91,8 +91,13 @@ namespace pcpp
 		std::shared_ptr<PcapRemoteAuthentication> m_RemoteAuthentication;
 
 		// c'tor is private, as only PcapRemoteDeviceList should create instances of it, and it'll create only one for
-		// every remote interface
```
seladb, line 94: Why remove this line? 🤔

seladb (approved):
Please see a few minor comments, otherwise LGTM

Pcap++/header/PcapRemoteDevice.h
Dimi1010: huh, good catch. did not mean that to be removed.

Pcap++/header/PcapLiveDevice.h
Dimi1010: Sure, its not strictly required, but I think its fine to stay, right?

Pcap++/header/PcapRemoteDevice.h
Dimi1010: Fixed d9bef35

Pcap++/header/PcapLiveDevice.h
seladb: Sure we can leave it 🙂

Issue #1499, opened 2024-07-17
PcapLiveDevice stores IP addresses in structured that may be overwritten.


Dimi1010:
During initialization of `PcapLiveDeviceList` information of all devices is fetched via `pcap_findalldevs`. This information is used to populate `PcapLiveDevice` objects and then the list is freed. This results in an issue as the device instance holds its IP address information in `pcap_addr_t` structs that are technically dangling, but still working somehow.

The fix to this issue will be implemented as a two stage process.
- [x] Add new getter that returns `std::vector<IPAddress>` for all IP addresses and deprecate `PcapLiveDevice::getAddresses` as that returns a :vector of `pcap_addr_t` structs.
(#1500)
- [x] Remove `PcapLiveDevice::getAddresses` and refactor `PcapLiveDevice` to hold its IP information in a vector of `IPAddress` objects. (#1497)

seladb:
This issue was fixed in #1500

Dimi1010:
> This issue was fixed in #1500

It hasn't been fixed. That was only Task 1.
The underlying IP is still stored in `pcap_addr_t` which I can't remove until `PcapLiveDevice::getAddresses()` is removed.

Dimi1010:
Fixed by #1497

PR #1500, opened 2024-07-17
Add getter to fetch all IP addresses of a live device as IPAddress objects.

Dimi1010:
Stage one of #1499

- Deprecated `getAddresses()` in `PcapLiveDevice.h`, advising to use `getIPAddresses()` instead for better functionality.
- Implemented `getIPAddresses()` in `PcapLiveDevice.cpp`, which returns a vector of `IPAddress` objects for both IPv4 and IPv6 addresses.
- Updated `LiveDeviceTests.cpp` to include tests for the new `getIPAddresses()` method, ensuring it correctly identifies and returns IP addresses.


Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -1168,6 +1168,31 @@ void PcapLiveDevice::setDefaultGateway()
...
 
+std::vector<IPAddress> PcapLiveDevice::getIPAddresses() const
+{
+	std::vector<IPAddress> results;
+	for (const auto& address : m_Addresses)
+	{
+		in_addr* ipv4Addr = internal::try_sockaddr2in_addr(address.addr);
+		if (ipv4Addr != nullptr)
+		{
+			results.push_back(IPv4Address(ipv4Addr->s_addr));
```
tigercosmos, line 1179: Could be `emplace_back`, but it's also fine now.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -1168,6 +1168,31 @@ void PcapLiveDevice::setDefaultGateway()
...
+			PCPP_LOG_DEBUG("IPv4 address found: " << results.back());
```
tigercosmos, line 1180: Do we need this log?

tigercosmos (approved):

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: Not sure. Added it for consistency as both `getIPv4Address` and `getIPv6Address` output address logs.

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: As in this?
```suggestion
			results.emplace_back(ipv4Addr->s_addr);
```

tigercosmos: I think `push_back(std::move(IPv4Address))` is equal to `emplace_back(IPv4Address`

how about this?
```
results.emplace_back(ipv4Addr->s_addr);
```
But it might be too trivial to distinguish the difference.

Dimi1010: Tested it on VS. That works for the IPv4, due to the `uint32_t` constructor. The IPv6 gets confused tho coz both v4 and v6 have a constructor from pointer.

Dimi1010: > But it might be too trivial to distinguish the difference.

Eh, it technically can save a single move constructor if the compiler does not optimize it away anyway.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -1168,6 +1168,31 @@ void PcapLiveDevice::setDefaultGateway()
 #endif
 }
 
+std::vector<IPAddress> PcapLiveDevice::getIPAddresses() const
+{
+	std::vector<IPAddress> results;
+	for (const auto& address : m_Addresses)
```
seladb, line 1174: it might be simpler to use `sockaddr2string` to convert the address to string, and the call `IPAddress` c'tor with this string

Pcap++/src/PcapLiveDevice.cpp
seladb: This is very old code, I think we can remove the logs 🙂

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: Eh, wouldn't that not achieve the same but with extra steps? We add encoding to string and then decoding from string for not much benefit. Atm, most operations are read/check operations.

Maybe it would be more concise and skip the if checks in the current function, but I don't think the current implementatios is that bad.

seladb: What I mean is that you can do something like this which is shorter and more concise:
```cpp
for (const auto& address : m_Addresses)
{
	std::array<char, INET6_ADDRSTRLEN> addrAsString;
	internal::sockaddr2string(addrIter.addr, addrAsString.data(), addrAsString.size());
	try
	{
		results.push_back(IPAddress(addrAsString));
	}
	catch(...) {}
}
```

Dimi1010: True, but looking at the way it will be implemented adds a lot of overhead that is avoided by the current code.

This is  the current implementation of the `IPAddress` string constructor.
```
IPAddress::IPAddress(const std::string& addrAsString)
{
	if (IPv4Address::isValidIPv4Address(addrAsString))  //  <-- string parse
	{
		m_Type = IPv4AddressType;
		m_IPv4 = IPv4Address(addrAsString);  //  <-- potential string parse
	}
	else if (IPv6Address::isValidIPv6Address(addrAsString))  //  <-- string parse
	{
		m_Type = IPv6AddressType;
		m_IPv6 = IPv6Address(addrAsString);  //  <-- potential string parse
	}
	else
	{
		throw std::invalid_argument("Not a valid IP address: " + addrAsString);
	}
}
```
This is between 2 and 3 parses of the string address to get the IP object in addition to the initial encode of the IP to a string. The current implementation, while slightly longer is at most 4 nullptr checks, 2 integer equality checks and a buffer copy.

seladb: ok, we can go with your approach 🙂

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: Removed 4100859

seladb (approved):

Issue #1463, opened 2024-06-24
Improve the CI time


tigercosmos:
Currently, we have too many GitHub action items, which results in a long building and testing time.
We need to refactor the current CI files and make the CI faster.

seladb:
how can we improve it? Maybe we can list the action items so anyone can work on it

tigercosmos:
@seladb let's brainstorm some ideas to improve this. cc @Dimi1010 @egecetin  @clementperon

clementperon:
I didn't look at the pipeline but using a cache + ccache could avoid to rebuild some files. 
Else limiting the number of OS and run a nightly pipeline with all the OS / configurations

egecetin:
My suggestion is that if the cache is not enough, we can distribute the workflow to several providers to prevent reaching compute limits. For example,

- Cirrus: only run doxy
- Azure: only windows runners
- Github: linux runners
- and so on ...

It should make it harder to maintain a bit, but we can handle it if we have to

tigercosmos:
@egecetin  I think it could be the last step. The issue is the current CI script is not efficient enough. Just like that we need to improve the algorithm before trying to use multithread(?)

egecetin:
@tigercosmos Yes of course. If possible, I agree we should improve the pipeline first. To be honest I did not check all workflows but most of the runs between 5-8 mins (approx 2-3 mins from build so it is like 30%-50% of run is build), right? So, I'm not sure we can dramatically improve CI runtime (even with caching) unless we limit the runner OS/version/config or improve parallelism. I think for "multiplatform" definition the code should be tested as much as possible. So, I don't know what we can do.

egecetin:
Something that comes to my mind, testing consumes approx 1 minute. If I'm not mistaken current test suite does not support multithreading (@seladb correct me if I'm wrong pls). So maybe using a test library supporting multithreading might lead better testing times? (eg. integrating GoogleTest + ctest to run tests parallel to efficiently saturate CPU)

tigercosmos:
@egecetin  true, I  think the parallelism of test cases may be the key. Probably we should focus on it in the next step.

tigercosmos:
@seladb  what do you think?

seladb:
The "Check installation" step re-builds the entire library, although it was built just a few steps earlier. Maybe we can use cache to prevent that?

[...]

PR #1507, opened 2024-07-21
Added Common++ unit test project based on Gtest.

Dimi1010:
## Summary
This PR introduces dedicated unit tests for the Common++ module based on the GoogleTest framework as well as several minor improvements. The dependency is resolved via `find_package` and conditionally via `FetchContent` if `find_package` fails.

### New dependency - GoogleTest
Added GoogleTest as dependency to the project when `PCAPPP_BUILD_TESTS` is set to `ON`. The dependency resolution happens in one of two ways: `find_package` or `FetchContent`. The project prioritizes the result of `find_package`, to allow environments to supply their own.

#### Restricted FetchContent dependency resolution.
For environments that disallow or discourage the use of `FetchContent` (e.g. brew, vcpkg, etc) an additional project option `PCAPPP_ALLOW_FETCH_CONTENT` (default: `ON`) is added. Setting the option to `OFF` would disable the automatic fetching and require all dependencies that are optionally supplied by fetch content be resolvable by `find_package` in the build environment.

**NB:** The `googletest` dependency is still declared inside the `FetchContent` infrastructure, but the option disables calling `FetchContent_MakeAvailable` effectively disabling the fetch operation.

### New Unit tests
- A new project `Test/Common++` has been added with unit tests.
  - Tested Classes
    - `IPAddress`, `IPv4Address`, `IPv6Address` - tests provided in `IPAddressTests.cpp`
    - `IPNetwork`, `IPv4Network`, `IPv6Network` - tests provided in `IPAddressTests.cpp`
    - `MacAddress` - tests provided in `MacAddressTests.cpp`
    - `PointereVector<T>` - tests provided in `PointerVectorTests.cpp`
    - `LRUList<T>` - tests provided in `LRUListTests.cpp`
  - Tested Functions
    - `byteArrayToHexString` - tests provided in `GeneralUtilsTests.cpp`
    - `hexStringToByteArray` - tests provided in `GeneralUtilsTests.cpp`
    - `cross_platform_memmem` - tests provided in `GeneralUtilsTests.cpp`
    - `align` - tests provided in `GeneralUtilsTests.cpp`
- Added machinery to fetch `googletest` (v1.16.0) during the cmake build process.
- Enabled `googletest` macro definitions for pre-commit's `cppcheck` step.
- Added listener that can be used to extend a unit test to check for memory leaks.

### Fixes and additions
- Added `#include <array>` to `MacAddress.h` to fix transitive include.


Tests/Common++Test/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ include(FetchContent)
 FetchContent_Declare(
     googletest
     GIT_REPOSITORY https://github.com/google/googletest.git
-    GIT_TAG release-1.11.0
+    GIT_TAG release-1.12.0
```
tigercosmos, line 8: I don't know that we already have gtest.

clementperon: We didn't the fetch content will pull it from internet

tigercosmos: I know. I meant I am suprised that there is gtest fetching in current master because we are not using gtest now.

Dimi1010: > I am suprised that there is gtest fetching in current master

Wdym, @tigercosmos ? There isn't one in the master branch?

clementperon: @tigercosmos you're looking at a specific commit and not the whole MR

Dimi1010:
So far, it hasn't been integrated into the workflow or the codecov as I'm not really sure how precisely that is achieved.

seladb:
@Dimi1010 @tigercosmos @egecetin @clementperon let's discuss this change: if we want to move away from the existing tests infra I think it should be a consternated effort to migrate all tests to Gtest. It'll take time, but may be feasible if all of us are working on it together. The question is: can we all commit to work on it before the next version?

egecetin:
@seladb What is the expected date of the new release? If it will be about 25.04 it is easy to work on it by planning. The more important question is, we talked about the gtest in #1463 to improve performance. Is everyone thinking moving Gtest is better, or anyone have any doubts (for performance, maintainability or maybe interface related)? It is better to share it if have any, before starting to work.

In my opinion, it will reduce maintaining requirements because of gtest has lots of developers and they already fix their bugs and has a good interface for a C++ code and have integration with the external tools/extensions but for performance I don't expect a drastically improvement. So, I think moving to gtest is feasible.

seladb:
> @seladb What is the expected date of the new release? If it will be about 25.04 it is easy to work on it by planning. The more important question is, we talked about the gtest in #1463 to improve performance. Is everyone thinking moving Gtest is better, or anyone have any doubts (for performance, maintainability or maybe interface related)? It is better to share it if have any, before starting to work.
> 
> In my opinion, it will reduce maintaining requirements because of gtest has lots of developers and they already fix their bugs and has a good interface for a C++ code and have integration with the external tools/extensions but for performance I don't expect a drastically improvement. So, I think moving to gtest is feasible.

I hope we can release faster, hopefully before 25.04. To be honest, I don't think performance is a good enough reason to switch to Gtest. The main reason would be having more standardize code and less custom-made stuff. But as I mentioned, it will require a lot of work (and most of it is boring - rewrite tests...), so the question is - can we commit to that and do we think it's worth the effort?

Dimi1010:
I think it will be beneficial overall. I think the rollout can be done in phases (project by project) similar to the way we did the clang-format roll out. I could probably work through it slowly as I get free time.

Also, a lot of the current tests could probably benefit by the expanded non-critical failure support, I know we have something similar in the PTF framework but its not exactly the same. In GTest a non-critical failure (EXPECT_*) the test is still marked as failed, but continues to run to catch other errors in the unit test if they exist. Fatal failures (ASSERT_*) are mostly used for conditions that invalidate the rest of the unit test. The fact that the current framework only catches 1 error per unit test has been somewhat annoying to me.

seladb:
> I think it will be beneficial overall. The rollout can be done in phases (project by project) similar to the way we did the clang-format roll out. I could probably work through it slowly as I get free time.

The difference between this and clang-format is that this project will be much slower and more tedious because we need to go over all the tests and port them one by one. Maybe we can open a separate branch for it so we don't merge unfinished work to `dev` and block `master`?

Anyway, this will require more time and commitment from everyone, and I wonder if we can/want to do it.

However, if we open a separate branch and it doesn't block anything then the risk is quite minimal 🤔 


> Also, a lot of the current tests could probably benefit by the expanded non-critical failure support, I know we have something similar in the PTF framework but its not exactly the same. In GTest a non-critical failure (EXPECT_*) the test is still marked as failed, but continues to run to catch other errors in the unit test if they exist. Fatal failures (ASSERT_*) are mostly used for conditions that invalidate the rest of the unit test. The fact that the current framework only catches 1 error per unit test has been somewhat annoying to me.

tigercosmos:
@Dimi1010  Do you want to continue this?

Dimi1010:
> @Dimi1010 Do you want to continue this?

Eh... eventually... Don't really have time for it now tho, so gonna close it for the moment.

Tests/Common++Test/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+cmake_minimum_required(VERSION 3.14)
+
+include(FetchContent)
+
+# TODO: This may have issues with brew as it doesn't allow the use of FetchContent.
+FetchContent_Declare(googletest GIT_REPOSITORY https://github.com/google/googletest.git GIT_TAG v1.16.0)
```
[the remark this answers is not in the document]
Dimi1010: Most recent is 1.17, but that requires cpp17

tigercosmos (approved):

Tests/Common++Test/header/MemoryLeakDetectorFixture.hpp
```diff
@@ -0,0 +1,61 @@
...
+		void TearDown() override
+		{
+			std::size_t memLeakCount = 0;
+			std::uint64_t memLeakSize = 0;
+			MemPlumber::memLeakCheck(memLeakCount, memLeakSize);
+			MemPlumber::stopAndFreeAllMemory();
+
+			if (memLeakCount > 0 || memLeakSize > 0)
+			{
+				FAIL() << "Memory leak found! " << memLeakCount << " objects and " << memLeakSize << " [bytes] leaked";
```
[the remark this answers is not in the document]
Dimi1010: Fixed 2b9a0bafd5213677e5b45462a6cb9e2eb488b73f

Tests/Common++Test/Tests/LoggerTests.cpp
```diff
@@ -0,0 +1,230 @@
...
+			{
+				throw std::runtime_error("Log Trampoline Error: Log callback not set");
+			}
+		}
+
+		static std::unique_ptr<LogCallbackMock> m_LogCallbackMock;
+		Logger& m_Logger;
+	};
+
+	std::unique_ptr<LogCallbackMock> LoggerTest::m_LogCallbackMock = nullptr;
```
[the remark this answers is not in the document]
Dimi1010: Fixed 36bea655643c1916f9815803c7357d4a117bdeb6

Dimi1010:
@seladb @egecetin @tigercosmos @clementperon 

So... minor issue with the google test integration. The framework allocates to heap during unit tests.
Due to that when the per-test memory leak tracking is added like we have in the current tests, it gives false positives. Another thing is that the `MemPlumber::stopAndFreeMemory` after the test essentially corrupts the heap, leading to access violations when the framework searches for what it has allocated during the test.

I don't think we can run the leak tracking on per test level as is. It might be workable on a global level though.

I am open to suggestions on how to proceed.

tigercosmos:
I am okay to move the memory-leaking test to global as long as it helps to find out issues.

Dimi1010:
Update on the memory leak infrastructure: It still doesn't work directly on global level as gtest uses thread local variables that end up lazy initializing at runtime, but get released on program exit. :/

tigercosmos:
Not sure if I fully understood. Aren't all tests passed now?

Dimi1010:
> Not sure if I fully understood. Aren't all tests passed now?

Tests pass because the memleak is disabled in the gtest project currently. [[1]](https://github.com/seladb/PcapPlusPlus/pull/1507/files#diff-b200b92ca48807fec6d8612efdafed6e1f786ecc093b90f8bdc72728027bdfd3R31-R35)

Dimi1010:
@tigercosmos In my tests, it didn't work having the memory tracking encompass all the tests as a single start/stop pair, either. 

That is because gtest appears to allocate on heap from thread local variables during the tests. That memory isn't released until the main thread actually exits when the program shuts down, and gets detected as a false positive. :/ (and leads to crash when the thread actually tries to deallocate)


This is the culprit. From googletest v1.16.0 [gtest-port.h](https://github.com/google/googletest/blob/6910c9d9165801d8827d628cb72eb7ea9dd538c5/googletest/include/gtest/internal/gtest-port.h#L1573-L1599)
```
// Implements thread-local storage on Windows systems.
//
//   // Thread 1
//   ThreadLocal<int> tl(100);  // 100 is the default value for each thread.
//
//   // Thread 2
//   tl.set(150);  // Changes the value for thread 2 only.
//   EXPECT_EQ(150, tl.get());
//
//   // Thread 1
//   EXPECT_EQ(100, tl.get());  // In thread 1, tl has the original value.
//   tl.set(200);
//   EXPECT_EQ(200, tl.get());
//
// The template type argument T must have a public copy constructor.
// In addition, the default ThreadLocal constructor requires T to have
// a public default constructor.
//
// The users of a TheadLocal instance have to make sure that all but one
// threads (including the main one) using that instance have exited before
// destroying it. Otherwise, the per-thread objects managed for them by the
// ThreadLocal instance are not guaranteed to be destroyed on all platforms.
//
// Google Test only uses global ThreadLocal objects.  That means they
// will die after main() has returned.  Therefore, no per-thread
// object managed by Google Test will be leaked as long as all threads
// using Google Test have exited when main() returns.
```

Tests/Common++Test/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
```
seladb, line 7: As the `TODO` comment mentions, package managers don't approve packages that use `FetchContent_Declare`, we already had this experience before...

Dimi1010: Yeah. Tbh, we could possibly have a copy of the lib in `3rdParty` to build the tests with.

I am unsure if we could use the brew version since we need to use up to 1.16 as 1.17 (current newest) requires cpp17.

seladb: yeah, putting it in `3rdParty` is the only option to keep PcapPlusPlus in package managers.

Though I'm still not sure that we want to merge this PR. Having 2 test frameworks doesn't look like a good idea, and converting all of our tests to Gtest will be an extremely hard task

Dimi1010: I did start another branch [[1]](https://github.com/Dimi1010/PcapPlusPlus/tree/feature/gtest-framework-packet%2B%2B) off of this one, for the Packet++ tests, to see how it would go. You can take a look, if you want. The new project is `Tests/Packet++Test-google`.

It actually mostly involves trivial changes, like macro replacements.
- `PTF_TEST_CASE` -> `TEST(Fixture, TestName)`
- `PTF_ASSERT_EQUAL` -> `ASSERT_EQ` / `EXPECT_EQ`
- `PTF_ASSERT_NULL` -> `ASSERT_EQ(statement, nullptr)`
- `PTF_ASSERT_NOT_NULL` -> `ASSERT_NE(statement, nullptr)`

The only non-trivial assertion so far has been `PTF_ASSERT_BUF_COMPARE` which was replaced by `EXPECT_TRUE(test::BuffersMatch(actualBuf, actualBufSize, expectedBuf, expectedBufSize))` which ended up being a more comprehensive check IMO. [[2]](https://github.com/Dimi1010/PcapPlusPlus/blob/feature/gtest-framework-packet%2B%2B/Tests/Packet%2B%2BTest-google/Utils/AssertionExtensions.cpp#L31-L94).

The loading utilities needed a revamp, as I didn't want to use macros everywhere and the current ones had a limitation of always using the current working directory (has issues with out of source builds).

For the majority of the cases the code: 
```cpp
timeval time;
gettimeofday(&time, nullptr);

READ_FILE_AND_CREATE_PACKET(1, "PacketExamples/TcpPacketNoOptions.dat");
// rest of test case
```
ended up being a one liner:
```cpp
// std::unique_ptr<RawPacket>
auto rawPacket1 = test::createPacketFromHexResource("PacketExamples/TcpPacketWithOptions3.dat", /* optional packet factory */, /* optional data loader */);
```

seladb: I'm not sure how we'll be able to review these huge PRs...

Dimi1010: That is why the Packet++ port is in a different target. It will allow incremental PRs as tests are ported.

A way I found to diff the changes between the tests is to diff the `Packet++Test/Tests/testname.cpp` and `Packet++Test-google/Tests/testname.cpp`. Don't think it can be done on github's PR interface, so it will probably require pulling to local, but it gives quite a clean diff.

image

seladb: I guess we can do it incrementally, file by file (or few files in a PR), if you have the patience to do all this work 🙂

Dimi1010: I don't mind doing it. I think it will be a worthwhile change.

seladb: Ok sure 👍 
I'll try to review this PR in the coming days

seladb: Actually, even this PR is pretty huge. Maybe we can break it into smaller PRs to make it easier to review?

Dimi1010: I can split the files in the Tests folder, to separate into different PRs. Don't think I can separate the rest tho.

Dimi1010: > As the `TODO` comment mentions, package managers don't approve packages that use `FetchContent_Declare`, we already had this experience before...

Should be fixed now. 789a17eebc4d7349fbe421f3b085929fb495ba29

seladb: @Dimi1010 adding the entire GTest repo bloats this repo... unfortunately, GTest isn't a header-only library and using `FetchContent_Declare` isn't possible.
I asked ChatGPT what are our options 🙂:

> To integrate a test framework in a C++14 library without using FetchContent (which is disallowed by package managers), we considered several options:
>
> - Single-header test libraries (✅ Recommended): Use a lightweight, header-only framework like doctest or Catch2 v2 by vendoring just the main header file under `tests/third_party/`. Keeps the repo small, package-manager-friendly, and easy to use in CI or locally.
> - Git submodules: Add the test framework as a submodule, keeping it out of the main source tree. Requires users to initialize submodules manually.
> - ExternalProject_Add: Use this older CMake feature to build test dependencies separately, avoiding bundling in the source tree but adding some CMake complexity.
> - CI-only downloads: Download the test framework during CI runs, keeping the repo clean but requiring more CI setup and complicating local testing.
> - System-installed test frameworks: Rely on system packages (e.g. gtest via apt or Homebrew), which keeps the repo clean but may hurt portability and reproducibility.
> Recommendation: Use a single-header test framework like doctest, vendored under tests/third_party/, and isolate tests from the main install using a BUILD_TESTING CMake option. > This approach balances simplicity, compatibility, and package policy compliance.


Maybe we should consider using Catch2 instead of GTest? Catch2 v2 is header only and v3 is one header + one source file 🤔

Dimi1010: Hmm, I am not familiar with Catch2 that much, will have to look into it.

Also, something I noticed. The issue with `FetchContent` is that we can't use it for main dependencies for the built targets. The benchmark in examples seem to be doing fine, due to the fact that it is not built in the package managers.

I am unsure how desirable this suggestion is, but we could just not build the tests for the package build? They don't appear to be exported in the install? From what I could find there are other projects that don't build their tests in the install on brew.

seladb: Perhaps you're right; I'm not familiar with the requirements of the different package managers. Maybe we can build the library without the tests, which will require changing the recipe in each package. To be on the safe side, I'd prefer using a header-only library if possible

Dimi1010: @seladb Looked at catch2. Version 2's last release 3 years ago and it seems to have only received minor backports since then. I am somewhat reluctant to port to a version that may need to be ported again in the future. :/

Version 3 (currently 3.9)[[1]](https://github.com/catchorg/Catch2/tree/devel/src/catch2) is a lot more than one source file. It is a regular static library at this point, so we have the same issue as with `googletest`.

I looked at the package managers we support. It seems we build tests for vcpkg and homebrew, while we don't for conan. Tbh, both homebrew and vcpkg have `gtest` available so we could specify it as a dep. Only issue is that homebrew would need to compile `gtest` with cpp17 as the version they have on `homebrew-core` requires it.

IMO, it would be better to keep the dependency tree clean in the managers and try to disable the tests as they aren't actually installed or used during the install process. It seems allowed by the managers. [[2]](https://github.com/search?q=repo%3Amicrosoft%2Fvcpkg+build_tests%3Doff&type=code) [[3]](https://github.com/search?q=repo%3AHomebrew%2Fhomebrew-core+build_tests%3Doff&type=code)

@clementperon Any thoughts on this? You found the issue with fetch content on homebrew last time.

cc: @tigercosmos @egecetin

tigercosmos: I am sure that Gtest can be installed during `make` runtime. I have an example in my GitHub: https://github.com/tigercosmos/googletest-tutorial

Dimi1010: > I am sure that Gtest can be installed during `make` runtime. I have an example in my GitHub: https://github.com/tigercosmos/googletest-tutorial

@tigercosmos  The issue at hand is that the package managers don't like FetchContent, so we either have to not use it, or not build the tests when running the package manager build + install (which aren't used there anyway), so it does not get called there.

tigercosmos: Got it. Thanks for the explanation. In this case, I think putting the source code in `3rdParty` should be fine? Sorry if I miss any context.

seladb: @Dimi1010 my understanding is that Catch2 is only 2 files, one header and one source: https://github.com/catchorg/Catch2/releases/tag/v3.9.1

Am I missing anything? 🤔

Dimi1010: Hmm... I missed that they offered amalgamated versions. I guess we can do that?

Upon first glance they mostly appear to be offered for migration compatibility from v2. The maintainers do recommend moving away from those headers in the migration instructions and inside the amalgamated header, as they apparently increase the compilation times significantly compared to the non-amalgamated version. https://github.com/catchorg/Catch2/blob/v3.9.1/docs/migrate-v2-to-v3.md#how-to-migrate-projects-from-v2-to-v3

seladb: Maybe we can re-consider using Gtest, however we need to make sure that:
1. All package managers are ok with using `FetchContent` if it's not used as part of building the package, otherwise we'll do this change for nothing. Any idea how we can verify that?
2. What's the overhead in fetching and compiling Gtest? How much time does it add to building PcapPlusPlus with the tests?

Dimi1010: > All package managers are ok with using FetchContent if it's not used as part of building the package, otherwise we'll do this change for nothing. Any idea how we can verify that?

A simple way to verify that it works is that they currently like our packages as is. Version 25.05 uses `FetchContent` in examples under PcapPlusPlus-benchmark [[1]](https://github.com/seladb/PcapPlusPlus/blob/v25.05/Examples/PcapPlusPlus-benchmark/CMakeLists.txt). Since we don't build the benchmarks under the package release and it runs fine. That implicitly means that it is fine to exist, as long as it is not executed.

For `homebrew`. The research I did ended up to this issue: https://github.com/Homebrew/brew/pull/17310
It essentially traps `FetchContent_MakeAvailable` when it is called. It shouldn't have any issues if its not called.

For `conan` we don't build the tests anyway, so if we place the fetch directive under the Tests directory. It shouldn't run at all, so it would have been as it does not exist.

Unsure of the specifics for `vcpkg`, it should also be fine to have it and not run it, since as mentioned above, the examples do just that.

I suppose a useful thing we can also try to have a CI workflow that attempts to build the release packages for the managers to validate the compatibility on a more regular basis instead of only when publishing a release? I know we have a "package and release" workflow, but that only does cmake package. Perhaps something similar, that runs only like once a week?


> What's the overhead in fetching and compiling Gtest? How much time does it add to building PcapPlusPlus with the tests?

The current PR's CI does include Gtest and Common++.

For CI timing comparisons I will use the following CI runs:
- commit d7dedbf in the master - Reference - [[CI Actions]](https://github.com/seladb/PcapPlusPlus/actions/runs/16718701510)
- commit 532678a in this PR - The last commit that fetches GTest via fetch content [[CI Actions]](https://github.com/seladb/PcapPlusPlus/actions/runs/16709469565)
- commit c98dff8 in this PR - Full run with gtest embedded in 3rdParty. [[CI Actions]](https://github.com/seladb/PcapPlusPlus/actions/runs/16721227098)

The comparison will be on `(ubuntu2404, OFF)` job.
Total run:
| Name | Time |
| ------ | ----- |
| master | 9m 28s |
| fetch + gtest | 9m 53s |
| gtest | 9m 51s |

CMake configure step timings. Fetch Content runs here.
| Name | Time |
| ------ | ----- |
| master | 6s |
| fetch + gtest | 8s |
| gtest | 6s |

Build step timings. The PR builds have 2 extra targets to build (`gtest`, `Common++`)
| Name | Time |
| ------ | ----- |
| master | 4m 24s |
| fetch + gtest | 4m 32s |
| gtest | 4m 37s |

seladb: @Dimi1010 I agree with you and think that using `FetchContent` should probably be safe with the package managers as long as we don't run it for building the pacakge.

Regarding the overhead - I checked the CI logs and couldn't find the part that runs `FetchContent` and fetches Gtest 🤔

Dimi1010: > Regarding the overhead - I checked the CI logs and couldn't find the part that runs FetchContent and fetches Gtest

That is normal tbh. As far as I know, `FetchContent` does not output log messages on a successful run. I can add messages manually around the calls if you want.

seladb: Ah ok, that makes sense. I just wanted to confirm that we're not missing anything.

So I guess we decided to go with Gtest?

Dimi1010: @seladb  I guess so.

IMO, the first thing to do would be to disable building the tests in `vcpkg` and `homebrew` install scenarios, via a revision to the portfile / bottle, so we know it can work fine as is, before adding fetch content to the test infrastructure.

Unless anyone has any objections? @egecetin @clementperon @tigercosmos

seladb: Yes I think that's a good idea 👍

Dimi1010: Opened a PR to disable the tests in `vcpkg` here https://github.com/microsoft/vcpkg/pull/46941.

Also found a bug when examining the build. It placed the test executables inside the `vcpkg_installed/vcpkg/blds/pcapplusplus/src/***.clean` which is a bug, IMO.

Our test projects don't really support out of source builds currently, so I plan on fixing this in the Gtest rework.

Dimi1010: Homebrew has also been updated https://github.com/Homebrew/homebrew-core/pull/233719

Conan needs no update as we already don't build tests there on install.

Dimi1010:
@seladb Updated the PR description, so you might want to reread that. Main update is `New dependency - GoogleTest` under `What's Changed?`

Tests/Common++Test/Tests/GeneralUtilsTests.cpp
```diff
@@ -0,0 +1,37 @@
+#include "pch.h"
+
+#include <cstring>
+
+#include "GeneralUtils.h"
+
+namespace pcpp
```
Dimi1010: Should probably change the namespace to `pcpp_tests`...

PR #1512, opened 2024-07-26
Added overload to getAndDetach that accepts a constant reference.

Dimi1010:
This PR addresses the issue of values being unable to be detached from an iterator temporary.

Example code: ```pVector.getAndDetach(pVector.begin())```


tigercosmos:
I wonder where will be this new function called.

Dimi1010:
> I wonder where this new function will be called.

I wanted to use the example code in #1507's unit tests for Pointer Vector. Adding it as a separate PR as it was not strictly related.

seladb (approved):

PR #1515, opened 2024-07-31
Broke up PcapLiveDeviceList initialization into steps.

Dimi1010:
This PR breaks down the `init()` method of `PcapLiveDeviceList` into two static methods `fetchAllLocalDevices` and `fetchDnsServers`, allowing cleaner separation between general fetching logic and the object initialization.

### DIFF
--- a/Pcap++/src/PcapLiveDeviceList.cpp
+++ b/Pcap++/src/PcapLiveDeviceList.cpp
@@ -26,14 +26,25 @@
 
 namespace pcpp
 {
+	namespace 
+	{
+		void syncPointerVectors(std::vector<std::unique_ptr<PcapLiveDevice>> const& mainVector, std::vector<PcapLiveDevice*>& viewVector)
+		{
+			viewVector.resize(mainVector.size());
+			// Full update of all elements of the view vector to synchronize them with the main vector.
+			std::transform(mainVector.begin(), mainVector.end(), viewVector.begin(),
+			               [](const std::unique_ptr<PcapLiveDevice>& ptr) { return ptr.get(); });
+		}
+	}
 
-PcapLiveDeviceList::PcapLiveDeviceList()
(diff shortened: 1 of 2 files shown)


Pcap++/src/PcapLiveDeviceList.cpp
```diff
@@ -26,14 +26,25 @@
 
 namespace pcpp
 {
+	namespace
```
tigercosmos, line 29: why is this anonymous namespace needed?

tigercosmos (approved):
LGTM, thanks!

Pcap++/src/PcapLiveDeviceList.cpp
Dimi1010: For organizational purposes and to mark the content functions as internal linkage, since they are only used as helpers in the current translation unit.

Dimi1010: Initially the `fetch*` functions were also as free functions in this namespace, but got changed to static functions as the private ctors of the device classes did not allow construction inside free functions.

tigercosmos: Then, in this case, I think the convention is to let it be under `pcpp::internal`?

Dimi1010: Then it would not link internally tho. Anonymous namespaces are special as they specify their linkage as internal. The `internal` namespace is more for things that are used throughout the library (therefor need external linkage) and we don't want the user to use. An anonymous namespace is more like "I don't even want the other parts (translation units) of the library know this exists because they don't need it".

tigercosmos: I got your intention. Thanks for the explanation.

seladb (approved):

PR #1525, opened 2024-08-03
Pin actions with hash instead of tag

egecetin:


tigercosmos (approved):

egecetin:
MacOS runners can be fixed with upgrading codecov-action. Check runners of #1521 or #1524

seladb (approved):

PR #1529, opened 2024-08-08
support RawSocketDevice with timeout in millisecond precision.

### LINKED ISSUE #1528: Support millisecond resolution timeout in RawSocketDevice
Hello,
 
Is it possible to offer an enhancement to the receive() operations in the RawSocketDevice class?
It will be great to have the option to set timeout in a milliseconds resolution instead of seconds - helpful when we need to perform a lot of receive() calls with some timeout without it taking forever to perform.

will gladly create a PR with the needed adaptions :)

tigercosmos:
Fix #1528. Changed the timeout from `int` to `double` to allow millisecond precision. No new tests are needed as the current ones already cover this change.

### DIFF
--- a/Pcap++/header/RawSocketDevice.h
+++ b/Pcap++/header/RawSocketDevice.h
@@ -86,7 +86,8 @@ namespace pcpp
 		 * @param[in] blocking Indicates whether to run in blocking or non-blocking mode. Default value is blocking
 		 * @param[in] timeout When in blocking mode, specifies the timeout [in seconds] to wait for a packet. If timeout expired
 		 * and no packets were captured the method will return RawSocketDevice#RecvTimeout. Zero or negative values mean no
-		 * timeout. The default value is no timeout
+		 * timeout. The default value is no timeout. The timeout precision is in milliseconds, for example a timeout of 0.123
+		 * means 123 milliseconds.
 		 * @return The method returns one on the following values:
 		 *  - RawSocketDevice#RecvSuccess is returned if a packet was received successfully
 		 *  - RawSocketDevice#RecvTimeout is returned if in blocking mode and timeout expired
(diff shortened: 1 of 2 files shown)


Pcap++/src/RawSocketDevice.cpp
```diff
@@ -216,26 +217,34 @@ RawSocketDevice::RecvPacketResult RawSocketDevice::receivePacket(RawPacket& rawP
...
 
-	long curSec, curNsec;
-	clockGetTime(curSec, curNsec);
+	int64_t timeoutMilli = timeout * 1000;
 
 	int packetCount = 0;
 	failedRecv = 0;
 
-	long timeoutSec = curSec + timeout;
+	auto start = std::chrono::system_clock::now();
```
Dimi1010, line 233: I recommend using `steady_clock` or `high_resolution_clock` as `system_clock` may not be steady and cause inaccuracies when used to calculate a wait duration.

Pcap++/src/RawSocketDevice.cpp
```diff
@@ -216,26 +217,34 @@ RawSocketDevice::RecvPacketResult RawSocketDevice::receivePacket(RawPacket& rawP
...
 
-	while (curSec < timeoutSec)
+	while (true)
 	{
+		auto now = std::chrono::system_clock::now();
```
Dimi1010, line 237: Ditto, `system_clock`

Pcap++/src/RawSocketDevice.cpp
tigercosmos: Changed. Thanks.

Dimi1010 (approved):

Pcap++/header/RawSocketDevice.h
```diff
@@ -86,26 +86,28 @@ namespace pcpp
 		 * @param[in] blocking Indicates whether to run in blocking or non-blocking mode. Default value is blocking
 		 * @param[in] timeout When in blocking mode, specifies the timeout [in seconds] to wait for a packet. If timeout expired
 		 * and no packets were captured the method will return RawSocketDevice#RecvTimeout. Zero or negative values mean no
-		 * timeout. The default value is no timeout
+		 * timeout. The default value is no timeout. The timeout precision is in milliseconds, for example a timeout of 0.123
```
seladb, line 89: Maybe we should mention that the timeout is in seconds but with precision in ms?

seladb (approved):
One small comment on documentation, otherwise LGTM

Pcap++/header/RawSocketDevice.h
tigercosmos: Note that L87 already mentions that.

PR #1516, opened 2024-07-31
Clang-format for Pcap++

Dimi1010:

PR #1532, opened 2024-08-12
Added a RAII PcapHandle wrapper class that manages a Pcap handle.

Dimi1010:
This PR adds a class `PcapHandle` that wraps a `pcap_t*` handle and manages the lifecycle of the handle via RAII mechanism, in a similar fashion to `std::unique_ptr`.

PS: The diff of PcapDevice.cpp is a bit messed up because the formatter ran on it... might be easier to review after #1516 has been merged.


Pcap++/header/PcapDevice.h
```diff
@@ -18,6 +18,79 @@ namespace pcpp
...
+			}
+
+			bool operator==(std::nullptr_t) const
+			{
+				return !hasValue();
+			}
+			bool operator!=(std::nullptr_t) const
+			{
+				return hasValue();
+			}
```
seladb, line 85: do we use this anywhere?

Pcap++/header/PcapDevice.h
```diff
@@ -18,6 +18,79 @@ namespace pcpp
 	//Forward Declaration - required for IPcapDevice::matchPacketWithFilter
 	class GeneralFilter;
 
+	namespace internal
+	{
+		/// @class PcapHandle
+		/// @brief A wrapper class for pcap_t* which is the libpcap packet capture descriptor.
+		/// This class is used to manage the lifecycle of the pcap_t* object
```
seladb, line 25: nit: can we change all doxygen comments to be `/* ... */` to be consistent with the rest of the code?

Pcap++/header/PcapDevice.h
Dimi1010: Quite a lot of places actually. It is to allow existing comparisons like `if (pcapDescriptor == nullptr)` without too much changes.

Pcap++/header/PcapDevice.h
Dimi1010: Sure, tho I am of the opinion that the `/// ...` look better with the less line bloat of extra lines above and below each description.

Btw, #1357 also has `/// ...` type comments.

tigercosmos: `///` is also a doxygen style. I think we can use both. In addition, I also agree with @Dimi1010  that `///` is simple when there are only a few lines.

Pcap++/header/PcapDevice.h
```diff
@@ -18,6 +18,79 @@ namespace pcpp
...
+			/// @brief Check if the handle is not null.
+			/// @return True if the handle is not null, false otherwise.
+			bool hasValue() const
+			{
+				return m_PcapDescriptor != nullptr;
+			}
+
+			/// @brief Access the underlying pcap descriptor.
+			/// @return The pcap descriptor.
+			pcap_t* value() const
```
tigercosmos, line 53: Does it make more sense to rename as `get()`?

Pcap++/header/PcapDevice.h
```diff
@@ -18,6 +18,79 @@ namespace pcpp
...
+
+			PcapHandle& operator=(const PcapHandle&) = delete;
+			PcapHandle& operator=(PcapHandle&& other) noexcept;
+			PcapHandle& operator=(std::nullptr_t) noexcept;
+
+			~PcapHandle();
+
+			/// @brief Check if the handle is not null.
+			/// @return True if the handle is not null, false otherwise.
+			bool hasValue() const
```
tigercosmos, line 46: How about `isValid`?

Pcap++/header/PcapDevice.h
```diff
@@ -18,6 +18,79 @@ namespace pcpp
...
+			PcapHandle() = default;
+			/// @brief Creates a handle from the provided pcap descriptor.
+			/// @param pcapDescriptor The pcap descriptor to wrap.
+			explicit PcapHandle(pcap_t* pcapDescriptor);
+
+			PcapHandle(const PcapHandle&) = delete;
+			PcapHandle(PcapHandle&& other) noexcept;
+
+			PcapHandle& operator=(const PcapHandle&) = delete;
+			PcapHandle& operator=(PcapHandle&& other) noexcept;
```
tigercosmos, line 39: No one is using it? Can it be deleted?

Dimi1010: Eh, it can, but I would prefer to keep it for completeness sake. The assignment to nullptr is used, so it would make little sense to have that but not regular assignment.

Dimi1010: Looking at it, we need to keep the copy assignment as deleted anyway, as it will be implicitly declared otherwise. At that point the only thing we can remove is the move assignment, which I personally would prefer to keep even if its not used at the moment due to rule of 5. This being a lifecycle class, I think it would be beneficial to have all the ctor/assignment semantics be strictly defined from the start.

tigercosmos: Sorry for confusing you, I am not against the moving assignment declaration. I meant we should mark it as "delete" since it seems to be not used by anyone.

Dimi1010: I see. Hmm, we have a move constructor tho? Wouldn't having that but not move assignment be confusing? I usually consider those 2 a pair.

tigercosmos: It seems that the move c'tor is not being used as well... Anyway, I am fine with them. I just though maybe we don't need them.

Dimi1010: I tried to have the ownership semantics of a unique_ptr, so a move operation would be a transfer of ownership.

tigercosmos: I see. No problem.

Pcap++/header/PcapDevice.h
```diff
@@ -18,6 +18,79 @@ namespace pcpp
...
+			pcap_t* release();
+
+			/// @brief Helper function to retrieve the last error string for this handle.
+			/// @return The last error string.
+			std::string getLastError() const;
+
+			/// @brief Helper function to retrieve a view of the last error string for this handle.
+			/// @return A view of the last error string.
+			/// @remarks This function is more efficient than getLastError() as it does not copy the string.
+			char const* getLastErrorView() const;
```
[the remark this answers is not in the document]
Dimi1010: Hmm, it would probably be better to keep the `char const*` version as it avoids a copy and the user can always convert it to a string if it needs to be saved.

Pcap++/header/PcapDevice.h
Dimi1010: Sure, changed 1b44021

Pcap++/header/PcapDevice.h
Dimi1010: Changed ba319a7

Pcap++/header/PcapDevice.h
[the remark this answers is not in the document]
Dimi1010: Removed `getLastError` in c1d0af8. Left the `getLastErrorView` with the `View` suffix to emphasize that the returned pointer is a view and should be saved if keeping it is required.

Pcap++/header/PcapDevice.h
seladb: I prefer consistency in the code base as much as possible, especially because this file contains another class `IPcapDevice` with the other documentation style `/** ... */`

Pcap++/header/PcapDevice.h
```diff
@@ -18,6 +18,75 @@ namespace pcpp
...
+			PcapHandle(PcapHandle&& other) noexcept;
+
+			PcapHandle& operator=(const PcapHandle&) = delete;
+			PcapHandle& operator=(PcapHandle&& other) noexcept;
+			PcapHandle& operator=(std::nullptr_t) noexcept;
+
+			~PcapHandle();
+
+			/// @brief Check if the handle is not null.
+			/// @return True if the handle is not null, false otherwise.
```
seladb, line 45: If both `@brief` and `@return` contain the same information, we usually omit `@brief` and keep only `@return`.

Same for the other methods in this class

Pcap++/header/PcapDevice.h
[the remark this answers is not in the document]
seladb: Why don't we want to return a string?
Errors are usually not in the critical path so coping a `char*` to `std::string` shouldn't be an issue...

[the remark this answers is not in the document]
Dimi1010: I chose to keep the view as its the more performant version as it does not have a heap allocation. Converting the view `char*` to string is trivial if we need it somewhere, and the string version could always be added back if its needed in too many places.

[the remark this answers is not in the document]
seladb: So let's rename it to `getLastError()`?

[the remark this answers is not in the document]
Dimi1010: I guess we can do that. If we need to add a string return type later  it will need renaming again, but I guess its not a big issue since its an internal class.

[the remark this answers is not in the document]
Dimi1010: Changed e481d1c

Pcap++/header/PcapDevice.h
Dimi1010: Fixed e5bfda6

Pcap++/header/PcapDevice.h
Dimi1010: Fixed e481d1c and febc6f4

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -409,8 +412,8 @@ namespace pcpp
 			return;
 		}
 
-		bool sameDescriptor = (m_PcapDescriptor == m_PcapSendDescriptor);
-		pcap_close(m_PcapDescriptor);
+		bool sameDescriptor = (m_PcapDescriptor.get() == m_PcapSendDescriptor);
+		m_PcapDescriptor = nullptr;
```
tigercosmos, line 416: We probably don't need this line because it will be destroyed automatically?

Dimi1010: `close` can be called in other places instead of only in the destructor, therefor we can't assume that `m_pcapDescriptor` will be destroyed and rely on that.

tigercosmos: Not a fan for setting `nullptr` as the way to close the handle. Maybe we should make the close function public and call `m_PcapDescriptor.close()`?

Dimi1010: Sure, we can expose a public function to close the handle.

Dimi1010: Added a function `reset()` mirroring `std::unique_ptr`'s implementation, that can be used to close the handle.

tigercosmos (approved):

seladb (approved):

PR #1539, opened 2024-08-18
[Do not merge] CI test

egecetin:
Do not merge! Only for testing CI.

With these changes it is possible to use pre-commit.ci to speed up pre-commit action in CI. This PR is for discussion and test we can merge or close it later


.pre-commit-config.yaml
```diff
@@ -1,32 +1,26 @@
...
   - repo: https://github.com/psf/black
     rev: 24.4.2
     hooks:
       - id: black
-  - repo: https://github.com/pocc/pre-commit-hooks
-    rev: v1.3.5
-    hooks:
-      - id: clang-format
-        args: ["--style=file"]  # Use the .clang-format file for configuration
-        files: ^(Common\+\+|Packet\+\+|Pcap\+\+|Tests|Examples)/.*\.(cpp|h)$
```
tigercosmos, line 27: But now if an user run `pre-commit --all`, he will not get the error from clang-format?

egecetin: Unfortunately yes, but pre-commit.ci has better performance. It is a drawback unfortunately

tigercosmos: It's okay to remove it from pre-commit. Perhaps we could provide a pre-check script for the users to run pre-commit, formatting checks, and any other checks before they commit.

egecetin: Agreed 👍. It is better to have a script for them. Still trying different things. When everything is settled, I can convert them to scripts

PR #1540, opened 2024-08-18
Parse ASN.1 root record in x509 certificates

seladb:
In the future we might add code to parse the entire certificate data, but for now we only parse the root record, which allows access to all the sub-records to get the (raw) data

### DIFF
--- a/Packet++/header/SSLHandshake.h
+++ b/Packet++/header/SSLHandshake.h
@@ -3,6 +3,7 @@
 #include <utility>
 #include "SSLCommon.h"
 #include "PointerVector.h"
+#include "Asn1Codec.h"
 
 /**
  * @file
@@ -302,6 +303,12 @@ namespace pcpp
 		}
 
 		/**
+		 * @return The root ASN.1 record of the certificate data. All of the certificate data will be under this record.
+		 * If the Root ASN.1 record is malformed, an exception is thrown
+		 */
+		Asn1SequenceRecord* getRootAsn1Record();
+
+		/**
 		 * Certificate messages usually spread on more than 1 packet. So a certificate is likely to split between 2
 		 * packets or more. This method provides an indication whether all certificate data exists or only part of it
 		 * @return True if this data contains all certificate data, false otherwise
@@ -312,6 +319,7 @@ namespace pcpp
 		}
 
 	private:
+		std::unique_ptr<Asn1Record> m_Asn1Record;
 		uint8_t* m_Data;
 		size_t m_DataLen;
 		bool m_AllDataExists;
--- a/Packet++/src/SSLHandshake.cpp
+++ b/Packet++/src/SSLHandshake.cpp
@@ -1216,6 +1216,20 @@ namespace pcpp
 		return result;
 	}
 
+	// --------------------------
+	// SSLx509Certificate methods
+	// --------------------------
+
+	Asn1SequenceRecord* SSLx509Certificate::getRootAsn1Record()
+	{
+		if (m_Asn1Record == nullptr)
+		{
+			m_Asn1Record = Asn1Record::decode(m_Data, m_DataLen);
+		}
+
+		return m_Asn1Record->castAs<Asn1SequenceRecord>();
+	}
+
 	// ---------------------------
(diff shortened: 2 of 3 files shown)


Packet++/header/SSLHandshake.h
```diff
@@ -301,6 +302,12 @@ namespace pcpp
 			return m_DataLen;
 		}
 
+		/**
+		 * @return The root ASN.1 record of the certificate data. All of the certificate data will be under this record.
+		 * If the Root ASN.1 record is malformed, an exception is thrown
```
Dimi1010, line 307: Probably better to use `@throws` special command instead, as it can also show the type of exception.

Dimi1010 (approved):
Minor note about the documentation, but LGTM.

Packet++/header/SSLHandshake.h
tigercosmos, line 307: nit: Root -> root

Packet++/src/SSLHandshake.cpp
```diff
@@ -1216,6 +1216,20 @@ namespace pcpp
 		return result;
 	}
 
+	// --------------------------
+	// SSLx509Certificate methods
+	// --------------------------
+
+	Asn1SequenceRecord* SSLx509Certificate::getRootAsn1Record()
+	{
+		if (m_Asn1Record == nullptr)
```
tigercosmos, line 1225: I am not familiar with the protocol, but I wonder what if `m_Asn1Record` has something, do we want to update it?

Packet++/header/SSLHandshake.h
seladb: The problem with `@throws` is that it expects the exception type. While I can search for all the possible exception types and mention them, it's not a [recommend pattern](https://stackoverflow.com/questions/4529810/how-to-document-all-exceptions-a-function-might-throw), and the best practice is only to specify exceptions thrown by the method itself.

I think we have 2 options:
- Keep the generic message without `@throws`
- Don't mention anything about an exception

Packet++/src/SSLHandshake.cpp
seladb: This `SSLx509Certificate` class is "read only", meaning updating the instance is impossible.

This pattern is for lazy loading - only load it if the user asks

tigercosmos (approved):

Packet++/header/SSLHandshake.h
Dimi1010: I guess we can keep the generic message for now then.

I looked through the `Asn1Codec.cpp` file. A thing I noticed is a lot of `std::invalid_argument` during decoding, that are not really about the specific argument being invalid and more about the decode procedure being invalid in that specific situation.

Perhaps we can add a custom `Asn1DecodeError` exception to use for ASN.1 decode failures, with potential derived exceptions if needed. It would also simplify the current issue with `@throws` where it can just be a notice that the function throws that base exception if the parse fails.

Dimi1010: Expanding on what I meant on the `std::invalid_argument` not being used precisely.

Look at this code [Asn1Codec.cpp#L232](https://github.com/seladb/PcapPlusPlus/blob/master/Packet%2B%2B/src/Asn1Codec.cpp#L232)
```
Asn1Record* Asn1Record::decodeTagAndCreateRecord(const uint8_t* data, size_t dataLen, int& tagLen)
{
	if (dataLen < 1)
	{
		throw std::invalid_argument("Cannot decode ASN.1 record tag");
	}
...
```
In this segment, neither `data` or `dataLen` are invalid per se. An invalid argument would be `data` being a `nullptr` or something. I suppose an argument could be made of `dataLen` of size 0 being invalid. but later in the code the same exception is raised with `if (dataLen < 2)` condition, so that exception is also thrown when size is other than 0.

seladb: The `std::invalid_argument` exception is to signal that the input data byte buffer is malformed, hence decoding failed.  Maybe we can consider a more specific error for it, but I'm not sure 🤔 

Anyway I can merge this PR as is, and we can decide on it later

PR #1543, opened 2024-08-21
Add new pre-commit hooks

egecetin:


.pre-commit-config.yaml
```diff
@@ -23,7 +31,7 @@ repos:
     rev: v1.3.5
     hooks:
       - id: clang-format
-        args: ["--style=file"]  # Use the .clang-format file for configuration
+        args: ["--style=file", "-i"]  # Use the .clang-format file for configuration and apply all fixes
```
tigercosmos, line 34: Wait, why do we need "-i"? We will need pre-commit to report unformatted code, and with this flag clang-format will not report error on the CI.

cmake/setup_dpdk.py
tigercosmos, line 1: Empty?

egecetin: Change of permission 644 -> 755

.pre-commit-config.yaml
egecetin: It should still give errors but for now local pre-commit runs also fixes formatting issues.

tigercosmos: Can you confirm this? Maybe I remember wrong.

egecetin: @tigercosmos https://github.com/seladb/PcapPlusPlus/actions/runs/10502860257/job/29095177138?pr=1543

.pre-commit-config.yaml
```diff
@@ -38,3 +46,12 @@ repos:
       - id: typos
         args: ['--config=typos-config.toml']
         pass_filenames: false
+  - repo: https://github.com/Yelp/detect-secrets
+    rev: v1.5.0
+    hooks:
+      - id: detect-secrets
+        args: ['--exclude-files', 'Tests/*']
```
seladb, line 53: do we really have secrets that we need this pre-commit hook? It's an open source project 🙂

.pre-commit-config.yaml
```diff
@@ -10,11 +10,19 @@ repos:
...
     hooks:
-      - id: trailing-whitespace
       - id: check-case-conflict
+      - id: check-executables-have-shebangs
+      - id: check-json
+      - id: check-shebang-scripts-are-executable
+      - id: check-merge-conflict
+      - id: check-symlinks
+      - id: destroyed-symlinks
+      - id: detect-private-key
```
seladb, line 20: ditto

.pre-commit-config.yaml
```diff
@@ -10,11 +10,19 @@ repos:
...
     rev: v4.6.0
     hooks:
-      - id: trailing-whitespace
       - id: check-case-conflict
+      - id: check-executables-have-shebangs
+      - id: check-json
+      - id: check-shebang-scripts-are-executable
+      - id: check-merge-conflict
+      - id: check-symlinks
+      - id: destroyed-symlinks
```
seladb, line 19: we don't have symlinks because Windows won't support them...

.pre-commit-config.yaml
seladb: so why do we need the `-i` argument?

tigercosmos: So we a user run pre-commit, the wrong format will be fixed automatically. Currently we need to manually run clang-format.

seladb: oh ok 👍

.pre-commit-config.yaml
egecetin: It is actually for prevent adding secret keys. Since it is an open source project there should be no secret key, so it should pass. Otherwise (if someone tried to add) it will be fail.

.pre-commit-config.yaml
egecetin: This hooks aims to detect them. But since we dont have any symlink I think i can remove

Dimi1010: > we don't have symlinks because Windows won't support them...

Fun fact: Windows versions since Vista do have symbolic link support. It is just that they are generally not that widely used as creating them required admin elevation and creating them is hard and obscure.

.pre-commit-config.yaml
clementperon: Could be also like Token or Password, it's not because it's open source that there is no security / authentification.

seladb: I don't have a strong opinion about adding it, however, I'm doubtful that we'll add secrets that we have (mostly used in CI) to our code-base

.pre-commit-config.yaml
seladb: Yeah, I guess we can just remove this check and save a few seconds in CI 🙂

egecetin: Removed with 8046a5f

.pre-commit-config.yaml
egecetin: Removed with 8046a5f

.pre-commit-config.yaml
egecetin: Removed with 8046a5f

seladb (approved):

PR #1545, opened 2024-08-21
Disable jobs of old pushes to speed up PR runners

egecetin:
With this change, runners from old commits will be automatically cancelled if there is a newer commit to release CI resources


tigercosmos (approved):
Brilliant!

seladb (approved):

PR #1547, opened 2024-08-23
Fixes various MSVC compiler warnings.

Dimi1010:


Packet++/src/IPv6Layer.cpp
```diff
@@ -133,6 +134,10 @@ namespace pcpp
 			}
 			else
 			{
+				if (curExt == nullptr)
+				{
+					throw std::logic_error("curExt is nullptr");
+				}
 				curExt->setNextHeader(newExt);
 				curExt = curExt->getNextHeader();
 			}
```
Dimi1010, line 143: Hmm, this could possibly be an assert instead of an exception but idk...

tigercosmos (approved):
Nice refactoring, LGTM

egecetin (approved):

seladb (approved):

PR #1548, opened 2024-08-25
Run static-analysis for only changed files and move to pre-commit.ci

egecetin:
- Remove pre-commit action since pre-commit.ci has better performance. As a drawback pre-commits require external tools should be removed.
- Introduce static-analysis job to run all static analyzes including removed tools from pre-commit.
- Run static-analysis tools for only changed files for faster execution
- Add "needs" directive to unit-test jobs. If formatting/static analysis results are failed, prevent unnecessary CI resource usage

Check every static-analysis steps really failing for,
- [x] "changed"
- [x] "all"


egecetin:
To enable pre-commit just sign in [pre-commit.ci](https://pre-commit.ci/) and add an installation to repo. It should start to trig pre-commit action for all new commits/PRs

egecetin:
Sorry for early review request. Looks like clang-tidy failed with different error. https://github.com/seladb/PcapPlusPlus/actions/runs/10630947277/job/29470868443

seladb:
@egecetin I'm not sure I understand the context. Why do we want to remove pre-commit and add shell scripts instead? 🤔

egecetin:
@seladb pre-commit.ci has better performance (both latency and runtime) for hooks but it has a fixed container which does not include tools like cppcheck etc. So, the hooks that require external tools (except pip-installable) should be removed from pre-commit config. And I try to make this to an advantage by modifying scripts to check only modified (from dev branch) files which drastically improves to CI performance for static-analysis (especially for clang-tidy). Old pre-commit job takes approx 8 mins to compile (because of clang-tidy), now it takes just seconds if there are no high number of changes.

I just remove the manual jobs from workflow which can be controlled by pre-commit.ci (both run and autoupdate)

seladb:
@egecetin I'm not sure where I can see the pre-commit.ci runs and results?

egecetin:
@seladb Unfortunately I can't enable it from PR, also you can't enable partially (for a PR for example) you should login to pre-commit.ci to enable for whole repo but if you are just curious about the interface you can check runs from my repo [here](https://results.pre-commit.ci/repo/github/480123583)

egecetin:
@seladb I managed to trig it for my fork. Check here https://github.com/egecetin/PcapPlusPlus/pull/13

egecetin:
Instead of replacing pre-commit just changing clang-tidy command significantly improves CI performance so it is better to keep pre-commit because of it is easy for developers. For this reason I'm closing the PR. Check #1602 for clang-tidy script changes

PR #1550, opened 2024-08-25
Fixes PointerVector's slicing issue for polymorphic types.

### LINKED ISSUE #1549: PointerVector's copy constructor/assignment can potentially introduce object slicing.
The current implementation of `PointerVector<T>` contains a copy constructor/assignment that does a deep copy of all elements.

The issue is that the element wise copy is done via `T* x = new T(*elemIt)` which would slice off any polymorphic types held in the vector via a pointer to a base type.

A potential twofold solution can be:
- Removal of the copy constructor/assignment of `PointerVector<T>`, making the class similar to a collection of `std::unique_ptr`.
- Adding a SFINAE enabled `clone()` method, allowing the copy of the vector provided the base class `T` provides a `clone()` method.

Dimi1010:
Fixes #1549 

- Adds a helper Copier struct that facilitates the copy of polymorphic objects via a `clone()` member function.

NB: A valid `T::clone()` method can return either a `T*` or a `std::unique_ptr<T>`.

An demo of the changes:
```
void fn()
{
	PointerVector<int> pvI;
	auto pvI2 = pvI;  // works fine - int is a non-polymorphic type

	struct A
	{
		int y;
	};

	PointerVector<A> pvA1;
	auto pvA2 = pvA1;  // works fine - A is a non-polymorphic type

	struct B
	{
		int y;
		virtual void a() const
		{}

		virtual B* clone() const
		{
			return new B(*this);
		}
	};

	PointerVector<B> pvB1;
	auto pvB2 = pvB1;  // works fine - B is a polymorphic type with a clone() method

	struct C
	{
		int y;
		virtual void a() const
		{}
	};

	PointerVector<C> pvC1;
        auto pvC2 = pvC1;  // compile time error - C is a polymorphic type without a clone method.
}
```


Common++/header/PointerVector.h
seladb, line 1: @Dimi1010 this implementation is quite complex and IMHO hard to read. Maybe we can simplify it to something like this that keep the copy constructor and equality operator:

```cpp
// Primary template for non-polymorphic types
template<typename T, typename Enable = void>
struct DeepCopier
{
    static T* copy(const T* obj)
    {
        return new T(*obj);
    }
};

// Specialization for polymorphic types
template<typename T>
struct DeepCopier<T, typename std::enable_if<std::is_polymorphic<T>::value>::type>
    {
    static T* copy(const T* obj) 
    {
        return obj->clone();
    }
};

class PointerVector
{
    ...

    static std::vector<T*> deepCopy(std::vector<T*> const& origin)
    {
        for (const auto iter : origin)
        {
             T* objCopy = DeepCopier<T>::copy(iter);
             ...
        }
    }
}


```

Common++/header/PointerVector.h
```diff
@@ -343,21 +372,16 @@ namespace pcpp
 		static std::vector<T*> deepCopyUnsafe(std::vector<T*> const& origin)
```
seladb, line 372: Maybe we can change the method name to just `deepCopy()`?

Dimi1010: It is specifically named unsafe as the returned vector contains allocated objects that need to be manually freed. same with `freeVectorUnsafe` where the pointers in the vector refer to invalid values after the function return.

Packet++/header/RawPacket.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <stdint.h>
+#include <memory>
```
seladb, line 4: Why is this include needed? 🤔

seladb (approved):
One small comment, otherwise LGTM!

Packet++/header/RawPacket.h
Dimi1010: Its not. It's left over from when I was prototyping the clone method as i was initially thinking of using unique_ptr but that had issues with unique pointers of Derived types not being covariant of unique pointers to base types.

tigercosmos (approved):

Packet++/header/RawPacket.h
Dimi1010, line 4:
```suggestion
```

PR #1553, opened 2024-08-27
Fix Undefined-shift in pcpp::SmtpRequestLayer::getCommand

egecetin:


Packet++/src/SmtpLayer.cpp
```diff
@@ -19,6 +19,12 @@ namespace pcpp
 
 		for (size_t idx = 0; idx < field.size(); ++idx)
 		{
+			// Truncate the value if it is longer than 8 bytes
+			if (idx >= 8)
```
Dimi1010, line 23: Can we not use std::min between field.size and 8?

egecetin: Good idea!

seladb (approved):

Issue #1554, opened 2024-08-28
Add WireGuard Packet Parsing Support


nadongjun:
@seladb I've observed that PcapPlusPlus currently does not support parsing WireGuard packets (parsing only, no decryption), which I needed for my project. To address this, I have implemented functionality to detect and parse various WireGuard message types, including Handshake Initiation, Handshake Response, Cookie Reply, and Transport Data.

I believe this addition could be valuable to PcapPlusPlus users, and I would like to contribute it. If you are open to this contribution, I would appreciate it if you could open a ticket to track the process.

tigercosmos:
Hi @nadongjun any contribution is welcome. If your feature is too big, it is suggested to split it into multiple PRs. Since you are more familiar with the feature, feel free to open issues for your new feature, or you can reuse this issue can add more items here.

nadongjun:
@tigercosmos Thank you for your quick response and guidance. I will work on splitting the feature into smaller PRs and will open new issues if needed.

clementperon:
@nadongjun can we close the issue ?

nadongjun:
@clementperon  Yes, I believe it's done. I'll go ahead and close this issue.

PR #1557, opened 2024-08-31
Add WireGuard protocol support

nadongjun:
#### Reference 
- #1554
-  [WireGuard Protocol](https://www.wireguard.com/papers/wireguard.pdf)

## Summary:
- This PR introduces WireGuard protocol parsing capabilities to the `parseNextLayer` function within the UDP layer. The following features are included in this enhancement:

### WireGuard Message Structures
- Common Header (`wg_common_header`): Base structure for all WireGuard messages containing the message type and reserved fields.
- Handshake Initiation (`wg_handshake_initiation`): Structure representing the Handshake Initiation message with fields such as sender index, ephemeral keys, and MACs.
- Handshake Response (`wg_handshake_response`): Structure representing the Handshake Response message with fields such as sender and receiver indexes, ephemeral keys, and MACs.
- Cookie Reply (`wg_cookie_reply`): Structure representing the Cookie Reply message with fields such as receiver index, nonce, and encrypted cookie.
- Transport Data (`wg_transport_data`): Structure representing the Transport Data message with fields such as receiver index, counter, and encrypted data.
### WireGuardLayer Class Methods
- Constructor: Initializes the WireGuardLayer object with raw data, data length, previous layer, and packet pointer.
- Message Getters: Methods to get pointers to specific WireGuard messages (`Handshake Initiation, Handshake Response, Cookie Reply, Transport Data`).
### Static Methods
- `isWireguardPorts(uint16_t portSrc, uint16_t portDst)`: Checks if the given ports match the WireGuard port (51820).
- `isWireGuard(uint8_t* data, size_t dataLen)`: Validates if the provided data represents a WireGuard message based on its type.
### Layer Methods
- `parseNextLayer()`: No operation as WireGuard does not have subsequent layers.
- `getHeaderLen()`: Returns the length of data
- `computeCalculateFields()`: No fields to compute; method is left empty.
- `toString()`: Converts the WireGuard layer to a string representation for easier debugging and logging.
- `getOsiModelLayer()`: Returns the OSI model layer corresponding to the Network layer.
### UDP Layer function
- `UDPLayer::parseNextLayer()`: Integrates the ability to recognize and parse WireGuard messages
## Example Usage:
```c++
#include <iostream>
#include <cstring>
#include <cstdint>
#include "UdpLayer.h"
#include "WireGuardLayer.h"
#include "Packet.h"
#include "PcapFileDevice.h"

int main(int argc, char* argv[])
{
    if (argc < 2)
    {
        std::cerr << "Usage: " << argv[0] << " <pcap file>" << std::endl;
        return 1;
    }

    std::string pcapFile = argv[1];

    pcpp::PcapFileReaderDevice reader(pcapFile);
    if (!reader.open())
    {
        std::cerr << "Error opening pcap file" << std::endl;
        return 1;
    }

    pcpp::RawPacket rawPacket;
    while (reader.getNextPacket(rawPacket))
    {
        pcpp::Packet parsedPacket(&rawPacket);

        if (parsedPacket.isPacketOfType(pcpp::IPv4))
        {
            pcpp::UdpLayer* udpLayer = parsedPacket.getLayerOfType<pcpp::UdpLayer>();
            if (udpLayer != nullptr)
            {
                // Parse the next layer from the UDP layer (in this case, WireGuard layer).   
                udpLayer->parseNextLayer();

                // Check if the parsed next layer is a WireGuard layer.
                pcpp::WireGuardLayer* wgLayer = dynamic_cast<pcpp::WireGuardLayer*>(udpLayer->getNextLayer());
                if (wgLayer != nullptr)
                {
                    // Output the WireGuard layer information as a string.
                    std::cout << wgLayer->toString() << std::endl;
                }
            }
        }
    }

    reader.close();
    return 0;
}
```
### Example Usage::Result
- `INPUT`: [WireGuard Pcap file](https://gitlab.com/wireshark/wireshark/-/blob/master/test/captures/wireguard-ping-tcp.pcap)
- `OUTPUT`: [result.log](https://github.com/user-attachments/files/16823801/result.log)

image


Dimi1010:
Looks good. Can you please add tests for the layer in `Test/Packet++`?

nadongjun:
@Dimi1010 Sure, I'm working on adding the tests for the layer in Test/Packet++ and will push the commits shortly.

Tests/Packet++Test/PacketExamples/WireGuardHandshakeInitiation.dat
seladb, line 1: Can you add a pcap file that contains all these sample packets?

Tests/Packet++Test/Tests/WGTests.cpp
seladb, line 1: I'd rename it to `WireGuardTests.cpp` to make it clear it's testing WireGuard

nadongjun: Sure, I've added a pcap file that contains all the sample packets and renamed the test file to WireGuardTests.cpp

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,258 @@
...
+		 *
+		 * @param data Pointer to the raw data representing the WireGuard layer
+		 * @param dataLen Length of the data
+		 * @param prevLayer Pointer to the previous layer in the packet (if any)
+		 * @param packet Pointer to the packet this layer belongs to
+		 */
+		WireGuardLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
+		    : Layer(data, dataLen, prevLayer, packet)
+		{
+			m_Protocol = WIREGUARD;
```
Dimi1010, line 123: Btw, newest head on the dev branch can pass the protocol to the Layer constructor.

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,258 @@
...
+		 * @param data Pointer to the raw data
+		 * @param dataLen Length of the data
+		 * @return True if the data starts with a valid WireGuard message type, false otherwise
+		 */
+		static inline bool isWireGuard(const uint8_t* data, size_t dataLen);
+
+		/**
+		 * No operation required for parsing the next layer since WireGuard does not have a next layer.
+		 */
+		void parseNextLayer()
```
Dimi1010, line 189: nit: Could you add an `override` specifier?

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,258 @@
...
+		{
+			// No next layer to parse for WireGuard, do nothing
+		}
+
+		/**
+		 * Calculates the length of the header based on the message type.
+		 *
+		 * @return Size of the header in bytes. For TransportData, returns the total data length.
+		 */
+		size_t getHeaderLen() const;
```
Dimi1010, line 199: nit, ditto: `override`

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,258 @@
...
+
+		/**
+		 * No fields to compute or update, so this method is left empty.
+		 */
+		void computeCalculateFields()
```
Dimi1010, line 204: nit, ditto: `override`

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,258 @@
...
+			// Since WireGuard headers have fixed lengths and no fields to compute (like checksums or lengths),
+			// this method does not need to perform any operations. It's left empty.
+		}
+
+		/**
+		 * Converts the WireGuard layer to a string representation.
+		 *
+		 * @return String representation of the WireGuard layer
+		 */
+		std::string toString() const;
```
Dimi1010, line 215: nit, ditto: `override`

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,258 @@
...
+
+		/**
+		 * Returns the OSI model layer that this protocol belongs to.
+		 *
+		 * @return OSI model layer corresponding to the Network layer
+		 */
+		OsiModelLayer getOsiModelLayer() const
```
Dimi1010, line 222: nit, ditto: `override`

Packet++/header/WireGuardLayer.h
nadongjun: Thanks for the review. I’ve applied your suggestions, including adding the `override` specifier to all relevant methods in the `WireGuardLayer` class and updating the constructor parameters.

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,257 @@
...
+		 * @param dataLen Length of the data
+		 * @param prevLayer Pointer to the previous layer in the packet (if any)
+		 * @param packet Pointer to the packet this layer belongs to
+		 */
+		WireGuardLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
+		    : Layer(data, dataLen, prevLayer, packet, WIREGUARD)
+		{
+			m_Data = data;
+			m_DataLen = dataLen;
+		}
```
Dimi1010, line 125: Why is `m_data` and `m_DataLen` being set again? Passing the pointers to the `Layer` ctor already sets the fields in:
https://github.com/seladb/PcapPlusPlus/blob/2270bd73521e20340a57792a9eda16d63638b058/Packet%2B%2B/header/Layer.h#L214-L217

nadongjun: Good catch, I’ll remove the redundant assignments.

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,254 @@
+#pragma once
+
+#include "Layer.h"
+#include "IpAddress.h"
+#include "MacAddress.h"
+#include <string.h>
```
Dimi1010, line 6: CI is failing because `std::memcmp` used in the tests is not guaranteed to be declared in `string.h` as it is in `cstring`. Either use `cstring` or use the memcmp from the global namespace.

Dimi1010: IIRC, Note that while `<cstring>` guarantees that the functions would be in the `std::` namespace, it does not guarantee them being in the global namespace.

nadongjun: Thank you for the review.

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,254 @@
+#pragma once
+
+#include "Layer.h"
+#include "IpAddress.h"
+#include "MacAddress.h"
+#include <sstream>
+#include <cstring>
```
Dimi1010, line 7: Are those 2 includes actually used directly in the `WireGuardLayer` header?

If they aren't it would be better to remove them to avoid including them in every translation unit that includes this header.

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,151 @@
...
+namespace pcpp
+{
+	std::string WireGuardLayer::toString() const
+	{
+		std::stringstream ss;
+
+		if (m_DataLen < sizeof(wg_common_header))
+		{
+			ss << "WireGuard header (incomplete)";
+			return ss.str();
```
Dimi1010, line 21: Generally a `stringstream::str()` allocates a new string buffer for the returned sting.

Since the return value if this branch is taken is constant `"WireGuard header (incomplete)"`, it would be better to return it directly to avoid allocating twice (once for `stringstream` and once for the return `string`).

`stringstream`'s declaration can also be moved below the current branch to avoid initializing it altogether if the branch is taken.

nadongjun: Thanks for your feedback. I’ve removed the unused includes and refactored the stringstream usage to avoid unnecessary allocations.

Dimi1010 (approved):
LGTM

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,252 @@
...
+ * \namespace pcpp
+ * \brief The main namespace for the PcapPlusPlus lib
+ */
+namespace pcpp
+{
+
+	/**
+	 * WireGuard message types
+	 */
+	enum WireGuardMessageType
```
Dimi1010, line 19: Minor thing, could you change that to `enum class`. We have been trying to use `enum classes` going forward, and porting an existing `enum` to `enum class` is a pain after it is already in use while keeping backwards compatibility.

nadongjun: Thanks for your feedback.

Packet++/header/ProtocolType.h
```diff
@@ -347,6 +347,11 @@ namespace pcpp
 	 */
 	const ProtocolType LDAP = 55;
 
+	/*
+	 * Wireguard protocol
+	 */
+	const ProtocolType WIREGUARD = 56;
```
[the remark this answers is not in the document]
nadongjun: Sure, I will make an update. Thanks for your feedback.

Packet++/header/WireGuardLayer.h
seladb, line 1: Can you add this protocol to [`README.md`](https://github.com/seladb/PcapPlusPlus?tab=readme-ov-file#supported-network-protocols)?

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,253 @@
...
+
+	/**
+	 * WireGuard message types
+	 */
+	enum class WireGuardMessageType
+	{
+		HandshakeInitiation = 1,  ///< Handshake Initiation message
+		HandshakeResponse = 2,    ///< Handshake Response message
+		CookieReply = 3,          ///< Cookie Reply message
+		TransportData = 4         ///< Transport Data message
```
seladb, line 24: very nit: some comments are inline (like here), and some are on top of the attribute (like in `wg_common_header`). Maybe move all to be like `wg_common_header` which is what we usually do?

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,253 @@
...
+#pragma pack(pop)
+
+	/**
+	 * @struct wg_handshake_initiation
+	 * Represents the Handshake Initiation message
+	 */
+#pragma pack(push, 1)
+	struct wg_handshake_initiation
+	{
+		wg_common_header common;               ///< Common header for all WireGuard messages
```
seladb, line 48: `wg_handshake_initiation` can also inherit from `wg_common_header`. Ditto in the next structs

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,253 @@
...
+
+		/**
+		 * Gets a pointer to the Handshake Initiation message.
+		 *
+		 * @return Pointer to the Handshake Initiation message, or nullptr if data is invalid
+		 */
+		const wg_handshake_initiation* getHandshakeInitiation() const
+		{
+			return reinterpret_cast<const wg_handshake_initiation*>(m_Data);
+		}
```
seladb, line 130: A single packet can include one of the 4 possible messages or another unknown message (which may indicate the packet is invalid). But what if the user doesn't know which type of packet it is? How would he know which getter to call?

We usually solve this problem in one of two ways:
1. Create a base and derived classes for each message type. A good example is [`BgpLayer`](https://github.com/seladb/PcapPlusPlus/blob/master/Packet%2B%2B/header/BgpLayer.h)
2. Include all messages in the base class, closer to this implementation like we do in [`IcmpLayer`](https://github.com/seladb/PcapPlusPlus/blob/master/Packet%2B%2B/header/IcmpLayer.h), although this is less common and we rarely do it lately

If you choose option (2), which is closer to your implementation (but less OOP IMO), I'd expect the following:
- Before doing `reinterpret_cast` check if the message is of the right type and if the length is valid, otherwise return `nullptr`
- Provide another method to get the type, like: `WireGuardMessageType getWireGuardType()` so users can know the packet type before calling the relevant method

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,253 @@
...
+	public:
+		/**
+		 * Constructs a WireGuardLayer object.
+		 *
+		 * @param data Pointer to the raw data representing the WireGuard layer
+		 * @param dataLen Length of the data
+		 * @param prevLayer Pointer to the previous layer in the packet (if any)
+		 * @param packet Pointer to the packet this layer belongs to
+		 */
+		WireGuardLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
```
seladb, line 118: It'd be nice to be able to create WG packets also and not only parse them. Here we also have 2 options:
1. If we create a class per message type, it's easy because each class will have its own c'tor
2. If we decide to keep one `WireGuardLayer` class we can do what we do in [`IcmpLayer`](https://github.com/seladb/PcapPlusPlus/blob/master/Packet%2B%2B/header/IcmpLayer.h) which is add an empty c'tor, and have setter methods for each layer type which extend or shorten the layer and then set the correct data

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "IPv6Layer.h"
+#include "WireGuardLayer.h"
+#include "Logger.h"
+#include <sstream>
+#include <iomanip>
+#include <cstring>
+
+namespace pcpp
+{
+	std::string WireGuardLayer::toString() const
```
seladb, line 14: We try to make `toString()` output only one line, so we can print a full packet with one row per layer

seladb (commented):
@nadongjun I have some comments about the current implementation, so I didn't review the tests yet. Once we finalize the implementation I'll review the tests

Packet++/header/WireGuardLayer.h
nadongjun: I will add the protocol to the README.md

Packet++/header/WireGuardLayer.h
nadongjun: Thanks for the good suggestion. I will move all the comments above the attributes.

README.md
```diff
@@ -242,43 +242,44 @@ PcapPlusPlus currently supports parsing, editing and creation of packets of the
...
 
 ### Transport Layer (L4)
 
-25. COTP
-26. GTP (v1)
-27. IPSec AH & ESP - parsing only (no editing capabilities)
-28. TCP
-29. TPKT
-30. UDP
+26. COTP
```
tigercosmos, line 249: The numbering here seems to be very unfriendly for new layers. How about changing them all into the non-number list? @seladb

Packet++/header/WireGuardLayer.h
nadongjun: Thank you for the feedback. I have implemented the first approach by creating a base class and derived classes for each message type, similar to the BgpLayer implementation.

Packet++/header/WireGuardLayer.h
Dimi1010: I would like to propose option 3, which is tagged constructors. Basically we have a special structure thàt is the MessageTag, that is passed as first param to the constructor to differentiate which overloads are to be used even if the overloads from different messages use similar input parameters.

nadongjun: @seladb I've implemented the feature to create WireGuard messages using constructors, following option 1.

@Dimi1010 Thank you for your suggestion regarding tagged constructors. However, I have already implemented the functionality using option 1. If any additional modifications are needed, I will be happy to consider them.

README.md
seladb: I think that it's nice to see the overall number of protocols, but otherwise no real reason...

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,423 @@
...
+		{
+			/** Message type field */
+			uint8_t messageType;
+			/** Reserved field (3 bytes) */
+			uint8_t reserved[3];
+		};
+#pragma pack(pop)
+
+		WireGuardLayer()
+		{}
```
seladb, line 53: This is redundant?

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,423 @@
...
+		/**
+		 * Checks if the given data represents a WireGuard message.
+		 *
+		 * @param data Pointer to the raw data
+		 * @param dataLen Length of the data
+		 * @return True if the data starts with a valid WireGuard message type, false otherwise
+		 */
+		static bool isDataValid(const uint8_t* data, size_t dataLen);
+
+		static WireGuardLayer* parseWireGuardLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet);
```
seladb, line 93: Let's add doxygen documentation here and in the other public methods that don't have it?

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,423 @@
...
+
+		std::string getMessageTypeAsString() const;
+
+		uint32_t getMessageType()
```
seladb, line 97: This method can be `const`

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,423 @@
...
+		{
+			return getBasicHeader()->messageType;
+		}
+
+		uint8_t* getReserved()
```
seladb, line 102: 1. `reserved` is always 3 bytes, so maybe it should return `uint32_t`?
2. This method can be `const`

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,324 @@
...
+#include <cstring>
+
+namespace pcpp
+{
+	void WireGuardLayer::parseNextLayer()
+	{
+		size_t headerLen = getHeaderLen();
+		if (m_DataLen <= headerLen || headerLen == 0)
+			return;
+		m_NextLayer = WireGuardLayer::parseWireGuardLayer(m_Data, m_DataLen, this, m_Packet);
```
seladb, line 19: `parseNextLayer` should always generate a layer. If `m_DataLen <= headerLen` (and it's not 0) or if `WireGuardLayer::parseWireGuardLayer()` returns `nullptr`, this method should generate a `PayloadLayer`

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,423 @@
...
+		 * in the packet to determine if another WireGuard message exists. If so, it parses the next message
+		 * as a new WireGuard layer.
+		 */
+		void parseNextLayer() override;
+
+		/**
+		 * Calculates the length of the header based on the message type.
+		 *
+		 * @return Size of the header in bytes. For TransportData, returns the total data length.
+		 */
```
seladb, line 119: When I look at the implementation, it always returns `m_DataLen`, regardless of the message type 🤔

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,423 @@
...
+		size_t getHeaderLen() const override;
+
+		/**
+		 * No fields to compute or update, so this method is left empty.
+		 */
+		void computeCalculateFields() override
+		{
+			// Since WireGuard headers have fixed lengths and no fields to compute (like checksums or lengths),
+			// this method does not need to perform any operations. It's left empty.
```
seladb, line 128: nit: This comment is not really needed

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,324 @@
...
+		case static_cast<uint8_t>(WireGuardMessageType::TransportData):
+			return "Transport Data";
+		default:
+			return "Unknown";
+		}
+	}
+
+	std::string WireGuardLayer::toString() const
+	{
+		return "WireGuardLayer, " + getMessageTypeAsString() + " message";
```
seladb, line 64: nit: add a space:
```cpp
return "WireGuard Layer, " + getMessageTypeAsString() + " message";
```

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,423 @@
...
+	 * @class WireGuardLayer
+	 * Represents a WireGuard protocol layer
+	 */
+	class WireGuardLayer : public Layer
+	{
+	public:
+		/**
+		 * WireGuard message types
+		 */
+		enum class WireGuardMessageType
```
seladb, line 26: I think we need to add an `Unknown` value in case we try to parse an invalid message

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,324 @@
...
+		if (m_DataLen <= headerLen || headerLen == 0)
+			return;
+		m_NextLayer = WireGuardLayer::parseWireGuardLayer(m_Data, m_DataLen, this, m_Packet);
+	}
+
+	WireGuardLayer* WireGuardLayer::parseWireGuardLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
+	{
+		if (dataLen < sizeof(WireGuardLayer::wg_common_header))
+			return nullptr;
+		wg_common_header* wgHeader = reinterpret_cast<wg_common_header*>(data);
```
seladb, line 26: If we have a method like `getMessageType()` that returns `WireGuardMessageType`, maybe we should use it here?

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,324 @@
...
+		default:
+			return nullptr;
+		}
+	}
+
+	std::string WireGuardLayer::getMessageTypeAsString() const
+	{
+		// Assuming you have a method to retrieve the message type from the header
+		wg_common_header* wgHeader = reinterpret_cast<wg_common_header*>(this->getData());
+		switch (wgHeader->messageType)
```
seladb, line 47: If we have a method like `getMessageType()` that returns `WireGuardMessageType`, maybe we should use it here?

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,324 @@
...
+		const size_t messageLen = sizeof(wg_handshake_initiation);
+		m_DataLen = messageLen;
+		m_Data = new uint8_t[messageLen];
+		memset(m_Data, 0, messageLen);
+
+		wg_handshake_initiation* msgHdr = reinterpret_cast<wg_handshake_initiation*>(m_Data);
+
+		msgHdr->messageType = static_cast<uint8_t>(WireGuardMessageType::HandshakeInitiation);
+
+		std::memset(msgHdr->reserved, 0, sizeof(msgHdr->reserved));
```
seladb, line 101: This is redundant because we already use `memset` in line 95.

Same in the other c'tors

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+		msgHdr->senderIndex = senderIndex;
```
seladb, line 103: I think we should use `htobe32()` here to convert from host order to BigEndian order.

Same in the other c'tors

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,324 @@
...
+		std::memcpy(msgHdr->encryptedTimestamp, encryptedTimestamp, 28);
+		std::memcpy(msgHdr->mac1, mac1, 16);
+		std::memcpy(msgHdr->mac2, mac2, 16);
+
+		m_Protocol = Wireguard;
+	}
+
+	uint32_t WireGuardHandshakeInitiationLayer::getMessageType() const
+	{
+		return getHandshakeInitiationHeader()->messageType;
```
seladb, line 115: This method should use `be32toh()` to convert from BigEndian to Host order.

Same in the other properties that return `uint32_t`

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,423 @@
...
+		                                const uint8_t mac1[16], const uint8_t mac2[16]);
+
+		uint32_t getMessageType() const;
+		const uint8_t* getReserved() const;
+		uint32_t getSenderIndex() const;
+		uint32_t getReceiverIndex() const;
+		const uint8_t* getResponderEphemeral() const;
+		const uint8_t* getEncryptedEmpty() const;
+		const uint8_t* getMac1() const;
+		const uint8_t* getMac2() const;
```
seladb, line 285: Maybe these properties should return `std::array` and copy the data from the packet to the array?

Same in properties in the other classes that return `uint8_t*`

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,423 @@
...
+		{
+			return (wg_handshake_response*)getBasicHeader();
+		}
+
+		// implement abstract methods
+
+		WireGuardMessageType getWireGuardMessageType() const
+		{
+			return WireGuardMessageType::HandshakeResponse;
+		}
```
seladb, line 297: There is no virtual abstract method in `WireGuardLayer`... once we add it this method and the others should be marked as `override`

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,423 @@
...
+		 * @param prevLayer Pointer to the previous layer in the packet (if any)
+		 * @param packet Pointer to the packet this layer belongs to
+		 */
+		WireGuardLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
+		    : Layer(data, dataLen, prevLayer, packet, Wireguard)
+		{}
+
+		wg_common_header* getBasicHeader() const
+		{
+			return (wg_common_header*)m_Data;
```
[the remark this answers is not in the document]
seladb: There are still C-Style casts in this file, can you change all of them to `static_cast` or `reinterpret_cast`?

Packet++/header/WireGuardLayer.h
seladb: This c'tor is still there...

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,463 @@
...
+		 * @param[in] mac1 The MAC1 field
+		 * @param[in] mac2 The MAC2 field
+		 */
+		WireGuardHandshakeInitiationLayer(uint32_t senderIndex, const uint8_t initiatorEphemeral[32],
+		                                  const uint8_t encryptedInitiatorStatic[48],
+		                                  const uint8_t encryptedTimestamp[28], const uint8_t mac1[16],
+		                                  const uint8_t mac2[16]);
+
+		uint32_t getMessageType() const;
+		uint32_t getReserved() const;
```
seladb, line 244: Why are those 2 methods implemented in every derived class?

seladb:
@nadongjun I resolved the comments that were fixed and kept open those that weren't. Can you please go over the open comments and resolve them?

nadongjun:
@seladb I’ll take a look at the open comments and address them soon. Thank you for your detailed feedback.

Packet++/header/WireGuardLayer.h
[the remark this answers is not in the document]
nadongjun: Sure, I've made the changes.

Packet++/header/WireGuardLayer.h
nadongjun: Sure, I've made the changes.

Packet++/header/WireGuardLayer.h
nadongjun: The subclasses of WireGuardLayer (WireGuardMessages) require a default constructor to be instantiated without initial data. Removing it leads to compilation errors.

Packet++/src/WireGuardLayer.cpp
nadongjun: Sure, I've made the changes.

Packet++/header/WireGuardLayer.h
nadongjun: Sure, I've made the changes.

Packet++/src/WireGuardLayer.cpp
nadongjun: This function is static, so it cannot call the protected function getMessageType(). Therefore, getMessageType is not used in this function.

Packet++/src/WireGuardLayer.cpp
nadongjun: Sure, I've made the changes.

Packet++/header/WireGuardLayer.h
nadongjun: Sure, I've made the changes.

Packet++/header/WireGuardLayer.h
nadongjun: I've removed the duplicate implementations of those two methods from the derived classes.

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,596 @@
...
+		/**
+		 * Get the basic header common to all WireGuard messages.
+		 *
+		 * @return Pointer to the common header structure.
+		 */
+		wg_common_header* getBasicHeader() const
+		{
+			return reinterpret_cast<wg_common_header*>(m_Data);
+		}
+
```
seladb, line 45: nit: private/protected methods or structs don't need doxygen documentation (same in the other classes in this file)

Packet++/header/WireGuardLayer.h
seladb: You're right! Let's move it to the `protected` area so it won't be public?

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,596 @@
...
+		                                  const uint8_t encryptedInitiatorStatic[48],
+		                                  const uint8_t encryptedTimestamp[28], const uint8_t mac1[16],
+		                                  const uint8_t mac2[16]);
+
+		/**
+		 * Get the sender index from the Handshake Initiation message.
+		 *
+		 * @return The sender index as a 32-bit integer.
+		 */
+		uint32_t getSenderIndex() const;
```
seladb, line 245: We have getter methods for all fields (in this class and the other classes), but we don't have setter methods. Do we want to add setter methods also?

For example:
```cpp
void setSenderIndex(uint32_t senderIndex);
void setInitiatorEphemeral(std::array<uint8_t, 32> initEphemeral);
...
void setReceiverIndex(uint32_t receiverIndex);
...

```

Tests/Packet++Test/PacketExamples/WireGuard.pcap
seladb, line 1: This pcap file doesn't include a Cookie Reply message. I searched online but couldn't find anything. Maybe you have an example?

Tests/Packet++Test/Tests/WireGuardTests.cpp
seladb, line 1: Please include tests for all methods and properties in the wireguard layer classes

Tests/Packet++Test/PacketExamples/WireGuard.pcap
nadongjun: I also looked for a pcap related to the Cookie Reply, but I couldn't find one. Since this message is generated during DDoS attacks, it may not be publicly available, and I do not have it either. For Cookie Reply, how about creating packets using the WireGuardCookieReplyLayer class and adding separate test code that includes parsing tests?

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,309 @@
...
+namespace pcpp
+{
+	void WireGuardLayer::parseNextLayer()
+	{
+		size_t headerLen = getHeaderLen();
+		if (m_DataLen <= headerLen && headerLen != 0)
+			m_NextLayer = new PayloadLayer(m_Data, m_DataLen, this, m_Packet);
+		else
+		{
+			m_NextLayer = WireGuardLayer::parseWireGuardLayer(m_Data, m_DataLen, this, m_Packet);
```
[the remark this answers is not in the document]
seladb: So in that case `WireGuardLayer::parseNextLayer()` shouldn't do anything, no?
```cpp
/**
 * Does nothing for this layer (WireGuard layer is always last)
 */
void parseNextLayer() override
{}
```

Tests/Packet++Test/PacketExamples/WireGuard.pcap
seladb: sure, you can generate a packet, check it in Wireshark to make sure it's valid, then add it to the test 👍

Packet++/header/WireGuardLayer.h
nadongjun: Sure, I've made the changes.

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,596 @@
...
+		 *
+		 * @return String representation of the message type.
+		 */
+		std::string getMessageTypeAsString() const;
+
+		/**
+		 * Get the message type of the WireGuard message.
+		 *
+		 * @return The message type as an unsigned 32-bit integer.
+		 */
```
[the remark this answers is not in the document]
nadongjun: Sure, I've made the changes.

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,309 @@
...
+#include "UdpLayer.h"
+#include "PayloadLayer.h"
+#include "IPv4Layer.h"
+#include "IPv6Layer.h"
+#include "WireGuardLayer.h"
+#include "Logger.h"
+#include "EndianPortable.h"
+#include <sstream>
+#include <iomanip>
+#include <cstring>
```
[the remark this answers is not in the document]
nadongjun: Sure, I've made the changes.

Packet++/src/WireGuardLayer.cpp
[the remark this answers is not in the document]
nadongjun: Sure, I've made the changes.

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,596 @@
...
+
+		/**
+		 * Get the Handshake Initiation message header.
+		 *
+		 * @return A pointer to the Handshake Initiation message structure.
+		 */
+		wg_handshake_initiation* getHandshakeInitiationHeader() const
+		{
+			return reinterpret_cast<wg_handshake_initiation*>(getBasicHeader());
+		}
```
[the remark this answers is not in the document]
nadongjun: Sure, I've made the changes.

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,596 @@
...
+
+		/**
+		 * Get the Transport Data message header.
+		 *
+		 * @return A pointer to the Transport Data message structure.
+		 */
+		wg_transport_data* getTransportHeader() const
+		{
+			return reinterpret_cast<wg_transport_data*>(getBasicHeader());
+		}
```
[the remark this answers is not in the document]
nadongjun: Sure, I've made the changes.

Tests/Packet++Test/Tests/WireGuardTests.cpp
nadongjun: Sure, I've made the changes.

Packet++/header/WireGuardLayer.h
nadongjun: I added documentation to the protected struct to avoid CI errors.

Packet++/header/WireGuardLayer.h
```diff
@@ -0,0 +1,596 @@
...
+
+		/**
+		 * Get the Cookie Reply message header.
+		 *
+		 * @return A pointer to the Cookie Reply message structure.
+		 */
+		wg_cookie_reply* getCookieReplyHeader() const
+		{
+			return reinterpret_cast<wg_cookie_reply*>(getBasicHeader());
+		}
```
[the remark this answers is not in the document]
nadongjun: Sure, I've made the changes.

Packet++/header/WireGuardLayer.h
seladb: CI won't fail if private/protected methods don't have doxygent documentation, it will only fail if public method don't have it

Packet++/src/WireGuardLayer.cpp
```diff
@@ -0,0 +1,292 @@
+#define LOG_MODULE PacketLogModuleWireGuardLayer
+
+#include "UdpLayer.h"
+#include "PayloadLayer.h"
```
seladb, line 4: `PayloadLayer` is not used anymore so we can remove this `#include`

Packet++/header/WireGuardLayer.h
seladb: yes sure 👍

Tests/Packet++Test/PacketExamples/WireGuard.pcap
seladb: Thanks! 🙏 
Can you please add this new packet to `WireGuard.pcap`? You can use `mergecap` which is an executable included in Wireshark to merge the existing `WireGuard.pcap` with the pcap that includes this new packet

Tests/Packet++Test/Tests/WireGuardTests.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+	pcpp::Packet wgHandShakeInitPacket(&rawPacket1);
+
+	PTF_ASSERT_TRUE(wgHandShakeInitPacket.isPacketOfType(pcpp::WireGuard));
+
+	pcpp::WireGuardLayer* wgLayer = wgHandShakeInitPacket.getLayerOfType<pcpp::WireGuardLayer>();
+	PTF_ASSERT_NOT_NULL(wgLayer);
+
+	pcpp::WireGuardHandshakeInitiationLayer* wgHandShakeInitLayer =
+	    wgHandShakeInitPacket.getLayerOfType<pcpp::WireGuardHandshakeInitiationLayer>();
```
seladb, line 24: nit: you can use `auto` here.

Same in the other tests - let's use `auto` wherever possible?

Tests/Packet++Test/Tests/WireGuardTests.cpp
```diff
@@ -0,0 +1,336 @@
...
+	PTF_ASSERT_NOT_NULL(wgHandShakeInitLayer);
+
+	PTF_ASSERT_EQUAL(wgHandShakeInitLayer->getMessageTypeAsString(), "Handshake Initiation");
+	PTF_ASSERT_EQUAL(wgHandShakeInitLayer->toString(),
+	                 "WireGuard Layer, " + wgHandShakeInitLayer->getMessageTypeAsString() + " message");
+
+	PTF_ASSERT_TRUE(wgHandShakeInitLayer->getWireGuardMessageType() ==
+	                pcpp::WireGuardLayer::WireGuardMessageType::HandshakeInitiation);
+
+	PTF_ASSERT_EQUAL(wgHandShakeInitLayer->getSenderIndex(), be32toh(818952152));
```
seladb, line 34: no need to use `be32toh` here, we can just put the value from Wireshark: `0x30d037d8`

Tests/Packet++Test/Tests/WireGuardTests.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+	PTF_ASSERT_EQUAL(wgHandShakeResponseLayer->getMessageTypeAsString(), "Handshake Response");
+	PTF_ASSERT_EQUAL(wgHandShakeResponseLayer->toString(),
+	                 "WireGuard Layer, " + wgHandShakeResponseLayer->getMessageTypeAsString() + " message");
+
+	PTF_ASSERT_TRUE(wgHandShakeResponseLayer->getWireGuardMessageType() ==
+	                pcpp::WireGuardLayer::WireGuardMessageType::HandshakeResponse);
+
+	PTF_ASSERT_EQUAL(wgHandShakeResponseLayer->getSenderIndex(), be32toh(2877158406));
+	PTF_ASSERT_EQUAL(wgHandShakeResponseLayer->getReceiverIndex(), be32toh(818952152));
```
seladb, line 87: ditto: no need to use `be32toh`, we can use the value from Wireshark

Tests/Packet++Test/Tests/WireGuardTests.cpp
```diff
@@ -0,0 +1,336 @@
...
+	PTF_ASSERT_EQUAL(wgCookieReplyaLayer->getMessageTypeAsString(), "Cookie Reply");
+	PTF_ASSERT_EQUAL(wgCookieReplyaLayer->toString(),
+	                 "WireGuard Layer, " + wgCookieReplyaLayer->getMessageTypeAsString() + " message");
+
+	PTF_ASSERT_TRUE(wgCookieReplyaLayer->getWireGuardMessageType() ==
+	                pcpp::WireGuardLayer::WireGuardMessageType::CookieReply);
+
+	uint32_t receiverIndex = 2877158406;
+
+	PTF_ASSERT_EQUAL(wgCookieReplyaLayer->getReceiverIndex(), receiverIndex);
```
seladb, line 135: We can just do: `PTF_ASSERT_EQUAL(wgCookieReplyaLayer->getReceiverIndex(), 2877158406);`

Tests/Packet++Test/Tests/WireGuardTests.cpp
```diff
@@ -0,0 +1,336 @@
...
+	PTF_ASSERT_EQUAL(wgTransportDataLayer->toString(),
+	                 "WireGuard Layer, " + wgTransportDataLayer->getMessageTypeAsString() + " message");
+
+	PTF_ASSERT_TRUE(wgTransportDataLayer->getWireGuardMessageType() ==
+	                pcpp::WireGuardLayer::WireGuardMessageType::TransportData);
+
+	PTF_ASSERT_EQUAL(wgTransportDataLayer->getReceiverIndex(), be32toh(2877158406));
+
+	uint64_t expectedCounter = 0x0000000000000000;
+	PTF_ASSERT_EQUAL(wgTransportDataLayer->getCounter(), be32toh(expectedCounter));
```
seladb, line 173: ditto: we can do `PTF_ASSERT_EQUAL(wgTransportDataLayer->getCounter(), 0);`

Tests/Packet++Test/Tests/WireGuardTests.cpp
seladb, line 1: Can we add tests for:
- `getMessageType()`
- `getReserved()`
- `getHeaderLen()`

Tests/Packet++Test/Tests/WireGuardTests.cpp
```diff
@@ -0,0 +1,336 @@
...
+	PTF_ASSERT_TRUE(memcmp(newHandshakeInitMessage.getInitiatorEphemeral().data(), expectedPublicKeyInit, 32) == 0);
+	PTF_ASSERT_TRUE(memcmp(newHandshakeInitMessage.getEncryptedInitiatorStatic().data(), expectedStaticKeyInit, 48) ==
+	                0);
+	PTF_ASSERT_TRUE(memcmp(newHandshakeInitMessage.getEncryptedTimestamp().data(), expectedTimestampInit, 28) == 0);
+	PTF_ASSERT_TRUE(memcmp(newHandshakeInitMessage.getMac1().data(), expectedMac1Init, 16) == 0);
+	PTF_ASSERT_TRUE(memcmp(newHandshakeInitMessage.getMac2().data(), expectedMac2Init, 16) == 0);
+
+	PTF_ASSERT_TRUE(wgHandshakeInitPacket.addLayer(&newHandshakeInitMessage));
+
+	PTF_ASSERT_EQUAL(wgHandshakeInitPacket.getRawPacket()->getRawDataLen(), bufferLength1);
```
seladb, line 240: We can also test the full buffer:
```cpp
PTF_ASSERT_BUF_COMPARE(newHandshakeInitMessage.getData(), origHandshakeInitMessage->getData(), origHandshakeInitMessage->getDataLen());
```

This will ensure all the data in the new packet is correct.

Same goes for the other packets we create later in the test

Tests/Packet++Test/Tests/WireGuardTests.cpp
```diff
@@ -0,0 +1,336 @@
...
+	PTF_ASSERT_NOT_NULL(origHandshakeInitMessage);
+	PTF_ASSERT_EQUAL(newHandshakeInitMessage.getDataLen(), origHandshakeInitMessage->getDataLen());
+
+	PTF_ASSERT_EQUAL(newHandshakeInitMessage.getSenderIndex(), 818952152);
+	PTF_ASSERT_TRUE(memcmp(newHandshakeInitMessage.getInitiatorEphemeral().data(), expectedPublicKeyInit, 32) == 0);
+	PTF_ASSERT_TRUE(memcmp(newHandshakeInitMessage.getEncryptedInitiatorStatic().data(), expectedStaticKeyInit, 48) ==
+	                0);
+	PTF_ASSERT_TRUE(memcmp(newHandshakeInitMessage.getEncryptedTimestamp().data(), expectedTimestampInit, 28) == 0);
+	PTF_ASSERT_TRUE(memcmp(newHandshakeInitMessage.getMac1().data(), expectedMac1Init, 16) == 0);
+	PTF_ASSERT_TRUE(memcmp(newHandshakeInitMessage.getMac2().data(), expectedMac2Init, 16) == 0);
```
seladb, line 236: Instead of `memcmp` we can use `PTF_ASSERT_BUF_COMPARE`.

Same goes for other instances of `memcmp` in this file

Tests/Packet++Test/Tests/WireGuardTests.cpp
```diff
@@ -0,0 +1,336 @@
...
+		                                  0x3e, 0x93, 0x7f, 0x56, 0x93, 0x49, 0xde, 0x28, 0x56, 0xdc,
+		                                  0x5f, 0x26, 0x16, 0x76, 0x3e, 0xee, 0xaf, 0xc0 };
+
+	uint8_t expectedMac1Init[16] = { 0x53, 0x3b, 0x01, 0xdd, 0x96, 0x5e, 0x7e, 0xc7,
+		                             0x69, 0x76, 0xe2, 0x8f, 0x68, 0x3d, 0x67, 0x12 };
+
+	uint8_t expectedMac2Init[16] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+		                             0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
+
+	pcpp::WireGuardHandshakeInitiationLayer newHandshakeInitMessage(818952152, expectedPublicKeyInit,
```
seladb, line 221: nit: can we use hexadecimal numbers instead of decimals? I find it easier to read.

Same goes for other decimals in this test

seladb (commented):
@nadongjun overall looks good!

Please find a few more comments, we're getting close to getting this PR ready to merge 🙂

nadongjun:
@seladb Thank you! I’ll address the remaining feedback quickly.

Tests/Packet++Test/PacketExamples/WireGuard.pcap
nadongjun: Sure, I've made the changes.

seladb (approved):

seladb:
Thank you so much @nadongjun for working on it, much appreciated! 🙏 🥇

tigercosmos:
@nadongjun  nice job :)

Issue #1531, opened 2024-08-12
Add the protocol into the Layer constructor argument 


tigercosmos:
Currently the `Layer` constructors are:
```cpp
		Layer()
		    : m_Data(NULL), m_DataLen(0), m_Packet(NULL), m_Protocol(UnknownProtocol), m_NextLayer(NULL),
		      m_PrevLayer(NULL), m_IsAllocatedInPacket(false)
		{}

		Layer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
		    : m_Data(data), m_DataLen(dataLen), m_Packet(packet), m_Protocol(UnknownProtocol), m_NextLayer(NULL),
		      m_PrevLayer(prevLayer), m_IsAllocatedInPacket(false)
		{}
```

It doesn't make sense that a layer inherited `Layer` cannot set the protocol in the initialization list.

Consider adding a new constructor:
```
Layer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet, ProtocolType protocolType)
		    : m_Data(data), m_DataLen(dataLen), m_Packet(packet), m_Protocol(protocolType), m_NextLayer(NULL),
		      m_PrevLayer(prevLayer), m_IsAllocatedInPacket(false)
		{}
```

Dimi1010:
Closed by #1559

PR #1559, opened 2024-08-31
Added ProtocolType to Layer constructor.

Dimi1010:
Implements #1531.

This PR essentially replaces most of the duplicated `m_Protocol = <ProtocolTypeHere>;` with a new parameter `ProtocolType protocol` in Layer's constructor (default value: `UnknownProtocol`).

NB: This PR does not modify cases where the layer is constructed via internal `init` method.

### DIFF
--- a/Packet++/header/Layer.h
+++ b/Packet++/header/Layer.h
@@ -211,8 +211,8 @@ namespace pcpp
 		      m_PrevLayer(nullptr), m_IsAllocatedInPacket(false)
 		{}
 
-		Layer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
-		    : m_Data(data), m_DataLen(dataLen), m_Packet(packet), m_Protocol(UnknownProtocol), m_NextLayer(nullptr),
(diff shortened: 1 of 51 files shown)


Packet++/header/Layer.h
```diff
@@ -211,8 +211,8 @@ namespace pcpp
 		      m_PrevLayer(nullptr), m_IsAllocatedInPacket(false)
 		{}
 
-		Layer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
-		    : m_Data(data), m_DataLen(dataLen), m_Packet(packet), m_Protocol(UnknownProtocol), m_NextLayer(nullptr),
+		Layer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet, ProtocolType protocol = UnknownProtocol)
```
seladb, line 214: Why do we need a default value of `UnknownProtocol`? 🤔

Dimi1010: To keep compatibility as some constructors still use the non-protocol version to initialize to `UnknownProtocol` and set the `m_protocol` manually. Usually in some kind of `init` method that gets executed in the body of the constructor.

seladb (approved):

PR #1574, opened 2024-09-12
Fix cmake warning about future support removal

rndx21033:
Simple fix of CMake warning about version < 3.5 future support removal.


rndx21033:
recreating for dev

PR #1576, opened 2024-09-12
Fix crash with parseuntil when top layer is more than the chosen parseuntil OSI layer.

rndx21033:
Fix crash with pcpp::Packet parseuntil of OSI layer that is less than the top layer in the packet.


rndx21033:
recreating for dev

Issue #1562, opened 2024-09-01
Apply Ruff for Python Code


tigercosmos:
[Ruff](https://github.com/astral-sh/ruff) is the fastest linter and formatter written in Rust for the Python code. Since we already introduce clang-format in the project for the C++ part, the next step is to start formatting and lintting on the Python side.

seladb:
@tigercosmos do you suggest we use it instead of Black as a formatter or just for linting?

tigercosmos:
@seladb  Ruff functions as both a formatter and a linter, and I chose it because it appears to be the fastest option. However, I'm open to discussing which tool we should use.

tigercosmos:
@Dimi1010  @egecetin @clementperon please also share your opinions

Dimi1010:
I don't have a strong opinion, so I am fine with either.

egecetin:
I'm also fine with all options since the python code base is just for testing and really short in terms of line

seladb:
I'm also okay with using it for both. It's a lot faster than Black / Pylint but we'll probably not notice it because our Python code base is very small

tigercosmos:
Okay, all of us are fine with either tool. I will try to use Ruff first. :)

seladb:
@tigercosmos can we close this issue?

PR #1584, opened 2024-09-19
add Ruff for Python file format and lint

tigercosmos:
As discussed in  https://github.com/seladb/PcapPlusPlus/issues/1562, use Ruff for Python file formatting and linting.


.pre-commit-config.yaml
```diff
@@ -21,10 +35,6 @@ repos:
       - id: mixed-line-ending
         args: ['--fix=lf']
       - id: trailing-whitespace
-  - repo: https://github.com/psf/black
```
tigercosmos, line 24: no need black now.

ci/check-ruff-version.py
```diff
@@ -0,0 +1,18 @@
+import subprocess
+
+EXPECTED_RUFF_VERSION = "0.6.5"
+
```
tigercosmos, line 4: check the version.

3rdParty/OUIDataset/create_oui_data.py
```diff
@@ -75,7 +75,7 @@ def convert_line(line: str) -> list[str]:
 
 
 def parse_mac_and_vendor(line_parts: list[str]) -> Optional[LineElements]:
-    if line_parts == None or len(line_parts) < 3:
+    if line_parts is None or len(line_parts) < 3:
```
tigercosmos, line 78: lint errors from Ruff.

tigercosmos:
@seladb  could you take a look?

.pre-commit-config.yaml
```diff
@@ -7,6 +7,20 @@ repos:
...
+      - id: ruff
+        name: ruff
+        entry: ruff check
+        types_or: [python]
+        language: system
+      - id: ruff-format
+        name: ruff-format
+        entry: ruff format --check
+        types_or: [python]
+        language: system
```
seladb, line 23: Shouldn't we use [`ruff-pre-commit`](https://github.com/astral-sh/ruff-pre-commit)?

It can make it easier to upgrade `ruff` if we need to

tigercosmos: Originally I think it's more flexible to do it on our side, but I think it also works to use other's scripts.

.github/workflows/build_and_test.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Install dependencies
         run: |
           apk update && apk add cppcheck python3-dev
-          python3 -m pip install cmake-format clang-format==18.1.6
+          python3 -m pip install cmake-format clang-format==18.1.6 ruff==0.6.5
```
seladb, line 37: Now that we use `ruff-pre-commit` I don't think this is needed

.pre-commit-config.yaml
```diff
@@ -7,6 +7,10 @@ repos:
         name: Check clang-format version
         entry: python3 ./ci/check-clang-format-version.py
         language: system
+      - id: check-ruff-format-version
+        name: Check ruff-format version
+        entry: python3 ./ci/check-ruff-version.py
+        language: system
```
seladb, line 13: Now that we use `ruff-pre-commit` this is not needed

seladb (approved):

PR #1587, opened 2024-09-21
Remove ifdef inside cpp and format

clementperon:
We should let CMake choose if this CPP should be compiled or not


tigercosmos:
I agree, it's better to manage this at CMake.

tigercosmos (approved):

Dimi1010 (approved):

seladb (approved):

PR #1588, opened 2024-09-22
Bump to Alpine 3.20, Fedora 40, FreeBSD 13.4 and 14.1

clementperon:


Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -1011,15 +1011,15 @@ namespace pcpp
...
-		if (sysctl(mib, 6, buf, &len, nullptr, 0) < 0)
+		if (sysctl(mib, 6, buf.data(), &len, nullptr, 0) < 0)
 		{
 			PCPP_LOG_DEBUG("Error in retrieving MAC address: sysctl 2 error");
 			return;
 		}
 
-		struct if_msghdr* ifm = (struct if_msghdr*)buf;
+		struct if_msghdr* ifm = (struct if_msghdr*)buf.data();
 		struct sockaddr_dl* sdl = (struct sockaddr_dl*)(ifm + 1);
```
Dimi1010, line 1023: Do we actually need the `struct` keywords here? Also possibly switch the casts to reinterpret casts?

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -1011,15 +1011,15 @@ namespace pcpp
 			return;
 		}
 
-		uint8_t buf[len];
+		std::vector<uint8_t> buf(len);
 
-		if (sysctl(mib, 6, buf, &len, nullptr, 0) < 0)
+		if (sysctl(mib, 6, buf.data(), &len, nullptr, 0) < 0)
```
Dimi1010, line 1016: Would probably be more correct for the length to be taken from `buf.size()`.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -1011,15 +1011,15 @@ namespace pcpp
 			return;
 		}
 
-		uint8_t buf[len];
+		std::vector<uint8_t> buf(len);
```
Dimi1010, line 1014: Tbh, since `len` is a non-constexpr variable I have... questions about how this compiled before...

Dimi1010 (approved):
LGTM. Made a couple of comments on the changes in the cpp, but they aren't critical.

Pcap++/src/PcapLiveDevice.cpp
clementperon: This is a VLA allocated on stack

Pcap++/src/PcapLiveDevice.cpp
clementperon: I don't think so, but this commit is only their to fix VLA build issue with new FreeBSD version, not to clean this part of code.

tigercosmos (approved):
LGTM, but maybe we should do the upgrade before the release though. 😂

clementperon:
> LGTM, but maybe we should do the upgrade before the release though. 😂

Yes we could have done it sooner, but priority was Ubuntu 2

seladb (approved):

Issue #1189, opened 2023-09-06
Question regarding the drop of nanosecond precision timestamp


Teles1:
The nanosecond precision is lost while re-assembling a TCP stream. I needed to know the timestamp for when a portion of a given stream arrived and unfortunately the precision was lost. Was this a design decision or if a pull request comes in to change this to timespec for example would it get merged in? Thanks
https://github.com/seladb/PcapPlusPlus/blob/666e2927569c685fd69aaf3560322287676cabaa/Packet%2B%2B/src/TcpReassembly.cpp#L29

seladb:
Hi @Teles1 ! I found the commit where it was changed: https://github.com/seladb/PcapPlusPlus/commit/2583f64ef62816a6a887c594e7aabb2b5482000a#diff-654f2a86f6954fcb6e711df2d4550f16028c3f79c08b2a7bc851dfa227dc37d2

As you can see, it was added when the packet timestamp became more accurate and changed from `timeval` to `timespec`. I assume we didn't want to change the `ConnectionData` and `TcpStreamData` structs and keep them backward compatible so we decided to just convert `timespec` back to `timeval`. 

However, I think it can be a nice change to deprecate `timeval` and move to a struct with higher precision. It can be either `timespec`  or a more C++11-style structure like `std::chrono::time_point`.

In order to stay backward compatible I'd keep the `timeval` properties in `ConnectionData` and `TcpStreamData`, mark them as deprecated and add new `timespec` / `std::chrono::time_point` properties. In the next version we can remove the `timeval` properties completely.

Would you consider implementing this change?

Teles1:
Sure. Will do.

seladb:
Fixed in https://github.com/seladb/PcapPlusPlus/pull/1591

PR #1591, opened 2024-09-26
Add nanosecond precision to TCP reassembly

seladb:
Should address this issue: https://github.com/seladb/PcapPlusPlus/issues/1189


Packet++/header/TcpReassembly.h
```diff
@@ -118,22 +123,16 @@ namespace pcpp
...
-		 * Set endTime of Connection
-		 * @param[in] endTimeValue integer value
+		 * Set the end time of the connection
+		 * @param[in] endTimeValue timestamp value
 		 */
-		void setEndTime(const timeval& endTimeValue)
-		{
-			endTime = endTimeValue;
-		}
+		void setEndTime(const std::chrono::time_point<std::chrono::system_clock>& endTimeValue);
```
Dimi1010, line 135: Why are the timepoints locked to `system_clock`? Generally I would prefer `steady_clock` for being monotonic, or `high_resolution_clock` for being the clock with the highest resolution on the system. Ideally, it should be templated, but that might be more trouble than its worth.

seladb: What we need here is a data structure that holds a timestamp in nanosecond precision. The timestamp itself comes from the packet: https://github.com/seladb/PcapPlusPlus/blob/634a78654a94805298fde5850670d5b7788aaeda/Packet%2B%2B/src/TcpReassembly.cpp#L142

Does it matter if it's `system_clock`, `steady_clock` or `high_resolution_clock`? 🤔

Dimi1010: Not sure tbh, but there are a some methods of timepoint that depend on the passed clock type. It probably won't really affect the current usage, as nowhere is `Clock::now()` called. so idk.

Packet++/header/TcpReassembly.h
```diff
@@ -106,10 +107,14 @@ namespace pcpp
...
 		/** A 4-byte hash key representing the connection */
 		uint32_t flowKey;
-		/** Start TimeStamp of the connection */
+		/** Start timestamp of the connection with microsecond precision */
 		timeval startTime;
-		/** End TimeStamp of the connection */
+		/** End timestamp of the connection with microsecond precision */
 		timeval endTime;
+		/** Start timestamp of the connection with nanosecond precision */
+		std::chrono::time_point<std::chrono::system_clock> startTimePrecise;
```
clementperon, line 115: It's not guarantee that std::chrono::time_point<std::chrono::system_clock> have nano seconds resolution.
It depends on Clock::duration. On macOS, system_clock::duration is microseconds.

IMO i would change all the timeval to timespec which is C11 posix and better defined.

Dimi1010: Tbh, using timespec might be simpler, as the timepoints in the chrono library are more specialized to be used with a clock, and we get the time externally anyway to we don't really have one.

Packet++/header/TcpReassembly.h
seladb: I tested it, apparently `system_clock` doesn't have nanosecond precision on all platforms, however, `high_resolution_clock` does so I think it's safe to use

Packet++/header/TcpReassembly.h
seladb: Please see: https://github.com/seladb/PcapPlusPlus/pull/1591#discussion_r1782353418

Packet++/header/TcpReassembly.h
clementperon: I mean we whould explicit the duration like this: 

using time_point_ns = std::chrono::time_point<std::chrono::high_resolution_clock, std::chrono::nanoseconds>;

Packet++/header/TcpReassembly.h
clementperon: I meant we whould explicit the duration like this:

```
using time_point_ns = std::chrono::time_point<std::chrono::high_resolution_clock, std::chrono::nanoseconds>;
time_point_ns startTimePrecise;
time_point_ns endTimePrecise;

```

This will also work for std::chrono::system_clock

seladb: Why do we need to be explicit? If we do:
```cpp
std::chrono::time_point<std::chrono::high_resolution_clock>
```

Then the user can decide which precision they want, for example, if they want nano sec they can do the following:
```
std::chrono::duration_cast<std::chrono::nanoseconds>(stats.begin()>second.connData.startTimePrecise.time_since_epoch())
```

But they can also get any other precision

Tests/Pcap++Test/Tests/TcpReassemblyTests.cpp
```diff
@@ -965,8 +999,16 @@ PTF_TEST_CASE(TestTcpReassemblyIPv6MultConns)
 	PTF_ASSERT_EQUAL(iter->second.connData.srcPort, 35999);
 	PTF_ASSERT_EQUAL(stats.begin()->second.connData.startTime.tv_sec, 1147551795);
 	PTF_ASSERT_EQUAL(stats.begin()->second.connData.startTime.tv_usec, 526632);
-	PTF_ASSERT_EQUAL(stats.begin()->second.connData.endTime.tv_sec, 0);
-	PTF_ASSERT_EQUAL(stats.begin()->second.connData.endTime.tv_usec, 0);
+	PTF_ASSERT_EQUAL(std::chrono::duration_cast<std::chrono::nanoseconds>(
+	                     stats.begin()->second.connData.startTimePrecise.time_since_epoch())
+	                     .count(),
```
tigercosmos, line 1004: Can we have a better format here?

seladb: so just ignore `clang-format`? 🤔

tigercosmos: I think you can ignore it here as the format seems not perfect.

seladb: Done in this commit https://github.com/seladb/PcapPlusPlus/pull/1591/commits/b0836404dad63980918c07ad5adeaa918f912977, please let me know if the format looks ok now

tigercosmos (approved):

Packet++/src/TcpReassembly.cpp
```diff
@@ -23,13 +23,44 @@
...
 
-	static timeval timespecToTimeval(const timespec& in)
+	static timeval timePointToTimeval(const std::chrono::time_point<std::chrono::high_resolution_clock>& in)
 	{
-		timeval out;
-		TIMESPEC_TO_TIMEVAL(&out, &in);
+		auto duration = in.time_since_epoch();
+
+		auto seconds = std::chrono::duration_cast<std::chrono::seconds>(duration);
+		auto nanoseconds = std::chrono::duration_cast<std::chrono::nanoseconds>(duration) % 1000000000;
```
Dimi1010, line 31: nit: Could possibly be done without the magic number by casting the seconds to nanosecond precision and substracting that from the main duration timepoint to get the remaining nanoseconds, but I guess this works too.

Packet++/src/TcpReassembly.cpp
```diff
@@ -23,13 +23,44 @@
...
+
+		struct timeval out;
+		out.tv_sec = seconds.count();
+		out.tv_usec = nanoseconds.count() / 1000;
```
Dimi1010, line 35: Why do we calculate the nanosecond precision here if we are just going to convert them to microseconds? As far as I see, the nanosecond precision duration object is not used anywhere else in this function.

Dimi1010 (approved):
A couple of comments, but LGTM otherwise.

Packet++/src/TcpReassembly.cpp
tigercosmos: Ture, I think Dimi's method will be more efficient.

Dimi1010: > Ture, I think Dimi's method will be more efficient.

Looked through it in [godbolt.org](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGe1wAyeAyYAHI%2BAEaYxCCSAJykAA6oCoRODB7evnrJqY4CQSHhLFEx8baY9vkMQgRMxASZPn5cFVXptfUEhWGR0bEJCnUNTdmtQ109xaUDAJS2qF7EyOwc5gDMwcjeWADUJutuyAjEAqgH2CYaAIJXt2abVFhUuwD6ALJCbq8AatgASnchsQvA5do42AA3MR3EwAdisN12yN2Q3QIBAXmCBHWZleBHBkNeSlM60R1xRqII6Mx2Nx%2BMJry8JIO5PhABFWbCHq48FRYTcIZhobRwXg2MpUNiACqoaXi4Via4QNAMIZUmnHU4MVAYoWvZLYg5uNEYrVnDEIPDABCvYiYBSeLzVV7bVDIADWF3MADZdsFZrCEXdKczgsBdsw2ApEkwVhqzSczlybgK4kxnahduglkxqvt1uz/QwAHT61IMFavTDJY4QQNktMZoiozCq9AKAtFnPEPPpV1MIbG00oJM6jEkgQdi4QHt9gQNtlIlHNrPMHWThgdrvZ3Mu0RDw4j83jkDrlJtqcKGdz6qzXYAWl3vf3g4Iw%2Bpie1urPhgv7evdZsAgTcO0XENKRRCDkSFEVdkWd9G1TDQ4gQssiRJHdQIUEs0C8QR6xTW4ULQggiWZNsd3PbDcMWAj7wAel2LgNFYojKXtAglgYeDnSIjl%2BMFBU4KFSUZTlYSxAsFUBHVY8xx/fVDUED9NQUy1rVte1HW8fdaHdL0gN9YtA1TYNl2RMMjEjVgHVjeN5O/QTiPTTNn3nHiDiLYIywVYktkwatawQQikJc1dWwAndb37A9EJNT9R2/CdLy3QDgJihdnKuVyW2o1Lty89zX0PBK1OS38NwK9LZz3dJ73MABWZjWNatr2Mg5FoLFKExF4xCl1Q510OJSiipovD6I65FSPIzCivygDaPwgh612JiWLYsKOMwLjiB4hD%2BLhTkwrubFdhYJhgkIsylwpFc3NUHdRKlQRZXlXraGVRyLRAK0bTtB0nT0gyMR1AB3etwIs3YItULhnoVMS3okz7pJ%2B09/q0oHdNi/TPTB1BIdmaHiPu5FOO43YNCOzkbg4eZaE4RreD8bheF1Dg3GsaxUUWZZMH2B4eFIAhNAZ%2BYEEwJgsBietSA9EBGo0EtGrMMwuDiXEuCkAAOXWzB9fROEkVnxdITneAUEANFF8X5jgWAYEQFBUBYRI6GichKDQd3PZibZDGATWWL4OgCGia2IAic2ImCeoAE9OBFuPmGIBOAHkIm0NsxfZ0hfbYQQM4YWgk44LRSCwCIvGANwxFoa386wS6jHECveHwe0HDwSEHXNzBVDbZ1VhF7FKnN2g8AiXt048LBzYIYhxWT3g%2B%2BICIL3ZTBW%2BAKejHtvgDGABQfjwTBwYzxJGFXmRBBEMR2CkO/5CUNRzd0VoDAP0xLGsfRp7W1gFGdgVAWAEFIOvEAS98IehJhbRI1Qm4PgAOoN0fGiLyv8rCWC4HCR8GczAc3XsvLAQD5Z2FzukFwW5RgtFIIEYIvQSj9FaLkNIAg6E5BSBwhgUw%2BgxHGJUKhAhOgjCdGMNoIiajDG6Ew6YrDbCyK4eMWR/CWGCPmI6JYKwJCM2ZmbDuFtOC7FULrH0D4fSSF2IHaymsSxcBLBoXYEBcCEBIELdYXBZi8DzloeBUsZb9HlorZWxsOCm1ICwEAPoVY%2BjiBoSQwsfRmEarrdYatSBs0rpbWwNs7Ydwds7CASAEKJGdN7CAmB8BEFIXofg99RDiGfg01%2BKh1BGM/qQcGvZEir30RwFmWTzacwzs6cpBJUAvDMRYqxNjv4RnsY45xEAPB%2B3oMQTx3jfH20ltLWWlABmROiesOIJZ1goUanCH06S4SSC4FY3WwyjG5OtrbPxEsFYDBLJIOEcJdYaB9P83WjUdbXPCesQxOTOA7MKQMohUSJCsWedCjgsL/HzHXqkZwkggA%3D%3D%3D). It actually does not matter under optimizations. Both compile down to mostly the same assembly.

tigercosmos: Amazing compiler

seladb: I agree! Fixed in https://github.com/seladb/PcapPlusPlus/pull/1591/commits/91f6b390a2c23412c301f6183ba634491106feec

Packet++/src/TcpReassembly.cpp
seladb: Good catch! Fixed in https://github.com/seladb/PcapPlusPlus/pull/1591/commits/91f6b390a2c23412c301f6183ba634491106feec

PR #1592, opened 2024-09-28
WinPcapLiveDevice.h : explicit override

Doekin:
fix `error: 'getDeviceType' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]` when compiling with `mingw-llvm 19.1.0`

### DIFF
--- a/Pcap++/header/WinPcapLiveDevice.h
+++ b/Pcap++/header/WinPcapLiveDevice.h
@@ -32,7 +32,7 @@ namespace pcpp
 		WinPcapLiveDevice& operator=(const WinPcapLiveDevice& other);
 
 	public:
-		virtual LiveDeviceType getDeviceType() const override
+		LiveDeviceType getDeviceType() const override
 		{
 			return WinPcapDevice;
 		}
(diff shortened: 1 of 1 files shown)


Pcap++/header/WinPcapLiveDevice.h
```diff
@@ -32,17 +32,17 @@ namespace pcpp
 		WinPcapLiveDevice& operator=(const WinPcapLiveDevice& other);
 
 	public:
-		virtual LiveDeviceType getDeviceType() const
+		virtual LiveDeviceType getDeviceType() const override
```
Dimi1010, line 35: The `virtual` can be removed if an `override` keyword is present.

Dimi1010 (approved):

Pcap++/header/WinPcapLiveDevice.h
Doekin: Yes, `virtual` could be omitted here.

tigercosmos (approved):
Nice catch!

seladb (approved):

PR #1602, opened 2024-10-03
Improve pre-commit workflow in CI

egecetin:
- Run clang-tidy for only changed files
- Use gersemi for cmake formatting
- Use venv for pre-commit workflow


.github/workflows/build_and_test.yml
```diff
@@ -39,18 +39,15 @@ jobs:
       # TODO: investigate how to run pre-commit with `venv`
       - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
 
-      - name: CMake format
-        run: |
-          ./ci/cmake-format-all.sh
-          git diff --exit-code
+      - name: Fetch dev branch
+        run: git fetch origin dev
```
seladb, line 43: Maybe we want to run it as part of `clang-tidy-all.sh`?

ci/clang-tidy-all.sh
```diff
@@ -0,0 +1,34 @@
...
+# Check if there are any files to process
+if [ -z "$files" ]; then
+    echo "No files to process."
+    exit 0
+fi
+
+# Process each file
+echo "$files" | while IFS= read -r file; do
+    echo "Checking: $file"
+    clang-tidy "$file" -p $BUILD_DIR --fix --checks=modernize-use-nullptr,modernize-use-override,performance-unnecessary-value-param
```
seladb, line 33: Do we want to install `clang-tidy` as a python package like we do for `cmake-format` and `clang-format`?
That way we may be able to control the `clang-tidy` version we're using.

Also - do we still need to install `cmake-format` as part of our CI?

tigercosmos:
Currently we only use clang-tidy to check "--checks=modernize-use-nullptr,modernize-use-override,performance-unnecessary-value-param". Maybe we should apply more checks? I wonder why we only check a few items?

egecetin:
It is possible to update Alpine to 3.20 now but it requires some fixes so it can be done in a separate PR

ci/clang-tidy-all.sh
egecetin: fixed

.github/workflows/build_and_test.yml
egecetin: fixed

egecetin:
> Currently we only use clang-tidy to check "--checks=modernize-use-nullptr,modernize-use-override,performance-unnecessary-value-param". Maybe we should apply more checks? I wonder why we only check a few items?

It is better to open a different PR for it. It might raise lots of warnings/errors

.codespellrc
```diff
@@ -1,4 +1,4 @@
 [codespell]
-skip = *.dat,typos-config.toml,.git,./ci,./Dist,./mk,./Tests/ExamplesTest/expected_output,./Tests/ExamplesTest/pcap_examples,./Tests/Packet++Test/PacketExamples,./Tests/Pcap++Test/PcapExamples,./3rdParty,./Examples/PcapSearch/dirent-for-Visual-Studio
+skip = *.dat,typos-config.toml,.git,venv,./ci,./Dist,./mk,./Tests/ExamplesTest/expected_output,./Tests/ExamplesTest/pcap_examples,./Tests/Packet++Test/PacketExamples,./Tests/Pcap++Test/PcapExamples,./3rdParty,./Examples/PcapSearch/dirent-for-Visual-Studio
```
clementperon, line 2: .venv no ?

clementperon:
You also need to remove the .cmake-format file

.pre-commit-config.yaml
```diff
@@ -37,6 +37,11 @@ repos:
         files: ^(Common\+\+|Packet\+\+|Pcap\+\+|Tests|Examples)/.*\.(cpp|h)$
       - id: cppcheck
         args: ["--std=c++11", "--language=c++", "--suppressions-list=cppcheckSuppressions.txt", "--inline-suppr", "--force"]
+  - repo: https://github.com/BlankSpruce/gersemi
+    rev: 0.15.1
+    hooks:
+    - id: gersemi
+      args: ['--line-length', '80']
```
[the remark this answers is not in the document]
clementperon: Also maybe prefer to introduce a .gersemirc file

.codespellrc
egecetin: Actually all CI workflows uses venv instead of .venv. I'll update it

.pre-commit-config.yaml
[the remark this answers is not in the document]
egecetin: I forgot .cmake-format file and checked the default so set to 80. I'll update it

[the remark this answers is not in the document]
tigercosmos: This tool is cool.

.github/workflows/build_and_test.yml
```diff
@@ -34,23 +34,22 @@ jobs:
       - name: Install dependencies
         run: |
           apk update && apk add cppcheck python3-dev
-          python3 -m pip install cmake-format clang-format==18.1.6
+          python3 -m venv .venv
+          . .venv/bin/activate
+          python3 -m pip install setuptools clang-format==18.1.6 clang-tidy==18.1.8 pre-commit==3.8.0
```
tigercosmos, line 39: Maybe use the latest pre-commit 4.0.1?

tigercosmos (approved):

.github/workflows/build_and_test.yml
egecetin: I removed version. I think it is better because pre-commit itself can be updated since this won't affect the formatting.

tigercosmos: I think it's always better to stick with the specific version of the CI. We don't know whether the future version can still be compatible with the current files.

tigercosmos: If pre-commit doesn't have much breaking changes, then it is fine to stick with the latest version.

CMakeLists.txt
seladb, line 1: Do we want to remove all the code related to `PCAPPP_ENABLE_CLANG_TIDY` in this file?

.github/workflows/build_and_test.yml
```diff
@@ -34,23 +34,22 @@ jobs:
...
       - name: Configure PcapPlusPlus for Static analysis
-        run: CXX=clang++ CC=clang cmake -DLIGHT_PCAPNG_ZSTD=ON -DPCAPPP_ENABLE_CLANG_TIDY=ON -S . -B "$BUILD_DIR"
+        run: cmake -S . -B "$BUILD_DIR"
 
-      - name: Build PcapPlusPlus and check any diff
+      - name: Run clang-tidy on changed files
         run: |
-          cmake --build "$BUILD_DIR" -j
-          git diff --exit-code
+          git fetch origin dev
```
seladb, line 51: don't we want to run `git fetch origin dev` inside `./ci/clang-tidy-all.sh`?

.gersemirc
```diff
@@ -0,0 +1,12 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/BlankSpruce/gersemi/master/gersemi/configuration.schema.json
+
+cache: true
+color: false
+definitions: []
```
seladb, line 5: do we need to add it if it's an empty list?

.gersemirc
```diff
@@ -0,0 +1,12 @@
...
+indent: 2
+line_length: 120
```
seladb, line 7: why not 80 like we usually have?

.gersemirc
```diff
@@ -0,0 +1,12 @@
...
+list_expansion: favour-inlining
+quiet: false
+unsafe: false
+warn_about_unknown_commands: false
+workers: 16
```
seladb, line 12: Maybe we should avoid it? The default is: `[default: number of CPUs on this system]`

CMakeLists.txt
egecetin: I'm not sure about it since it is also a good feature, check while compiling. The only downside is it is slower for CI. So, someone can still use it in their local builds maybe.

.gersemirc
egecetin: `.cmake-format` define it as 120 currently so I get the same length from there

CMakeLists.txt
seladb: I guess that users of this library don't need to use this flag, and we (or other contributors) have the CI to run `clang-tidy`. I'm not sure this duplication is needed 🤔

CMakeLists.txt
```diff
@@ -123,6 +123,8 @@ set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_EXTENSIONS ON)
 # Set Position Independent Code for static libraries
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+# Export compile commands for external tools
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
seladb, line 127: Why is this needed? 🤔

seladb (approved):

CMakeLists.txt
egecetin: It is because of clang tidy. It requires compilation database (or commands) to be exported since clang-tidy is now externally running.

PR #1604, opened 2024-10-04
Remove unused includes

seladb:


Dimi1010:
Just FYI, CI is failing because that all the headers like <cstring> only guarante to declare their contents within the `std` namespace, while the old c headers like <string.h> only guarantee to have their contents in the global namespace. So some functions might need adjustments to use the `std::` version.

seladb:
> Just FYI, CI is failing because that all the headers like `<cstring>` only guarantee to declare their contents within the `std` namespace, while the old c headers like `<string.h>` only guarantee to have their contents in the global namespace. So some functions might need adjustments to use the `std::` version.

Yes, I'm still working on this PR 🙂

egecetin:
@seladb What about also adding https://github.com/include-what-you-use/include-what-you-use after fix all of them. I don't have experience with it, but do you already know or use it?

tigercosmos (approved):
LGTM

seladb:
> @seladb What about also adding https://github.com/include-what-you-use/include-what-you-use after fix all of them. I don't have experience with it, but do you already know or use it?

I've never used it, but I can look into it. Probably in a separate PR?

egecetin:
@seladb Separate PR is better, I think. Because it might need additional config which will mix this PR.

PR #1609, opened 2024-10-05
Bump macos version

egecetin:
Macos 12 is deprecated. Replace it with macos-15.

https://github.com/actions/runner-images/issues/10721


clementperon:
@egecetin also bump the packaging please

egecetin:
@clementperon Packaging already runs on MacOS 14. There is no need to change at least for now. But looks like Android runs on MacOS which i forgot to update.

tigercosmos (approved):

egecetin:
I need help for gradle. I don't have experience with android builds unfortunately

tigercosmos:
The error is caused by this:
```
  General error during conversion: Unsupported class file major version 65
  
  java.lang.IllegalArgumentException: Unsupported class file major version 65
```
cc @seladb

clementperon:
> The error is caused by this:
> 
> ```
>   General error during conversion: Unsupported class file major version 65
>   
>   java.lang.IllegalArgumentException: Unsupported class file major version 65
> ```
> 
> cc @seladb

This is because new MacOS comes with Java 21 -> need to upgrade graddle

clementperon:
> @clementperon Packaging already runs on MacOS 14. There is no need to change at least for now. But looks like Android runs on MacOS which i forgot to update.

I would be in favor to keep both sync no? If we test under macos15 then why not releasing a package for MacOS15 ?

egecetin:
@clementperon Sure, sync is better but is there any idea why older versions removed earlier or not added @seladb (maybe for some reason) ?

tigercosmos:
> maybe for some reason ?

I think we just missed them. Let's fix it.

tigercosmos (approved):

.github/workflows/build_and_test.yml
```diff
@@ -628,11 +632,11 @@ jobs:
           - run-on-os: ubuntu-20.04
             target: x86
             api-version: 30
-          - run-on-os: macos-12
+          - run-on-os: macos-15
             target: arm64-v8a
             cmake_configure: "-DCMAKE_OSX_ARCHITECTURES=arm64"
             api-version: 30
-          - run-on-os: macos-12
+          - run-on-os: macos-15
```
seladb, line 639: To unblock this PR until we figure out the Android build, can we change it to a working version, maybe `macos-13` or `macos-14`?

seladb (approved):

.github/workflows/build_and_test.yml
egecetin: @seladb No luck even macos-13 fails but I just noticed, android packaging only runs with ubuntu-latest for different target architectures. What about just testing with ubuntu and targeting archs also?

seladb: So maybe move it back to macos 11 until we figure it out?

clementperon: @seladb I have open a MR to bump gradle https://github.com/seladb/ToyVpn-PcapPlusPlus/pull/10/files

This is the latest gradle version that support Android SDK 30.
But I think we should consider bumping ToyVpn SDK target

seladb: @clementperon unfortunately ToyVpn doesn't have CI so the only way to know if your PR works is use the PR branch in PcapPlusPlus CI. It should be pretty easy to do using this change in [`build_and_test.yml`](https://github.com/seladb/PcapPlusPlus/blob/9f31f713951d34f1c600b44c113d985b519c6531/.github/workflows/build_and_test.yml#L660C1-L665C27):
```yaml
- name: Checkout ToyVpn-PcapPlusPlus
   uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938 # master
   with:
     repository: clementperon/ToyVpn-PcapPlusPlus
     ref: bump_gradle
     path: ./ToyVpn-PcapPlusPlus
     submodules: true
```

BTW, I'm considering re-writing the ToyVpn example using modern Kotlin and build tools, but this will take time because I'm not familiar with Android development...

PR #1614, opened 2024-10-17
Do not check whether u8 value is greater than zero in LdapLayer.cpp

ol-imorozko:
This comparison is always true due to limited range of data type


tigercosmos (approved):

seladb (approved):

seladb:
Thank you @ol-imorozko for this fix! 🙏

PR #1621, opened 2024-10-22
PcapSplitter: Fix typos (avoid mixing split-param and split_param)

fxlb:
Using split_param with a '_' like in split_method.


seladb (approved):

seladb:
Thank you @fxlb for this fix, much appreciated! 🙏

PR #1637, opened 2024-11-05
Fix getRawPacketReadOnly() const-correctness #1573

### LINKED ISSUE #1573: getRawPacketReadOnly() const-correctness
Hi,

It seems that both the [read-only](https://github.com/seladb/PcapPlusPlus/blob/4af4b8b04e6d5dd40b8e69cc30c5c1f62e7fe368/Packet%2B%2B/header/Packet.h#L181) and [non-read-only](https://github.com/seladb/PcapPlusPlus/blob/4af4b8b04e6d5dd40b8e69cc30c5c1f62e7fe368/Packet%2B%2B/header/Packet.h#L155) getters of the raw packet from the parsed packet have the exact same signature and behaviour:

```
RawPacket* getRawPacket() const;
RawPacket* getRawPacketReadOnly() const;
```

Wouldn't it be more correct to declare the read-only getter as returning a pointer-to-const, and similarly, the non-read-only method not be marked as const method as it can alter the underlying object ?

```
RawPacket* getRawPacket();
const RawPacket* getRawPacketReadOnly() const;
```

Best regards,

egecetin:
Fix #1573

Requires #1638


seladb:
@egecetin can you fix `clang-format`?

egecetin:
Hmm I wasn't expecting this while changing clang-tidy script. It can't find header because this container doesn't have kni (or same for dpdk etc). But since the file is changed it tries to analyze it.

egecetin:
@seladb I'll split PR for clang-tidy script changes if it passes. Just pushed here for testing

seladb (approved):

seladb:
> @seladb I'll split PR for clang-tidy script changes if it passes. Just pushed here for testing

Why ignore those files altogether? Can't we just ignore those specific issues?

egecetin:
@seladb There is no way to ignore clang-diagnostic-error unfortunately because it is a compile error. Old clang-tidy check was based on full compile, so these headers/source files are not checked since they are not enabled. But now it just trying to check if anything is changed

seladb:
> @seladb There is no way to ignore clang-diagnostic-error unfortunately because it is a compile error. Old clang-tidy check was based on full compile, so these headers/source files are not checked since they are not enabled. But now it just trying to check if anything is changed

yes, I just saw this stackoverflow response 😕 : https://stackoverflow.com/a/67487680

PR #1641, opened 2024-11-15
Fix: sockaddr2string() is only the debug function. Throwing an except…

KangLin:
Fix: sockaddr2string() is only the debug function. Throwing an exception will cause the guide program to break. So change it to print debugging information.

### DIFF
--- a/Common++/src/IpUtils.cpp
+++ b/Common++/src/IpUtils.cpp
@@ -100,7 +100,7 @@ namespace pcpp
 				break;
 			}
 			default:
-				throw std::invalid_argument("Unsupported sockaddr family. Family is not AF_INET or AF_INET6.");
+				PCPP_LOG_ERROR("Unsupported sockaddr family. Family is not AF_INET or AF_INET6.");
 			}
 		}
 


tigercosmos:
We throw `std::exception` on purpose. You should add try-catch in your own program.

seladb:
@KangLin this was changed quite recently, in this PR: https://github.com/seladb/PcapPlusPlus/pull/1426

The idea is to throw exceptions for invalid input in order to make the interface a bit more straightforward

KangLin:
I understand what you think. But there is no need to throw an exception here.
When I have a non-IP address network interface, the program is breaked, which is not as expected.

seladb:
@KangLin this is a change we made recently, and I understand there can be different opinions, but we currently don't plan to change it back. I'm sorry it doesn't work for you...

Dimi1010:
@KangLin 
If I understand this correctly, your main issue is the [debug message](https://github.com/seladb/PcapPlusPlus/blob/4af4b8b04e6d5dd40b8e69cc30c5c1f62e7fe368/Pcap%2B%2B/src/PcapLiveDeviceList.cpp#L321-L326) in `PcapLiveDeviceList` itself, instead of the actual exception?

KangLin:
Yes.

















At 2024-11-20 16:53:16, "Dimitar Krastev" ***@***.***> wrote:

@KangLin
If I understand this correctly, your main issue is the debug message in PcapLiveDeviceList itself that calls the method that produces the exception, instead of the actual exception?

KangLin:
> this is a change we made recently, and I understand there can be different opinions, but we currently don't plan to change it back. I'm sorry it doesn't work for you...

It's not a difference of opinion. It's a bug.

Dimi1010:
> Yes.

That debug message has already been removed in the the master branch by #1497.

So I don't see a reason to change the exception being thrown from socket2string in the dev branch, as the main issue (the debug message) no longer exists.

KangLin:
As you guys think，You can also fix this bug([debug message](https://github.com/seladb/PcapPlusPlus/blob/4af4b8b04e6d5dd40b8e69cc30c5c1f62e7fe368/Pcap%2B%2B/src/PcapLiveDeviceList.cpp#L288-L293)) as follows. But it needs to be changed in many places:

    try{
        if (Logger::getInstance().isDebugEnabled(PcapLogModuleLiveDevice) && addressInfo.addr != nullptr)
	{
		std::array<char, INET6_ADDRSTRLEN> addrAsString;
               internal::sockaddr2string(addressInfo.addr, addrAsString.data(), addrAsString.size());
		PCPP_LOG_DEBUG("Searching address " << addrAsString.data());
	}
    } cathc(...) {
            PCPP_LOG_DEBUG("exceptio");
    }

KangLin:
> > Yes.
> 
> That debug message has already been refactored in the the master branch by #1497.
> 
> So I don't see a reason to change the exception being thrown from socket2string in the dev branch, as the main issue (the call in the debug message) no longer exists.

@Dimi1010 You can't assume that the user knows. Users can't use debugging in the main branch?

KangLin:
Don't use it for the sake of using exceptions. When using it, choose the appropriate method

tigercosmos:
I don't see the point. I will say this is a no-fix issue.

KangLin:
@tigercosmos 我有非IP的设备，当我使用调试时，([debug message](https://github.com/seladb/PcapPlusPlus/blob/4af4b8b04e6d5dd40b8e69cc30c5c1f62e7fe368/Pcap%2B%2B/src/PcapLiveDeviceList.cpp#L288-L293)) 触发异常，导致程序中断，这与预期不一样（程序应该不受调试信息的影响而继续）。
因为你这里仅是用于调试的，所以不应该因为使用了异常处理而影响主程序的正常执行。
你也不能假设用户在使用时不使用调试。

seladb:
> @KangLin If I understand this correctly, your main issue is the [debug message ](https://github.com/seladb/PcapPlusPlus/blob/4af4b8b04e6d5dd40b8e69cc30c5c1f62e7fe368/Pcap%2B%2B/src/PcapLiveDeviceList.cpp#L288-L293) in `PcapLiveDeviceList` itself that calls the method that produces the exception, instead of the actual exception?

@KangLin , as @Dimi1010 mentioned - the code was refactored recently. Do you still see the same issue when pulling the code from the `master` branch?

KangLin:
> > @KangLin If I understand this correctly, your main issue is the [debug message ](https://github.com/seladb/PcapPlusPlus/blob/4af4b8b04e6d5dd40b8e69cc30c5c1f62e7fe368/Pcap%2B%2B/src/PcapLiveDeviceList.cpp#L288-L293) in `PcapLiveDeviceList` itself that calls the method that produces the exception, instead of the actual exception?
> 
> @KangLin , as @Dimi1010 mentioned - the code was refactored recently. Do you still see the same issue when pulling the code from the `master` branch?

Yes. It is a bug!

KangLin:
I use v24.09.

Dimi1010:
Can you build the library from the master branch and see if that fixes your issue?

Issue #1640, opened 2024-11-13
RawSocketDevice  cannot bind NIC on the centos7.9


vectorzjl:
When I use RawSocketDevice to capture data, binding to the network card does not take effect, and it captures data from all available network cards. Perhaps I should use bind in the implementation to bind to the network card. Adding the following code should allow normal binding to the network card for data capture.
```
struct sockaddr_ll device;
memset(&device, 0, sizeof(device));
device.sll_family = AF_PACKET;
device.sll_protocol = htons(ETH_P_IP);
device.sll_ifindex = if_nametoindex(ifaceName.c_str());  // 替换为你的网络接口

// 绑定到网络接口
if (bind(fd, (struct sockaddr*)&device, sizeof(device)) < 0)
{
	PCPP_LOG_ERROR("Cannot bind raw socket to interface '" << ifaceName << "'");
	::close(fd);
	return false;
}
```

tigercosmos:
Could you provide the minimum compilable sample code? Also please provide the NIC information if possible (`ifconfig` or `ip address`).

vectorzjl:
> Could you provide the minimum compilable sample code? Also please provide the NIC information if possible (`ifconfig` or `ip address`).

ok , here is my sample code:
```

int main()
{
	const std::string ip_str = "192.168.199.1";
	pcpp::IPAddress ip_addr = pcpp::IPAddress(ip_str);
	auto rawSocket = std::make_shared<pcpp::RawSocketDevice>(ip_addr);

	if (!rawSocket->open())
	{
		return -1;
	}
	while (true)
	{
		pcpp::RawPacket packet;
		int ret = rawSocket->receivePacket(packet);
		if (pcpp::RawSocketDevice::RecvSuccess == ret)
		{
			pcpp::Packet parsePacket(&packet);

			auto* macLayer = parsePacket.getLayerOfType<pcpp::EthLayer>();
			if (macLayer != nullptr)
			{
				// print source and dest MAC addresses
				std::cout << "Source MAC address: " << macLayer->getSourceMac() << std::endl
				          << "Destination MAC address: " << macLayer->getDestMac() << std::endl;
			}
			auto* ipLayer = parsePacket.getLayerOfType<pcpp::IPv4Layer>();
			if (ipLayer != nullptr)
			{
				std::cout << "Source IP address: " << ipLayer->getSrcIPAddress() << std::endl
				          << "Destination IP address: " << ipLayer->getDstIPAddress() << std::endl
[...]

PR #1642, opened 2024-11-15
Use `bind()` instead of `setsockopt()` for binding raw sockets on Linux

seladb:
Fixes https://github.com/seladb/PcapPlusPlus/issues/1640 (implementing the solution proposed in the issue)


seladb:
Replaced by https://github.com/seladb/PcapPlusPlus/pull/1646

PR #1646, opened 2024-11-27
Use `bind()` instead of `setsockopt()` for binding raw sockets on Linux

seladb:
Fixes https://github.com/seladb/PcapPlusPlus/issues/1640 (implementing the solution proposed in the issue)


Pcap++/src/RawSocketDevice.cpp
```diff
@@ -505,12 +505,15 @@ namespace pcpp
...
-		memset(&ifr, 0, sizeof(ifr));
-		snprintf(ifr.ifr_name, sizeof(ifr.ifr_name), "%s", ifaceName.c_str());
-		if (setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, (void*)&ifr, sizeof(ifr)) == -1)
+		struct sockaddr_ll saddr;
+		memset(&saddr, 0, sizeof(saddr));
+		saddr.sll_family = AF_PACKET;
+		saddr.sll_protocol = htons(ETH_P_ALL);
+		saddr.sll_ifindex = if_nametoindex(ifaceName.c_str());
+
+		if (bind(fd, reinterpret_cast<struct sockaddr*>(&saddr), sizeof(saddr)) < 0)
```
Dimi1010, line 514: nit: Do we need the `struct` keyword here and on L508?

Pcap++/src/RawSocketDevice.cpp
```diff
@@ -456,7 +456,7 @@ namespace pcpp
 		int fd = socket(AF_PACKET, SOCK_RAW, htobe16(ETH_P_ALL));
 		if (fd < 0)
 		{
-			PCPP_LOG_ERROR("Failed to create raw socket. Error code was " << errno);
+			PCPP_LOG_ERROR("Failed to create raw socket. Error code was " << strerror(errno));
```
egecetin, line 459: Not checked the full code so maybe not important but `strerror` is not thread safe

Pcap++/src/RawSocketDevice.cpp
```diff
@@ -505,12 +505,15 @@ namespace pcpp
...
+		struct sockaddr_ll saddr;
+		memset(&saddr, 0, sizeof(saddr));
+		saddr.sll_family = AF_PACKET;
+		saddr.sll_protocol = htons(ETH_P_ALL);
+		saddr.sll_ifindex = if_nametoindex(ifaceName.c_str());
+
+		if (bind(fd, reinterpret_cast<struct sockaddr*>(&saddr), sizeof(saddr)) < 0)
 		{
-			PCPP_LOG_ERROR("Cannot bind raw socket to interface '" << ifaceName << "'");
+			PCPP_LOG_ERROR("Cannot bind raw socket to interface '" << ifaceName << "': " << strerror(errno));
```
egecetin, line 516: same `strerror`

Pcap++/src/RawSocketDevice.cpp
seladb: Fixed in https://github.com/seladb/PcapPlusPlus/pull/1646/commits/c43e11019c27c046629d01382c59b4a70a4c778e

Pcap++/src/RawSocketDevice.cpp
egecetin: Actually errno is thread since it is thread local but there are other strerror functions in library so better to fix it all later

egecetin (approved):

Pcap++/src/RawSocketDevice.cpp
seladb: yes, I think you're right. I'll open a separate PR for it

PR #1647, opened 2024-11-27
Reformat Common++ documentation to use triple slash.

Dimi1010:
A draft reformat of the documentation comments in Common++ from `/** */` format to `///` format.

The change has the following effects:
- Reduced line bloat. Each documentation block is 2 lines shorter on average.
This can compound to over 100 lines on some of the larger files. (e.g. "IPAddress.h")
- Makes usages of comment blocks `/* */` easier to manage during refactoring, as they won't be terminated early if there are documentation comments inside them.


Dimi1010:
> @Dimi1010 did you use any automated tool to do this reformatting?

2 regexes and then manually cleaning up a blank lines.

Common++/header/LRUList.h
```diff
@@ -115,26 +102,22 @@ namespace pcpp
...
+		std::size_t getSize() const
 		{
 			return m_CacheItemsMap.size();
 		}
 
 	private:
 		std::list<T> m_CacheItemsList;
 		std::unordered_map<T, ListIterator> m_CacheItemsMap;
-		size_t m_MaxSize;
+		std::size_t m_MaxSize;
```
seladb, line 120: Why replacing `size_t` with `std::size_t`? 🤔

Common++/header/LRUList.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <list>
+#include <cstddef>
```
seladb, line 4: Why adding this include?

Common++/header/TablePrinter.h
```diff
@@ -1,61 +1,47 @@
 #pragma once
 
+#include <string>
```
seladb, line 3: Why adding this include?

Common++/header/LRUList.h
Dimi1010: If you look at the CI fails from the first 2 commits in this PR, they fail because of unrecoginized `size_t` and `string`.

Common++/header/LRUList.h
Dimi1010: See comment about CI fails. This is the header in which std::size_t is guaranteed to be defined by the standard.

Common++/header/TablePrinter.h
Dimi1010: See comment about CI fails.

Common++/header/LRUList.h
seladb: These are all `clang-tidy` errors 😕 
It's probably because these files haven't been changed in a long time. Maybe we should suppress this error? We don't use `std::size_t` anywhere else and it complies and runs ok...

egecetin: If I'm not mistaken including `cstddef` fixes it. Just clang-tidy can't find definition

Dimi1010: I prefer the error be fixed instead of suppressed, but I am fine either way.
I usually prefer using `std::size_t` in C++ due to the standard containers using it and usually defining it.

Both `size_t` and `std::size_t` alias to the same type. The slight difference is that `size_t` is from the C runtime and `std::size_t` is from the C++ runtime, and its a bit iffy which of the 2 are defined where.

For instance:
- when using CPP style headers `<cstring>`, `<cstddef>`, etc `std::size_t` is defined.
- when using C style headers `<string.h>`, `<stddef.h>` `size_t` is defined.

seladb: It's just that we're using `size_t` everywhere else, so it's more consistent

tigercosmos (approved):
Love this, much cleaner.

Common++/header/LRUList.h
Dimi1010: Sure

Dimi1010: Changed it back to `size_t` 
72dd3b3

seladb (approved):

Issue #1656, opened 2024-12-09
Add config support for libpcap hardware timestamp


vcomito-apexai:
libpcap supports the use of hardware timestamp, but it is not possible to control this option in PcapPlusPlus. 

Add a new option that lets the user enable this setting

vcomito-apexai:
Working on a PR for adding this support

tigercosmos:
@vcomito-apexai  That would be great!

vcomito-apexai:
> @vcomito-apexai That would be great!

thanks for the interest! PR is here: https://github.com/seladb/PcapPlusPlus/pull/1657

seladb:
Added in #1657

PR #1657, opened 2024-12-10
Add support for timestamp related options (#1656)

vcomito-apexai:
As described by https://github.com/seladb/PcapPlusPlus/issues/1656 

@seladb would it be possible to enable the CI?


Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -308,6 +340,34 @@ namespace pcpp
 		}
 #endif
 
+		if (config.timestampProvider)
+		{
+			ret = pcap_set_tstamp_type(pcap, timestampProviderMap(config.timestampProvider));
```
seladb, line 345: According to [libpcap docs](https://www.tcpdump.org/manpages/libpcap-1.10.5/pcap-tstamp.7.html) we should first call [`pcap_list_tstamp_types`](https://www.tcpdump.org/manpages/libpcap-1.10.5/pcap_list_tstamp_types.3pcap.html) to figure out which type are possible for this device. I'm not sure what happens if we set a type that isn't supported...

vcomito-apexai: @seladb the behavior is described here:
 https://www.tcpdump.org/manpages/libpcap-1.10.5/pcap_set_tstamp_type.3pcap.html
 
 if I read it correctly, this means that:
   1. the error will be printed 
   2. the default timestamp type will be used. 
   
   Let me know if you prefer to add the check and e.g. throw an exception if the timestamp is not supported

seladb: Thanks @vcomito-apexai ! I think it's ok to log an error

Pcap++/header/PcapLiveDevice.h
```diff
@@ -186,6 +186,38 @@ namespace pcpp
 			PCPP_OUT
 		};
 
+		/**
+		 * Set which source provides timestamps associated to each captured packet
+		 * (you can read more here: <https://www.tcpdump.org/manpages/pcap-tstamp.7.html>)
+		 */
+		enum TimestampProvider
```
Dimi1010, line 193: Can we use `enum class` here? we are trying to move away from plain enums where possible and its a pain to migrate after it has already been set.

Pcap++/header/PcapLiveDevice.h
```diff
@@ -186,6 +186,38 @@ namespace pcpp
...
+			TimestampProviderAdapterUnsynced,
+			/** host-provided, high precision, not synced with the system clock */
+			TimestampProviderHostHighPrecUnsynced
+		};
+
+		/**
+		 * Set the precision of timestamps associated to each captured packet
+		 * (you can read more here: <https://www.tcpdump.org/manpages/pcap-tstamp.7.html>)
+		 */
+		enum TimestampPrecision
```
Dimi1010, line 213: ditto: `enum class`

Pcap++/header/PcapLiveDevice.h
vcomito-apexai: done

Pcap++/header/PcapLiveDevice.h
vcomito-apexai: done

vcomito-apexai:
@seladb @Dimi1010 

It seems that clang-tidy finds issues unrelated to this PR https://github.com/seladb/PcapPlusPlus/pull/1657/checks

How may I address those? In a separate PR?

Pcap++/src/PcapLiveDevice.cpp
vcomito-apexai: added @seladb

vcomito-apexai:
> @seladb @Dimi1010
> 
> It seems that clang-tidy finds issues unrelated to this PR https://github.com/seladb/PcapPlusPlus/pull/1657/checks
> 
> How may I address those? In a separate PR?

ah, appearently they are fixed after rebase :ok_hand:

vcomito-apexai:
> > @seladb @Dimi1010
> > It seems that clang-tidy finds issues unrelated to this PR https://github.com/seladb/PcapPlusPlus/pull/1657/checks
> > How may I address those? In a separate PR?
> 
> ah, appearently they are fixed after rebase 👌

Hello @seladb @Dimi1010  

Any more changes I should apply to the PR :) ?

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -84,6 +89,120 @@ namespace pcpp
...
+		if (numSupportedTstampTypes < 0)
+		{
+			std::cerr << "Error retrieving timestamp types: " << pcap_geterr(pcap) << " - default Host will be used"
+			          << std::endl;
+			isSupported = false;
+		}
+		else if (numSupportedTstampTypes == 1)
+		{
+			// If 1 is returned, then the only available typestamp is TimestampProvider::Host;
+			return timestampProvider == PcapLiveDevice::TimestampProvider::Host;
```
Dimi1010, line 152: Do we not have to free `supportedTstampTypes` here?

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -84,6 +89,120 @@ namespace pcpp
...
+		}
+		return PCAP_TSTAMP_PRECISION_MICRO;
+	}
+
+	static bool isTimestampProviderSupportedByDevice(pcap_t* pcap,
+	                                                 const PcapLiveDevice::TimestampProvider timestampProvider)
+	{
+		int tstampType = timestampProviderMap(timestampProvider);
+		int* supportedTstampTypes = nullptr;
+		const int numSupportedTstampTypes = pcap_list_tstamp_types(pcap, &supportedTstampTypes);
```
Dimi1010, line 140: It might be useful to wrap `supportedTstampTypes` into a std::unique_ptr with a custom deleter.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -84,6 +89,120 @@ namespace pcpp
...
+		case PcapLiveDevice::TimestampPrecision::Nanoseconds:
+			return PCAP_TSTAMP_PRECISION_NANO;
+		}
+		return PCAP_TSTAMP_PRECISION_MICRO;
+	}
+
+	static bool isTimestampProviderSupportedByDevice(pcap_t* pcap,
+	                                                 const PcapLiveDevice::TimestampProvider timestampProvider)
+	{
+		int tstampType = timestampProviderMap(timestampProvider);
```
seladb, line 138: nit: you can use `auto` instead of `int`

Dimi1010: It doesn't actually save you much in this case tho, does it? 🤔

seladb: not really, it's just nicer code 🙂

seladb:
> > > @seladb @Dimi1010
> > > It seems that clang-tidy finds issues unrelated to this PR https://github.com/seladb/PcapPlusPlus/pull/1657/checks
> > > How may I address those? In a separate PR?
> > 
> > 
> > ah, appearently they are fixed after rebase 👌
> 
> Hello @seladb @Dimi1010
> 
> Any more changes I should apply to the PR :) ?

@vcomito-apexai can you please address the rest of the comments so we can merge this PR?

Pcap++/header/PcapLiveDevice.h
```diff
@@ -186,6 +186,38 @@ namespace pcpp
 			PCPP_OUT
 		};
 
+		/**
```
tigercosmos, line 189: Since we are now refactoring comments to use three slashes `///` in other places, I think it's better to use `///` here directly, right? @Dimi1010

seladb: Yes, I agree 👍

Common++/src/SystemUtils.cpp
```diff
@@ -200,7 +200,9 @@ namespace pcpp
 
 	bool directoryExists(const std::string& dirPath)
 	{
-		struct stat info{};
+		struct stat info
+		{
```
tigercosmos, line 204: can we revert the format?

Packet++/src/DnsResourceData.cpp
```diff
@@ -204,7 +204,8 @@ namespace pcpp
 	{
 		if (m_DataLen == 0 || m_Data == nullptr)
 		{
-			PCPP_LOG_ERROR("Input data is null or illegal" << "|m_DataLen:" << m_DataLen);
+			PCPP_LOG_ERROR("Input data is null or illegal"
+			               << "|m_DataLen:" << m_DataLen);
```
tigercosmos, line 208: any reason to change the format?

Pcap++/header/PcapLiveDevice.h
```diff
@@ -169,6 +169,34 @@ namespace pcpp
 			PCPP_OUT
 		};
 
+		/// Set which source provides timestamps associated to each captured packet
+		/// (you can read more here: <https://www.tcpdump.org/manpages/pcap-tstamp.7.html>)
+		enum class TimestampProvider
+		{
+			/** host-provided, unknown characteristics, default */
```
tigercosmos, line 176: let use `///` directly.

Pcap++/header/PcapLiveDevice.h
```diff
@@ -169,6 +169,34 @@ namespace pcpp
...
+			Host = 0,
+			/** host-provided, low precision, synced with the system clock */
+			HostLowPrec,
```
tigercosmos, line 179: `Prec` seems not to be a common abbreviation. Maybe it's better to spell out the full `Precision`.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -86,6 +91,124 @@ namespace pcpp
...
+			return PCAP_TSTAMP_ADAPTER;
+		case PcapLiveDevice::TimestampProvider::AdapterUnsynced:
+			return PCAP_TSTAMP_ADAPTER_UNSYNCED;
+		case PcapLiveDevice::TimestampProvider::HostHighPrecUnsynced:
+			return PCAP_TSTAMP_HOST_HIPREC_UNSYNCED;
+		}
+		return PCAP_TSTAMP_HOST;
+	}
+
+	static int timestampPrecisionMap(const PcapLiveDevice::TimestampPrecision timestampPrecision)
```
tigercosmos, line 125: Maybe rename this `timestampPrecisionMap`?  like `GetPcapPrecision`?

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -310,6 +433,16 @@ namespace pcpp
 		}
 #endif
 
+		if (config.timestampProvider != PcapLiveDevice::TimestampProvider::Host)
```
tigercosmos, line 436: Does it make sense to always bind `config.timestampProvider` and `config.timestampPrecision` together?

vcomito-apexai:
> @vcomito-apexai are you planning to address the remaining comments? Please note that CI fails because `pcap_list_tstamp_types` doesn't exist in all libpcap versions 🤔

yes, thanks for the reminder. Addressing this....

Pcap++/header/PcapLiveDevice.h
vcomito-apexai: @tigercosmos I tried to keep the same name of the libpcap enums, but I will apply this if preferred :ok_hand:

Pcap++/src/PcapLiveDevice.cpp
vcomito-apexai: renaming it, but notice that there is already a `directionTypeMap` function, so I named the new functions like that

Pcap++/src/PcapLiveDevice.cpp
vcomito-apexai: handled via the unique_ptr

Pcap++/src/PcapLiveDevice.cpp
vcomito-apexai: Not sure what you mean by binding them together :) ?

These two are separate params that can be used indipendently. E.g. also in tcpdump the user can specify just one of them: https://www.tcpdump.org/manpages/tcpdump.1.html

vcomito-apexai:
@seladb could you review and possibly run the workflow?

seladb:
> @seladb could you review and possibly run the workflow?

I just re-ran the workflow, let's make sure it passes and I'll review a bit later

vcomito-apexai:
updated the doxygen docs...

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -310,6 +440,16 @@ namespace pcpp
...
+		{
+			setTimestampProvider(pcap, config.timestampProvider);
+		}
+
+		if (config.timestampPrecision != PcapLiveDevice::TimestampPrecision::Nanoseconds)
```
seladb, line 448: Shouldn't this be `if (config.timestampPrecision != PcapLiveDevice::TimestampPrecision::Microseconds)`?

Because microsecond precision is the default...

Pcap++/header/PcapLiveDevice.h
seladb, line 1: Can you think of any way we can write tests for this feature? 🤔

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -86,6 +91,131 @@ namespace pcpp
...
+		std::unique_ptr<int[], void (*)(int*)> supportedTstampTypes(nullptr, [](int* ptr) {
+			if (ptr != nullptr)
+			{
+				pcap_free_tstamp_types(ptr);
+			}
+		});
+
+		auto rawSupportedTstampTypesPtr = supportedTstampTypes.release();
+		const int numSupportedTstampTypes = pcap_list_tstamp_types(pcap, &rawSupportedTstampTypesPtr);
+		supportedTstampTypes.reset(rawSupportedTstampTypesPtr);
```
Dimi1010, line 153: Two comments for this code fragment.

1. Can't we just use a raw pointer and then assign that to the `std::unique_ptr` directly skipping the `release` call?
2. IMO Functor style custom deleters are easier to read and do not require an explicit pointer to be passed to the constructor as the deleter is default constructed. They can also be optimized better as the deleter call can be inlined.

Here is an example deleter. If it won't be used outside the function, it can be declared as a local struct to the function scope.
```
struct PcapFreeTstampTypesDeleter
{
  void operator()(int* ptr) { 
    // no nullptr check as unique_ptr already checks internally and does not call the deleter if ptr == nullptr
    pcap_free_tstamp_types(ptr);
  }
};
```

seladb:
@vcomito-apexai will you have time to address the remaining comments so we can get this PR merged?

Pcap++/src/PcapLiveDevice.cpp
seladb: Fixed in this commit: https://github.com/seladb/PcapPlusPlus/pull/1657/commits/5251602c9b95ca0229c2a78fbda199c1816482d0

Pcap++/header/PcapLiveDevice.h
seladb: I think it'd be hard to test this feature without mocks. I'll try to introduce mocks in a separate PR and write tests for it

Pcap++/src/PcapLiveDevice.cpp
seladb: Fixed in this commit: https://github.com/seladb/PcapPlusPlus/commit/5251602c9b95ca0229c2a78fbda199c1816482d0

seladb (approved):

Pcap++/src/PcapLiveDevice.cpp
seladb: @Dimi1010 do you want to review the recent changes I made in this PR before we merge it?

Dimi1010 (approved):
LGTM

seladb:
Thank you @vcomito-apexai for contributing to this project! 🙏 ❤️ 
We've made some changes in your PR before meging it, please let me know if you have any concerns

Issue #1658, opened 2024-12-15
Reformat documentation to use triple-slash doxygen format.


Dimi1010:
## Rationale:
- Reduced line bloat. Each documentation block is 2 lines shorter on average.
This can compound to over 100 lines on some of the larger files. (e.g. "IPAddress.h")
- Makes usages of comment blocks /* */ easier to manage during refactoring, as they won't be terminated early if there are documentation comments inside them.

## Progress summary:
- [x] Common++
  - [x] Namespace and file docs (#1647)
  - [x] Class and Struct docs (#1647)
  - [x] Enums (unsure if the enums are formatted already)
- [x] Packet++
  - [x] Namespace and file docs (#1659)
  - [x] Class and Struct docs (#1659)
  - [x] Enums (#1659)
- [x] Pcap++
  - [x] Namespace and file docs (#1660)
  - [x] Class and Struct docs (#1660)
  - [x] Enums (#1660)

Dimi1010:
Discontinued due to issues with #1659 .

seladb:
@Dimi1010 I think we can close this issue, please let me know if you think otherwise

PR #1659, opened 2024-12-15
Reformat Packet++ documentation to use triple-slash doxygen format.

Dimi1010:
Part of #1658 
Includes reformatting of classes, structs and namespace documentation.


Dimi1010:
Ok, minor issue with the reformatting.
There are some `@verbatim` blocks that don't like working well with the `///` format, and I am not sure how to deal with that...
The issues are that I am unsure if the `///` will show in the actual verbatim text and the `///` block needs to be unbroken with those lines to count as a single block. Also, I am getting a parser error in the `SSLLayer` so the parser may also be getting confused.

So unless anyone has ideas, I guess, we have 3 options.
- Use the `/** */ ` style for those sections.
- Remove the verbatim blocks (undesirable imo, because they are helpful)
- Call it a day, revert the other 2 refactor PRs (#1660, #1647) and continue with the old `/** */` format. There haven't been many changes to the documentation yet, so the revert should be mostly clean if needed.

cc: @seladb @tigercosmos

Dimi1010:
Discontinued due to issues with doxygen parsing of verbatim example blocks.

Dimi1010:
Fix seems to work. Thanks @egecetin .

seladb (approved):

PR #1661, opened 2024-12-18
Removed VLA usage as it is not part of the C++ standard.

Dimi1010:
Variable Length Arrays, while widely supported as compiler extensions are not part of the C++ standard. Replaced a usage I found with a vector implementation.

### DIFF
--- a/Pcap++/src/XdpDevice.cpp
+++ b/Pcap++/src/XdpDevice.cpp
@@ -10,6 +10,7 @@
 #include <net/if.h>
 #include <sys/mman.h>
 #include <unistd.h>
+#include <vector>
 #include <functional>
 #include <algorithm>
 #include <poll.h>
@@ -144,6 +145,7 @@ namespace pcpp
 		pollfd pollFds[1];
 		pollFds[0] = { .fd = xsk_socket__fd(socketInfo->xsk), .events = POLLIN };
 
+		std::vector<RawPacket> receiveBuffer;
 		while (m_ReceivingPackets)
 		{
 			checkCompletionRing();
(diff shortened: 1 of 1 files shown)


ol-imorozko:
> Replaced a usage I found with a vector implementation.

I just want to note that there are also a few places where variable length arrays are used.

- Pcap++/src/DpdkDevice.cpp |932 col 30|
- Pcap++/src/DpdkDevice.cpp |975 col 30|
- Pcap++/src/DpdkDevice.cpp |1136 col 21|
- Pcap++/src/DpdkDevice.cpp |1138 col 35|
- Pcap++/src/DpdkDevice.cpp |1177 col 21|
- Pcap++/src/DpdkDevice.cpp |1178 col 35|
- Pcap++/src/DpdkDeviceList.cpp |124 col 31|
- Examples/DpdkExample-FilterTraffic/main.cpp |591 col 34|
- Examples/PfRingExample-FilterTraffic/main.cpp |409 col 29|
- Examples/PfRingExample-FilterTraffic/main.cpp |433 col 48|

tigercosmos (approved):
LGTM

Pcap++/src/XdpDevice.cpp
```diff
@@ -144,6 +145,7 @@ namespace pcpp
 		pollfd pollFds[1];
 		pollFds[0] = { .fd = xsk_socket__fd(socketInfo->xsk), .events = POLLIN };
 
+		std::vector<RawPacket> receiveBuffer;
```
egecetin, line 148: Isn't it better to move inside to while loop to decrease scope?

Dimi1010: No, because the buffer will need to be allocated every cycle. Currently it's only allocated once, and reallocated if expansion is needed. When its expanded, subsequent cycles can reuse the expanded size as is.

egecetin: @Dimi1010 Ok thanks for explanation

egecetin (approved):

seladb (approved):

PR #1662, opened 2024-12-18
Replace variable length arrays with std::vector

ol-imorozko:
Variable Length Arrays, while widely supported as compiler extensions are not part of the C++ standard. Replaced them with std::vector.

This PR extends [Dimi1010](https://github.com/Dimi1010)'s #1661

### DIFF
--- a/Examples/DpdkExample-FilterTraffic/main.cpp
+++ b/Examples/DpdkExample-FilterTraffic/main.cpp
@@ -162,7 +162,7 @@ void listDpdkPorts()
 void prepareCoreConfiguration(std::vector<pcpp::DpdkDevice*>& dpdkDevicesToUse,
                               std::vector<pcpp::SystemCore>& coresToUse, bool writePacketsToDisk,
                               const std::string& packetFilePath, pcpp::DpdkDevice* sendPacketsTo,
-                              AppWorkerConfig workerConfigArr[], int workerConfigArrLen, uint16_t rxQueues)
+                              std::vector<AppWorkerConfig>& workerConfigArr, int workerConfigArrLen, uint16_t rxQueues)
 {
 	// create a list of pairs of DpdkDevice and RX queues for all RX queues in all requested devices
 	int totalNumOfRxQueues = 0;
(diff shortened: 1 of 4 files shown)


Examples/DpdkExample-FilterTraffic/main.cpp
```diff
@@ -162,7 +162,7 @@ void listDpdkPorts()
 void prepareCoreConfiguration(std::vector<pcpp::DpdkDevice*>& dpdkDevicesToUse,
                               std::vector<pcpp::SystemCore>& coresToUse, bool writePacketsToDisk,
                               const std::string& packetFilePath, pcpp::DpdkDevice* sendPacketsTo,
-                              AppWorkerConfig workerConfigArr[], int workerConfigArrLen, uint16_t rxQueues)
+                              std::vector<AppWorkerConfig>& workerConfigArr, int workerConfigArrLen, uint16_t rxQueues)
```
Dimi1010, line 165: Doesn't this still work as an unknown bound array? It should be fine to be passed the `data()` ptr from the vector on the call site?

If it is kept as a vector ref, there is no point to the `workerConfigArrLen`.

Dimi1010 (approved):
LGTM, with 1 minor comment.

Examples/DpdkExample-FilterTraffic/main.cpp
ol-imorozko: Yes, we could pass the `data()` pointer, but there's no strong reason to do so. The `prepareCoreConfiguration` function iterates over `coresToUse` and fills the `workerConfigArr`, whose size is determined by `coresToUse.size()`. Thus, the logic of the code remains unchanged—we still iterate over its size and populate it.

As for `workerConfigArrLen`, it's an interesting case because it was unused even before this patch. Perhaps we should address this in a separate PR that removes unused parameters?

egecetin (approved):

seladb (approved):

seladb:
Thank you @ol-imorozko for this contribution, much appreciated! 🙏

PR #1666, opened 2024-12-19
Added exception when an unknown direction is supplied.

### LINKED ISSUE #1665: Compilation Error in PcapLiveDevice.cpp - Control Reaches End of Non-Void Function
### Question

I've encountered an issue with the Pcap++ library when I enabled the HAS_SET_DIRECTION_ENABLED option. The error I'm facing is related to control reaching the end of a non-void function, which is flagged as an error due to the -Werror=return-type compiler flag. This error is located in the PcapLiveDevice.cpp file, specifically on line 84

### Operating systems

_No response_

Dimi1010:
Should fix #1665.


egecetin (approved):

tigercosmos (approved):
LGTM

seladb (approved):

PR #1669, opened 2024-12-23
fixes a use-of-uninitialized-value  in light_pcapng.c

Shivam7-1:
This PR fixes a use-of-uninitialized-value vulnerability in light_pcapng.c. The changes ensure all allocated memory is properly initialized before use, mitigating potential crashes and undefined behavior.

Issue: https://issues.oss-fuzz.com/issues/42536106 
Reproducer Testcase: https://oss-fuzz.com/download?testcase_id=6639796737867776


Shivam7-1:
Hii @seladb Could You please Review this PR 
Thanks

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
```diff
@@ -85,56 +85,52 @@ static struct _light_option *__parse_options(uint32_t **memory, const int32_t ma
...
-         idb->reserved = (link_reserved >> 16) & 0xFFFF;
-         idb->snapshot_length = *local_data++;
-         current->block_body = (uint32_t*)idb;
-         local_offset = (size_t)local_data - (size_t)block_start;
-         opt = __parse_options((uint32_t **)&local_data, current->block_total_length - local_offset - sizeof(current->block_total_length));
-         current->options = opt;
-      }
-      break;
+void parse_by_block_type(struct _light_pcapng *current, const uint32_t *local_data, const uint32_t *block_start) {
+    switch (current->block_type) {
```
tigercosmos, line 89: It's better to follow the original format.

Shivam7-1: Thank you for reviewing Could you clarify if the concern is solely about formatting or if there are issues with the fix itself? If it’s just formatting, I’ll align it with the original style while keeping the vulnerability fix intact.

tigercosmos: Only about the formatting. We try to minimize the change in `3rdParty` because we may merge the latest changes from upstream in the future. However, vulnerability fixes are welcome. Please also leave a comment like `// PCPP patch` in your patch so we can easily identify the changes in the future.

Shivam7-1: Yes Done you can Review and Ready to Merge

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
```diff
@@ -85,56 +85,55 @@ static struct _light_option *__parse_options(uint32_t **memory, const int32_t ma
...
-      }
-      break;
+void parse_by_block_type(struct _light_pcapng *current, const uint32_t *local_data, const uint32_t *block_start) {
+    switch (current->block_type) 
+    {
+        case LIGHT_SECTION_HEADER_BLOCK: 
+		{ // PCPP patch
+            DPRINT_HERE(LIGHT_SECTION_HEADER_BLOCK);
+            struct _light_section_header *shb = calloc(1, sizeof(struct _light_section_header));
+            struct _light_option *opt = calloc(1, sizeof(struct _light_option));
```
Dimi1010, line 95: Doesn't that esentially leak the memory as `opt` is overwritten on L109?

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
```diff
@@ -85,56 +85,55 @@ static struct _light_option *__parse_options(uint32_t **memory, const int32_t ma
...
+            opt = __parse_options((uint32_t **)&local_data, current->block_total_length - local_offset - sizeof(current->block_total_length));
+            current->options = opt;
+        }
+        break;
+
+        case LIGHT_INTERFACE_BLOCK:
+		{ // PCPP patch
+            DPRINT_HERE(LIGHT_INTERFACE_BLOCK);
+            struct _light_interface_description_block *idb = calloc(1, sizeof(struct _light_interface_description_block));
+            struct _light_option *opt = calloc(1, sizeof(struct _light_option));
```
Dimi1010, line 118: Ditto: memory leak

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
```diff
@@ -85,56 +85,55 @@ static struct _light_option *__parse_options(uint32_t **memory, const int32_t ma
...
+            current->options = opt;
+        }
+        break;
 
       case LIGHT_ENHANCED_PACKET_BLOCK:
       {
          DPRINT_HERE(LIGHT_ENHANCED_PACKET_BLOCK);
          struct _light_enhanced_packet_block *epb = NULL;
-         struct _light_option *opt = NULL;
+         struct _light_option *opt = calloc(1, sizeof(struct _light_option));
```
Dimi1010, line 136: Ditto: memory leak

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
```diff
@@ -182,7 +181,7 @@ void parse_by_block_type(struct _light_pcapng *current, const uint32_t *local_da
       {
          DPRINT_HERE(LIGHT_CUSTOM_DATA_BLOCK);
          struct _light_custom_nonstandard_block *cnb = NULL;
-         struct _light_option *opt = NULL;
+         struct _light_option *opt = calloc(1, sizeof(struct _light_option));
```
Dimi1010, line 184: Ditto: memory leak

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
Shivam7-1: Yes you are correct for this here can be used free (opt) Memory Before Reassignment
I had created Commit below

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
Shivam7-1: Yes done some Improvement to memory free(opt) now there wouldn't be any leakage i think

tigercosmos:
@Shivam7-1  Could you only modify the necessary parts? Try to minimiaze the line of changes ...
Also please address the CI issue.

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
```diff
@@ -85,81 +85,92 @@ static struct _light_option *__parse_options(uint32_t **memory, const int32_t ma
...
-
-         memcpy(epb->packet_data, local_data, captured_packet_length); // Maybe actual_len?
-         local_data += actual_len / sizeof(uint32_t);
-         current->block_body = (uint32_t*)epb;
-         local_offset = (size_t)local_data - (size_t)block_start;
-         opt = __parse_options((uint32_t **)&local_data, current->block_total_length - local_offset - sizeof(current->block_total_length));
-         current->options = opt;
-      }
-      break;
+void parse_by_block_type(struct _light_pcapng *current, const uint32_t *local_data, const uint32_t *block_start) {
```
tigercosmos, line 88: Don't change the format. Same for all other lines.

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
Dimi1010: Why is it allocated here actually, if it will be freed in L110? I don't see any usages of the pointer in between?

Shivam7-1: I had updated Recent commit

Shivam7-1:
> @Shivam7-1 Could you only modify the necessary parts? Try to minimiaze the lines of changes ... Also please address the CI issue.

Hii where I could see CI issue or find out

seladb:
> @Shivam7-1 Could you only modify the necessary parts? Try to minimiaze the lines of changes ... Also please address the CI issue.

@Shivam7-1 as @tigercosmos commented, please try only to change specific lines and avoid changing the formatting. LightPcapNg is a 3rd party library so we try to make minimal change in it

Shivam7-1:
> > @Shivam7-1 Could you only modify the necessary parts? Try to minimiaze the lines of changes ... Also please address the CI issue.
> 
> @Shivam7-1 as @tigercosmos commented, please try only to change specific lines and avoid changing the formatting. LightPcapNg is a 3rd party library so we try to make minimal change in it

Hii @seladb Thanks For Reviewing I had Done this Changes and avoided any changes in formatting  it can reviewed here https://github.com/seladb/PcapPlusPlus/pull/1669/files

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
seladb, line 1: @Shivam7-1 maybe a better option would be to fix `__parse_options` to never return `NULL`?

Shivam7-1: Hii @seladb Yes I think modifying `__parse_options` to never return NULL would be another approach but it may introduce complexity in handling unknown options and ensuring a valid default return value.
The current approach which includes freeing the memory before reassignment is straightforward and ensures proper memory management.

seladb: `__parse_options` can only return `NULL` if `max_len <= 0` (which I guess is an invalid value and means the pcapng file is malformed), but this PR anyway handles (almost) all calls to this function, so why not fix it in the function itself? 🤔

Shivam7-1: Hi @seladb,

Thank you for the feedback. I understand your point about handling the invalid `max_len` case within `__parse_options`. 
Current changes is good I think because handling the NULL return explicitly in the calling code provides several benefits:

It makes the error condition clear and explicit which aids in debugging and understanding the code.
It would ensures no unnecessary memory allocation is performed when max_len <= 0, which can help with resource management.
Returning a static empty option would essentially act like returning NULL, as the calling code would still need to handle it as an error condition.
I think this approach maintains clarity and flexibility while managing resources efficiently.
If there would be change in max_len <=0 in this statement if statement looks like this static struct _light_option emptyOption = {0};
        return &emptyOption; ultimately it would get return empty only i think

Dimi1010: The way I see it, the options are represented as a forward linked list. Where if the `next` pointer is null the sequence is terminated and there is no next option.

Also I have a couple of questions regarding the changes:
1. Why are we calling `calloc` on `opt` declaration if it is immediately then going to be freed a couple of lines later, without actually doing anything with it? That looks to me like a pointless round trip to the heap allocator.
2. The way I read the code, the variables `version` and `local_offset` are also assigned prior to use in the original code, like on lines L102 and L109. What changes by assigning them to 0 first and then assigning them again?

seladb: @Shivam7-1 I don't have a strong opinion, so the current approach seems good enough for me. However, as @Dimi1010 said:

> 1. Why are we calling `calloc` on `opt` declaration if it is immediately then going to be freed a couple of lines later, without actually doing anything with it? That looks to me like a pointless round trip to the heap allocator.

I think what we should do is call `opt = __parse_options` first, and if `opt` is `NULL` (which should be a very rare case), call `calloc`

Shivam7-1: Hii @seladb Thanks feedback. Yes you are right that the initial calloc allocation for opt is unnecessary since it is immediately freed before being reassigned by __parse_options. I will update the code to call __parse_options first and only allocate memory with calloc if opt is NULL. This will avoid the redundant allocation and improve code efficiency.

seladb: Yes, it looks good now! I'll let the CI run to make sure everything is ok and then we can merge the PR

Shivam7-1: Hii @seladb Thanks For Reviewing yes I can see all checks are passed

tigercosmos (approved):

seladb (approved):

seladb:
Thank you @Shivam7-1 for working on this fix, much appreciated! 🙏

Shivam7-1:
> Thank you @Shivam7-1 for working on this fix, much appreciated! 🙏

Hii Thanks to you also and Team for Continuous Support and feedback

PR #1670, opened 2024-12-25
Fix pre-commit configuration for gersemi

egecetin:
Currently all txt files are excluded which prevents fixing of CMake files (which not noticed by me in #1602 unfortunately). This PR should fix this issue


.pre-commit-config.yaml
```diff
@@ -41,6 +41,7 @@ repos:
     rev: 0.17.1
     hooks:
     - id: gersemi
+      args: ["-c"]
```
egecetin, line 44: This ensures that gersemi run in check mode which returns 1 if any formatting required and 0 otherwise

seladb (approved):

Issue #1682, opened 2025-01-05
Can not build gratuitous ARP request properly


wangchong2023:
### Bug description

**Describe the bug**
RFC 5227 stipulates that gratuitous ARP can be a request, which is also described in the wireshark [wiki](https://wiki.wireshark.org/Gratuitous_ARP).
The characteristics of gratuitous ARP are that the sending and source IP addresses are both the IP address of the prober, and the source MAC addresses is the MAC address of the prober and **target MAC is ff:ff:ff:ff:ff:ff(not all zero)**.
Currently, the function 'computeCalculateFields' in Arplayer implemented by PcapPlusPlusPlus will set the targetMacAddr of the ARP request message to all zeros, which will cause the scenario of constructing a gratuitous ARP message to not work properly.

**Code example to reproduce**
```cpp
int NeighAdvSendArpPacket(PcapLiveDevice* dev,
								const MacAddress& sourceMacAddr, const MacAddress& dstMacAddr, const MacAddress& targetMacAddr,
								const IPv4Address& senderIpAddr, const IPv4Address& targetIP, ArpOpcode arpOperCode)
	{
		Packet advPacket(100);

		// Ethernet layer
		EthLayer ethLayer(sourceMacAddr, dstMacAddr);
		if (!advPacket.addLayer(&ethLayer))
		{
			PCPP_LOG_ERROR("Couldn't build Eth layer for ARP packet");
			return 1;
		}

		//ARP layer
		ArpLayer arpLayer(arpOperCode, sourceMacAddr, targetMacAddr, senderIpAddr, targetIP);
		if (!advPacket.addLayer(&arpLayer))
		{
			PCPP_LOG_ERROR("Couldn't build ARP layer for ARP packet");
			return 1;
		}

		advPacket.computeCalculateFields();

		dev->sendPacket(&advPacket);

		return 0;
	}
```

**Expected behavior**
It is recommended to remove this default behavior (the targetMacAddr field of the ARP request type is set to all zeros by default) or to control whether to clear it through an additional field.

### PcapPlusPlus versions tested on

PcapPlusPlus master branch

### Operating systems tested on

Windows

### Compiler version

GCC 10.2.0

### Packet capture backend (if applicable)

WinPcap

tigercosmos:
[...]

PR #1684, opened 2025-01-07
Allow gratuitous ARP requests.

Dimi1010:
This PR is a potential implementation of #1682.

### Specific header constructors
`ArpLayer` has been expanded with constructors accepting the following structures that build a respective ARP header inside the layer.
- ArpRequest
- ArpReply
- GratuitousArpRequest
- GratuitousArpReply


Packet++/header/ArpLayer.h
```diff
@@ -52,6 +52,57 @@ namespace pcpp
...
+	 * @brief A struct representing the build data for an ARP request
+	 */
+	struct ArpRequest
+	{
+		MacAddress senderMacAddr;
+		IPv4Address senderIpAddr;
+		IPv4Address targetIpAddr;
+
+		ArpRequest(MacAddress senderMacAddress, IPv4Address senderIPAddress, IPv4Address targetIPAddress)
+		    : senderMacAddr(senderMacAddress), senderIpAddr(senderIPAddress), targetIpAddr(targetIPAddress) {};
```
tigercosmos, line 65: As you have copied the objects while they are passed as arguments, I think you can use `std::move` here.

Packet++/header/ArpLayer.h
```diff
@@ -83,6 +134,30 @@ namespace pcpp
 		ArpLayer(ArpOpcode opCode, const MacAddress& senderMacAddr, const MacAddress& targetMacAddr,
```
tigercosmos, line 134: As you added other constructors, I am unsure if we need to expose this one as public.

Dimi1010: I'm also unsure if we need it public. I left it for backwards compatibility. I suppose having it public could allow users to manually punch in unsupported protocols, but I agree that having only the new constructors would better specify which requests are available for construction by the layer. If they want more they could always add their own constructor with a custom struct.

Packet++/header/ArpLayer.h
Dimi1010: Ended up using references to remove the initial copy into the ctor. As both `IPAddress` and `MACAddress` are small and all on the stack, IMO it is better to remove a possible copy all together, since the a move does not offer that much performance over a copy in this case.

seladb:
@Dimi1010 the main issue I see with this approach is that it's quite strict and limits users to specific options to create ARP messages. Creating a non-standard ARP message is not possible unless they use the deprecated c'tor.

My suggestion [here](https://github.com/seladb/PcapPlusPlus/issues/1682#issuecomment-2577077624) keeps the existing c'tor which gives the most freedom while creating another c'tor for gratuitous ARP requests which AFAIK are quite rarely used (no one requested it until now, even though ARP exists for many years in PcapPlusPlus)

Dimi1010:
> @Dimi1010 the main issue I see with this approach is that it's quite strict and limits users to specific options to create ARP messages. Creating a non-standard ARP message is not possible unless they use the deprecated c'tor.
> 
> My suggestion [here](https://github.com/seladb/PcapPlusPlus/issues/1682#issuecomment-2577077624) keeps the existing c'tor which gives the most freedom while creating another c'tor for gratuitous ARP requests which AFAIK are quite rarely used (no one requested it until now, even though ARP exists for many years in PcapPlusPlus)

I mean, deprecating the current ctor is not really a requirement, I deprecated it because people might have been relying on the zeroing the target MAC functionality. The new private ctor can also be exposed to allow ppl to construct whatever packets they want.

seladb:
> ing the current ctor is not really a requirement, I deprecated it because people might have been relying on the zeroing the target MAC functionality.
> 
> The new private ctor can also be exposed to

When I think about it, maybe we can just remove zeroing out `targetMacAddr` in `computeCalculateFields()` because we don't do it anywhere else 🤔 

We probably need to update some tests and examples that use ARP (like Arping) which assume an ARP request always has a `targetMacAddr` of zero, but maybe that's the most flexible solution that provides a viable fix. The downside is that it's a breaking change for everyone who assumes it's zerod...

Dimi1010:
> When I think about it, maybe we can just remove zeroing out `targetMacAddr` in `computeCalculateFields()` because we don't do it anywhere else 🤔
> 
> We probably need to update some tests and examples that use ARP (like Arping) which assume an ARP request always has a `targetMacAddr` of zero, but maybe that's the most flexible solution that provides a viable fix. The downside is that it's a breaking change for everyone who assumes it's zerod...

I can make the `ArpLayer(ArpOpcode opCode, const MacAddress& senderMacAddr, const IPv4Address& senderIpAddr, const MacAddress& targetMacAddr, const IPv4Address& targetIpAddr);` constructor public. It would allow a gentler transition for those that use the zeroing, as the old one will be flagged as deprecated to allow them time to transition.

seladb:
> > When I think about it, maybe we can just remove zeroing out `targetMacAddr` in `computeCalculateFields()` because we don't do it anywhere else 🤔
> > We probably need to update some tests and examples that use ARP (like Arping) which assume an ARP request always has a `targetMacAddr` of zero, but maybe that's the most flexible solution that provides a viable fix. The downside is that it's a breaking change for everyone who assumes it's zerod...
> 
> I can make the `ArpLayer(ArpOpcode opCode, const MacAddress& senderMacAddr, const IPv4Address& senderIpAddr, const MacAddress& targetMacAddr, const IPv4Address& targetIpAddr);` constructor public. It would allow a gentler transition for those that use the zeroing, as the old one will be flagged as deprecated to allow them time to transition.
> 
> Also a standard ARP request should have the target mac zeroed. It is in the spec.

You convinced me 🙂 
If we keep the existing c'tor (not deprecated) then I guess we can add the additional c'tors with specific messages, and remove zeroing `targetMacAddr`

Packet++/header/ArpLayer.h
Dimi1010: https://github.com/seladb/PcapPlusPlus/pull/1684#issuecomment-2582135925 Decided to keep a version of the direct ctor public to not limit the ARP requests users can form. The structures are to be left as helpers instead of limiters.

Packet++/header/ArpLayer.h
```diff
@@ -72,17 +183,57 @@ namespace pcpp
...
+		 * @brief A constructor that creates an ARP header
+		 * @param[in] opCode ARP message type (ARP request or ARP reply)
+		 * @param[in] senderMacAddr The sender MAC address (will be put in arphdr#senderMacAddr)
+		 * @param[in] senderIpAddr The sender IP address (will be put in arphdr#senderIpAddr)
+		 * @param[in] targetMacAddr The target MAC address (will be put in arphdr#targetMacAddr)
+		 * @param[in] targetIpAddr The target IP address (will be put in arphdr#targetIpAddr)
+		 * @remarks No validation is done on the input parameters. The caller must ensure that the input creates a valid
+		 * header.
+		 */
+		ArpLayer(ArpOpcode opCode, const MacAddress& senderMacAddr, const IPv4Address& senderIpAddr,
```
seladb, line 196: Why do we need a new c'tor with the same params as the other c'tor but in a different order?

Tests/Packet++Test/Tests/EthAndArpTests.cpp
```diff
@@ -120,32 +120,129 @@ PTF_TEST_CASE(EthAndArpPacketParsing)
...
+		pcpp::arphdr* arpHeader = pArpLayer->getArpHeader();
+		PTF_ASSERT_EQUAL(arpHeader->hardwareSize, 6);
+		PTF_ASSERT_EQUAL(arpHeader->protocolType, htobe16(PCPP_ETHERTYPE_IP));
+
+		PTF_ASSERT_EQUAL(bufferLength1, argRequestPacket.getRawPacket()->getRawDataLen());
+		PTF_ASSERT_BUF_COMPARE(argRequestPacket.getRawPacket()->getRawData(), buffer1, bufferLength1);
+	}
+
+	{
+		// TODO: Add an actual packet to test against.
```
seladb, line 183: You can use [ArpResponsePacket.dat](https://github.com/seladb/PcapPlusPlus/blob/1e981935e15bee8c639499bd7582364beb66fb7d/Tests/Packet++Test/PacketExamples/ArpResponsePacket.dat)

seladb (commented):
There are also ARP layers generated in 3 places in ArpSpoofing that we can modify: https://github.com/seladb/PcapPlusPlus/blob/feba0b7124a4536ec2a0a947bc96c146ada099d7/Examples/ArpSpoofing/main.cpp

Packet++/header/ArpLayer.h
Dimi1010: The new ctor is because the old one still zeroes the targetMacAddress to keep people who are expecting that from forming invalid arp requests. It just does it in the ctor instead of the `calculateFields`.

The new constructor is there because:
- The targetMacAddr is no longer zeroed. All parameters are directly written in the header.
- Parameters are reordered because you cant have two constructors with identical signature. Also they are now ordered in the way they appear in the header.

Packet++/header/ArpLayer.h
```diff
@@ -95,6 +246,13 @@ namespace pcpp
 			return reinterpret_cast<arphdr*>(m_Data);
 		}
 
+		/**
+		 * Get the ARP opcode
+		 * @return The ARP opcode
+		 * @remarks The opcode may not be one of the values in @ref ArpOpcode
+		 */
+		ArpOpcode getOpcode() const;
```
tigercosmos, line 254: btw, we have `isReply` function though

Dimi1010: The Op codes for a packet can technically be other values than request and reply that we have currently.

tigercosmos (approved):

Packet++/header/ArpLayer.h
seladb: ok sure, let's keep both of them for backward compatibility and remove the previous one in the next version 👍

Tests/Packet++Test/Tests/EthAndArpTests.cpp
Dimi1010: Sure, but why is the packet size 60 in `EthAndArpPacketParsing`? 14 (`EthLayer`) +28 (`ArpLayer`) makes 42, no?

seladb: Good question, the extra is Ethernet trailer, as you can see in [`ArpResponsePacket.pcap`](https://github.com/seladb/PcapPlusPlus/blob/1e981935e15bee8c639499bd7582364beb66fb7d/Tests/Packet++Test/PacketExamples/ArpResponsePacket.pcap). But it is still a valid ARP response packet

image

tigercosmos: ARP response can have random length padding at the end, which also confused me before.

Dimi1010: Added the packet to be used for the test.

seladb (approved):

PR #1690, opened 2025-01-17
Revert of #1660

Dimi1010:
This PR reverts the formatting changes made in #1660.

The reason for the revert is issues with #1659. Verbatim blocks have issues parsing with the `///` syntax. And since the PRs are linked, decided it was better to revert to the `/**` structure than risk breaking the documentation.

This reverts commit 91d20581a28821799c6be9df64bf274d951a3322.


tigercosmos:
Why don't you just revert the places that have issue?

Dimi1010:
> Why don't you just revert the places that have issue?

@tigercosmos  Uniformity mostly. The issue isn't in these PRs per se. The issue is in the Packet++ one (#1659), that still isn't merged. [Its an issue with doxygen getting confused with verbatim blocks with `///` comments](https://github.com/seladb/PcapPlusPlus/pull/1659#issuecomment-2581338453). The revert here is mostly so we aren't left with half the project in `///` format and the other in `/**` format.

tigercosmos:
IMO, I would like to have `///` most of the places, and only use `/** */` whenever there is `@verbatim`. In addition, only 5 files contain `@verbatim`, I think we can just treat them as special cases.

Dimi1010:
> IMO, I would like to have `///` most of the places, and only use `/** */` whenever there is `@verbatim`. In addition, only 5 files contain `@verbatim`, I think we can just treat them as special cases.

I am personally fine with either, tbf. As I mentioned in the linked comment, there are options.
I decided to do a revert PR due to the fact that so far we have tried to have the documentation uniform.

@seladb What do you say?

egecetin:
@Dimi1010 Not followed the issues so it might be off topic, but did you saw [this](https://web.archive.org/web/20240724083629/https://technicalwriting.dev/src/verbatim-wrangling.html)? Maybe it can fix the issues you encountered?

Dimi1010:
> @Dimi1010 Not followed the issues so it might be off topic, but did you saw [this](https://web.archive.org/web/20240724083629/https://technicalwriting.dev/src/verbatim-wrangling.html)? Maybe it can fix the issues you encountered?

@egecetin huh... that might work actually.

seladb:
@Dimi1010 I'm sorry for the delayed response. I can see that replacing `@verbatim` with `@code{.unparsed}` solves the issue, so that's great news! 🎉

PR #1692, opened 2025-01-18
Refactored MacAddress to use std::array internally.

Dimi1010:


Packet++/header/StpLayer.h
```diff
@@ -2,6 +2,7 @@
 
 #include "Layer.h"
 #include "MacAddress.h"
+#include <cstring>
```
tigercosmos, line 5: Is it neccessary?

Dimi1010: Yeah, a lot of translation units were apparently depending on `MacAddress.h` having `cstring`/`string.h` included instead of including it themselves for `memset`/`memcpy`/`memcmp`.

tigercosmos (approved):

seladb (approved):

PR #1693, opened 2025-01-19
Added static asserts to fixed length header structures.

Dimi1010:
The PR adds cpp11 `static_assert` statements after the declarations of fixed length structured used throughout Packet++ for encoding/decoding raw data.

### DIFF
--- a/Packet++/header/BgpLayer.h
+++ b/Packet++/header/BgpLayer.h
@@ -59,6 +59,7 @@ namespace pcpp
 			uint8_t messageType;
 		};
 #pragma pack(pop)
+		static_assert(sizeof(bgp_common_header) == 19, "bgp_common_header size is not 19 bytes");
 
 		/**
 		 * @return BGP message type
--- a/Packet++/header/CotpLayer.h
+++ b/Packet++/header/CotpLayer.h
@@ -11,7 +11,7 @@ namespace pcpp
  * Represents a COTP protocol header
  */
 #pragma pack(push, 1)
-	typedef struct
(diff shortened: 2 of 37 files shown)


seladb:
@Dimi1010 why are these static asserts needed? 🤔

seladb:
> > @Dimi1010 why are these static asserts needed? 🤔
> 
> They aren't needed per se, but they provide compile time safety that the header structures are precisely the specified fixed length, as they will raise an error at compile time if not.

I always wondered why static asserts are needed 🙂 
If someone decides to change a struct, it's probably for a good reason and they'll need to update the static assert as well.
Is there any real purpose for it?

Dimi1010:
> I always wondered why static asserts are needed 🙂 
> If someone decides to change a struct, it's probably for a good reason and they'll need to update the static assert as well.
> Is there any real purpose for it?

That is kinda the point. If you are changing the structs, it stos you by asking the question: "Do you really have a good reason to change it?".

My opinion in general about them is that it prevents accidental changes and also ensures that certain assumptions that the developers make about the code are correct, instead of only relying on the dev that the assumption holds.

tigercosmos:
I think this change is helpful. It prevents people changing the struct layout accidently. I think asssertion and test are always useful.

seladb (approved):

seladb:
@Dimi1010 @tigercosmos if both of you think it's helpul I'm ok with it and we can merge this PR, although I'm not 100% convinced why it is needed 🙂

PR #1694, opened 2025-01-19
Remove non-functional SSL 2 version code point

droe:
SSL 2 uses a version field of 0x0002, not 0x0200.  This is confirmed not only in the original Netscape spec [1] and RFC draft of the time [2], but also in major implementations such as OpenSSL [3] and Wireshark [4].

[1] https://www-archive.mozilla.org/projects/security/pki/nss/ssl/draft02.html
[2] https://datatracker.ietf.org/doc/html/draft-hickman-netscape-ssl-00
[3] https://github.com/openssl/openssl/blob/OpenSSL_0_9_6m/ssl/ssl2.h#L66-L71
[4] https://github.com/wireshark/wireshark/blob/release-4.4/epan/dissectors/packet-tls-utils.h#L266-L277

### DIFF
--- a/Packet++/header/SSLCommon.h
+++ b/Packet++/header/SSLCommon.h
@@ -100,8 +100,6 @@ namespace pcpp
 		/// SSL/TLS versions enum
 		enum SSLVersionEnum
 		{
-			/// SSL 2.0
-			SSL2 = 0x0200,
 			/// SSL 3.0
 			SSL3 = 0x0300,
 			/// TLS 1.0
--- a/Packet++/src/SSLCommon.cpp
+++ b/Packet++/src/SSLCommon.cpp
@@ -23,9 +23,6 @@ namespace pcpp
 				return static_cast<SSLVersion::SSLVersionEnum>(m_SSLVersionValue);
 		}
 
-		if (m_SSLVersionValue == 0x200)
-			return SSLVersion::SSL2;
-
 		return SSLVersion::Unknown;
 	}
 
@@ -81,8 +78,6 @@ namespace pcpp
 			return "TLS 1.3 (Facebook draft 26)";
 		case SSLVersion::Unknown:
 			return "Unknown";
-		case SSLVersion::SSL2:
-			return "SSL 2.0";
 		default:
 			return "Unknown";
 		}


tigercosmos:
It's just bytes order. PCPP here is using host order, and other libraries that you posted are using network order.

droe:
If it was byte order, as you suggest, then the constants for SSL 3 and TLS would be wrong.

droe:
I missed code in `Packet++/src/SSLCommon.cpp` that used `0x200` in the first attempt, second attempt also addresses that.

Dimi1010:
Close and reopen to run CI after base branch change.

tigercosmos:
@droe So is this still an issue?

droe:
Yes, certainly.  Code point 0x0200 is wrong, 0x0002 as proposed in this PR is correct.

tigercosmos:
@droe  In Wireshark the values are:
```
#define SSLV2_VERSION           0x0002 /* not in record layer, SSL_CLIENT_SERVER from
                                          http://www-archive.mozilla.org/projects/security/pki/nss/ssl/draft02.html */
#define SSLV3_VERSION          0x300
#define TLSV1_VERSION          0x301
```

And in PCPP, the current (master) values are:
```
			/** SSL 2.0 */
			SSL2 = 0x0200,
			/** SSL 3.0 */
			SSL3 = 0x0300,
			/** TLS 1.0 */
			TLS1_0 = 0x0301,
```

According to the byte order, I think `SSL2` should be `0x0020`, not `0x0002`?

droe:
It should be 0x0002 for SSL 2.  This is not a byte order problem (and you're swapping nibbles, not bytes).

tigercosmos:
I meant: major is 0x00, minor is 0x02. If you swap the bytes, 0x02 can be 0x2.

droe:
Not sure I understand what you're saying, uint8_t 0x02 and 0x2 are the same thing.  You are running the version uint16_t from the wire through be16toh() at deserialization time, so the code ends up seeing SSLVersion 0x0002 in host byte order for the SSL 2 version bytes `00 02` on the wire.

Packet++/header/SSLCommon.h
```diff
@@ -117,7 +117,7 @@ namespace pcpp
 		enum SSLVersionEnum
 		{
 			/** SSL 2.0 */
-			SSL2 = 0x0200,
+			SSL2 = 0x0002,
```
seladb, line 120: I wouldn't change the byte order here because it'll be different from the rest of the values in this enum

Packet++/header/SSLCommon.h
seladb, line 1: Do you happen to have a SSLv2 pcap that we can use in tests? If so, it'd be great to add a test for it!

When I wrote this code I couldn't find a pcap file...

Packet++/header/SSLCommon.h
tigercosmos: @seladb  Actually I think `0x0200` is wrong, and it should be `0x0020` based on other enum value byte order.

Packet++/header/SSLCommon.h
seladb: That's great, thanks! Would you consider adding a test with a single SSLv2 packet from this file?

droe: Fair request, I'll look into it.

droe:
```
SSL2RecordSSL2ClientHelloTest      : PASSED
SSL2RecordTLS1ClientHelloTest      : PASSED
```

Tests/Packet++Test/main.cpp
```diff
@@ -210,6 +210,8 @@ int main(int argc, char* argv[])
 	PTF_RUN_TEST(TLSCipherSuiteTest, "ssl");
 	PTF_RUN_TEST(ClientHelloTLSFingerprintTest, "ssl");
 	PTF_RUN_TEST(ServerHelloTLSFingerprintTest, "ssl");
+	PTF_RUN_TEST(SSL2RecordSSL2ClientHelloTest, "ssl;ssl2");
+	PTF_RUN_TEST(SSL2RecordTLS1ClientHelloTest, "ssl;ssl2");
```
seladb, line 214: nit: maybe make the names shorter?
```
SSL2ClientHelloTest
TLS1ClientHelloTest
```

Tests/Packet++Test/PacketExamples/ssl2rec-ssl2clienthello.dat
seladb, line 1: Can you add the pcap file as well for these 2 packets to make it easier to view them in Wireshark?

Tests/Packet++Test/Tests/SSLTests.cpp
```diff
@@ -753,3 +754,69 @@ PTF_TEST_CASE(ServerHelloTLSFingerprintTest)
...
+PTF_TEST_CASE(SSL2RecordSSL2ClientHelloTest)
+{
+	timeval time;
+	gettimeofday(&time, nullptr);
+
+	READ_FILE_AND_CREATE_PACKET(1, "PacketExamples/ssl2rec-ssl2clienthello.dat");
+
+	pcpp::Packet clientHelloPacket(&rawPacket1);
+
+	// PCPP does not know how to parse SSLv2 yet, so we find the version field manually.
```
seladb, line 801: I probably miss something here - the change you made in `SSLCommon` should cover parsing of SSLv2, am I wrong?

Why do we need to parse the packet manually then? 🤔

Tests/Packet++Test/main.cpp
droe: That would remove very important information:  These are a SSL 2 ClientHello in an SSL 2 record, and TLS 1 ClientHello in an SSL 2 record, respectively.  The "SSL 2 record" bit is crucial because the wire format is different.  TLS 1 handshake in SSL 2 record is different from a regular TLS 1 handshake in a TLS 1 record.

As of now, PCPP knows how to parse record format for SSL 3 through TLS 1.3, but not SSL 2 (neither plain SSL 2 nor SSL 2 compatible SSL 3 and later).

Tests/Packet++Test/PacketExamples/ssl2rec-ssl2clienthello.dat
droe: Done

Tests/Packet++Test/Tests/SSLTests.cpp
droe: Because of what is says in the comment:  PCPP does not know how to parse SSL 2 record format, nor SSL 2 handshake messages.  While SSL 3 developed from SSL 2 and has similar concepts and a similar structure, it is more useful to think of it as a different protocol.  SSL 2 support is _not_ just a matter of looking at the version field in SSL 3 and later.  SSL 2 uses a different record format (that has no version field!), different protocol layering, and handshake messages look different.  For instance, cipher suite code points are three bytes, not two bytes.  Telling the difference between SSL 2 and SSL 3 without context is not straightforward and takes heuristics.

The point of this PR is _only_ to fix the objectively wrong SSL 2 version code point from `0x0200` to the only code point you'll ever see on the wire for SSL 2: `0x0002`.

To bring full SSL 2 support to PCPP, much more work would be needed.

I've included the tests that you asked for only to settle the "endianness" misconceptions by demonstrating that with SSL 2 compatible SSL 3 / TLS handshakes (as per RFC 6101 Appendix E), you _will_ see 0x0002 for an SSL 2 ClientHello transported over an SSL 2 record, and 0x0301 for an TLS 1 ClientHello transported over an SSL 2 record.

Packet++/header/SSLCommon.h
tigercosmos: @droe  Please change the value to `0x0020`. 

Let take a look a [Wireshark code](https://github.com/wireshark/wireshark/blob/release-4.4/epan/dissectors/packet-tls-utils.h#L266-L277):
```
#define SSLV2_VERSION           0x0002 /* not in record layer, SSL_CLIENT_SERVER from
                                          http://www-archive.mozilla.org/projects/security/pki/nss/ssl/draft02.html */
#define SSLV3_VERSION          0x300
#define TLSV1_VERSION          0x301
```

and let's look at PCPP code:
```
			/** SSL 3.0 */
			SSL3 = 0x0300,
			/** TLS 1.0 */
			TLS1_0 = 0x0301,
```

It's obvious that `SSL2` should be `0x0020` instead of `0x0002`.

tigercosmos (changes requested):

Packet++/header/SSLCommon.h
droe: No, that does not make any sense.  You seem to be confused about some key aspect of this.  If you explain step by step how you determined the (wrong) value 0x0020, I might be able to tell what exactly you are confused about.

Tests/Packet++Test/Tests/SSLTests.cpp
seladb: I didn't realize SSLv2 is so different from SSLv3 (aka TLS). Maybe the solution is to actually remove SSLv2 from the enum if we don't actually support it?

I assume no one complained so far because they didn't use SSLv2, so even though it's a breaking change it's not really breaking anything because things are already broken... 😕 

@droe do you think it'd be worth the effort to implement a separate SSLv2 parser?

Packet++/header/SSLCommon.h
tigercosmos: @seladb could you help double check?

Tests/Packet++Test/Tests/SSLTests.cpp
droe: I concur, removing `SSL2` from the enum would also be a sensible fix.  Happy to update the branch accordingly if you prefer to go down that route.

I don't have an opinion on whether it's worth implementing full support for SSL 2 and/or SSL 2 compatible SSL 3 / TLS 1.x in PCPP.  For most use of PCPP, it's probably irrelevant.  For some niche use cases, it might be useful.  I'm not sure if it can be done without API-breaking changes.

I'm not going to be contributing full support for SSL 2 / SSL 2 compatible SSL 2 / TLS 1.x.  My mission is just to shepherd this change in order to fix or remove the wrong code point for SSL 2.

seladb: @droe I think we can remove SSL2 from the enum for now, and later consider whether to implement a full parser or not

Packet++/header/SSLCommon.h
seladb: @tigercosmos please see our discussion here: https://github.com/seladb/PcapPlusPlus/pull/1694#discussion_r1947974342

seladb (approved):

seladb:
Thank you @droe for working on this, and sorry for the delay in merging the PR 🙏

PR #1703, opened 2025-02-01
Fixed #1635 - xsk.h should be taken from libxdp

### LINKED ISSUE #1635: bpf/xsk.h file not found
https://github.com/seladb/PcapPlusPlus/blob/35f486cfa677c4a11a9d3742d7f32ab51da9870e/Pcap%2B%2B/src/XdpDevice.cpp#L8C1-L8C21
Didn't this file move to libxdp and should be under xdp/xsk?

hoxnox:
Added FindXDP.cmake, corrected CMakeLists and changed bpf/xsk.h to xdp/xsk.h.
This is fix for https://github.com/seladb/PcapPlusPlus/issues/1635.


Pcap++/src/XdpDevice.cpp
```diff
@@ -5,7 +5,7 @@
 #include "Logger.h"
 #include "Packet.h"
 #include <bpf/libbpf.h>
-#include <bpf/xsk.h>
+#include <xdp/xsk.h>
```
tigercosmos, line 8: Can you explain why we need "xdp" instead of "bpf" here?

hoxnox: > Can you explain why we need "xdp" instead of "bpf" here?

On my systems there is no such header. Same is here: https://github.com/seladb/PcapPlusPlus/issues/1635.

tigercosmos: @hoxnox  I think we may need to add some conditions here instead of changing the code directly. What's your system? Maybe just add a special case for handling your system?

hoxnox: > @hoxnox I think we may need to add some conditions here instead of changing the code directly. What's your system? Maybe just add a special case for handling your system?

I don't understand why in your system "xsk.h" present. There is no such at https://github.com/libbpf/libbpf and no such in kernel tree https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/lib/bpf?h=v6.14-rc1

tigercosmos: I think it might based on how and where you install it. I saw peopl using `<bpf/xsk.h>`, `<xdp/xsk.h>`, and `<linux/xsk.h>`. Maybe we need a better way to find the corret one for the user system.

hoxnox: > I think it might based on how and where you install it. I saw people using `<bpf/xsk.h>`, `<xdp/xsk.h>`, and `<linux/xsk.h>`. Maybe we need a better way to find the corret one for the user system.

I'll try to figure it out and return with more accurate patch.

hoxnox: Between 5.15.178 and 6.1.128 xsk.h was moved from `tools/lib/bpf/bpf.h` to `net/xdp/xsk.h` and it is not part of linux-headers. So regular user with kernel 6+ doesn't have that header.

seladb: @hoxnox I have a few questions:
- Are you sure `libxdp` is necessary to work with AF_XDP? I managed to get it work without it, at least in older kernels
- The [XDP build](https://github.com/seladb/PcapPlusPlus/actions/runs/13086459644/job/36518189771?pr=1703) fails because it can't find `libxdp`. I think it'd be great if we can support both options, and then we can add a build where libxdp is installed

seladb: @hoxnox I'm not sure if you saw my previous comment ☝️ 
Please let me know if you're interested on continuing working on this PR

hoxnox: > @hoxnox I'm not sure if you saw my previous comment ☝️ Please let me know if you're interested on continuing working on this PR

I had to patch both libxdp and pcapplusplus. I've created related conan packages libbpf, libxdp and pcapplusplus: https://github.com/hoxnox/conan-center-index/tree/missing/my/recipes. It successfully builds on all my systems - gentoo, ubuntu and alpine. Unfortunately I have no time to make it accurate enough to contribute. Sorry.

seladb: Got it, thanks @hoxnox ! I'll close this PR now, we may want to add libxdp support in the future if there's a demand for it

PR #1718, opened 2025-02-18
Fix getCommandOption for command-only packets

s-genereux:
Pcap++ crashes when you attempt to getCommandOption() of a command-only FTP request packet. I traced the issue to SingleCommandTextProtocol::getCommandOptionInternal(). 

To check for command-only packets, it compares the offset to the option field returned by getArgumentFieldOffset() to the raw data length minus 1. This should be minus 2;

1. -1 to account for the fact getArgumentFieldOffset() returns the position of the CR in the CRLF terminator in the case of command-only packets.
2. -1 to account for the fact we are comparing a length value to a zero-indexed position value. 

This PR makes the necessary change in SingleCommandTextProtocol.cpp and adds a command-only FTP request packet testcase to Packet++Test


egecetin (approved):
Fails due to brownouts. Need to run CI later

seladb (approved):

seladb:
Thank you @s-genereux for working on this fix! 🙏

PR #1738, opened 2025-03-23
Refactor reentrant flag to be set during channel creation.

Dimi1010:
Split of #1668 

This PR aims to standardize `m_ReentrantMode` to be set during the channel opening phase instead of the capture initialization phase as it is determined by the opening flags. The currently used implementation essentially assumes the what the reentrant mode is based on the function being called to start the capture, instead of recording the correct value.

### DIFF
--- a/Pcap++/src/PfRingDevice.cpp
+++ b/Pcap++/src/PfRingDevice.cpp
@@ -75,6 +75,8 @@ namespace pcpp
 		{
 			PCPP_LOG_DEBUG("Succeeded opening device [" << m_DeviceName << "]");
 			m_NumOfOpenedRxChannels = 1;
+			// Set reentrant mode to false as the channel is opened without the PF_RING_REENTRANT flag.
+			m_ReentrantMode = false;
 			m_DeviceOpened = true;
 			return true;
 		}
@@ -92,7 +94,7 @@ namespace pcpp
 		return openMultiRxChannels(channelIds, 1);
 	}
 
-	int PfRingDevice::openSingleRxChannelImpl(const char* deviceName, pfring** ring)
+	int PfRingDevice::openSingleRxChannelImpl(const char* deviceName, pfring** ring, bool useReenterant)
 	{
 		if (m_DeviceOpened)
 		{
@@ -101,6 +103,11 @@ namespace pcpp
 		}
 
 		uint32_t flags = PF_RING_PROMISC | PF_RING_HW_TIMESTAMP | PF_RING_DNA_SYMMETRIC_RSS;
+		if (useReenterant)
+		{
+			flags |= PF_RING_REENTRANT;
+		}
+
 		*ring = pfring_open(deviceName, DEFAULT_PF_RING_SNAPLEN, flags);
 
 		if (*ring == nullptr)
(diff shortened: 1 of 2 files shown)


Pcap++/src/PfRingDevice.cpp
```diff
@@ -176,7 +183,9 @@ namespace pcpp
 			std::string ringName = ringNameStream.str();
 			PCPP_LOG_DEBUG("Trying to open device [" << m_DeviceName << "] on channel [" << channelId
 			                                         << "]. Channel name [" << ringName << "]");
-			int res = openSingleRxChannel(ringName.c_str(), &m_PfRingDescriptors[i]);
+			// todo: Shouldn't we use the reentrant mode here? We are opening multiple channels?
+			// todo: Potentially only open in reenterant mode if creating N > 1 channels?
+			int res = openSingleRxChannelImpl(ringName.c_str(), &m_PfRingDescriptors[i]);
```
Dimi1010, line 188: @seladb Can you check please this code? 

Why is the implementation opening multiple Rx channels in non-reentrant mode here, when the other multiple channel function opens them in reentrant mode?

seladb:
@Dimi1010 this code was written many years ago so I don't remember all the details, but I _think_ the `PF_RING_REENTRANT` flag should be set if multiple threads are going to read from the same PF_RING device. However, if we know that a single thread is going to read from a a PF_RING device then this flag is redundant because it has a performance cost.

The problem with the current implementation is that when opening the channels we don't know how many threads will read from them: 
- We could open multiple channels, but only a single thread will read from them - the flag is redundant
- We could open multiple channels, and multiple threads will read from each of them - the flag is needed
- We could open a single channel, and multiple threads will read from it - the flag is needed
- We could open a single channel, and a single thread will read from it- the flag is redundant

In addition - if opening in reentrant mode then zero copy can't be used, hence the `m_ReentrantMode` private member.

We may need a bigger refactor to allow the user to choose when opening the device whether multiple threads will read from each channel (and set the flag accordingly), and when start capturing verify that we don't allow multiple threads to read from the same channel.

But this is something that needs further testing with a PF_RING environment to test in - do you have such an environment?

Dimi1010:
> We may need a bigger refactor to allow the user to choose when opening the device whether multiple threads will read from each channel (and set the flag accordingly), and when start capturing verify that we don't allow multiple threads to read from the same channel.

This would work. I think a bool in the `openChannel` functions would be enough. Something like `multithreadedCaptureSupport` that controls if the flag is used.
 
> But this is something that needs further testing with a PF_RING environment to test in - do you have such an environment?

Not fully. I have a WSL to confirm compilation, but not sure if that will work for actual capture testing.

seladb:
> This would work. I think a bool in the `openChannel` functions would be enough. Something like `multithreadedCaptureSupport` that controls if the flag is used.

Yes, something like that could work

> Not fully. I have a WSL to confirm compilation, but not sure if that will work for actual capture testing.

I'm not sure it'll work well with WSL, you may need to set up an Ubuntu VM. I use VirtualBox on Windows...

Dimi1010:
Closing this as stale for now. Might work on it again eventually.

PR #1740, opened 2025-03-23
Add stop token contract to encapsulate shutdown signal interface.

Dimi1010:
Split of #1668 

- Adds `StopToken` and `StopTokenSource` to encapsulate a stop token contract.
- Replace `PfRingDevice::m_stopThread` with a stop token source.
- Remove the check of the startup block `state` variable to stop the capture thread early due to an error.
- Added waiting for all started threads to shutdown if an exceptional state happened during capture thread startup.

### DIFF
--- a/Pcap++/header/PfRingDevice.h
+++ b/Pcap++/header/PfRingDevice.h
@@ -6,6 +6,7 @@
 #include "MacAddress.h"
 #include "SystemUtils.h"
 #include "Packet.h"
+#include "StopToken.h"
 #include <array>
 #include <thread>
 #include <mutex>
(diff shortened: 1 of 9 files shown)


Pcap++/header/PfRingDevice.h
```diff
@@ -59,7 +60,8 @@ namespace pcpp
 		MacAddress m_MacAddress;
 		int m_DeviceMTU;
 		std::array<CoreConfiguration, MAX_NUM_OF_CORES> m_CoreConfiguration;
-		bool m_StopThread;
+		// An empty stop token source is used to indicate that no capture is running
+		internal::StopTokenSource m_StopTokenSource{ internal::NoStopStateTag{} };
```
seladb, line 64: Wouldn't it be simpler to replace `bool m_StopThread` with `std::atomic<bool> m_StopThread` instead of building this complex stop-token infra?

Dimi1010: Possibly. The stop token was necessary for the split in #1668 due to the main capture function being separated from the `PfRingDevice` class. If we don't do that, either is fine technically.

seladb: Even if we separate the main capture function, we can pass is a `std::shared_ptr` of the `std::atomic<bool>` which is basically what the stop token is doing but simpler and less code.

But keeping the capture function inside the device is also ok with me.

Dimi1010: Sure, that works. I will close this PR for now and extract the fixes in this PR to a separate one. (like waiting for all capture threads to be stopped in case of failed initialization in the `startCaptureMultiThread`)

PR #1748, opened 2025-03-27
Templated next layer sub-construction.

Dimi1010:
With regards to this comment: https://github.com/seladb/PcapPlusPlus/pull/1741#discussion_r2013486816

This proof of concept PR provides a templated helper functions that automate the boilerplate of validating and constructing the next layer.
The templated functions also implicitly require Layer classes to provide `bool isDataValid(...)` static member. If the member function is not provided `tryConstructNextLayer` (and derivatives) won't compile.


Packet++/header/Layer.h
```diff
@@ -189,8 +212,96 @@ namespace pcpp
...
+		///
+		/// See the documentation for tryConstructNextLayer<T1, T2, T...>(data, dataLen, packet) for more information.
+		///
+		/// @tparam T1 The first type in the template parameter pack.
+		/// @param[in] data The data to construct the layer from.
+		/// @param[in] dataLen The length of the data.
+		/// @param[in] packet The packet the layer belongs to.
+		/// @return The constructed layer or nullptr if the data is invalid for all types in the pack.
+		/// @todo Replace with a fold expression in Cpp17
+		template <typename T1> Layer* tryConstructNextLayerChain(uint8_t* data, size_t dataLen, Packet* packet)
```
Dimi1010, line 258: This might be possible to be merged with `tryConstructNextLayer`?

Packet++/header/Layer.h
seladb, line 1: I'm not sure we need all 3 methods, I'd create just one:
```cpp
template <typename T, typename TFallback>
		Layer* tryConstructNextLayer(uint8_t* data, size_t dataLen, Packet* packet)
```

And enforce each layer to have a static `isDataValid` method. The reason some layers don't have it is because they were written before this concept evovled

Dimi1010: I think the `constructNextLayer` is useful in some situations, like the default statement of the switch constructing a payload layer directly. In my opinion, it's cleaner and safer than to have a manual assignment there when all the other code in the switch uses `tryConstructNextLayer`.

I did the `tryConstructNextLayer` recursively to allow potentially more than one fallback class. I can also see a situation where zero fallbacks are preferred and then a manual fallback construction if all the construction tries fail. 

T::isDataValid is still enforced on all layers tried as a next layer with this function, even the fallbacks, as they are all treated the same.

seladb: > I think the `constructNextLayer` is useful in some situations, like the default statement of the switch constructing a payload layer directly. In my opinion, it's cleaner and safer than to have a manual assignment there when all the other code in the switch uses `tryConstructNextLayer`.

I guess we can keep `constructNextLayer()`, my main concern is that people will be tempted to use it just because it's there instead of using `tryConstructNextLayer()` which is (almost) always the right method to use.


> I did the `tryConstructNextLayer` recursively to allow potentially more than one fallback class. I can also see a situation where zero fallbacks are preferred and then a manual fallback construction (via `constructNextLayer`) if all the construction tries fail.

I don't think we have a use-case of more than one fallback class (please correct me if I'm wrong). I'd make the `Layer` class API simpler and only require one fallback class. If we could, I'd even prefer not to require any fallback class and always use `PayloadLayer`, but this will create a circular dependency between the two classes so we can't do that. If we need more fallback classes in the future we can always create the recursive method

Dimi1010: > I don't think we have a use-case of more than one fallback class (please correct me if I'm wrong).

I originally thought about using it in something like this:
https://github.com/seladb/PcapPlusPlus/pull/1748/files#diff-3a0071b735e3e99d13ec73349061b34d89b143b2c91315b2584e5b497bdd312cR274-R291

Which would work after the changes to IPv4 and IPv6 `isDataValid` to check the version constants properly.
https://github.com/seladb/PcapPlusPlus/pull/1748/files#diff-58e1d5efbbfb9271a94beb63f7424b1fa8699c7fce35eef3d7d51609b272ccb1L629-R629
https://github.com/seladb/PcapPlusPlus/pull/1748/files#diff-77bd9e9cde4bf8f5aa68ec17947fd1956de7b96699fa1d3714ba0b13d81d52f6L247-R248

seladb: Hmm while it does seem to be a valid use-case, I'm still concerned about how it's going to be used. I don't have a super-strong argument against it, more like a gut feeling that it's not the right thing to do.

If you insist on keeping it I guess I'll agree 🙂

seladb: sure, sounds good 👍

Dimi1010: @seladb Refactored it. Ended up cutting the chain functionality altogether to allow for variable arguments passed to the layer constructor.

Packet++/header/IPLayer.h
```diff
@@ -30,5 +30,8 @@ namespace pcpp
 
 		/// An empty destructor
 		virtual ~IPLayer() = default;
+
+		/// A static method that gets a pointer to the IP data header and returns the IP version
```
seladb, line 34: Since it's a public method (although used internally), let's add the params and the return type to the doxygen documentation?
```cpp
/// A static method that gets a pointer to the IP data header and returns the IP version
/// @param[in] data ...
/// @param[in] dataLen ...
/// @return ...
```

Packet++/header/Layer.h
```diff
@@ -28,6 +30,22 @@ namespace pcpp
...
+		///
+		/// @tparam T The type to reinterpret as
+		/// @param data The data to check
+		/// @param dataLen The length of the data
+		/// @return True if the data is large enough to reinterpret as T, false otherwise
+		template <typename T> bool canReinterpretAs(const uint8_t* data, size_t dataLen)
+		{
+			return data != nullptr && dataLen >= sizeof(T);
+		}
+	}  // namespace internal
```
seladb, line 47: Maybe instead of a "standalone" internal function, it could be a protected static method in `Layer`?

Packet++/src/EthLayer.cpp
```diff
@@ -39,43 +39,38 @@ namespace pcpp
...
-			                  : static_cast<Layer*>(new PayloadLayer(payload, payloadLen, this, m_Packet));
+			tryConstructNextLayer<WakeOnLanLayer>(payload, payloadLen, m_Packet);
 			break;
-		default:
-			m_NextLayer = new PayloadLayer(payload, payloadLen, this, m_Packet);
 		}
+
+		// If no next layer was constructed, assume it's a payload layer
+		if (!hasNextLayer())
+			constructNextLayer<PayloadLayer>(payload, payloadLen, m_Packet);
```
seladb, line 73: My concern with this approach is that people might miss it in future PRs and we might also miss it in PR reviews. 

I'd prefer `tryConstructNextLayer` to be complete and take care of the fallback. It won't be pretty to always pass it `PayloadLayer` but I don't see a way around it...

Packet++/header/Layer.h
```diff
@@ -189,6 +207,54 @@ namespace pcpp
...
+		///
+		/// @tparam T The type of the layer to construct
+		/// @tparam Args The types of the extra arguments to pass to the layer constructor
+		/// @param[in] data The data to construct the layer from
+		/// @param[in] dataLen The length of the data
+		/// @param[in] packet The packet the layer belongs to
+		/// @param[in] extraArgs Extra arguments to be forwarded to the layer constructor
+		/// @return The constructed layer or nullptr if the data is invalid
+		template <typename T, typename... Args>
+		Layer* tryConstructNextLayer(uint8_t* data, size_t dataLen, Packet* packet, Args&&... extraArgs)
```
seladb, line 250: Does this method need to return the layer? I don't see we use it anywhere 🤔

Packet++/header/Layer.h
```diff
@@ -189,6 +207,54 @@ namespace pcpp
...
+		/// Construct the next layer in the protocol stack. No validation is performed on the data.
+		/// @tparam T The type of the layer to construct
+		/// @tparam Args The types of the arguments to pass to the layer constructor
+		/// @param[in] data The data to construct the layer from
+		/// @param[in] dataLen The length of the data
+		/// @param[in] packet The packet the layer belongs to
+		/// @param[in] extraArgs Extra arguments to be forwarded to the layer constructor
+		/// @return The constructed layer
+		template <typename T, typename... Args>
+		Layer* constructNextLayer(uint8_t* data, size_t dataLen, Packet* packet, Args&&... extraArgs)
```
seladb, line 225: ditto

Packet++/src/EthLayer.cpp
```diff
@@ -39,43 +39,38 @@ namespace pcpp
...
 		case PCPP_ETHERTYPE_IPV6:
-			m_NextLayer = IPv6Layer::isDataValid(payload, payloadLen)
-			                  ? static_cast<Layer*>(new IPv6Layer(payload, payloadLen, this, m_Packet))
-			                  : static_cast<Layer*>(new PayloadLayer(payload, payloadLen, this, m_Packet));
+			tryConstructNextLayer<IPv6Layer>(payload, payloadLen, m_Packet);
 			break;
 		case PCPP_ETHERTYPE_ARP:
-			m_NextLayer = new ArpLayer(payload, payloadLen, this, m_Packet);
+			// todo: change to tryConstructNextLayer
+			constructNextLayer<ArpLayer>(payload, payloadLen, m_Packet);
```
seladb, line 49: We already add `isDataValid` to some layers, why not add to others? 🤔

Packet++/src/IPv4Layer.cpp
seladb, line 1: There are more layers that parse the next layers, why modifying `EthLayer` and `IPv4Layer` in this PR and not the others? 🤔 

I agree that updating everything in one PR would create a huge and messy PR, maybe we can start with just `EthLayer`?

Dimi1010: Mostly because it was proof of concept, and those 2 had a high amount of "next" layers that can parse.

Packet++/header/Layer.h
Dimi1010: It can be used to check for nullptr for construction, so that is a way to check if the construction succeeded.

Packet++/header/Layer.h
Dimi1010: We can place it there, but this is more of a general utility function, no?

Packet++/src/EthLayer.cpp
Dimi1010: The issue now is that because some layers require extra params, the method needs to accept a parameter pack. Example:

`tryConstructNextLayer<VrrpV3Layer>(payload, payloadLen, m_Packet, IPAddress::IPv4AddressType);`

It's more of a design issue currently. If the method accepts a fallback layer, too, what do you do with the extra params? Do you pass them to the fallback, or do you discard them?

That is why I decided to cut the fallback and do a check afterward.

IMO, even though the check will need to be accounted for, that is offset by the switch statement internals having much cleaner code, making the read through straightforward.

Packet++/src/EthLayer.cpp
Dimi1010: Added 3bff18a3ab91e719bc807a85ed38c3a16906e3ac

Packet++/header/Layer.h
seladb: It's a general utility function that resides in `Layer.h` and is only used in the `Layer` classes 🙂 
I think it's fine to have it as a protected method of the `Layer` class

Packet++/src/EthLayer.cpp
seladb: I think we can decide not to pass the extra params to the fallback layer. We know it's always going to be `PayloadLayer` which doesn't have extra params anyway

Packet++/header/Layer.h
seladb: I guess we can leave the return value if you insist, although we never use it...

Packet++/src/IPv4Layer.cpp
seladb: ok 👍

Packet++/src/EthLayer.cpp
Dimi1010: I guess we can do that.

Although, if we have a fallback explicitly built-in then it isn't really a `tryConstruct`, imo. :/

In all the APIs I have seen `tryX` methods, the primary convention of these methods is no-op on failure. They usually attempt the operation and if they fail they just report the failure. They don't usually execute an alternative action. For example, say a `tryAdd` method, it will attempt to add the element, and if it can't, it will return `false`, it won't add another element.

The `tryConstruct` chain earlier was called that more because it tries the layers in a chain, until the first one succeeds., rather than providing an explicit fallback layer mechanism. If none of the layers had passed validation there, it would have still been a no-op.

Packet++/header/Layer.h
Dimi1010: Fixed f6e1529fa4d1eb83b773190f80fee5025e7c3609

Packet++/header/Layer.h
Dimi1010: There are some situations that can be useful, like single check `tryConstruct` with following fallback.

Example from CotpLayer:
```
if (S7CommLayer::isDataValid(payload, payloadLen))
	m_NextLayer = new S7CommLayer(payload, payloadLen, this, m_Packet);
else
```

```
if (!tryConstructNextLayer<S7CommLayer>(payload, payloadLen, m_Packet))
	constructNextLayer<PayloadLayer>(payload, payloadLen, m_Packet);
```

Packet++/src/EthLayer.cpp
seladb: I agree that the terminology is not accurate. Maybe we can rename the method to be:
```cpp
template <typename T, typename TFallback, typename... Args>
Layer* constructNextLayer(uint8_t* data, size_t dataLen, Packet* packet, Args&&... extraArgs);
```

We can move the existing `tryConstructNextLayer` to be a private member of `Layer` and have `constructNextLayer` use it. Let me know what you think

Packet++/header/Layer.h
seladb: If we decide to go with the approach discussed [here](https://github.com/seladb/PcapPlusPlus/pull/1748#discussion_r2026144949), then it's probably less relevant?

Packet++/src/EthLayer.cpp
Dimi1010: Ended up with `tryConstructNextLayerWithFallback` to distinguish from the standard `tryConstruct` method and expectation. The existing `construct` method performs no data validation, so having an overload with validation with might be confusing. 

What do you think?

Packet++/header/Layer.h
```diff
@@ -189,6 +191,93 @@ namespace pcpp
...
+		///
+		/// @tparam T The type of the layer to construct
+		/// @tparam Args The types of the extra arguments to pass to the layer constructor
+		/// @param[in] data The data to construct the layer from
+		/// @param[in] dataLen The length of the data
+		/// @param[in] packet The packet the layer belongs to
+		/// @param[in] extraArgs Extra arguments to be forwarded to the layer constructor
+		/// @return The constructed layer or nullptr if the data is invalid
+		template <typename T, typename... Args>
+		Layer* tryConstructNextLayer(uint8_t* data, size_t dataLen, Packet* packet, Args&&... extraArgs)
```
seladb, line 234: We probably don't want people to use this method so let's move it to be private?

Dimi1010: Fixed 954f8e8

Packet++/header/ArpLayer.h
```diff
@@ -271,6 +271,11 @@ namespace pcpp
 		{
 			return OsiModelNetworkLayer;
 		}
+
+		static bool isDataValid(const uint8_t* data, size_t dataLen)
+		{
+			return canReinterpretAs<arphdr>(data, dataLen);
+		}
```
seladb, line 278: Let's add doxygen documentation?

Same for the rest of the `isDataValid` methods that were added in the other layers.
I'd suggest a full doxygen documentation that includes the params:

```cpp
/// A static method that takes a byte array and detects whether it is a valid XXXX
/// @param[in] data A byte array
/// @param[in] dataSize The byte array size (in bytes)
/// @return True if the data looks like a valid XXXX layer
static bool isDataValid(const uint8_t* data, size_t dataSize);
```

Packet++/header/Layer.h
```diff
@@ -189,6 +191,94 @@ namespace pcpp
...
+		}
+
+		/// Construct the next layer in the protocol stack. No validation is performed on the data.
+		/// @tparam T The type of the layer to construct
+		/// @tparam Args The types of the arguments to pass to the layer constructor
+		/// @param[in] data The data to construct the layer from
+		/// @param[in] dataLen The length of the data
+		/// @param[in] packet The packet the layer belongs to
+		/// @param[in] extraArgs Extra arguments to be forwarded to the layer constructor
+		/// @return The constructed layer
```
seladb, line 207: nit: private/protected methods don't need doxygen documentation (applies to this method and the other new methods below)

Packet++/header/Layer.h
```diff
@@ -189,6 +191,94 @@ namespace pcpp
...
+		Layer* tryConstructNextLayerWithFallback(uint8_t* data, size_t dataLen, Packet* packet, Args&&... extraArgs)
+		{
+			if (tryConstructNextLayer<T>(data, dataLen, packet, std::forward<Args>(extraArgs)...))
+			{
+				return m_NextLayer;
+			}
+			else
+			{
+				return constructNextLayer<TFallback>(data, dataLen, packet);
+			}
```
seladb, line 244: very nit: the `else` clause is not needed:
```cpp
if (tryConstructNextLayer<T>(data, dataLen, packet, std::forward<Args>(extraArgs)...))
{
	return m_NextLayer;
}

return constructNextLayer<TFallback>(data, dataLen, packet);
```

Packet++/src/IPv4Layer.cpp
```diff
@@ -248,104 +248,108 @@ namespace pcpp
...
-			ipVersion = *payload >> 4;
-			if (ipVersion == 4 && IPv4Layer::isDataValid(payload, payloadLen))
-				m_NextLayer = new IPv4Layer(payload, payloadLen, this, m_Packet);
-			else if (ipVersion == 6 && IPv6Layer::isDataValid(payload, payloadLen))
-				m_NextLayer = new IPv6Layer(payload, payloadLen, this, m_Packet);
-			else
-				m_NextLayer = new PayloadLayer(payload, payloadLen, this, m_Packet);
+		{
+			// todo: no tests for this case
+			switch (IPLayer::getIPVersion(payload, payloadLen))
```
seladb, line 276: Let's add a `default` clause that generates a `PayloadLayer`?

Packet++/src/IPv4Layer.cpp
```diff
@@ -248,104 +248,108 @@ namespace pcpp
...
 		case PACKETPP_IPPROTO_GRE:
-			greVer = GreLayer::getGREVersion(payload, payloadLen);
-			if (greVer == GREv0 && GREv0Layer::isDataValid(payload, payloadLen))
-				m_NextLayer = new GREv0Layer(payload, payloadLen, this, m_Packet);
-			else if (greVer == GREv1 && GREv1Layer::isDataValid(payload, payloadLen))
-				m_NextLayer = new GREv1Layer(payload, payloadLen, this, m_Packet);
-			else
-				m_NextLayer = new PayloadLayer(payload, payloadLen, this, m_Packet);
+		{
+			switch (GreLayer::getGREVersion(payload, payloadLen))
```
seladb, line 289: ditto: Let's add a `default` clause that generates a `PayloadLayer`?

Packet++/src/IPv4Layer.cpp
```diff
@@ -248,104 +248,108 @@ namespace pcpp
...
+			bool igmpQuery = false;
+			ProtocolType igmpVer = IgmpLayer::getIGMPVerFromData(
 			    payload, std::min<size_t>(payloadLen, be16toh(getIPv4Header()->totalLength) - hdrLen), igmpQuery);
-			if (igmpVer == IGMPv1)
-				m_NextLayer = new IgmpV1Layer(payload, payloadLen, this, m_Packet);
-			else if (igmpVer == IGMPv2)
-				m_NextLayer = new IgmpV2Layer(payload, payloadLen, this, m_Packet);
-			else if (igmpVer == IGMPv3)
+
+			switch (igmpVer)
```
seladb, line 306: ditto

Packet++/src/IPv4Layer.cpp
```diff
@@ -248,104 +248,108 @@ namespace pcpp
...
 		case PACKETPP_IPPROTO_VRRP:
 		{
-			auto vrrpVer = VrrpLayer::getVersionFromData(payload, payloadLen);
-			if (vrrpVer == VRRPv2)
-				m_NextLayer = new VrrpV2Layer(payload, payloadLen, this, m_Packet);
-			else if (vrrpVer == VRRPv3)
-				m_NextLayer = new VrrpV3Layer(payload, payloadLen, this, m_Packet, IPAddress::IPv4AddressType);
-			else
-				m_NextLayer = new PayloadLayer(payload, payloadLen, this, m_Packet);
+			switch (VrrpLayer::getVersionFromData(payload, payloadLen))
```
seladb, line 336: ditto

Packet++/header/Layer.h
Dimi1010: Sure, don't need them. Is there a drawback to having them if they are written already, tho?

seladb: no real drawback, just fewer lines of code 🙂

Dimi1010: I think having docstrings on some of the non-trivial protected/private methods might be beneficial?

seladb: it's a nit comment, so it's up to you to decide, and I won't delay approving the PR because of it. So whatever you decide 🙂

Packet++/header/ArpLayer.h
Dimi1010: Added 9f95e8e

Packet++/header/Layer.h
Dimi1010: Fixed ee4899c1996cc63b8c9629b9196053a72933eb3a

Packet++/src/IPv4Layer.cpp
Dimi1010: Fixed a634fa45657bf6ecdf84437dd55c516f1a0eccd6

Packet++/src/IPv4Layer.cpp
Dimi1010: Fixed a634fa45657bf6ecdf84437dd55c516f1a0eccd6

Packet++/src/IPv4Layer.cpp
Dimi1010: Fixed a634fa45657bf6ecdf84437dd55c516f1a0eccd6

Packet++/src/IPv4Layer.cpp
Dimi1010: Fixed a634fa45657bf6ecdf84437dd55c516f1a0eccd6

Packet++/src/IPv4Layer.cpp
```diff
@@ -248,104 +248,120 @@ namespace pcpp
...
+			}
 			break;
 		}
-		default:
-			m_NextLayer = new PayloadLayer(payload, payloadLen, this, m_Packet);
 		}
+
+		// If no next layer was constructed, assume it's a payload layer
+		if (!hasNextLayer())
+			constructNextLayer<PayloadLayer>(payload, payloadLen, m_Packet);
```
seladb, line 364: Instead of this `if` clause, maybe we can add a `default` clause to the main `switch`?

Packet++/header/Layer.h
```diff
@@ -189,6 +191,92 @@ namespace pcpp
...
+		/// Construct the next layer in the protocol stack. No validation is performed on the data.
+		/// @tparam T The type of the layer to construct
+		/// @tparam Args The types of the arguments to pass to the layer constructor
+		/// @param[in] data The data to construct the layer from
+		/// @param[in] dataLen The length of the data
+		/// @param[in] packet The packet the layer belongs to
+		/// @param[in] extraArgs Extra arguments to be forwarded to the layer constructor
+		/// @return The constructed layer
+		template <typename T, typename... Args>
+		Layer* constructNextLayer(uint8_t* data, size_t dataLen, Packet* packet, Args&&... extraArgs)
```
seladb, line 209: I think `data` can be of type `const uint_8*`. The same in the rest of the methods: `tryConstructNextLayerWithFallback`, `tryConstructNextLayer`

seladb (approved):
A few more small comments, otherwise LGTM

Packet++/header/Layer.h
Dimi1010: Layer ctor takes a non-const uint8_t*

Packet++/src/IPv4Layer.cpp
Dimi1010: I think its better to have the if clause standalone, as that can catch if any of the other cases did not construct a layer somehow.

seladb: Ok I guess we can keep it

Issue #1711, opened 2025-02-10
Segmentation fault (Heap use after free) in GtpV1Layer.addExtension()


RShvadron:
### Bug description

**Describe the bug**
When adding a new GTPv1 extension to a GTPv1 layer object (that was not assigned to a packet yet) with the GtpV1Layer::addExtension() method I'm receiving SEG Fault when accessing the header pointer (header->extensionHeaderFlag = 1) that was assigned at function start (GtpLayer.cpp:418 for git tag v24.09).
After a short debug it seems that you guys are deleting the original header buffer and allocating a new one at Layer::extendLayer() but forgot to update the header pointer to point into the new memory that was allocated. 

**Code example to reproduce**
	uint32_t teid = 0x12345678; /* Tunnel endpoint identifier */
	uint8_t ext_type = 0x85;
	uint8_t msg_type = 0xFF; /* G-PDU (255) */
	uint16_t ext_content = 0x1234;
	pcpp::GtpV1Layer *gtp_layer = new pcpp::GtpV1Layer(pcpp::GtpV1MessageType::GtpV1_VersionNotSupported, teid, true, 1, false, 0);
	gtp_layer->getHeader()->messageType = msg_type;
	pcpp::GtpV1Layer::GtpExtension ext = gtp_layer->addExtension(ext_type, ext_content);

        Segmentation fault is being thrown on the above line.

**Expected behavior**
No segmentation...

### PcapPlusPlus versions tested on

PcapPlusPlus master branch

### Operating systems tested on

Linux, Other (specify in the next question)

### Other operation systems (if applicable)

Ubuntu 22.04 LTS

### Compiler version

GCC 11.2.0

seladb:
Thanks for reporting this issue @RShvadron ! Fixed in #1749

PR #1749, opened 2025-03-28
Fix segmentation fault in `GtpV1Layer.addExtension()`

seladb:
Fix issue https://github.com/seladb/PcapPlusPlus/issues/1711


Packet++/src/GtpLayer.cpp
```diff
@@ -415,6 +414,9 @@ namespace pcpp
 		else
 		{
 			// mark extension flags in the layer
+			header = getHeader();
+			headerExtra = getHeaderExtra();
```
Dimi1010, line 418: I assume we don't set the `header`/`headerExtra` in the other branch because the extension is metadata is stored in the previous extension item, right?

Dimi1010 (approved):
LGTM

Packet++/src/GtpLayer.cpp
seladb: I'm not sure I understand the question...

The reason we fetch the `header` and `headerExtra` here is because we previously used `extendLayer()` which changed `m_Data` under the hood, so the previous pointers aren't valid anymore

PR #1752, opened 2025-03-28
Moved `setFilter` pcap implementation to `PcapHandle`.

Dimi1010:
This PR moves the implementation of `IPcapDevice::setFilter` to `PcapHandle::setFilter` in a move to encapsulate common pcap operations on the handle under the `PcapHandle` wrapper.


Pcap++/header/PcapDevice.h
```diff
@@ -138,10 +150,10 @@ namespace pcpp
 		/// @param[in] filterAsString The filter to be set in Berkeley Packet Filter (BPF) syntax
 		/// (http://biot.com/capstats/bpf.html)
 		/// @return True if filter set successfully, false otherwise
-		virtual bool setFilter(std::string filterAsString);
+		bool setFilter(std::string filterAsString) override;
```
tigercosmos, line 153: cpp signature is `bool PcapHandle::setFilter(std::string const& filter)`?

Dimi1010: Yes. What is the question here?

`IFilterableDevice` has the filter as a value param in the pure virtual. So the override must match.

tigercosmos: Sorry, I meant the `const`. You put `const&` in the cpp, but not in the header file.

tigercosmos: ops, my bad.

tigercosmos (approved):

seladb (approved):

PR #1753, opened 2025-03-28
Refactored PcapHandle to utilize std::unique_ptr internally.

Dimi1010:


Pcap++/header/PcapDevice.h
Dimi1010, line 1: Should we have an `/internal/` folder for headers that need to be public for compilation but shouldn't be included by users directly?

Dimi1010: It feels cleaner to have the "internal" headers separated. That way, we don't have to worry about breaking changes if files are renamed or moved around due to people directly including them. And if they still do, that is on them.

tigercosmos: Well, I agree with you. Just think we can put effort on more important parts (if there are such things)

Dimi1010: Eh, it's possibly not the .ost important. But since this PR exposes a new public header (`PcapUtils.h`), that only defines internal symbols, I would prefer to have that now instead of maybe having to do a transition later, when it has already been placed next to the user API headers.

tigercosmos: Got it. I don't have strong opion on this. Let see how @seladb  says.

Pcap++/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ set(
   header/PcapFilter.h
   header/PcapLiveDevice.h
   header/PcapLiveDeviceList.h
+  header/PcapUtils.h
```
seladb, line 37: I think we should avoid exposing internal header files like `PcapUtils.h`. 

The reason it is needed is because we use `internal::PcapCloseDeleter` inside a "public" header file, and we should find a way around it.

I tried using forward declaration of `internal::PcapCloseDeleter` but couldn't make it work so far. @Dimi1010 maybe you can make it work?

Other options are:
- Move `internal::PcapCloseDeleter` to `PcapDevice.h` - not ideal, but I think it'll work
- Define `m_PcapDescriptor` as "no type" like `void*` and assign the type in the `.cpp` file. However it's a pretty ugly workaround

Dimi1010: See this as a potential solution. I think this is the cleanest, tbh.
https://github.com/seladb/PcapPlusPlus/pull/1753#discussion_r2018533325

`internal::PcapCloseDeleter` can't work as a forward declaration here as it is default constructed in the `std::unique_ptr` so it needs a concrete type.

seladb: I think exposing an `internal` folder of headers is not a good solution either because exposing header files just because we have some technical limitations is not something we want. I think we should either find a workaround (even if it's not very nice), or just leave the code as it is

Dimi1010: I have seen that pattern be used to separate the stable public headers from the unstable ones, that can change at any moment. Most notably I have seen it in many boost libraries ("details" folders).

Moving `internal::PcapCloseDeleter` to `PcapDevice.h` creates a dependency between `PcapFilter.cpp` and `PcapDevice.h`, which I would prefer to not have.

I guess we can leave this as is for now tho.

seladb: If it was "stable" vs "unstable" public API I'd agree that separating them into different folders could work. But exposing a "private API" (internal code) is probably not the best way to go.

I guess leaving it as is is fine if we can find a better solution...

Issue #1762, opened 2025-04-02
Add Cmake switch to compile only Common++ and Packet++.


Dimi1010:
`Common++` and `Packet++` do not have required external dependencies unlike `Pcap++`.

The new CMake switch should do the following:
- Skip the check for `pcap` dependency.
- Exclude the `Pcap++` subproject from the compilation.
- Default behaviour should be configured to build all targets `Common++`, `Packet++`, `Pcap++`.

 _Originally posted by @seladb in [#1744](https://github.com/seladb/PcapPlusPlus/issues/1744#issuecomment-2770595198)_

PR #1763, opened 2025-04-02
Added cmake switch to control the building of Pcap++ subproject.

Dimi1010:
Implements #1762 

- Added new switch `PCAPPP_BUILD_PCAPPP` to enable `Pcap++` target.
- Refactored the root CMakeLists.txt to skip `Pcap++` related dependencies when building `Pcap++` is disabled.
- Temporarily disabled building of examples and tutorials, if building `Pcap++` is disabled.
- Disabled building of `Pcap++Test` if `Pcap++` is disabled.


Dimi1010:
Might also be a good idea to add a CI that builds with this switch turned off?

seladb:
@clementperon do you think you'll be able to review this PR? You wrote most of the CMake files for this project and you're definitely more expert than me on CMake

CMakeLists.txt
```diff
@@ -111,6 +111,7 @@ if(PCAPPP_LOG_LEVEL)
 endif()
 
 # Build options (Turn on Examples and Tests if it's the main project)
+option(PCAPPP_BUILD_PCAPPP "Build Pcap++ project. Turning this off will build only Common++ and Packet++" ON)
```
clementperon, line 114: Would be nice to have a test in CI that build with this option turn to off

CMakeLists.txt
```diff
@@ -247,20 +176,93 @@ if(Git_FOUND)
...
 endif()
 
-if(PCAPPP_USE_XDP)
-  find_package(BPF)
-  if(NOT BPF_FOUND)
-    message(FATAL_ERROR "libbpf not found!")
+# Required dependencies only for PCap++ project.
+if(PCAPPP_BUILD_PCAPPP)
+  # Usually on Windows PCAP_ROOT and Packet_ROOT are at the same location
+  if(WIN32 AND PCAP_ROOT AND NOT Packet_ROOT)
```
clementperon, line 188: If all this is only required for Pcap++ maybe it should be move to Pcap++/CMakelists.txt no ?

CMakeLists.txt
clementperon: @Dimi1010 just saw your comment, I agree then yes :)

CMakeLists.txt
Dimi1010: Generally, I agree that is should probably be moved there if its exclusive to that module.

Do you think it better to be moved in this PR or another one?

clementperon: Fine for me to make it in another PR as it was already here at the moment

clementperon (approved):

tigercosmos (approved):
LGTM

CMakeLists.txt
Dimi1010: Do we want one for all OS types, or just a single ubuntu CI enough?

seladb: I think a single Ubuntu test should be sufficient 🤔

seladb: @Dimi1010 I think this is the last comment to address to merge this PR, do you think you'll have time to work on it?

Dimi1010: If I figure out how the CI works, sure. I have very little experience with Github's CI.

seladb: I can help you if neeed, please let me know

Dimi1010: @seladb  Added the new CI to the linux matrix. If you could check it?

ci/run_tests/run_tests.py
```diff
@@ -32,39 +86,20 @@ def main():
     )
     args = parser.parse_args()
 
-    ip_address = get_if_addr(args.interface)
-
-    print("IP address is: %s" % ip_address)
+    for test_suite in args.test_suites:
+        if test_suite not in ("packet", "pcap"):
+            print(f"Unknown test suite: {test_suite}")
```
seladb, line 91: If we add the `choices` param to `parser.add_argument` we don't need this check:
```python
parser.add_argument(
    "--test-suites",
    ...
    choices=["packet", "pcap"],
    ...
```

ci/run_tests/run_tests.py
```diff
@@ -1,17 +1,71 @@
...
+        tcpreplay_proc.kill()
+
+
+def run_packet_tests(args: list[str], use_sudo: bool):
+    cmd_line = []
+    if use_sudo:
+        cmd_line += ["sudo"]
+
+    cmd_line += [os.path.join("Bin", "Packet++Test")]
+    cmd_line += args
```
seladb, line 30: nit: this is a bit nicer:
```python
cmd_line = ["sudo"] if use_sudo else []
cmd_line += [os.path.join("Bin", "Packet++Test"), *args]
```

ci/run_tests/run_tests.py
```diff
@@ -1,17 +1,71 @@
...
+def run_pcap_tests(interface: str, tcpreplay_dir: str, args: list[str], use_sudo: bool):
+    ip_address = get_if_addr(interface)
+    print(f"IP address is: {ip_address}")
+
+    with tcp_replay_worker(interface, tcpreplay_dir):
+        cmd_line = []
+        if use_sudo:
+            cmd_line += ["sudo"]
+        cmd_line += [os.path.join("Bin", "Pcap++Test"), "-i", ip_address]
+        cmd_line += args
```
seladb, line 47: ditto:
```python
cmd_line = ["sudo"] if use_sudo else []
cmd_line += [os.path.join("Bin", "Pcap++Test"), "-i", ip_address, *args]
```

ci/run_tests/run_tests.py
Dimi1010: Fixed c32039b

ci/run_tests/run_tests.py
Dimi1010: Fixed 5a9ab70

ci/run_tests/run_tests.py
Dimi1010: Fixed 5a9ab70

.github/workflows/build_and_test.yml
```diff
@@ -130,9 +135,10 @@ jobs:
       - name: Test PcapPlusPlus
         run: |
           . .venv/bin/activate
-          python3 ci/run_tests/run_tests.py --interface eth0 ${{ matrix.test-flags }}
+          python3 ci/run_tests/run_tests.py --interface eth0 ${{ matrix.additional-test-flags }}
 
       - name: Test Examples
+        if: (!contains(matrix.additional-flags, '-DPCAPPP_BUILD_PCAPPP=OFF'))
```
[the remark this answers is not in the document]
Dimi1010: Fixed 8d14fcd

.github/workflows/build_and_test.yml
```diff
@@ -175,6 +175,82 @@ jobs:
           path: ${{ env.CCACHE_DIR }}
           key: ${{ steps.ccache-restore.outputs.cache-primary-key }}
 
+  onlypacket:
```
seladb, line 178: nit: I'd call it `packet-only`

.github/workflows/build_and_test.yml
```diff
@@ -175,6 +175,82 @@ jobs:
...
+    runs-on: ubuntu-latest
+    container: seladb/${{ matrix.image }}
+    strategy:
+      matrix:
+        include:
+          - image: ubuntu2404
+            config-zstd: OFF
+            additional-flags: -DPCAPPP_BUILD_PCAPPP=OFF
+            additional-test-flags: --test-suites "packet"
```
seladb, line 187: If we only want to run one image, there's no need for `strategy` and `matrix`. It can be simpler, and we don't need `config-zstd` (which defaults to `OFF`), `additional-flags` and `additional-test-flags`

Dimi1010: So, should usages of those just be hardcoded/removed from the workflow steps then? There are a couple of places where `{matrix.image}` are used?

seladb: Instead of `{matrix.image}` we can use an `env`:
```yml
env:
  IMAGE: ubuntu2404
steps:
  ...

  - name: Upload Coverage Results
    uses: codecov/codecov-action@0565863a31f2c772f9f0395002a31e3f06189574 # v5.4.0
    with:
      files: ./coverage.xml
      flags: ${{ env.IMAGE }},unittest
      fail_ci_if_error: false
      verbose: true
    env:
      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```

.github/workflows/build_and_test.yml
Dimi1010: Fixed 1c04601

.github/workflows/build_and_test.yml
Dimi1010: Should be fixed now.

seladb (approved):

PR #1767, opened 2025-04-04
Change Cmake to use target_compiler_features than CMAKE_CXX_STANDARD to define cpp standard.

Dimi1010:
- Added `target_compiler_features` clause to `Common++`, `Packet++` and `Pcap++`. Linking projects will automatically select a CPP standard that satisfies all required features by dependent targets.
- Made `CMAKE_CXX_STANDARD` optionally defined only if not set externally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,8 +139,12 @@ option(PCAPPP_USE_XDP "Setup PcapPlusPlus with XDP")
 option(PCAPPP_INSTALL "Install Pcap++" ${PCAPPP_MAIN_PROJECT})
 option(PCAPPP_PACKAGE "Package Pcap++ could require a recent version of CMake" OFF)
 
-# Set C++11
-set(CMAKE_CXX_STANDARD 11)
+# Set C++11 if not defined
+if(NOT DEFINED CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+endif()
+
 # popen()/pclose() are not C++ standards
 set(CMAKE_CXX_EXTENSIONS ON)
 # Set Position Independent Code for static libraries
--- a/Common++/CMakeLists.txt
+++ b/Common++/CMakeLists.txt
@@ -34,6 +34,8 @@ set(
 # Set the public header that will be installed
 set_property(TARGET Common++ PROPERTY PUBLIC_HEADER ${public_headers})
 
+target_compile_features(Common++ PUBLIC cxx_std_11)
+
 target_include_directories(
   Common++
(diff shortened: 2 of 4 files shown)


clementperon:
@Dimi1010 please explain the motivation or issue you face with the actual implementation please

CMakeLists.txt
```diff
@@ -139,8 +139,12 @@ option(PCAPPP_USE_XDP "Setup PcapPlusPlus with XDP")
 option(PCAPPP_INSTALL "Install Pcap++" ${PCAPPP_MAIN_PROJECT})
 option(PCAPPP_PACKAGE "Package Pcap++ could require a recent version of CMake" OFF)
 
-# Set C++11
-set(CMAKE_CXX_STANDARD 11)
+# Set C++11 if not defined
+if(NOT DEFINED CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
clementperon, line 145: Why is it needed ?

clementperon (approved):

Dimi1010:
> @Dimi1010 please explain the motivation or issue you face with the actual implementation please

Sorry, for late response. 

The main idea of the PR is that it encodes the required c++ standard into the targets themselves. The way I understand it, it can be configured to propagate the required standard based on CMake target linking. For example, if target A has `cxx_std_11` compiler feature, set to `PUBLIC`/`INTERFACE`, targets linking to targets A will also inherit that requirement, configuring cmake to compile that target with cpp11 flags, provided no higher standard is required by the linking target itself or another linked target.

clementperon:
> > @Dimi1010 please explain the motivation or issue you face with the actual implementation please
> 
> Sorry, for late response.
> 
> The main idea of the PR is that it encodes the required c++ standard into the targets themselves. The way I understand it, it can be configured to propagate the required standard based on CMake target linking. For example, if target A has `cxx_std_11` compiler feature, set to `PUBLIC`/`INTERFACE`, targets linking to targets A will also inherit that requirement, configuring cmake to compile the linking target with cpp11 flags, provided no higher standard is required by the linking target itself or another linked target.

So no issue, just to make it proper right?

Dimi1010:
> So no issue, just to make it proper right?

@clementperon Yes, it is just a refactor. There is no specific issue I am having with the current implementation.

CMakeLists.txt
Dimi1010: Mostly so it does not override the global standard if the library is imported as a subdirectory to another project during the build process.

The override will happen implicitly if needed via the compiler feature directives on the targets, but it should be fine to compile with higher standard if available, no?

seladb (approved):

PR #1772, opened 2025-04-13
Remove libpcap checks for options available since libpcap >=0.9

seladb:
- Remove libpcap options: `PCAPPP_ENABLE_PCAP_IMMEDIATE_MODE` and `PCAPPP_ENABLE_PCAP_IMMEDIATE_MODE` and enable them automatically if supported by libpcap/WinPcap/Npcap
- Remove the check for `pcap_setdirection` because it's supported since libpcap 0.9 so should be supported on all platforms
- Remove the check for `pcap_lib_version` because it's supported since libpcap 0.8 so should be supported on all platforms


Dimi1010 (approved):
LGTM

clementperon (approved):

clementperon:
@seladb I agree but the MR title is not really intuituive: "Remove redundant CMake options"

I would more say : "Remove libpcap check for options available since libpcap <=0.9"

seladb:
> @seladb I agree but the MR title is not really intuituive: "Remove redundant CMake options"
> 
> I would more say : "Remove libpcap check for options available since libpcap <=0.9"

Thanks @clementperon ! I updated the PR title

PR #1773, opened 2025-04-14
Removed excessive use of std::endl from examples.

Dimi1010:
Removed excessive use of `std::endl` in examples.

Each `std::endl` call flushes the stream in addition to appending a new line ('\n') character if the stream is buffered (`std::cout`). Usages of it in the middle of a long string block can cause performance overhead where the stream is repeatedly flushed without a need for it. 

This PR addresses this issue by replacing calls to `std::endl` in the middle of string blocks with '\n'. In addition the PR groups adjacent C-string literals, by removing the `<<` operators between them. Under the C++ standard, adjacently declared C-string literals are packed into a single C-string during compilation.

### DIFF
--- a/Examples/IPFragUtil/main.cpp
+++ b/Examples/IPFragUtil/main.cpp
@@ -84,9 +84,8 @@ void printUsage()
  */
 void printAppVersion()
 {
-	std::cout << pcpp::AppName::get() << " " << pcpp::getPcapPlusPlusVersionFull() << std::endl
-	          << "Built: " << pcpp::getBuildDateTime() << std::endl
-	          << "Built from: " << pcpp::getGitInfo() << std::endl;
+	std::cout << pcpp::AppName::get() << " " << pcpp::getPcapPlusPlusVersionFull()
+	          << "\nBuilt: " << pcpp::getBuildDateTime() << "\nBuilt from: " << pcpp::getGitInfo() << std::endl;
 	exit(0);
 }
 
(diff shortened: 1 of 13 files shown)


Examples/ArpSpoofing/main.cpp
```diff
@@ -33,18 +33,16 @@ static struct option L3FwdOptions[] = {
...
-	          << pcpp::AppName::get() << " [-hv] -i interface_ip -c victim_ip -g gateway_ip" << std::endl
-	          << std::endl
-	          << "Options:" << std::endl
-	          << std::endl
-	          << "    -i interface_ip   : The IPv4 address of interface to use" << std::endl
-	          << "    -c victim_ip      : The IPv4 address of the victim" << std::endl
-	          << "    -g gateway_ip     : The IPv4 address of the gateway" << std::endl
-	          << "    -h                : Displays this help message and exits" << std::endl
-	          << "    -v                : Displays the current version and exists" << std::endl
+	std::cout << "\nUsage:"
```
tigercosmos, line 36: For the strings like here, the performace doesn't matter actually. Personally, I prefer the original `std::endl` which looks clearer.

tigercosmos:
I looked over the PR. It seems that there is no permance intensive code that `std::endl` can slow down the performance. Most of the cases are just printing the info or usages. I would suggest keeping the original code.

Examples/ArpSpoofing/main.cpp
Dimi1010: I find the '\n' version significantly cleaner to read and visualize how the string will be formatted on the console.

Issue #1770, opened 2025-04-12
Add support for DPDK 23.11 (LTS) and 24.11 (LTS)


seladb:
https://core.dpdk.org/download/

We can also drop support for DPDK 19.11 and 20.11

PR #1774, opened 2025-04-16
Few fixes to support DPDK 23.11 and 24.11

seladb:
The first PR to address issue https://github.com/seladb/PcapPlusPlus/issues/1770 (there'll be a follow-up PR).

I'm working on adding DPDK 23.11 and 24.11. These fixes are needed to support these versions. They should all be backward-compatible and work on older DPDK versions as well.

Once this PR is merged, I'll add Docker images for DPDK 23.11 and 24.11 (here is a draft PR for it: https://github.com/seladb/PcapPlusPlus-DockerImages/pull/49) and open another PR to update our CI with the new images.

### DIFF
--- a/Pcap++/src/DpdkDevice.cpp
+++ b/Pcap++/src/DpdkDevice.cpp
@@ -105,6 +105,18 @@ namespace pcpp
 		0x6D, 0x5A, 0x6D, 0x5A, 0x6D, 0x5A, 0x6D, 0x5A, 0x6D, 0x5A, 0x6D, 0x5A,
 	};
 
+	static bool getDeviceInfo(uint16_t devId, rte_eth_dev_info& devInfo)
+	{
+		auto ret = rte_eth_dev_info_get(devId, &devInfo);
+		if (ret < 0)
+		{
+			PCPP_LOG_ERROR("Couldn't get device info, error was: " << rte_strerror(ret) << " (" << ret << ")");
+			return false;
+		}
+
+		return true;
+	}
+
 	DpdkDevice::DpdkDevice(int port, uint32_t mBufPoolSize, uint16_t mBufDataSize)
 	    : m_Id(port), m_MacAddress(MacAddress::Zero),
 	      m_MBufDataSize(mBufDataSize < 1 ? RTE_MBUF_DEFAULT_BUF_SIZE : mBufDataSize)
@@ -348,10 +360,8 @@ namespace pcpp
 	bool DpdkDevice::initQueues(uint8_t numOfRxQueuesToInit, uint8_t numOfTxQueuesToInit)
 	{
 		rte_eth_dev_info devInfo;
-		auto ret = rte_eth_dev_info_get(m_Id, &devInfo);
-		if (ret < 0)
+		if (!getDeviceInfo(m_Id, devInfo))
 		{
-			PCPP_LOG_ERROR("Couldn't get device info, error was: " << rte_strerror(ret) << " (" << ret << ")");
 			return false;
 		}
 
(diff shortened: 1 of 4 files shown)


Pcap++/src/DpdkDevice.cpp
```diff
@@ -273,6 +273,8 @@ namespace pcpp
 		{
 			PCPP_LOG_DEBUG("PMD '" << m_PMDName << "' doesn't support RSS, setting RSS hash functions to 0");
 			m_Config.rssHashFunction = RSS_NONE;
+			m_Config.rssKey = nullptr;
+			m_Config.rssKeyLength = 0;
```
seladb, line 277: In DPDK 23.11/24.11 it won't let setting the RSS has function to `RSS_NONE` but have something in `rssKey`

Pcap++/src/DpdkDevice.cpp
```diff
@@ -346,7 +348,11 @@ namespace pcpp
 	bool DpdkDevice::initQueues(uint8_t numOfRxQueuesToInit, uint8_t numOfTxQueuesToInit)
 	{
 		rte_eth_dev_info devInfo;
-		rte_eth_dev_info_get(m_Id, &devInfo);
+		if (rte_eth_dev_info_get(m_Id, &devInfo) < 0)
```
seladb, line 351: In recent DPDK 24.11 they added [`__rte_warn_unused_result`](https://github.com/DPDK/dpdk/blob/ca1690ebd224f148268285b15b97441ccdbdd07e/lib/ethdev/rte_ethdev.h#L3538) to `rte_eth_dev_info_get` and `rte_eth_link_get` so we have to consider the return value

Pcap++/src/MBufRawPacket.cpp
```diff
@@ -333,9 +333,9 @@ namespace pcpp
 			return;
 		}
 
-		m_MBuf = mBuf;
 		RawPacket::setRawData(rte_pktmbuf_mtod(mBuf, const uint8_t*), rte_pktmbuf_pkt_len(mBuf), timestamp,
 		                      LINKTYPE_ETHERNET);
+		m_MBuf = mBuf;
```
seladb, line 338: This was a bug that was introduced somewhere, not sure where: when calling `RawPacket::setRawData()` it calls `MBufRawPacket::clear()` which sets `m_MBuf` to `nullptr`, so we have to set the mbuf after calling `RawPacket::setRawData()`, otherwise we get a segmentation fault

seladb:
@clementperon @egecetin @tigercosmos @Dimi1010 can one of you review this PR?

Pcap++/src/DpdkDevice.cpp
clementperon: maybe add PCPP_LOG_ERROR no ?

Pcap++/src/MBufRawPacket.cpp
clementperon: this is a good explanation that should go in the commit log of "Fix a bug in MBufRawPacket::setMBuf"

clementperon (approved):

Pcap++/src/MBufRawPacket.cpp
seladb: We run "Squash and merge" so these commits go away after merging. That's why it's better to put them as comments, so anyone can go to the PR that made the change and look

Pcap++/src/DpdkDevice.cpp
seladb: Yes I agree, update in this commit: https://github.com/seladb/PcapPlusPlus/pull/1774/commits/6a44d0885da211c613acd7bba9f81ff62cc19dc3

PR #1790, opened 2025-04-27
Added `internal::DeviceListBase` as a common base for device list classes.

Dimi1010:
Split of #1488

Overview of changes:
- Added the base class `internal::DeviceListBase` that provides storage and common iterator API for a list of devices.
- Converted `PcapLiveDeviceList` to utilize `internal::DeviceListBase`.

### DIFF
--- a/Pcap++/header/DeviceListBase.h
+++ b/Pcap++/header/DeviceListBase.h
@@ -23,6 +23,10 @@ namespace pcpp
 			DeviceListBase& operator=(DeviceListBase&&) = default;
 
 		public:
+			using value_type = T*;
+			using size_type = std::size_t;
+			using difference_type = std::ptrdiff_t;
+
 			using iterator = typename PointerVector<T, Deleter>::VectorIterator;
 			using const_iterator = typename PointerVector<T, Deleter>::ConstVectorIterator;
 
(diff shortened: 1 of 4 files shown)


tigercosmos (approved):
LGTM

Pcap++/header/DeviceListBase.h
```diff
@@ -0,0 +1,87 @@
...
+			// Protected destructor to disallow deletion of derived class through a base class pointer
+			~DeviceListBase() = default;
+
+			DeviceListBase& operator=(DeviceListBase const&) = default;
+			DeviceListBase& operator=(DeviceListBase&&) = default;
+
+		public:
+			using value_type = T*;
+			using size_type = std::size_t;
+			using difference_type = std::ptrdiff_t;
```
seladb, line 28: Why do we need these typedefs? 🤔

seladb (approved):

Pcap++/header/DeviceListBase.h
Dimi1010: Mostly to mimic standard container. Originally there were other accessor methods beside the iterator, but I cut those from this PR, so they aren't used atm.

seladb: I'm not sure we need them but I don't have a strong opinion so I'll let you decide

Dimi1010: I guess we can remove them and readd them later if needed.

PR #1791, opened 2025-04-27
Simplified function names for fetching pcap live devices.

Dimi1010:
This PR shortens all `PcapLiveDeviceList::getPcapLiveDeviceBy***` methods to `PcapLiveDeviceList::getDeviceBy***`.

IMO, having `PcapLiveDevice` in the method name is redundant as the list name already informs the user of the type of devices it contains.

### DIFF
--- a/Pcap++/header/PcapLiveDeviceList.h
+++ b/Pcap++/header/PcapLiveDeviceList.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "DeprecationUtils.h"
 #include "IpAddress.h"
 #include "PcapLiveDevice.h"
 #include <vector>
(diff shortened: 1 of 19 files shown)


tigercosmos (approved):
LGTM

Dimi1010:
@tigercosmos It shouldn't be a breaking change? The old functions are still there with a deprecation warning.

Dimi1010:
> @Dimi1010  I kinda forget the defination of using this "breaking change" label. Not sure if we should use this label before or after removing the old methods?

The way I understand it, the label means a breaking change in the API. Thus, the code from the old version won't compile with the new version.

I don't think deprecations need the label, as the old code will still compile, just with a deprecation warning. Just the actual removal of old methods.

Pcap++/header/PcapLiveDeviceList.h
```diff
@@ -54,33 +55,69 @@ namespace pcpp
...
+		/// @param[in] name The name of the interface (e.g eth0)
+		/// @return A pointer to the live device if this name exists. nullptr otherwise
+		PcapLiveDevice* getDeviceByName(const std::string& name) const;
+
 		/// Get a pointer to the live device by its IP address or name
 		/// @param[in] ipOrName An IP address or name of the interface
 		/// @return A pointer to the live device if exists, nullptr otherwise
+		PCPP_DEPRECATED("Use `getDeviceByIpOrName`")
 		PcapLiveDevice* getPcapLiveDeviceByIpOrName(const std::string& ipOrName) const;
+		/// Get a pointer to the live device by its IP address or name
```
seladb, line 117: nit: we can add a new line between the 2 methods

Pcap++/src/PcapLiveDeviceList.cpp
```diff
@@ -266,18 +266,26 @@ namespace pcpp
 	}
 
 	PcapLiveDevice* PcapLiveDeviceList::getPcapLiveDeviceByIp(const IPAddress& ipAddr) const
+	{
+		return getDeviceByIp(ipAddr);
+	}
+	PcapLiveDevice* PcapLiveDeviceList::getDeviceByIp(const IPAddress& ipAddr) const
```
seladb, line 272: nit: ditto: add a new line here and in the other methods below as well

seladb (approved):

tigercosmos:
Sure, let's add the "breaking change" label when we removed deprecated methods.

Pcap++/src/PcapLiveDeviceList.cpp
Dimi1010: Fixed 195398b

Pcap++/header/PcapLiveDeviceList.h
Dimi1010: Fixed 195398b

Issue #1348, opened 2024-03-26
Migrate LightPcapNg to the forked repo


tigercosmos:
Currently we copy LightPcapNg code directly in `seladb/PcapPlusPlus `

The code lives at here:  [PcapPlusPlus](https://github.com/seladb/PcapPlusPlus/tree/1ff377c5de22caf9907fcc31f3f71ea138346155)/[3rdParty](https://github.com/seladb/PcapPlusPlus/tree/1ff377c5de22caf9907fcc31f3f71ea138346155/3rdParty)/[LightPcapNg](https://github.com/seladb/PcapPlusPlus/tree/1ff377c5de22caf9907fcc31f3f71ea138346155/3rdParty/LightPcapNg)

We now use the [current_commit.git](https://github.com/seladb/PcapPlusPlus/blob/1ff377c5de22caf9907fcc31f3f71ea138346155/3rdParty/LightPcapNg/current_commit.git) file to record the upstream git commit, but this way is difficult to sync the code from the upstream (basically, manual picking).

Therefore we created a forked LightPcapNg for esiast maintenance.

Here are the steps:
1. Migrate the current code to https://github.com/PcapPlusPlus/LightPcapNg (Manually picking and copying)
2. Remove the current copied code of LightPcapNg.
3. Use CMake `ExternalProject_Add` feature to download the source code from `PcapPlusPlus/LightPcapNg` when doing the first run of `cmake`. 
4. Try to push our change in `PcapPlusPlus/LightPcapNg` back to the upstream

seladb:
@tigercosmos I don't think we should use git submodule. Unfortunately submodules are not very convenient because the user needs to remember to fetch them (a simple `git clone` won't fetch them by default).
I think we can keep what we do now, just sync the git commit against our fork instead of the upstream repo

tigercosmos:
@seladb  I think we can just write in the document to use `git clone --recurse-submodules`?

> I think we can keep what we do now, just sync the git commit against our fork instead of the upstream repo

Do you mean to copy the code from the forked repo? I think it doesn't make the maintaining sync issue easier.

seladb:
> @seladb I think we can just write in the document to use `git clone --recurse-submodules`?
> 
> > I think we can keep what we do now, just sync the git commit against our fork instead of the upstream repo
> 
> Do you mean to copy the code from the forked repo? I think it doesn't make the maintaining sync issue easier.

[...]

PR #1792, opened 2025-04-28
Use forked `LightPcapNg` instead of including the sources in `3rdParty/`

seladb:
Addresses issue https://github.com/seladb/PcapPlusPlus/issues/1348

It uses a branch of the forked LightPcapNg that contains all the code we added: https://github.com/PcapPlusPlus/LightPcapNg/pull/1. Once this is all working I'll merge the LightPcapNg PR.

@clementperon @Dimi1010 @tigercosmos I'll need your help in this PR, it contains some sophisticated CMake tricks I'm not familiar with...

PR #1793, opened 2025-04-28
Test `FetchContent` for fetching `LightPcapNg` from fork.

Dimi1010:
Potential implementation of #1348

- Uses `FetchContent` to pull `light_pcapng` dependency from [PcapPlusPlus/LightPcapNg](https://github.com/PcapPlusPlus/LightPcapNg)
- Builds `light_pcapng` as an OBJECT library.
- Simplified linking to `light_pcapng` as CMake 3.12+ supports linking to `OBJECT` libraries with `target_link_libraries`.


Dimi1010:
@seladb Kinda a duplicate of your #1792, but it was easier to prototype on my own PR. Sorry.

seladb:
> @seladb Kinda a duplicate of your #1792, but it was easier to prototype on my own PR. Sorry.

No worries, thanks for working on it! In my PR I managed to make it work when LightPcapNg is part of Pcap++. The only thing that is not yet working is Zstd, I'm not sure why...

clementperon:
@Dimi1010 what's the advantage of fetchcontent vs submodule ?

Dimi1010:
> @Dimi1010 what's the advantage of fetchcontent vs submodule ?

Overall, `FetchContent` is built upon `ExternalProject` to allow the projects to be pulled at configure time, instead of build time.

Haven't really gone into depth about differences with submodule and but a couple of differences I can think of:
- `FetchContent` is integrated in the build system and not the repo, so it is automatically pulled on configure, without the need to update git submodules.
- It can allow content to be fetched by other places than git. (It uses `ExternalProject`'s infrastructure so it is whatever that can pull)
- It has integrations with CMake's `find_package` directives and the general CMake ecosystem.
-- By default `FetchContent_MakeAvailable` only fetches the declared package, if an earlier `find_package` hasn't already found it.
-- There is also a switch that redirects `find_package` to internally call `FetchContent_MakeAvailable`

Can't think of much else off the top of my head, but there are probably other differences too.

PR #1796, opened 2025-04-30
Updated `DpdkDeviceList` to use `internal::DeviceListBase`.

Dimi1010:
Changes:
- `PointerVector`'s `front()`, `back()` and `at()` changed to return T* even in `const` on the rationale that modifying the pointee does not modify the vector's contents (the pointer in the vector isn't modified).
- Updated `DpdkDeviceList` to use `internal::DeviceListBase`.

### DIFF
--- a/Common++/header/PointerVector.h
+++ b/Common++/header/PointerVector.h
@@ -218,25 +218,13 @@ namespace pcpp
 		}
 
 		/// @return A pointer of the first element in the vector
-		T* front()
-		{
-			return m_Vector.front();
-		}
-
-		/// @return A pointer to the first element in the vector
-		T const* front() const
+		T* front() const
 		{
 			return m_Vector.front();
 		}
 
 		/// @return A pointer to the last element in the vector
-		T* back()
-		{
-			return m_Vector.back();
-		}
-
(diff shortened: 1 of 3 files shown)


Common++/header/PointerVector.h
```diff
@@ -218,25 +218,13 @@ namespace pcpp
 		}
 
 		/// @return A pointer of the first element in the vector
-		T* front()
-		{
-			return m_Vector.front();
-		}
-
-		/// @return A pointer to the first element in the vector
-		T const* front() const
```
seladb, line 227: Why remove these methods? 🤔

seladb (approved):

Common++/header/PointerVector.h
Dimi1010: Mostly because of this [line](https://github.com/seladb/PcapPlusPlus/pull/1796/files#diff-b60eea790b374a2552afcce8cc861ba224111482acd8579c9373ac715a7af73fR221).

Although it actually might be more correct to change the DeviceList containers to have the `getXXX` functions that return non-`const` pointers not be marked as `const`.

PR #1797, opened 2025-04-30
Updated `KniDeviceList` to use `internal::DeviceListBase`.

Dimi1010:
Changes:
- Added `reserve` and `capacity` to `PointerVector`
- Updated `KniDeviceList` to use `internal::DeviceListBase`.

### DIFF
--- a/Common++/header/PointerVector.h
+++ b/Common++/header/PointerVector.h
@@ -217,6 +217,22 @@ namespace pcpp
 			return m_Vector.size();
 		}
 
+		/// @brief Get the current capacity of the vector.
+		/// @return The number of elements that can be held in the vector without requiring a reallocation.
+		size_t capacity() const
+		{
+			return m_Vector.capacity();
+		}
+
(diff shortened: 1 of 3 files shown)


Common++/header/PointerVector.h
```diff
@@ -217,6 +217,22 @@ namespace pcpp
 			return m_Vector.size();
 		}
 
+		/// @brief Get the current capacity of the vector.
+		/// @return The number of elements that can be held in the vector without requiring a reallocation.
+		size_t capacity() const
+		{
+			return m_Vector.capacity();
+		}
```
seladb, line 225: Is this needed?

Dimi1010: Not at the moment, but since a `reserve` is provided, it seems nice to also provide a way to actually check the capacity of the vector, no?

seladb: I guess it's ok and we can keep it...

seladb (approved):

PR #1804, opened 2025-05-04
Simplified the implementation of `PcapNgFileWriterDevice` opening.

Dimi1010:
The purpose of this PR is streamlining of the `PcapNgFileWriterDevice` infrastructure for opening the devices. The current opening code has many duplicated fragments as different modes require only minor changes to the main opening logic. In addition it also provides a common struct for storing `pcapng` metadata that can be read/written by pcpp.

`PcapNgFileReaderDevice` changes:
- Added new `PcapNgMetadata` struct that can hold the metadata of PcapNg file that can be read/write by pcpp.
- Added a new method `PcapNgFileReaderDevice::getMetadata` that returns `PcapNgMetadata` object.

`PcapNgFileWriterDevice` changes:
- Consolidated all device opening logic under `PcapNgFileWriterDevice::openImpl` to remove duplicated opening logic.
- Added `override` keyword to overridden methods.
- Added a new `open(PcapNgMetadata const&)` method that opens the device in write mode and writes the metadata to the header. This method is supposed to be an alternative to `open(const std::string& os, const std::string& hardware, const std::string& captureApp, const std::string& fileComment)`.

### DIFF
--- a/Pcap++/header/PcapFileDevice.h
+++ b/Pcap++/header/PcapFileDevice.h
@@ -220,6 +220,21 @@ namespace pcpp
 		void close();
 	};
 
+	/// @struct PcapNgMetadata
+	/// @brief A struct for holding the metadata of a pcap-ng file. The metadata includes the operating system,
+	/// hardware, capture application and file comment.
+	struct PcapNgMetadata
+	{
+		/// The operating system that was used for capturing the packets
+		std::string os;
+		/// The hardware that was used for capturing the packets
+		std::string hardware;
+		/// The capture application that was used for capturing the packets
+		std::string captureApplication;
+		/// The comment that was written inside the file
+		std::string comment;
+	};
+
 	/// @class PcapNgFileReaderDevice
 	/// A class for opening a pcap-ng file in read-only mode. This class enable to open the file and read all packets,
 	/// packet-by-packet
@@ -453,22 +468,6 @@ namespace pcpp
 			close();
 		}
 
-		/// Open the file in a write mode. If file doesn't exist, it will be created. If it does exist it will be
-		/// overwritten, meaning all its current content will be deleted. As opposed to open(), this method also allows
-		/// writing several metadata attributes that will be stored in the header of the file
-		/// @param[in] os A string describing the operating system that was used to capture the packets. If this string
-		/// is empty or null it will be ignored
-		/// @param[in] hardware A string describing the hardware that was used to capture the packets. If this string is
-		/// empty or null it will be ignored
-		/// @param[in] captureApp A string describing the application that was used to capture the packets. If this
-		/// string is empty or null it will be ignored
-		/// @param[in] fileComment A string containing a user-defined comment that will be part of the metadata of the
-		/// file. If this string is empty or null it will be ignored
-		/// @return True if file was opened/created successfully or if file is already opened. False if opening the file
-		/// failed for some reason (an error will be printed to log)
-		bool open(const std::string& os, const std::string& hardware, const std::string& captureApp,
-		          const std::string& fileComment);
-
 		/// The pcap-ng format allows adding a user-defined comment for each stored packet. This method writes a
 		/// RawPacket to the file and adds a comment to it. Before using this method please verify the file is opened
 		/// using open(). This method won't change the written packet or the input comment
@@ -501,7 +500,7 @@ namespace pcpp
 		/// overwritten, meaning all its current content will be deleted
 		/// @return True if file was opened/created successfully or if file is already opened. False if opening the file
 		/// failed for some reason (an error will be printed to log)
-		bool open();
+		bool open() override;
 
 		/// Same as open(), but enables to open the file in append mode in which packets will be appended to the file
 		/// instead of overwrite its current content. In append mode file must exist, otherwise opening will fail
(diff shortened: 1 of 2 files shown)


Pcap++/header/PcapFileDevice.h
```diff
@@ -220,6 +220,21 @@ namespace pcpp
 		void close();
 	};
 
+	/// @struct PcapNgMetadata
+	/// @brief A struct for holding the metadata of a pcap-ng file. The metadata includes the operating system,
+	/// hardware, capture application and file comment.
+	struct PcapNgMetadata
```
seladb, line 226: I'm not sure we need to expose this struct. Maybe we can keep it internal?

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -919,6 +863,35 @@ namespace pcpp
...
+	{
+		PcapNgMetadata metadata;
+		metadata.os = os;
+		metadata.hardware = hardware;
+		metadata.captureApplication = captureApp;
+		metadata.comment = fileComment;
+		return openImpl(false, &metadata);
+	}
+
+	bool PcapNgFileWriterDevice::openImpl(bool appendMode, PcapNgMetadata const* metadata)
```
seladb, line 890: I'm not sure `openImpl` is significantly simpler than the previous implementation, IMO it's actually more complex...

Dimi1010: I think its conceptually simpler when reasoning about the overloads of `open`. It also removes duplicated code across the different `open` methods.

In the old structure we have 3 open methods that are roughly standalone-ish:
- `open()` - standard open in write mode.
- `open(bool appendMode)` - this one is a bit weird, as it redirects to `open()` if append mode is false, which is fair but it makes for weird call chain to follow, IMO.
- `open(/* header params */)` - this one is completely standalone and it was even written and declared in a separate place from the other methods.

The new structure essentially removes the code duplication of shared sections where everything is forwarded to `openImpl()` and the different `open(...)` methods just act as a public interface. That way all the code to open a file is in one place and people don't have to look through and/or change 3 methods if something changes. If there were drastic procedure differences in opening the files like in `PcapFileWriterDevice` then inner `openWrite` and `openAppend` would have been better, but as the procedures are mostly similar, I find it simpler to have it all in one function.

Pcap++/header/PcapFileDevice.h
Dimi1010: I am curious as to why? I could see reasoning for people to want to pull the entire metadata in one call?

seladb: The API with this change is a bit confusing because we have `getMetadata()` but also `getOS()`, `getHardware()`, `getCaptureApplication()` and `getCaptureFileComment()`. I think the existing API is good enough and this duplication is not needed

Pcap++/src/PcapFileDevice.cpp
seladb: With the new approach, we move the complexity from the different overloads of `open()` to `openImpl()`. IMO it doesn't make things a lot simpler or more readable. It does remove some code duplication, but not too much...

Dimi1010: The main benefit I see is that it makes only 1 point we need to maintain instead of 3.

Pcap++/header/PcapFileDevice.h
Dimi1010: Sure, I can see the reasoning. I will remove it then, along with the new `open` overload and the changes to the `getOS`, etc, methods.

Pcap++/src/PcapFileDevice.cpp
seladb: I think most of the implementation of the 3 overloads is not duplicated, so the price we pay on some code duplication is not that bad

Dimi1010: Idk, looking at them side by side, they do look pretty similar.

The only real difference is:
- if `light_pcapng_open_write` or `light_pcapng_open_append` is used 
- if `info` is created with `light_create_default_file_info` or `light_create_file_info`.

The rest is just boilerplate.

image

seladb: I think the difference is significant enough to keep them separate and the duplication is just a few lines of code

Dimi1010: Maybe do something similar to the `openAppend` and `openWrite` instead of keeping `openImpl` similar to the PcapFileWriter then?
Since IMO, it feels weird to have the `open(bool appendMode)` to redirect to `open()` implementation. Usually I expect overloads with less parameters to defer to overloads with more parameters and just set parameters, instead of deferring to the overload with no parameters.

Something like this
```cpp

// public methods
open() { return openWrite(); }
open(bool appendMode) { return appendMode ? openAppend() : openWrite(); }
open(/* header data */)
{
  PcapMetadata data = /* header setup */
  return openWrite(&data);
}

// private methods
openWrite(PcapMedatada* metadata = nullptr);
openAppend();
```

Pcap++/header/PcapFileDevice.h
Dimi1010: Removed the `getMetadata` function and moved `PcapNgMetadata` to be private struct under `PcapNgFileWriterDevice`.

Pcap++/src/PcapFileDevice.cpp
seladb: I think this approach is better 👍

Pcap++/header/PcapFileDevice.h
```diff
@@ -521,23 +505,58 @@ namespace pcpp
...
+		/// is empty or null it will be ignored
+		/// @param[in] hardware A string describing the hardware that was used to capture the packets. If this string is
+		/// empty or null it will be ignored
+		/// @param[in] captureApp A string describing the application that was used to capture the packets. If this
+		/// string is empty or null it will be ignored
+		/// @param[in] fileComment A string containing a user-defined comment that will be part of the metadata of the
+		/// file. If this string is empty or null it will be ignored
+		/// @return True if file was opened/created successfully or if file is already opened. False if opening the file
+		/// failed for some reason (an error will be printed to log)
+		bool open(const std::string& os, const std::string& hardware, const std::string& captureApp,
```
seladb, line 523: nit: is there any reason to move this method?

seladb (approved):
One small nit comment, otherwise LGTM

Pcap++/header/PcapFileDevice.h
Dimi1010: So all the overloads of open are grouped together for easier overview from the header.

PR #1806, opened 2025-05-04
Cleanup of `PcapFileWriterDevice::open`

Dimi1010:
This PR aims to improve the readability of the code, by moving opening logic to `openWrite` and `openAppend` and changing `open()` to forward to `open(false)`.

This allows a single point of dispatch for all `open` calls that can be used to check for preconditions (device already opened, etc...).

### DIFF
--- a/Pcap++/src/PcapFileDevice.cpp
+++ b/Pcap++/src/PcapFileDevice.cpp
@@ -673,7 +673,7 @@ namespace pcpp
 	{
 		if (isOpened())
 		{
-			// TODO: Ambuguity in API
+			// TODO: Ambiguity in API
 			//   If appendMode is required but the file is already opened in write mode.
 			PCPP_LOG_DEBUG("Pcap descriptor already opened. Nothing to do");
 			return true;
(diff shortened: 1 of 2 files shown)


Pcap++/src/PcapFileDevice.cpp
```diff
@@ -666,12 +666,33 @@ namespace pcpp
 
 	bool PcapFileWriterDevice::open()
 	{
-		if (m_PcapDescriptor != nullptr)
+		return open(false);
```
tigercosmos, line 669: why not `PcapFileWriterDevice::open(bool appendMode = false)`?

Dimi1010: Do you mean just having the `open(bool appendMode = false)` method and skip having the no param one?

tigercosmos: @Dimi1010  Yes, that's what I meant.

Dimi1010: @tigercosmos It can't be done, due to `open()` being a virtual method. Methods with defaulted parameters still have the full method signature and the defaults are embedded at the call site. Virtual methods need a precise signature match their overrides so the dynamic dispatch works correctly, and in this case that means a dedicated no param function.

seladb (approved):

PR #1785, opened 2025-04-23
Separated `PcapLiveDevice` statistics update worker under its own class.

Dimi1010:
The PR separates the logic related to the statistics update worker into its own subclass to improve readability.

The main changes include:
- Moved `PCapStats` to be declared under `internal` namespace.
- Added helper method `getStatistics` to `PcapHandle` that calls `pcap_stats` on the handle internally.
- Added `StatisticsUpdateWorker` as a nested class under PcapLiveDevice to encapsulate a running statistics update worker thread.
- Replaced member variables related to the statistics update worker with a `std::unique_ptr<StatisticsUpdateWorker>` member field, representing an active worker.

PR #1811, opened 2025-05-07
Extracted a separate interface for `getStatistics` methods.

Dimi1010:
This PR extracts a separate `IPcapStatisticsProvider` interface that defines the `getStatistics` method signature. It also adds a helper method for returning the statistics by return value rather than an `out` parameter.

Rationale: The motivation for this PR is the eventual detangling of `PcapFileDevice` classes. Currently all file device classes inherit from `PcapDevice` even though only 1 of 3 types actually uses pcap. This PR aims to extract a common functionality that is used amongst all 3 pcap file types into a separate interface to lessen the dependency on the `PcapDevice` class.


tigercosmos (approved):

seladb (commented):
I'm ok with this change, but it will conflict with this PR: https://github.com/seladb/PcapPlusPlus/pull/1785

Dimi1010:
> I'm ok with this change, but it will conflict with this PR: #1785

@seladb Why would it conflict?

seladb (approved):
Maybe we can merge this PR first and then resolve the conflict with https://github.com/seladb/PcapPlusPlus/pull/1785

PR #1812, opened 2025-05-11
Fix read overflow

Marti2203:
Fixes the following issue detected by OSS-Fuzz:
* https://issues.oss-fuzz.com/issues/390004170

This fix was generated by CodeRover-S, an LLM agent for fixing security vulnerabilities. More details can be found at:

* https://arxiv.org/pdf/2411.03346
* https://github.com/AutoCodeRoverSG/auto-code-rover

### DIFF
--- a/Packet++/src/RawPacket.cpp
+++ b/Packet++/src/RawPacket.cpp
@@ -136,6 +136,13 @@ namespace pcpp
 
 	void RawPacket::insertData(int atIndex, const uint8_t* dataToInsert, size_t dataToInsertLen)
 	{
+		// Check for overflow in the new length
+		if ((size_t)m_RawDataLen + dataToInsertLen < (size_t)m_RawDataLen)
+		{
+			PCPP_LOG_ERROR("RawPacket::insertData: dataToInsertLen causes overflow");
+			return;
+		}
+
 		// memmove copies data as if there was an intermediate buffer in between - so it allows for copying processes on
 		// overlapping src/dest ptrs if insertData is called with atIndex == m_RawDataLen, then no data is being moved.
 		// The data of the raw packet is still extended by dataToInsertLen


Packet++/src/RawPacket.cpp
```diff
@@ -136,6 +136,13 @@ namespace pcpp
 
 	void RawPacket::insertData(int atIndex, const uint8_t* dataToInsert, size_t dataToInsertLen)
 	{
+		// Check for overflow in the new length
+		if ((size_t)m_RawDataLen + dataToInsertLen < (size_t)m_RawDataLen)
```
Dimi1010, line 140: This is always `false`. The equation is equivalent to `a + x < a` where x∈[0, `size_t` max]. For it to be true, x needs to be less than 0, which is impossible due to `dataToInsertLen` is unsigned.

Packet++/src/RawPacket.cpp
```diff
@@ -136,6 +136,13 @@ namespace pcpp
...
+		{
+			PCPP_LOG_ERROR("RawPacket::insertData: dataToInsertLen causes overflow");
+			return;
+		}
```
Dimi1010, line 144: Can we throw an `std::length_error` instead of just returning with a error message? The error message will not indicate the error condition to the caller.

seladb: Will throwing an exception prevent the application from crashing? 🤔 

This method was called in `BgpLayer` so we should probably think how to prevent it from being called with malformed data

Dimi1010: You want a crash if overflow happens. Or at least you want an exception thrown to indicate the failure of the operation, and if the caller doesn't handle it, then to crash. It isn't really possible to recover from the situation as the system has reached the limit of how much data can be stored.

Silently continuing with just an error message in the log is worse as the program will assume the operation went file and probably produce corrupted data down the line. This is exactly the type of improbable situation the exception mechanism was made to handle.

seladb: I totally agree, however I don't think throwing an exception will resolve the OSS-Fuzz issue. We should probably throw an exception and fix the issue that caused it in `BgpLayer`

Marti2203: Hi, just to confirm - I will update the patch to throw an `std::length_error` . Would that be accepted by both of you? @Dimi1010 @seladb

seladb: @Marti2203 we can throw an `std::length_error` but should also fix the issue in `BgpLayer`. This exception alone won't solve the issue because the program will still crash because of the exception

Marti2203: Sure, will prepare a patch for that

Marti2203: Hi @seladb @Dimi1010 apologies for the delay, I have pushed a commit that I think handles it as discussed - an error is thrown from the raw packet which is caught in the `BGPLayer` to convert into a bool.

seladb: @Marti2203 thanks for working on it!
Catching exceptions is pretty expensive in C++. I'd avoid using `try..catch` and instead check the sizes and prevent calling `extendLayer` is we know it'd fail

Dimi1010: @seladb @Marti2203 

Generally, I find exceptions are fine as long as they aren't used to dictate common control flow. The current mainstream compilers use an architecture that has a "zero cost" try-catch blocks if the exception isn't thrown. The tradeoff is a significant overhead if an exception is thrown, so you only want them thrown for **rare** events.

The current usage seems ok to me? How often are overflows expected to happen?

Marti2203: Hi, I have been working on other things and have been checking out this PR again. I agree with sela's points and will prepare a patch after a confirmation from your side - 
* NO exception will be thrown in RawPacket, only it is logged and an error is returned. This is to ensure new clients of the code do not lead to a crash and code stylistically aligns with the rest. 
* Separately, will add a check for the error in the BGPLayer as it is the only client in the project that uses it.

Packet++/src/RawPacket.cpp
seladb: nit: maybe use `static_cast` instead of C-style casting?

Packet++/src/RawPacket.cpp
seladb: I think we can still throw an error in `insertData()` but I would avoid `try..catch` in `BgpLayer`  and instead check the sizes before calling `extendLayer()`. @Marti2203 let me know what you think

Marti2203: Sounds good. I will also cover the nitpick and send it tonight :)

Marti2203:
Pushed an update. Feedback is appreciated!

seladb:
> Pushed an update. Feedback is appreciated!

@Marti2203 CI fails, can you please take a look?

Marti2203:
Hi @seladb Apologies for the super late response, will take care of it today

Marti2203:
Thanks for the rerun, will check the results!

Marti2203:
@seladb I have added Packet.h to fix the forward declaration issue. Would be happy to discuss a different solution but I think that it should not be a problem

Packet++/src/BgpLayer.cpp
```diff
@@ -744,9 +745,25 @@ namespace pcpp
 
 		if (newNlriDataLen > curNlriDataLen)
 		{
+
+			// offsetInLayer, numOfBytesToExtend
+			//		int indexToInsertData = layer->m_Data + offsetInLayer - m_RawPacket->getRawData();
```
seladb: Can we remove commented-out code?

Packet++/src/BgpLayer.cpp
```diff
@@ -744,9 +745,25 @@ namespace pcpp
...
+			auto bytesToExtend = newNlriDataLen - curNlriDataLen;
+
+			if (m_Data != nullptr && m_Packet != nullptr)
+			{
+				auto raw_len = static_cast<size_t>(m_Packet->getRawPacket()->getRawDataLen());
```
seladb: Variable name convention is camleCase: `rawLen` instead of `raw_len`

Packet++/src/BgpLayer.cpp
```diff
@@ -744,9 +745,25 @@ namespace pcpp
...
+				if (raw_len + bytesToExtend < raw_len)
```
seladb: How can this be true? 🤔 
`bytesToExtend` is of type `size_t`, which can't be a negative number

Packet++/src/BgpLayer.cpp
seladb, line 1: Can you add the packet data that caused this issue to our regression tests, the same way it's done in [this PR](https://github.com/seladb/PcapPlusPlus/pull/1954)?

Packet++/src/BgpLayer.cpp
Marti2203: As previously discussed, this can overflow when unsigned going back to a lower value.

Packet++/src/BgpLayer.cpp
Marti2203: Sure, will do!

Packet++/src/BgpLayer.cpp
Marti2203: Yep, will do!

Packet++/src/BgpLayer.cpp
Marti2203: ah, had the wrong impression due to the `m_`, my bad

Packet++/src/BgpLayer.cpp
seladb: You're right! Can you add a comment above this line to remind future us why we did it? 🙂

Marti2203: Sure! 😄 It is an annoying edge-case

seladb:
@Marti2203 @danasana [This PR](https://github.com/seladb/PcapPlusPlus/pull/1954) should also cover this fix, am I wrong?

Marti2203:
@seladb , to my understanding it does similar checks. Can we test the example datapoint from clusterfuzz on the other PR to validate?

Marti2203:
By the way, the current build failures do not look like something caused by the PR, correct?

danasana:
> @Marti2203 @danasana [This PR](https://github.com/seladb/PcapPlusPlus/pull/1954) should also cover this fix, am I wrong?

@seladb Yes, I believe so. I have tested the problematic scenario on my PR, and no errors occurred

seladb:
> > @Marti2203 @danasana [This PR](https://github.com/seladb/PcapPlusPlus/pull/1954) should also cover this fix, am I wrong?
> 
> @seladb Yes, I believe so. I have tested the problematic scenario on my PR, and no errors occurred

Thanks @danasana ! 🙏

seladb:
#1954 was just merged. @Marti2203 @danasana should we close this PR?

Marti2203:
Hi @seladb , yes, in this case the functionality has been implemented on a more broader level. I will close the PR

Issue #1813, opened 2025-05-12
Add Modbus Protocol Support


yahyayozo:
Hello PcapPlusPlus maintainers,

I’d love to contribute to the library by implementing support for the Modbus protocol, a widely used communication protocol in industrial automation. 
Would this be a welcome addition to the project? If so, I'd appreciate any guidance or recommendations for implementing it according to PcapPlusPlus' standards.

Looking forward to your feedback!

seladb:
Sure @yahyayozo , every contribution is highly welcome and appreciated, thank you! 🙏 

Can you provide some info about the Modbus protocol? Mostly about the different PDUs and message types. Based on this, I can recommend a similar protocol you can use as a refrence

yahyayozo:
Modbus is a client/server data communications protocol in the application layer, used mainly in industrial automation. It is basically for reading and writing register values in the server.

The Frame starts by a MBAP (MODBUS Appicaiton Protocol Header) (7 bytes):
- **Transaction Identifier** (2 bytes): A unique ID for matching client requests with server responses.
- **Protocol Identifier** (2 bytes): Set to 0x0000 for Modbus protocol.
- **Length** (2 bytes): Specifies the length of the remaining bytes (Unit Identifier + PDU).
- **Unit Identifier** (1 byte): Identifies the target device.

The PDU consists of:
- **Function Code** (1 byte): Defines the operation (e.g., read/write)
- **Data (variable)**: Contains operation-specific parameters (e.g., register address, data values). For responses, it includes requested data or exception codes.

You can refer to this [official document](https://www.modbus.org/docs/Modbus_Application_Protocol_V1_1b3.pdf) from Modbus for its specifications. You can find all the function codes and their meanings there

Dimi1010:
Haven't looked at it in depth, but a way I can see it implemented in the framework can be something like this:

- A main `ModbusLayer` that parses the frame header.
- A collection of helper structures that represent known PDU configurations.

Each PDU must contain a method to serialize / deserialize itself to/from a byte span. The method must be standard across all PDUs. Deserialization is probably better to be a static member function that acts as a factory for the PDU.

The main layer can have the following methods:
- general frame header accessors/mutators.
- `getPDUFunctionCode` for parsing the first byte of the PDU.
[...]

PR #1820, opened 2025-05-18
[WIP] Attempt to flatten the `BgpLayer` inheritance chain.

Dimi1010:
## Overview
This PR is a draft attempt to flatten the inheritance chain of `BgpLayer` and separate the protocol layer parsing logic from the message payload parsing logic. 

In the current implementation each BGP message has a dedicated derived BGP layer class. This introduces limitations where the message payload cannot be switched to a different type without destroying the layer object and creating a new one, requiring the layer to be extracted from the stack and then readded to it. 

This PR aims to simplify the following operations by providing the following objects:
- "MessageView" proxies - thin wrappers over a `BgpLayer` that parse the payload as a specific message.
They exist in two flavors "View" and "ConstView" with the former allowing in-place data modifications and the latter being read-only.
- "Message" objects - lightweight struct objects that can be passed to the `BgpLayer`'s constructor or a dedicated method to completely replace the underlying message with a message of a different type. Similar to the ArpLayer's [helper structs](https://github.com/seladb/PcapPlusPlus/blob/a49a79e0b67b402ad75ffa96c1795def36df75c8/Packet%2B%2B/header/ArpLayer.h#L63-L139).

### The "View" object
The view object is a thin wrapper over a `BgpLayer` that holds a reference to the layer it has been created from. It provides accessors and mutators for reading and writing message specific data directly to the underlying buffer. Each view is intended to be passed by value and should be no larger than a pointer (the reference to the `BgpLayer`), if possible.

Each view should maintain message integrity. During construction the View should validate a minimally viable message of the type can be parsed from the buffer. Full validation can be deferred until use of accessors and/or mutators of the view object.

#### Improved support of full length update message path attributes.
The previous version supported path attributes with data up to 32 bytes. The new `PathAttribute` class supports the full spectrum of variable data lengths a path attribute block can contain specified by  [RFC 4271 Section 4.3](https://www.rfc-editor.org/rfc/rfc4271#section-4.3). 
The attribute can support a data buffer of:
- standard length attribute - up to 255 bytes
- extended length attribute - unlimited, restricted by total message length of 4096 bytes.

#### Addition of utility functions for `enum class` flags.
This PR adds the header `EnumFlagUtils.h` to the `Common++` package. The header provides operators for bitwise operators and helper functions of `enum class` objects when they satisfy the trait `EnableBitMaskOperators`. 

In addition the header defines the following macros:
- `PCPP_DECLARE_ENUM_FLAG` - to enable the operators to be used for a class.
- `PCPP_USING_ENUM_FLAG_OPERATORS`  - to declare aliases to all templated operators in the scope this macro is used to satisfy ADL.

### DIFF
--- a/Common++/CMakeLists.txt
+++ b/Common++/CMakeLists.txt
@@ -15,6 +15,7 @@ add_library(
 set(
   public_headers
   header/DeprecationUtils.h
+  header/EnumFlagUtils.h
   header/GeneralUtils.h
   header/IpAddress.h
   header/IpAddressUtils.h
--- a/Common++/header/EnumFlagUtils.h
+++ b/Common++/header/EnumFlagUtils.h
@@ -1,0 +1,74 @@
+#pragma once
+
+#include <type_traits>
+
+namespace pcpp
+{
+	namespace internal
+	{
+		template <typename EnumClass> struct EnableBitMaskOperators : std::false_type
+		{
+		};
+
+		template <typename EnumClass>
+		using EnableIfBitMask = typename std::enable_if<EnableBitMaskOperators<EnumClass>::value, EnumClass>::type;
+
+		template <typename EnumClass> constexpr EnableIfBitMask<EnumClass> operator|(EnumClass lhs, EnumClass rhs)
+		{
+			return static_cast<EnumClass>(static_cast<std::underlying_type<EnumClass>::type>(lhs) |
+			                              static_cast<std::underlying_type<EnumClass>::type>(rhs));
+		}
+
+		template <typename EnumClass> constexpr EnableIfBitMask<EnumClass> operator&(EnumClass lhs, EnumClass rhs)
+		{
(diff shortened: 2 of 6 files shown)


Dimi1010:
@seladb @tigercosmos @clementperon @egecetin Can I get some first impressions on the idea of this draft?

seladb:
@Dimi1010 is it worth it to rewrite `BgpLayer`? 🤔

Dimi1010:
> @Dimi1010 is it worth it to rewrite `BgpLayer`? 🤔 

Dunno yet. I have been mostly using it as a test bed for expanding the idea I had in the #1813 about ways it can allow direct read/write of fields different payloads without having to decode and copy the entire buffer out.

Dimi1010:
> @Dimi1010 Is it possible to split this PR? It seems the PR is too large to review.

Possibly, but this is still a WIP.

Dimi1010:
Might work on this later. Closing this for now so it doesn't clutter.

Issue #1754, opened 2025-03-30
Request new protocols


seladb:
This issue includes sub-tickets for reuqested protocols

PR #1823, opened 2025-05-18
#1754 Add Modbus Support

yahyayozo:
- Add ModbusLayer first implementation with header parsing only (no test included)

### DIFF
--- a/Packet++/header/ModbusLayer.h
+++ b/Packet++/header/ModbusLayer.h
@@ -1,0 +1,59 @@
+#pragma once
+
+#include "Layer.h"
+
+/// @file
+/// This file contains classes for parsing, creating and editing Modbus packets.
+
+/// @namespace pcpp
+/// @brief The main namespace for the PcapPlusPlus lib
+namespace pcpp
+{
+
+#pragma pack(push, 1)
+	/// @struct modbus_common_header
+	/// MODBUS Application Protocol header
+	struct modbus_common_header
+	{
+		/// For synchronization between messages of server and client
+		uint16_t transactionId;
+		/// 0 for Modbus/TCP
+		uint16_t protocolId;
+		/// Number of remaining bytes in this frame starting from the unit id
+		uint16_t length;
+		/// Unit identifier
+		uint8_t unitId;
+		/// Function code
+		uint8_t functionCode;
+	};
+#pragma pack(pop)
+
+	class ModbusLayer : public Layer
+	{
+	public:
+		static_assert(sizeof(modbus_common_header) == 8, "modbus_common_header size is not 8 bytes");
+
+		/// A constructor that creates the layer with empty fields
+		ModbusLayer();
+
+		/// @return A pointer to the MODBUS header
+		modbus_common_header* getModbusHeader() const;
+
+		/// @return MODBUS message type
+		uint16_t getTransactionId();
+
+		/// @return MODBUS protocol id
+		uint16_t getProtocolId();
+
+		/// @return MODBUS remaining bytes in frame starting from the unit id
+		/// @note This is the length of the MODBUS payload + unit_id, not the entire packet
+		uint16_t getLength();
+
+		/// @return MODBUS unit id
+		uint8_t getUnitId();
+
+		/// @return MODBUS function code
+		uint8_t getFunctionCode();
+	}
+
+}  // namespace pcpp
--- a/Packet++/header/ModbusLayer.h
+++ b/Packet++/header/ModbusLayer.h
@@ -27,11 +27,10 @@ namespace pcpp
 		uint8_t functionCode;
 	};
 #pragma pack(pop)
-
+static_assert(sizeof(modbus_common_header) == 8, "modbus_common_header size is not 8 bytes");
 	class ModbusLayer : public Layer
 	{
 	public:
-		static_assert(sizeof(modbus_common_header) == 8, "modbus_common_header size is not 8 bytes");
 
 		/// A constructor that creates the layer with empty fields
 		ModbusLayer();
--- a/Packet++/src/ModbusLayer.cpp
+++ b/Packet++/src/ModbusLayer.cpp
@@ -1,0 +1,52 @@
+
+#include "ModbusLayer.h"
+
+namespace pcpp
+{
+	ModbusLayer::ModbusLayer()
+	{
+		const size_t headerLen = sizeof(modbus_common_header);
+		m_DataLen = headerLen;
+		m_Data = new uint8_t[headerLen];
+		memset(m_Data, 0, headerLen);
+
+		// Initialize the header fields to default values
+		modbus_common_header* header = getModbusHeader();
+		header->transactionId = 0;
+		header->protocolId = 0;  // 0 for Modbus/TCP
+		header->length = 2;      // minimum length of the MODBUS payload + unit_id
+		header->unitId = 0;
+		header->functionCode = 0;
+	}
+
+	modbus_common_header* ModbusLayer::getModbusHeader() const
+	{
+		return (modbus_common_header*)m_Data;
+	}
+
+	uint16_t ModbusLayer::getTransactionId()
+	{
+		return getModbusHeader()->transactionId;
+	}
+
+	uint16_t ModbusLayer::getProtocolId()
+	{
+		return getModbusHeader()->protocolId;
+	}
+
+	uint16_t ModbusLayer::getLength()
+	{
+		return getModbusHeader()->length;
+	}
+
+	uint8_t ModbusLayer::getUnitId()
+	{
+		return getModbusHeader()->unitId;
+	}
+
+	uint8_t ModbusLayer::getFunctionCode()
+	{
+		return getModbusHeader()->functionCode;
+	}
+
+}  // namespace pcpp


yahyayozo:
@seladb I've prepared only 1 constrcutor which initiate the layer with default fields values
Can you please mention what are the other constructors I need to implement? also please mention if I need to follow on other protocols layers?

seladb:
@yahyayozo we usually have at least 2 c'tors, sometimes more:
1. As part of the packet parsing flow. For example: https://github.com/seladb/PcapPlusPlus/blob/a49a79e0b67b402ad75ffa96c1795def36df75c8/Packet%2B%2B/header/MplsLayer.h#L37 which is used here (and in some other places also): https://github.com/seladb/PcapPlusPlus/blob/a49a79e0b67b402ad75ffa96c1795def36df75c8/Packet%2B%2B/src/VlanLayer.cpp#L101
2. When the user wants to create a new layer of this type. This is the c'tor you already implemented, but usually we don't implement (only) c'tors with zero arguments. We tend to implement c'tors that allow the user to build the layer. In your case, I'd add arguments like `transactionId`, `unitId` and `functionCode`

yahyayozo:
Hi @seladb can you please guide me briefly on how to add tests for my ModbusLayer for parsing/crafting headers? As I want to pass to function types request/response after testing the header part.

yahyayozo:
@seladb I'll add tests for the headers crafting/parsing implementation and then you can a review before we move to adding the different PDU types

yahyayozo:
@seladb @Dimi1010 I've started adding some tests, but I have a question regarding the ".dat" files; how can I generate examples for some Modbus frames?

seladb:
> @seladb @Dimi1010 I've started adding some tests, but I have a question regarding the ".dat" files; how can I generate examples for some Modbus frames?

If you open the file in Wireshark, you can choose a packet and right click -> Copy -> ... as a Hex Stream:

image

seladb:
@yahyayozo CI fails, can you please look into it? 🙁

yahyayozo:
> @yahyayozo CI fails, can you please look into it? 🙁

yahyayozo:
@seladb I'll check further

yahyayozo:
@seladb all the pre-commit checks have passed + the build was successful
I think if the pipeline build fails, I need to check my compiler, maybe I'm using a wrong version

yahyayozo:
@seladb CI passed! can you please make a review of the already done work before moving forward?

yahyayozo:
@seladb Hi, can you please check the progress I've reached until now? please leave comments on any part that needs adjustement

seladb:
> @seladb Hi, can you please check the progress I've reached until now? please leave comments on any part that needs adjustement

@yahyayozo sure, I'm sorry for the long delay. I'll take a look shortly

Tests/Packet++Test/PacketExamples/ModbusRequest.dat
seladb, line 1: Can you add the pcap/pcapng file you used also?
It'll also help me understand the protocol better

Packet++/header/ProtocolType.h
seladb, line 1: Can you also add this protoclo to `README.md`?

Packet++/header/ModbusLayer.h
```diff
@@ -0,0 +1,125 @@
...
+		/// @brief  Check if a port is a valid MODBUS port
+		/// @param port
+		/// @return
+		static bool isModbusPort(uint16_t port)
+		{
+			return port == 502;
+		}
+
+		/// @return A pointer to the MODBUS header
+		modbus_common_header* getModbusHeader() const;
```
seladb: This could be a private method?

Packet++/header/ModbusLayer.h
```diff
@@ -0,0 +1,125 @@
...
+
+/// @namespace pcpp
+/// @brief The main namespace for the PcapPlusPlus lib
+namespace pcpp
+{
+
+#pragma pack(push, 1)
+	/// @struct modbus_common_header
+	/// MODBUS Application Protocol header
+	struct modbus_common_header
```
seladb: nit: I think we can call it just `modbus_header`?

Packet++/header/ModbusLayer.h
```diff
@@ -0,0 +1,125 @@
...
+
+		/// A constructor that creates the layer from user inputs
+		/// @param[in] transactionId Transaction ID
+		/// @param[in] unitId Unit ID
+		/// @param[in] functionCode Function code
+		ModbusLayer(uint16_t transactionId, uint8_t unitId, uint8_t functionCode);
+
+		/// @brief  Check if a port is a valid MODBUS port
+		/// @param port
+		/// @return
```
seladb: nit: complete the return documentation?

Packet++/header/ModbusLayer.h
```diff
@@ -0,0 +1,125 @@
...
+
+		/// @return A string representation of the layer most important data (should look like the layer description in
+		/// Wireshark)
+		std::string toString() const override
+		{
+			return "Modbus Layer, Transaction ID: " + std::to_string(getTransactionId()) +
+			       ", Protocol ID: " + std::to_string(getProtocolId()) + ", Length: " + std::to_string(getLength()) +
+			       ", Unit ID: " + std::to_string(getUnitId()) +
+			       ", Function Code: " + std::to_string(getFunctionCode());
+		}
```
seladb: 1. Maybe move the implementation to the `.cpp` file to keep the `.h` file cleaner?
2. I'm not sure how Modbus packets look in Wireshark, but probably contain less info? 🤔

Tests/Packet++Test/Tests/ModbusTests.cpp
```diff
@@ -0,0 +1,52 @@
...
+	gettimeofday(&time, nullptr);
+
+	READ_FILE_AND_CREATE_PACKET(1, "PacketExamples/ModbusRequest.dat");
+
+	pcpp::Packet packet(&rawPacket1);
+	PTF_ASSERT_TRUE(packet.isPacketOfType(pcpp::Modbus));
+
+	pcpp::ModbusLayer* modbusLayer = packet.getLayerOfType<pcpp::ModbusLayer>();
+	PTF_ASSERT_NOT_NULL(modbusLayer);
+	PTF_ASSERT_EQUAL(modbusLayer->getTransactionId(), 0);
```
seladb: Can you get a packet where transaction ID is not 0?

Tests/Packet++Test/Tests/ModbusTests.cpp
```diff
@@ -0,0 +1,52 @@
...
+	pcpp::ModbusLayer modbusLayer(12345, 1, 3);
+
+	PTF_ASSERT_EQUAL(modbusLayer.getTransactionId(), 12345);
+	PTF_ASSERT_EQUAL(modbusLayer.getProtocolId(), 0);
+	PTF_ASSERT_EQUAL(modbusLayer.getLength(), 2);  // minimum length of the MODBUS payload + unit_id
+	PTF_ASSERT_EQUAL(modbusLayer.getUnitId(), 1);
+	PTF_ASSERT_EQUAL(modbusLayer.getFunctionCode(), 3);
+	PTF_ASSERT_EQUAL(modbusLayer.getHeaderLen(), sizeof(pcpp::modbus_common_header));
+
+	PTF_ASSERT_EQUAL(modbusLayer.toString(),
```
seladb: I'd test `toString()` in `ModbusLayerParsingTest` instead of here

Tests/Packet++Test/Tests/ModbusTests.cpp
```diff
@@ -0,0 +1,52 @@
+#include "../TestDefinition.h"
+#include "../Utils/TestUtils.h"
+#include "Packet.h"
+#include "ModbusLayer.h"
+#include "EndianPortable.h"
+#include "SystemUtils.h"
+
+PTF_TEST_CASE(ModbusLayerCreationTest)
```
seladb, line 8: The best way to test layer creation is to compare it to a real packet. You can load `ModbusRequest.dat`, extract the modbus layer and compare it to the created packet in this test

yahyayozo:
@seladb pcap files are ignored, in the gitignore, how can I commit my modbus.pcap file?

seladb:
> @seladb pcap files are ignored, in the gitignore, how can I commit my modbus.pcap file?

@yahyayozo you can use `git add -f modbus.pcap` to force adding a pcap file despite of `.gitignore`

yahyayozo:
@seladb I've made the changes of your last review, and the pipeline is working
I've also included the pcap file for Modbus
Can you please review again?

README.md
```diff
@@ -287,6 +287,7 @@ PcapPlusPlus currently supports parsing, editing and creation of packets of the
 52. Telnet - parsing only (no editing capabilities)
 53. X509 certificates - parsing only (no editing capabilities)
 54. Generic payload
+55. Modbus
```
seladb: 1. The protocols in each layer are ordered alphabetically. It should probably go after `LDAP`
2. Should it be MODBUS or Modbus?

Tests/Packet++Test/PacketExamples/ModbusRequest.dat
seladb: Ths pcap file takes 9MB and contains 60K packets. Can you filter out only the single packet used in the test (`ModbusRequest.dat`)?

Packet++/header/ModbusLayer.h
seladb: This comment wasn't addressed

Packet++/header/ModbusLayer.h
```diff
@@ -0,0 +1,132 @@
...
+	enum modbus_function_code
+	{
+		MODBUS_READ_COILS = 1,
+		MODBUS_READ_DISCRETE_INPUTS = 2,
+		MODBUS_READ_HOLDING_REGISTERS = 3,
+		MODBUS_READ_INPUT_REGISTERS = 4,
+		MODBUS_WRITE_SINGLE_COIL = 5,
+		MODBUS_WRITE_SINGLE_REGISTER = 6,
+		MODBUS_WRITE_MULTIPLE_COILS = 15,
+		MODBUS_WRITE_MULTIPLE_REGISTERS = 16
```
seladb: 1. Can you rename `modbus_function_code` to `ModbusFunctionCode`?
2. Maybe change it to `enum class` that inherits from `uint8_t` and put it inside `ModbusLayer`?
3. Please add doxygen documentation to each enum value
4. I'd add another `Unknown` value that could be returned in `getFunctionCode()` if the packet value doesn't match any of the enum values

Packet++/header/ModbusLayer.h
seladb: Please complete the documentation for `port`:
```cpp
/// @param[in] port The port number to be checked
```

Packet++/header/ModbusLayer.h
seladb: > 2\. I'm not sure how Modbus packets look in Wireshark, but probably contain less info? 🤔

What do you think about it?

Packet++/header/ModbusLayer.h
```diff
@@ -0,0 +1,132 @@
...
+		/// @param[in] packet A pointer to the Packet instance where layer will be stored in
+		ModbusLayer(uint8_t* data, size_t dataLen, Layer* prevLayer, Packet* packet)
+		    : Layer(data, dataLen, prevLayer, packet, Modbus)
+		{}
+
+		/// A constructor that creates the layer from user inputs
+		/// @param[in] transactionId Transaction ID
+		/// @param[in] unitId Unit ID
+		/// @param[in] functionCode Function code
+		ModbusLayer(uint16_t transactionId, uint8_t unitId, uint8_t functionCode);
```
seladb: Since we now have an enum for function code, maybe use it here?

Packet++/src/ModbusLayer.cpp
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	uint8_t ModbusLayer::getUnitId() const
+	{
+		return getModbusHeader()->unitId;
+	}
+
+	modbus_function_code ModbusLayer::getFunctionCode() const
+	{
+		return static_cast<modbus_function_code>(getModbusHeader()->functionCode);
```
seladb: We should handle a case when `getModbusHeader()->functionCode` contains a value that is not included in the enum, and return `Unknown`

Packet++/header/ModbusLayer.h
```diff
@@ -0,0 +1,132 @@
...
+		{}
+
+		/// @brief Get the length of the MODBUS header
+		/// @return Length of the MODBUS header in bytes
+		size_t getHeaderLen() const override
+		{
+			return sizeof(modbus_header);
+		}
+
+		/// Each layer can compute field values automatically using this method. This is an abstract method
```
seladb: Please update the comment to:
```
/// Does nothing for this layer
```

Tests/Packet++Test/Tests/ModbusTests.cpp
```diff
@@ -0,0 +1,59 @@
...
+	// Transaction ID: 0, Unit ID: 10, Function Code: 17
+	READ_FILE_AND_CREATE_PACKET(1, "PacketExamples/ModbusRequest.dat");
+
+	pcpp::Packet realPacket(&rawPacket1);
+	PTF_ASSERT_TRUE(realPacket.isPacketOfType(pcpp::Modbus));
+	pcpp::ModbusLayer* modbusLayerFromRealPacket = realPacket.getLayerOfType<pcpp::ModbusLayer>();
+
+	pcpp::ModbusLayer modbusLayer(0, 10, 17);
+
+	PTF_ASSERT_BUF_COMPARE(modbusLayer.getData(), modbusLayerFromRealPacket->getData(), modbusLayer.getHeaderLen());
```
seladb: I'd also test that the lengths are similar:
```cpp
PTF_ASSERT_EQUAL(modbusLayer.getDataLen(), modbusLayerFromRealPacket->getDataLen());
```

Tests/Packet++Test/Tests/ModbusTests.cpp
seladb: This comment was not addressed

Tests/Packet++Test/Tests/ModbusTests.cpp
```diff
@@ -0,0 +1,59 @@
...
+	pcpp::Packet packet(&rawPacket1);
+	PTF_ASSERT_TRUE(packet.isPacketOfType(pcpp::Modbus));
+
+	pcpp::ModbusLayer* modbusLayer = packet.getLayerOfType<pcpp::ModbusLayer>();
+	PTF_ASSERT_NOT_NULL(modbusLayer);
+	PTF_ASSERT_EQUAL(modbusLayer->getTransactionId(), 0);
+	PTF_ASSERT_EQUAL(modbusLayer->getProtocolId(), 0);
+	PTF_ASSERT_EQUAL(modbusLayer->getLength(), 2);
+	PTF_ASSERT_EQUAL(modbusLayer->getUnitId(), 10);
+	PTF_ASSERT_EQUAL(modbusLayer->getFunctionCode(), 17);
```
seladb: Let's update it to compare to the enum value?

seladb:
@yahyayozo please scroll up and see all the comments that weren't addressed

Packet++/header/ModbusLayer.h
```diff
@@ -29,24 +29,54 @@ namespace pcpp
...
+
+			/** Report slave ID (0x11) */
+			MODBUS_REPORT_SLAVE_ID = 17,
+
+			/** Limit to check if the function code is valid */
+			MODBUS_FUNCTION_CODE_LIMIT,
+
+			/** Unknown or unsupported function code (0xFF) */
+			MODBUS_UNKNOWN_FUNCTION = 0xFF
+		};
```
Dimi1010: nit: Do we really need the `MODBUS_*` prefix on the values anymore? Those are used on pure `enum` to avoid conflicts due to the fact that it leaks the values into the namespace, but `enum class` doesn't do that and require enum qualification. So we always would end up with `ModbusFunctionCode::MODBUS_READ_COILS` instead of something like `ModbusFunctionCode::ReadCoils`.

seladb:
@yahyayozo the tests fail in CI...

yahyayozo:
@seladb should be fixed now

Packet++/src/ModbusLayer.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "EndianPortable.h"
+#include <iostream>
+#include <iomanip>
+#include <cstring>
+
+namespace pcpp
+{
+	ModbusLayer::ModbusLayer(uint16_t transactionId, uint8_t unitId, ModbusLayer::ModbusFunctionCode functionCode)
+	{
+		const int16_t pduSize = getFunctionDataSize(functionCode);
```
seladb: I can see that `getFunctionDataSize()` currently supports only one function code, which is `ModbusFunctionCode::READ_INPUT_REGISTERS`. In that case, maybe we can remove the `functionCode` parameter, remove the `getFunctionDataSize()` method, and set the function code in this constructor to be `ModbusFunctionCode::READ_INPUT_REGISTERS`

Packet++/src/ModbusLayer.cpp
seladb: This comment wasn't addressed

Packet++/header/ModbusLayer.h
```diff
@@ -0,0 +1,177 @@
...
+	class ModbusLayer : public Layer
+	{
+	public:
+		/**
+		 * @brief Enum class representing Modbus function codes.
+		 *
+		 * This enumeration defines the standard Modbus function codes used in request and response PDUs.
+		 * Each value corresponds to a specific operation defined by the Modbus protocol.
+		 *
+		 */
```
seladb: Please change the comment style to be `///` instead of `/* */`

Packet++/header/ModbusLayer.h
```diff
@@ -0,0 +1,177 @@
...
+		enum class ModbusFunctionCode : uint8_t
+		{
+			/** Read coil status (0x01) */
```
seladb: Ditto here and the rest of the enum values

Packet++/header/ModbusLayer.h
```diff
@@ -0,0 +1,177 @@
...
+			UNKNOWN_FUNCTION = 0xFF
+		};
+
+		/// @struct ModbusReadInputRegisters
+		/// Represents a Modbus request to read input registers.
+		struct ModbusReadInputRegisters
+		{
+			uint16_t startingAddress;  ///< Starting address of the input registers to read
+			uint16_t quantity;         ///< Number of input registers to read
+		};
```
seladb, line 79: This struct is only used in `getFunctionDataSize()`. If we remove it we can remove this struct as well

Packet++/src/ModbusLayer.cpp
```diff
@@ -0,0 +1,101 @@
...
+	void ModbusLayer::setUnitId(uint8_t unitId)
+	{
+		getModbusHeader()->unitId = unitId;
+	}
+
+	void ModbusLayer::setFunctionCode(ModbusLayer::ModbusFunctionCode functionCode)
+	{
+		if (functionCode >= ModbusLayer::ModbusFunctionCode::FUNCTION_CODE_LIMIT)
+		{
+			return;
```
seladb: I think we should log an error using `PCPP_LOG_ERROR`

Tests/Packet++Test/Tests/ModbusTests.cpp
```diff
@@ -0,0 +1,61 @@
...
+	READ_FILE_AND_CREATE_PACKET(1, "PacketExamples/ModbusRequest.dat");
+
+	pcpp::Packet realPacket(&rawPacket1);
+	PTF_ASSERT_TRUE(realPacket.isPacketOfType(pcpp::Modbus));
+	pcpp::ModbusLayer* modbusLayerFromRealPacket = realPacket.getLayerOfType<pcpp::ModbusLayer>();
+
+	pcpp::ModbusLayer modbusLayer(17, 255, pcpp::ModbusLayer::ModbusFunctionCode::READ_INPUT_REGISTERS);
+
+	PTF_ASSERT_BUF_COMPARE(modbusLayer.getData(), modbusLayerFromRealPacket->getData(), modbusLayer.getHeaderLen());
+	PTF_ASSERT_EQUAL(modbusLayer.getDataLen(), modbusLayerFromRealPacket->getDataLen());
```
seladb, line 23: nit: I'd move this test to be before `PTF_ASSERT_BUF_COMPARE`

Tests/Packet++Test/Tests/ModbusTests.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+	timeval time;
+	gettimeofday(&time, nullptr);
+
+	// Transaction ID: 0, Unit ID: 10, Function Code: 17
+	READ_FILE_AND_CREATE_PACKET(1, "PacketExamples/ModbusRequest.dat");
+
+	pcpp::Packet realPacket(&rawPacket1);
+	PTF_ASSERT_TRUE(realPacket.isPacketOfType(pcpp::Modbus));
+	pcpp::ModbusLayer* modbusLayerFromRealPacket = realPacket.getLayerOfType<pcpp::ModbusLayer>();
```
seladb: nit: you can use `auto*` here

Tests/Packet++Test/Tests/ModbusTests.cpp
```diff
@@ -0,0 +1,61 @@
...
+	modbusLayer.setTransactionId(54321);
+	PTF_ASSERT_EQUAL(modbusLayer.getTransactionId(), 54321);
+	modbusLayer.setUnitId(2);
+	PTF_ASSERT_EQUAL(modbusLayer.getUnitId(), 2);
+	modbusLayer.setFunctionCode(pcpp::ModbusLayer::ModbusFunctionCode::WRITE_SINGLE_REGISTER);
+	PTF_ASSERT_EQUAL(static_cast<uint8_t>(modbusLayer.getFunctionCode()),
+	                 static_cast<uint8_t>(pcpp::ModbusLayer::ModbusFunctionCode::WRITE_SINGLE_REGISTER));
+
+	// just to pass the codecov
+	modbusLayer.computeCalculateFields();
```
seladb: Why is this needed to pass codecov? 🤔

Packet++/src/ModbusLayer.cpp
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	ModbusLayer::ModbusFunctionCode ModbusLayer::getFunctionCode() const
+	{
+		ModbusLayer::ModbusFunctionCode functionCode =
+		    static_cast<ModbusLayer::ModbusFunctionCode>(getModbusHeader()->functionCode);
+		if (functionCode >= ModbusLayer::ModbusFunctionCode::FUNCTION_CODE_LIMIT)
+		{
+			return ModbusLayer::ModbusFunctionCode::UNKNOWN_FUNCTION;
+		}
```
seladb: I don't think this code would work because casting an enum value that's not define in the enum results in undefined behavior. What you should do is:
```cpp
switch (getModbusHeader()->functionCode)
{
case 1:
{
    return ModbusFunctionCode::READ_COILS
}
case 2:
{
    return ModbusFunctionCode::READ_DISCRETE_INPUTS;
}
...
default:
{
    return ModbusFunctionCode::UNKNOWN_FUNCTION;
}
}
```

The `FUNCTION_CODE_LIMIT` enum value is redundant.

Packet++/header/ModbusLayer.h
```diff
@@ -0,0 +1,173 @@
...
+	class ModbusLayer : public Layer
+	{
+	public:
+		// @brief Enum class representing Modbus function codes.
+		// This enumeration defines the standard Modbus function codes used in request and response PDUs.
+		// Each value corresponds to a specific operation defined by the Modbus protocol.
+		enum class ModbusFunctionCode : uint8_t
+		{
+			// Read coil status (0x01)
+			READ_COILS = 1,
```
seladb: Our enum value convention is PascalCase, for example: `ReadCoils`. Same for the rest of the enum values

Packet++/src/ModbusLayer.cpp
seladb: This comment wasn't addressed

Packet++/header/ModbusLayer.h
seladb: You changed it to `//` instead of `///`

Packet++/header/ModbusLayer.h
seladb: ditto: should be `///` instead of `//`

seladb:
@yahyayozo please also see older comments:
- https://github.com/seladb/PcapPlusPlus/pull/1823#discussion_r2265164208
- https://github.com/seladb/PcapPlusPlus/pull/1823#discussion_r2278509145
- https://github.com/seladb/PcapPlusPlus/pull/1823#discussion_r2287041041
- https://github.com/seladb/PcapPlusPlus/pull/1823#discussion_r2287041757
- https://github.com/seladb/PcapPlusPlus/pull/1823#discussion_r2278514017

seladb:
@yahyayozo I think the only comments left are:
- https://github.com/seladb/PcapPlusPlus/pull/1823#discussion_r2278509145
- https://github.com/seladb/PcapPlusPlus/pull/1823#discussion_r2278514017

Once you address those I think we can merge this PR

yahyayozo:
@seladb all the comments shall be addressed now
I kept the ModbusReadInputRegisters struct because I use in the constructor and also to keep it as an example for the future implementation of other functions Codes

yahyayozo:
@seladb can you please check why it shows forbid submodules in the failed pre-commit?

Dimi1010:
> @seladb can you please check why it shows forbid submodules in the failed pre-commit?

a7b926daa892f6d5242076be0e1f63c8301cdf95 this commit adds `deps/benchmark-src` to the repo. I think this is a mistake.

yahyayozo:
@seladb @Dimi1010 I removed it
I don't know how it was added as a submodule!!

seladb:
@yahyayozo thank you so much for working on this and contributing to PcapPlusPlus, it is much appreciated! 🙏

yahyayozo:
@seladb so, the other PDUs types will added in another PR?

Dimi1010:
Yes, it's easier to review that way as it results in smaller differential changes in PRs.

seladb:
> Yes, it's easier to review that way as it results in smaller differential changes in PRs.

Yes I agree. Smaller PRs are always better and easier to review

PR #1824, opened 2025-05-19
Improvements to `PcapLiveDevice::sendPacket` and `PcapLiveDevice::sendPackets`.

Dimi1010:
## Changes:
- Added `sendPacketDirect` helper method to encapsulate pure packet sending logic.
- Refactored `sendPacket` to improve readability and provide a final leaf of the call stack in the form of `sendPacketDirect`.
- Added `sendPacketsDirect` helper methods to allow overload of bulk send processing operations when no checks are required.
`WinPcapLiveDevice` will now use `pcap_sendqueue` on `sendPackets(RawPacket const*, int arrLength, bool checkMtu)` when `checkMtu` is false.
- Changed `sendPackets` to forward to `sendPacketsDirect` if `checkMtu` is set to `false`.
- Extracted the `sendPackets` message loop into templated `sendPacketsLoop` function.
- Added new overload `sendPacket(Packet const& packet, bool checkMtu)`.
- Deprecated `sendPacket(Packet const* packet, bool checkMtu)` and updated usages.
- Added `data()` accessor to `PointerVector` to allow usages in C style pointer + size API.

PR #1825, opened 2025-05-20
Improves the readability of `PcapLiveDevice::sendPacket`

Dimi1010:
Split of #1824 

This PR improves the readability of `sendPacket` by providing a designated leaf call `sendPacketDirect` that just handles data transmission.

Additionally the PR reduces indentation the indentation level of the functions by redirecting `if(!checkMtu)` early and avoiding having most of the function be under an if block.

### DIFF
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -602,6 +602,9 @@ namespace pcpp
 	protected:
 		internal::PcapHandle doOpen(const DeviceConfiguration& config);
 
+		// Sends a packet directly to the network.
+		bool sendPacketDirect(uint8_t const* packetData, int packetDataLength);
+
 	private:
 		bool isNflogDevice() const;
 	};
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -531,7 +531,7 @@ namespace pcpp
 		/// - Packet length is 0
 		/// - Packet length is larger than device MTU and checkMtu is true
 		/// - Packet could not be sent due to some error in libpcap/WinPcap/Npcap
-		bool sendPacket(Packet* packet, bool checkMtu = true);
+		bool sendPacket(Packet const* packet, bool checkMtu = true);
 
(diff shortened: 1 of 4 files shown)


Pcap++/header/PcapLiveDevice.h
seladb, line 1: These changes might break the API for users...

In addition, even though this is more correct, I find it less readable

Dimi1010: The `const` additions or the new helper function?

seladb: Sorry, I was referring to the `const` additions

Dimi1010: Ok, I can revert them for now.

Dimi1010: Also, how can it break the API? 

In my experience so far both MSVC and GCC have promoted T* to T const* seamlessly.
It also seems to be written in the standard that `const` qualification can be implicitly added, it is the removal of a qualification that breaks the API. [Link](https://en.cppreference.com/w/cpp/language/implicit_conversion#:~:text=A%20prvalue%20of%20type%20pointer%20to%20cv%2Dqualified%20type%20T%20can%20be%20converted%20to%20a%20prvalue%20pointer%20to%20a%20more%20cv%2Dqualified%20same%20type%20T%20(in%20other%20words%2C%20constness%20and%20volatility%20can%20be%20added).)

seladb: TBH, I'm not an expert when it comes to const 🙈 
Are you sure it can't break any existing code that uses this API (on all platforms)?

Dimi1010: Pretty sure, but we can leave it for another PR that is going to be cleaner to debug if it causes issues. I'm going to revert it here for now.

Dimi1010: Reverted 158f3427d789f70a3e0ce2cb35ffe46b68952197. Makes the PR diff a bit cleaner too. :)

seladb (approved):

Issue #1815, opened 2025-05-14
Crash when handling malformed packet


bratbiswas:
### Bug description

**Describe the bug**
IPReassembly::processPacket(...) function crashes when handling a malformed packet. The payload size indicated in the packet is incorrect - it is far larger than what is actually present. The code does not have a check for correctness of the payload size value in the IP header. That leads to the problem.

I am having trouble adding the PCAP for you to test. Let me know if I can send it through alternate means. Here is the hex dump of the offending packet:

0000   ac 85 3d b9 fe 0d ac 85 3d b9 fd e4 88 47 00 00   ..=.....=....G..
0010   8f 01 4f a6 0f 01 01 04 00 24 00 1d f7 e3 00 00   ..O......$......
0020   00 01 00 00 00 01 ff 03 c0 21 09 14 00 08 43 3e   .........!....C>
0030   f2 1d 00 00 00 00 00 00 00 00 00 00               ............

**Code example to reproduce**
```cpp
void process(pcpp::RawPacket &raw)
{
	pcpp::IPReassembly                     reassembler ;
	pcpp::IPReassembly::ReassemblyStatus   status ;
	pcpp::Packet                           parsedPacket(raw, pcpp::UDP) ;
	pcpp::Packet                          *assembled = r

[bad-packet.tar.gz](https://github.com/user-attachments/files/20201897/bad-packet.tar.gz)

eassembler.processPacket(&parsedPacket, status) ;
	:
	:
	:
	:
}
```

**Expected behavior**
IP reassembler should return MALFORMED_FRAGMENT.

### PcapPlusPlus versions tested on

PcapPlusPlus master branch

### Operating systems tested on

Linux

### Compiler version

GCC 12.3.0

### Packet capture backend (if applicable)

libpcap

bratbiswas:
I have fixed the problem on my setup and I can proceed with this. I'll attach it here - in case you want to take a look.

[IPReassembly.tar.gz](https://github.com/user-attachments/files/20202096/IPReassembly.tar.gz)

seladb:
Thank you @bratbiswas for reporting this issue! Would you consider opening a PR with the fix?

bratbiswas:
I would have loved to. But I am not very conversant with github and might end up messing up things. Should I try? Will you be able to undo if I make mistakes?

seladb:
Sure, I encourage you to try! Don't worry, you won't break or mess up anything 🙂 

[...]

PR #1829, opened 2025-05-21
Malformed packet with incorrect payload length in header caused exception

bratbiswas:
When a malformed IP packet looks like a fragment and contains an invalid payload length in the header, the IPRessembly class trusts the payload length in header and can cause an exception if the length is more than the real length of the payload. A check is introduced validate the lengths and return MALFORMED_PACKET if things are not right.


seladb:
@bratbiswas thanks for contributing this fix! 🙏 

Since you already have a malformed pcap that causes this crash (attached to the [issue](https://github.com/seladb/PcapPlusPlus/issues/1815)), can you add a test here? https://github.com/seladb/PcapPlusPlus/blob/master/Tests/Pcap%2B%2BTest/Tests/IPFragmentationTests.cpp

You can find more info on how to run tests here: https://pcapplusplus.github.io/docs/tests

Let me know if you need my help with writing the test.

Tests/Pcap++Test/Tests/IPFragmentationTests.cpp
```diff
@@ -1073,3 +1074,30 @@ PTF_TEST_CASE(TestIPFragWithPadding)
...
 	delete[] buffer;
 }  // TestIPFragWithPadding
+
+PTF_TEST_CASE(TestIPv4MalformedFragment)
+{
+	timeval time;
+	gettimeofday(&time, nullptr);
+
+	int bufferLength = 0;
+	uint8_t* buffer = readFileIntoBuffer("PcapExamples/ip4_bad_fragment.txt", bufferLength);
```
seladb, line 1084: I believe you also need to push the file `"PcapExamples/ip4_bad_fragment.txt"`?

Maybe instead you can push a pcap file that contains this packet?

Since `*.pcap` is in `.gitignore` you need to use the `-f` flag:
```sh
git add -f Tests/Pcap++Test/PcapExamples/ip4_bad_fragment.pcap
git commit -m "Add pcap file"
```

If you add a pcap file you can use `pcpp::PcapFileReaderDevice` to read it, like we do here: https://github.com/seladb/PcapPlusPlus/blob/51c9af39410c5eb899599357b439959cff00e299/Tests/Pcap%2B%2BTest/Tests/IPFragmentationTests.cpp#L82

seladb, line 1084: Do you also have an example of a malformed IPv6 packet?

bratbiswas: No, I don't have an IPv6 packet.

PR #1832, opened 2025-05-25
Refactored overrides of `WinPcapLiveDevice::startCapture`.

Dimi1010:
This PR replaces overrides of `WinPcapLiveDevice::startCapture` with a simpler `prepareCapture` extensions point.

The change reduces the maintenance burden, allowing derived classes to customize their own behaviour without redirecting the main methods, and allows the base class to exercise greater control on where exactly the extension point is called.

The PR also adds a `final` qualifier to `startCapture` overloads, as a transition step to converting the methods back to non-virtual methods. If anyone has created a derived class of `PcapLiveDevice` for some reason and overriden the method it will result in compile time error.

### DIFF
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -635,6 +635,16 @@ namespace pcpp
 		void getStatistics(IPcapDevice::PcapStats& stats) const override;
 
 	protected:
+		/// @brief Called before starting a capture to prepare the device for capturing packets.
+		///
+		/// This method can be overridden by derived classes to perform additional preparations before starting
+		/// the packet capture.
+		///
+		/// @param asyncCapture True if the capture is asynchronous (i.e. packets are captured in a separate thread),
+		/// @param captureStats True if statistics should be captured during the capture process.
+		virtual void prepareCapture(bool asyncCapture, bool captureStats)
+		{}
+
 		internal::PcapHandle doOpen(const DeviceConfiguration& config);
 
 		// Sends a packet directly to the network.
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -404,7 +404,8 @@ namespace pcpp
 		/// - Capture is already running
 		/// - Device is not opened
 		/// - Capture thread could not be created
-		virtual bool startCapture(OnPacketArrivesCallback onPacketArrives, void* onPacketArrivesUserCookie) final;
+		/// @remarks This method is planned for conversion to non-virtual in the future, so it should not be overridden.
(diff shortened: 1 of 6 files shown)


Pcap++/header/PcapLiveDevice.h
```diff
@@ -404,7 +404,7 @@ namespace pcpp
 		/// - Capture is already running
 		/// - Device is not opened
 		/// - Capture thread could not be created
-		virtual bool startCapture(OnPacketArrivesCallback onPacketArrives, void* onPacketArrivesUserCookie);
+		virtual bool startCapture(OnPacketArrivesCallback onPacketArrives, void* onPacketArrivesUserCookie) final;
```
seladb, line 407: Adding `final` might be a breaking change if someone derived from `PcapLiveDevice`, no? 🤔

Pcap++/header/PcapLiveDevice.h
```diff
@@ -635,6 +635,16 @@ namespace pcpp
...
 
 	protected:
+		/// @brief Called before starting a capture to prepare the device for capturing packets.
+		///
+		/// This method can be overridden by derived classes to perform additional preparations before starting
+		/// the packet capture.
+		///
+		/// @param asyncCapture True if the capture is asynchronous (i.e. packets are captured in a separate thread),
+		/// @param captureStats True if statistics should be captured during the capture process.
+		virtual void prepareCapture(bool asyncCapture, bool captureStats)
```
seladb, line 645: I think this method can return `bool` to decide if the preparation succeeded

Pcap++/header/PcapLiveDevice.h
Dimi1010: It will. It will raise a compile time error for overriding a final method. I did that because it is a better behavior than removing the virtual, which would silently continue to work wrongly.

I don't think user derived classes from PcapLiveDevice are really something many would do, nor something we want to support. None of the other device classes seem to support this. The virtuals are mainly to support the overrides of WinPcapDevice.

If we want to support custom behavior, IMO defining extension points is a better way than allowing the entire method to be overridden.

seladb: I agree we don't necessarily want to support it, but since it's already in the API - do we really want to break it?

Dimi1010: Eh, we can leave it virtual for now, and have a comment that specified that the `virtual` will be removed in the future?
The reason we had the function as virtual is gone and I think `startCapture` is a bit tightly integrated in the internal logic of `PcapLiveDevice` to allow full override, no?

Pcap++/header/PcapLiveDevice.h
seladb: It looks like you have a strong opinion on this one, so I guess we can go with your approach. It's part of the internal implementation of this class and shouldn't impact users

seladb (approved):

PR #1834, opened 2025-05-29
Extracted MTU check code into its own helper functions.

Dimi1010:
This PR separates the logic for checking packets against the device MTU size by moving validation code to `isPayloadWithinMtu` helper functions.

The change simplifies the control flow of `sendPacket` by changing the behaviour from:
```
1. sendPacket(uint8_t* data, size_t dataLen, true /* checkMtu */)
  1.1. sendPacket(RawPacket&, true /* checkMtu */)
    1.1.1. sendPacket(Packet&, true /* checkMtu */)
      1.1.1.1. doMtuCheck(/* ... */) // does check
      1.1.1.2. sendPacket(RawPacket&, false /* no mtu check */)
        1.1.1.2.1. sendPacket(uint8_t* data, size_t dataLen, false /* no mtu check */)
          1.1.1.2.1.1. sendPacketDirect(/* ... */) // does transmission
```
to:
```
1. sendPacket(uint8_t* data, size_t dataLen, true /* checkMtu */)
  1.1. isPayloadWithinMtu(uint8_t* data, size_t dataLen)
    1.1.1. isPayloadWithinMtu(RawPacket&)
      1.1.1.1. isPayloadWithinMtu(Packet&) // does check
  1.2. sendPacketUnchecked(/* ... */) // does transmission
```

It also renames `sendPacketDirect` to `sendPacketUnchecked` to improve readability.


Pcap++/header/PcapLiveDevice.h
```diff
@@ -637,8 +637,55 @@ namespace pcpp
 	protected:
 		internal::PcapHandle doOpen(const DeviceConfiguration& config);
 
+		/// @brief Checks whether the packetPayloadLength is larger than the device MTU.
```
Dimi1010, line 640:
```suggestion
		/// @brief Checks whether the packetPayloadLength is smaller or equal than the device MTU.
```

Pcap++/header/PcapLiveDevice.h
```diff
@@ -637,8 +637,55 @@ namespace pcpp
...
+		/// @param[in] payloadLength The length of the IP layer of the packet
+		/// @return True if the payloadLength is less than or equal to the device MTU
+		bool checkMtu(size_t payloadLength) const;
+
+		/// @brief Checks whether the packet's payload length is larger than the device MTU.
```
Dimi1010, line 645:
```suggestion
		/// @brief Checks whether the packet's payload length is smaller or equal than the device MTU.
```

Pcap++/header/PcapLiveDevice.h
```diff
@@ -637,8 +637,55 @@ namespace pcpp
...
+		/// If allowUnknown is set to true, packets which's length cannot be determined will be allowed.
+		/// In such cases if outPayloadLength is not nullptr, it will be set to the maximum possible payload length.
+		///
+		/// @param packet The packet to check
+		/// @param allowUnknown Controls whether packets with unknown payload length are allowed.
+		/// @param outPayloadLength If not nullptr, the payload length of the packet will be written to this pointer.
+		/// @return True if the packet's payload length is less than or equal to the device MTU.
+		bool checkMtu(Packet const& packet, bool allowUnknown = true, size_t* outPayloadLength = nullptr) const;
+
+		/// @brief Checks whether the payload length of a RawPacket is larger than the device MTU.
```
Dimi1010, line 656:
```suggestion
		/// @brief Checks whether the payload length of a RawPacket is smaller or equal than the device MTU.
```

Pcap++/header/PcapLiveDevice.h
```diff
@@ -637,8 +637,55 @@ namespace pcpp
...
+		/// In such cases, if outPayloadLength is not nullptr, it will be set to the maximum possible payload length.
+		///
+		/// @param[in] rawPacket The RawPacket to check.
+		/// @param[in] allowUnknown Controls whether packets with unknown payload length are allowed.
+		/// @param[out] outPayloadLength If not nullptr, the payload length of the packet will be written to this
+		/// pointer.
+		/// @return True if the packet's payload length is less than or equal to the device MTU.
+		bool checkMtu(RawPacket const& rawPacket, bool allowUnknown = true, size_t* outPayloadLength = nullptr) const;
+
+		/// @brief Checks whether the payload length of a packet's raw data is larger than the device MTU.
```
Dimi1010, line 668:
```suggestion
		/// @brief Checks whether the payload length of a packet's raw data is smaller or equal than the device MTU.
```

Pcap++/header/PcapLiveDevice.h
```diff
@@ -637,8 +637,55 @@ namespace pcpp
...
+		bool checkMtu(size_t payloadLength) const;
+
+		/// @brief Checks whether the packet's payload length is larger than the device MTU.
+		///
+		/// If allowUnknown is set to true, packets which's length cannot be determined will be allowed.
+		/// In such cases if outPayloadLength is not nullptr, it will be set to the maximum possible payload length.
+		///
+		/// @param packet The packet to check
+		/// @param allowUnknown Controls whether packets with unknown payload length are allowed.
+		/// @param outPayloadLength If not nullptr, the payload length of the packet will be written to this pointer.
```
Dimi1010, line 652:
```suggestion
		/// @param[in] packet The packet to check
		/// @param[in] allowUnknown Controls whether packets with unknown payload length are allowed.
		/// @param[out] outPayloadLength If not nullptr, the payload length of the packet will be written to this pointer.
```

Dimi1010 (commented):
Comment fixes.

Dimi1010:
> - It doesn't prevent code duplication because each `checkMtu()` override is only used once (unless I'm missing something)

The initial idea was to separate the logic for checking a packet against the device's MTU into its own function.

The new check functions can also be added as part of the public API of a device as pure checks. I didn't make them public initially to keep it simple for now.

IMO, `doMtuCheck` doesn't work that good as a pure check function in the API due to it raising an error message when the check fails. The user may want a silent check and do some other action on failure silently.

> - IMO, it doesn't really make the code easier to read

It somewhat simplifies all the `sendPacket` flow into:
1. Conditionally call `checkMtu` and log error on fail.
2. Call `sendPacketUnchecked`.

IMO, that is easier to read and understand, instead of having to have `sendPacket` overloads to call each other in circles.

> - Some of the `checkMtu()` overloads have a bit strange API, like the one that checks the MTU but also returns the packet length

Yeah, the out param is a bit strange, but it is the best way to optionally get the packet length that the packet was parsed to without needing to reparse the packet. That can be useful info if the check fails and the user wants to log the error. It was also the best way to keep the return value consistent across overloads.

Hope that answers your questions. :)

seladb:
Having both `doMtuCheck()` as a public method and `checkMtu()` as a protected method is even more confusing IMO. I'm ok with removing the error log in `doMtuCheck()` if you think it's a cleaner API

Dimi1010:
IMO, having the `checkMtu` as public and `doMtuCheck` possibly deprecated/removed would be better?
None of the other public methods have the naming structure of `doMtuCheck`, so it kinda sticks out to me.
The other `do***` methods are mostly protected implementation methods like `doOpen`.

Dimi1010:
Although on second thought, perhaps `checkMtu` is a bit ambiguous as it might be interpreted as a check for the MTU instead of against the MTU.

Perhaps changing the method name to `fitsInMtu` or something similar would be better?

Dimi1010:
@seladb Changed the function name to be more explicit and moved the new functions to public API.
Also added a new section to the PR post explaining how the change improves readability by simplifying the control flow of the device internals.

seladb:
@Dimi1010 I'm still not convinced this change is needed. We add multiple public methods to `PcapLiveDevice` which makes the API more complex for no apparent reason. Also - the existing logic (before this change) is clear and readable enough and I don't think we need this refactoring

Dimi1010:
@seladb IDK, about the previous internal logic being clear. I spent a significant amount of time trying to wrap my head around the original call chains of `sendPacket` initially, due to them going in both directions depending on the `checkMtu` flag.

When `checkMtu` is `false`, the overloads defer in the direction of `Packet` -> `RawPacket` -> `C array`.
When `checkMtu` is `true`, the overloads defer in the direction of `C array` -> `RawPacket` -> `Packet` -> `RawPacket` -> `C array`.

IMO, the MTU check's logic is complex enough that having it as a dedicated function family improves the readability as it abstracts away the "backwards" control flow in `sendPacket` from `C array` to `Packet` due to needing to parse the packet.

It provides two distinct logical units to think about:
- sendPacket - defers between overloads always go `Packet` -> `RawPacket` -> `C array`
- isPayloadWithinMtu - defers between overloads always go `C array` -> `RawPacket` -> `Packet`

Instead the previous implementation had a single logic unit that conditionally branches on every possible level like this:
- `C array` (A) -> Can sometimes go to (B) and sometimes transmits.
- `RawPacket` (B) -> Can sometimes go to (A) and sometimes go to (C)
- `Packet` (C) -> Can sometimes go to (B) and sometimes do MTU check and go to (A)


> We add multiple public methods to PcapLiveDevice which makes the API more complex for no apparent reason.

I have no strong opinion on if the helper methods that encapsulate the MTU validation logic are public or not. It is mostly because `doMtuCheck` is already `public` and we do provide the ability for users to perform a query against the MTU size, that the new functions are `public`.

In this respect, the extra overloads can be considered utility functions of convenience, so the query can be performed against more common library objects (i.e., `Packet`, `RawPacket`), instead of having the user manually have to extract the payload length.

seladb:
ok @Dimi1010 let's go with your suggestion. But can we make the new MTU check methods private? We can leave the existing `doMtuCheck()` public even though it's not ideal. Please let me know what you think

Dimi1010:
@seladb moved the declarations to protected, or do you want them explicitly in private?

seladb (approved):

Issue #1835, opened 2025-05-30
Parse X509 certificates


seladb:
Prerequisite PRs:
- https://github.com/seladb/PcapPlusPlus/pull/1836
- https://github.com/seladb/PcapPlusPlus/pull/1837
- https://github.com/seladb/PcapPlusPlus/pull/1850
- https://github.com/seladb/PcapPlusPlus/pull/1851
- https://github.com/seladb/PcapPlusPlus/pull/1894
- https://github.com/seladb/PcapPlusPlus/pull/1906

X509 decoding PRs:
- https://github.com/seladb/PcapPlusPlus/pull/1871
- https://github.com/seladb/PcapPlusPlus/pull/1882
- https://github.com/seladb/PcapPlusPlus/pull/1913

Parse X509 certificates in SSL/TLS messages:
- https://github.com/seladb/PcapPlusPlus/pull/1915

X509 parsing example:
- https://github.com/seladb/PcapPlusPlus/pull/1917

bhaskarbhar:
@seladb While looking for this issue, I went through this function:

```
SSLx509Certificate* SSLCertificateMessage::getCertificate(int index) const
	{
		if (index < 0 || index > (int)m_CertificateList.size())
		{
			PCPP_LOG_DEBUG("certificate index out of range: asked for index " << index << ", total size is "
			                                                                  << m_CertificateList.size());
			return nullptr;
		}

		return const_cast<SSLx509Certificate*>(m_CertificateList.at(index));
	}

```

is the bound check correct here? 
`if (index < 0 || index > (int)m_CertificateList.size())`

It should be 
`if (index < 0 || index >= (int)m_CertificateList.size())`
????

seladb:
> [@seladb](https://github.com/seladb) While looking for this issue, I went through this function in SSLHandshake.cpp
> 
> ```
> SSLx509Certificate* SSLCertificateMessage::getCertificate(int index) const
> 	{
> 		if (index < 0 || index > (int)m_CertificateList.size())
> 		{
> 			PCPP_LOG_DEBUG("certificate index out of range: asked for index " << index << ", total size is "
> 			                                                                  << m_CertificateList.size());
> 			return nullptr;
> 		}
> 
> 		return const_cast<SSLx509Certificate*>(m_CertificateList.at(index));
> 	}
> ```
> 
> is the bound check correct here? `if (index < 0 || index > (int)m_CertificateList.size())`
> 
[...]

PR #1836, opened 2025-05-30
Update `Asn1IntegerRecord` to support integers of any size

seladb:
As part of https://github.com/seladb/PcapPlusPlus/issues/1835, X509 certificates require support for ASN.1 integers of more than 4 bytes

### DIFF
--- a/Packet++/header/Asn1Codec.h
+++ b/Packet++/header/Asn1Codec.h
@@ -4,6 +4,7 @@
 #include <memory>
 #include <typeinfo>
 #include <stdexcept>
+#include <sstream>
 #include "PointerVector.h"
 
 /// @file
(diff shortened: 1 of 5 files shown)


Packet++/header/Asn1Codec.h
```diff
@@ -390,7 +415,64 @@ namespace pcpp
...
+			template <typename T, EnableIfUnsignedIntegral<T> = 0> bool canFit() const
+			{
+				return sizeof(T) >= (m_Value.size() + 1) / 2;
+			}
+
+			std::string toString() const;
+			std::vector<uint8_t> toBytes() const;
+
+		private:
+			std::string m_Value;
```
Dimi1010, line 463: Wouldn't it be better to store big int as vector of uint8_t with little endian encoding?

Packet++/header/Asn1Codec.h
```diff
@@ -390,7 +415,64 @@ namespace pcpp
...
+				return *this;
+			}
+			BigInt& operator=(const std::string& value);
+			size_t size() const;
+
+			template <typename T, EnableIfUnsignedIntegral<T> = 0> T getInt() const
+			{
+				if (!canFit<T>())
+				{
+					throw std::invalid_argument("Value cannot fit into requested int type");
```
Dimi1010, line 443: std::overflow_error would be a better fit here imo.

Dimi1010 (approved):
Minor comment but LGTM.

Packet++/header/Asn1Codec.h
seladb: I was contemplating a lot between `std::string` and `std::vector<uint8_t>` but decided to go with `std::string` because most of the numbers stored in ASN.1 records are actually small numbers <= 4 bytes, so the footprint of a string is actually slightly better than a vector. Also - we don't need any arithmetic operations on `BigInt`, we only need to convert between unsigned int and string, so there is no huge overhead in terms of runtime efficiency

PR #1837, opened 2025-05-30
Add decoding and encoding of `UTCTime` and `GeneralizedTime` ASN.1 records

seladb:
As part of https://github.com/seladb/PcapPlusPlus/issues/1835, X509 certificates require support for ASN.1 records of `UTCTime` and `GeneralizedTime`

### DIFF
--- a/Packet++/header/Asn1Codec.h
+++ b/Packet++/header/Asn1Codec.h
@@ -5,9 +5,8 @@
 #include <typeinfo>
 #include <stdexcept>
 #include <sstream>
-#include "PointerVector.h"
-
 #include <chrono>
+#include "PointerVector.h"
 
 /// @file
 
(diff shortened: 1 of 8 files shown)


Packet++/header/Asn1Codec.h
```diff
@@ -574,4 +576,89 @@ namespace pcpp
...
+	/// @class Asn1TimeRecord
+	/// An abstract class for representing ASN.1 time records (UTCTime and GeneralizedTime).
+	/// This class is not instantiable, users should use either Asn1UtcTimeRecord or Asn1GeneralizedTimeRecord
+	class Asn1TimeRecord : public Asn1PrimitiveRecord
+	{
+	public:
+		/// @return The time-point value of this record
+		/// @param[in] timezone A timezone string - should be in the format of "Z" for UTC or +=HHMM for other
+		/// timezones. The default value is UTC
+		/// @throws std::invalid_argument if timezone is not in the correct format
```
Dimi1010: Documentation order usually goes `brief`, `params`, `return`, `throws`. Return is above `params` currently.

```suggestion
		/// @param[in] timezone A timezone string - should be in the format of "Z" for UTC or +=HHMM for other
		/// timezones. The default value is UTC
		/// @return The time-point value of this record
		/// @throws std::invalid_argument if timezone is not in the correct format
```

Packet++/header/Asn1Codec.h
```diff
@@ -574,4 +576,89 @@ namespace pcpp
...
+			decodeValueIfNeeded();
+			return adjustTimezones(m_Value, "Z", timezone);
+		};
+
+		/// @param[in] format Requested value format
+		/// @param[in] timezone A timezone string - should be in the format of "Z" for UTC or +=HHMM for other
+		/// timezones. The default value is UTC
+		/// @param[in] includeMilliseconds Should Include milliseconds in the returned string
+		/// @throws std::invalid_argument if timezone is not in the correct format
+		/// @return The value as string
```
Dimi1010: ditto, documentation order. `throws` is before `return` here

Packet++/header/Asn1Codec.h
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/1837/commits/8c1fbf6b0ccd61579a25de8ee8cb1094e3de8ab9

Packet++/header/Asn1Codec.h
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/1837/commits/8c1fbf6b0ccd61579a25de8ee8cb1094e3de8ab9

seladb:
> Looks good code wise.
> 
> A minor issue with the timepoint params, tho. IMO, every method that takes a timepoint should be templated to allow usages of the dirrerent clocks provided by the standard via a SFINAE template as the current implementation requires usage of system clock timepoints. High res and steady clock timepoints can't be used there. This is fine for now, but we might want a to-do or issue open for that.

Sure, maybe we can address it later as needed

PR #1842, opened 2025-06-01
Cleanup of RawPacket construction.

Dimi1010:
This PR updates the construction implementation of RawPacket to utilize C++ 11 practices.

Changes include:
- `init()` method has been replaced with default member initializers.
- Changed constructor implementation to utilize member initialization list.

### DIFF
--- a/Packet++/header/RawPacket.h
+++ b/Packet++/header/RawPacket.h
@@ -262,7 +262,7 @@ namespace pcpp
 		int m_RawDataLen = 0;
 		int m_FrameLength = 0;
 		timespec m_TimeStamp{};  // Zero initialized
-		bool m_DeleteRawDataAtDestructor = false;
+		bool m_DeleteRawDataAtDestructor = true;
 		bool m_RawPacketSet = false;
 		LinkLayerType m_LinkLayerType = LinkLayerType::LINKTYPE_ETHERNET;
 
@@ -272,8 +272,10 @@ namespace pcpp
 		/// A default constructor that initializes class'es attributes to default value:
 		/// - data pointer is set to nullptr
 		/// - data length is set to 0
+		/// - frame length is set to 0
 		/// - deleteRawDataAtDestructor is set to 'true'
 		/// - timestamp is set to 0
+		/// - layer type is set to Ethernet
 		RawPacket() = default;
 
 		/// A constructor that receives a pointer to the raw data (allocated elsewhere). This constructor is usually
--- a/Packet++/header/RawPacket.h
+++ b/Packet++/header/RawPacket.h
@@ -258,17 +258,24 @@ namespace pcpp
 	class RawPacket
 	{
 	protected:
-		uint8_t* m_RawData;
-		int m_RawDataLen;
-		int m_FrameLength;
-		timespec m_TimeStamp;
-		bool m_DeleteRawDataAtDestructor;
-		bool m_RawPacketSet;
-		LinkLayerType m_LinkLayerType;
-		void init(bool deleteRawDataAtDestructor = true);
+		uint8_t* m_RawData = nullptr;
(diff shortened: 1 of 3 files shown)


Packet++/header/RawPacket.h
```diff
@@ -258,17 +258,26 @@ namespace pcpp
...
-		timespec m_TimeStamp;
-		bool m_DeleteRawDataAtDestructor;
-		bool m_RawPacketSet;
-		LinkLayerType m_LinkLayerType;
-		void init(bool deleteRawDataAtDestructor = true);
+		uint8_t* m_RawData = nullptr;
+		int m_RawDataLen = 0;
+		int m_FrameLength = 0;
+		timespec m_TimeStamp{};  // Zero initialized
+		bool m_DeleteRawDataAtDestructor = true;
```
Dimi1010, line 265: IMO, this should default to `false` as by default there is nothing to "own".

Dimi1010: Also, the flag itself might be better named as `m_OwnsDataBuffer` as at the end of the day that is what the flag represents?

Packet++/src/RawPacket.cpp
```diff
@@ -112,7 +102,7 @@ namespace pcpp
 	bool RawPacket::initWithRawData(const uint8_t* pRawData, int rawDataLen, timespec timestamp,
```
Dimi1010, line 100: Curious as to why this method exists, as opposed to using a constructor to initialize?

Dimi1010 (commented):
## Some other design notes on the class:

### `setRawData` is ambiguous in buffer ownership semantics.
During runtime the method `setRawData` does a conditional ownership transfer. The choice is controlled by `m_DeleteRawDataAtDestructor` which is set at construction time and the user has no way to query afterwards. If wants to replace the buffer on a `RawPacket` that has previously been in "view" mode, the packet will not take ownership over the provided buffer, and the user has no way to know that.

A recommended fix would be the split of setRawData into several methods(or overloads w/ tag dispatch) for the different situations:
#### Assign a buffer and transfer ownership of it to the `RawPacket` object.
In this case the buffer ownership is transferred to the packet and the class can handle it as it sees fit. The provided pointer should be considered invalid and the user should not use it directly. Instead the user should query the packet object via `getRawData`.

#### Copy an input buffer data into the `RawPacket` object.
The packet object will consider the input buffer as read-only and copy the contents to an internal buffer.

#### Enter "View" ("Proxy") mode over a buffer.
A "view" mode is considered a mode where the packet object will attempt to read/write from a buffer it does not own. When such an operation is called the `RawPacket` object will attempt to enter such mode with the selected input buffer. The success of that would be dependent on the specifics of the `RawPacket` implementation. For example `MBufRawPacket` would not support such a mode, as it is tightly integrated with DPDK's memory pools and as such can't work with regular memory.

### Separate `RawPacket` and `MBufRawPacket` and extract a common base class.
Currently `RawPacket` is made to utilize standard memory and `MBufRawPacket` is tightly integrated in the DPDK memory pool ecosystem. Their separation would improve code maintainability as it would remove the need of both classes to work around each other for their concrete implementations.

Packet++/src/RawPacket.cpp
seladb: TBH I don't remember 🙈 
But it's not used anywhere...

seladb (approved):

Packet++/src/RawPacket.cpp
```diff
@@ -7,37 +7,27 @@
 
 namespace pcpp
 {
-
-	void RawPacket::init(bool deleteRawDataAtDestructor)
+	namespace
 	{
-		m_RawData = nullptr;
```
tigercosmos, line 13: Cleaner now!

PR #1843, opened 2025-06-02
Wrap `timespec` <-> `timeval` conversion in helpers.

Dimi1010:
This PR adds helper functions `toTimespec` and `toTimeval` to encapsulate conversion logic.
The helpers are defined in `TimespecTimeval.h` and are expected to be in inlined when optimizations are enabled.


tigercosmos (approved):
LGTM

PR #1846, opened 2025-06-07
C style cast cleanup.

Dimi1010:
- Replaced or removed C style casts with C++ casts.
 - Replaced iterator based for loops with range based for loops.


tigercosmos (approved):
Nice work, thanks!

PR #1847, opened 2025-06-07
Added sized overloads for buffer construction of `IPv6Address` and `MacAddress`.

Dimi1010:
- Added new constructors: for when the buffer size isn't guaranteed.
  - `IPv6Address(const uint8_t* buf, size_t size)` 
  - `MacAddress(const uint8_t* buf, size_t size)`
- Added sized overloads of `copyTo`
  - `size_t IPv6Address::copyTo(uint8_t* buf, size_t size)`
  - `size_t MacAddress::copyTo(uint8_t* buf, size_t size)`
- Replaced allocating `copyTo(uint8_t** buf)` and `copyTo(uint8_t** buf, size_t& size)` with `copyToNewBuffer`.

### DIFF
--- a/3rdParty/LightPcapNg/LightPcapNg/src/light_platform.c
+++ b/3rdParty/LightPcapNg/LightPcapNg/src/light_platform.c
@@ -77,6 +77,7 @@ light_file light_open(const char *file_name, const __read_mode_t mode)
 	{
 		if (light_is_compressed_file(file_name))
 		{
+			free(fd);  // PCPP PATCH
 			return light_open_decompression(file_name, mode);
 		}
 		fd->file = fopen(file_name, "rb");
--- a/Common++/header/Logger.h
+++ b/Common++/header/Logger.h
@@ -27,8 +27,9 @@
 // Allows for conditional removal of unwanted log calls at compile time.
 #define PCPP_LOG_LEVEL_OFF 0
 #define PCPP_LOG_LEVEL_ERROR 1
-#define PCPP_LOG_LEVEL_INFO 2
-#define PCPP_LOG_LEVEL_DEBUG 3
+#define PCPP_LOG_LEVEL_WARN 2
+#define PCPP_LOG_LEVEL_INFO 3
+#define PCPP_LOG_LEVEL_DEBUG 4
 
 // All log messages built via a PCPP_LOG_* macro below the PCPP_ACTIVE_LOG_LEVEL will be removed at compile time.
 // Uses the PCPP_ACTIVE_LOG_LEVEL if it is defined, otherwise defaults to PCAP_LOG_LEVEL_DEBUG
@@ -146,6 +147,7 @@ namespace pcpp
 	{
 		Off = PCPP_LOG_LEVEL_OFF,      ///< No log messages are emitted.
 		Error = PCPP_LOG_LEVEL_ERROR,  ///< Error level logs are emitted.
+		Warn = PCPP_LOG_LEVEL_WARN,    ///< Warning level logs and above are emitted.
 		Info = PCPP_LOG_LEVEL_INFO,    ///< Info level logs and above are emitted.
 		Debug = PCPP_LOG_LEVEL_DEBUG   ///< Debug level logs and above are emitted.
 	};
@@ -428,6 +430,12 @@ namespace pcpp
 #	define PCPP_LOG_DEBUG(message) PCPP_LOG(pcpp::LogLevel::Debug, message)
 #else
 #	define PCPP_LOG_DEBUG(message) (void)0
+#endif
+
+#if PCPP_ACTIVE_LOG_LEVEL >= PCPP_LOG_LEVEL_WARN
+#	define PCPP_LOG_WARN(message) PCPP_LOG(pcpp::LogLevel::Warn, message)
+#else
+#	define PCPP_LOG_WARN(message) (void)0
 #endif
 
 #if PCPP_ACTIVE_LOG_LEVEL >= PCPP_LOG_LEVEL_INFO
--- a/Common++/src/SystemUtils.cpp
+++ b/Common++/src/SystemUtils.cpp
@@ -31,22 +31,18 @@
 #endif
 
 #ifdef _MSC_VER
+#	include <chrono>
 int gettimeofday(struct timeval* tp, struct timezone* tzp)
 {
-	// Note: some broken versions only have 8 trailing zero's, the correct epoch has 9 trailing zero's
-	static const uint64_t EPOCH = ((uint64_t)116444736000000000ULL);
 
-	SYSTEMTIME system_time;
-	FILETIME file_time;
-	uint64_t time;
+	auto now = std::chrono::system_clock::now();
+	auto duration = now.time_since_epoch();
 
-	GetSystemTime(&system_time);
-	SystemTimeToFileTime(&system_time, &file_time);
-	time = ((uint64_t)file_time.dwLowDateTime);
-	time += ((uint64_t)file_time.dwHighDateTime) << 32;
+	auto seconds = std::chrono::duration_cast<std::chrono::seconds>(duration);
+	auto microseconds = std::chrono::duration_cast<std::chrono::microseconds>(duration - seconds);
 
-	tp->tv_sec = (long)((time - EPOCH) / 10000000L);
-	tp->tv_usec = (long)(system_time.wMilliseconds * 1000);
+	tp->tv_sec = static_cast<long>(seconds.count());
+	tp->tv_usec = static_cast<long>(microseconds.count());
 	return 0;
 }
 #endif
@@ -211,63 +207,18 @@ namespace pcpp
 
 	int clockGetTime(long& sec, long& nsec)
 	{
-		sec = 0;
-		nsec = 0;
+		using namespace std::chrono;
 
-#if defined(_WIN32)
+		auto now = system_clock::now();
+		auto duration = now.time_since_epoch();
 
-#	define CLOCK_GETTIME_BILLION (1E9)
+		auto secondsDuration = duration_cast<seconds>(duration);
+		auto nanosecondsDuration = duration_cast<nanoseconds>(duration - secondsDuration);
 
-		static BOOL clock_gettime_first_time = 1;
-		static LARGE_INTEGER clock_gettime_counts_per_sec;
-
-		LARGE_INTEGER count;
-
-		if (clock_gettime_first_time)
-		{
-			clock_gettime_first_time = 0;
-
-			if (0 == QueryPerformanceFrequency(&clock_gettime_counts_per_sec))
-			{
-				clock_gettime_counts_per_sec.QuadPart = 0;
-			}
-		}
-
-		if ((clock_gettime_counts_per_sec.QuadPart <= 0) || (0 == QueryPerformanceCounter(&count)))
-		{
-			return -1;
-		}
-
-		sec = count.QuadPart / clock_gettime_counts_per_sec.QuadPart;
-		nsec = ((count.QuadPart % clock_gettime_counts_per_sec.QuadPart) * CLOCK_GETTIME_BILLION) /
-		       clock_gettime_counts_per_sec.QuadPart;
+		sec = static_cast<long>(secondsDuration.count());
+		nsec = static_cast<long>(nanosecondsDuration.count());
 
 		return 0;
-
-#elif defined(__APPLE__)
-
-		clock_serv_t cclock;
-		mach_timespec_t mts;
-		host_get_clock_service(mach_host_self(), CALENDAR_CLOCK, &cclock);
-		clock_get_time(cclock, &mts);
-		mach_port_deallocate(mach_task_self(), cclock);
-		sec = mts.tv_sec;
-		nsec = mts.tv_nsec;
-
-		return 0;
-
-#else  // Linux
-
-		timespec tspec{};
-		const int res = clock_gettime(CLOCK_REALTIME, &tspec);
-		if (res == 0)
-		{
-			sec = tspec.tv_sec;
-			nsec = tspec.tv_nsec;
-		}
-		return res;
-
-#endif
 	}
 
 	time_t mkUtcTime(std::tm& tm)
--- a/Examples/IcmpFileTransfer/IcmpFileTransfer-pitcher.cpp
+++ b/Examples/IcmpFileTransfer/IcmpFileTransfer-pitcher.cpp
@@ -316,7 +316,7 @@ void receiveFile(pcpp::IPv4Address pitcherIP, pcpp::IPv4Address catcherIP, int p
 
 			// if rate limit was set by the user, sleep between sending packets
 			if (packetPerSec > 1)
-				usleep(sleepBetweenPackets);
+				std::this_thread::sleep_for(std::chrono::microseconds(sleepBetweenPackets));
 			else if (packetPerSec == 1)
 				std::this_thread::sleep_for(std::chrono::seconds(1));
 
@@ -334,7 +334,7 @@ void receiveFile(pcpp::IPv4Address pitcherIP, pcpp::IPv4Address catcherIP, int p
 			{
 				sendIcmpRequest(dev, pitcherMacAddr, catcherMacAddr, pitcherIP, catcherIP, icmpId, ICMP_FT_ABORT,
 				                nullptr, 0);
-				usleep(SLEEP_BETWEEN_ABORT_MESSAGES);
+				std::this_thread::sleep_for(std::chrono::microseconds(SLEEP_BETWEEN_ABORT_MESSAGES));
 			}
 
 			file.close();
@@ -502,7 +502,7 @@ void sendFile(const std::string& filePath, pcpp::IPv4Address pitcherIP, pcpp::IP
 
 			// use usleep or sleep (see comment a few lines below)
 			if (packetPerSec > 1)
-				usleep(sleepBetweenPackets);
+				std::this_thread::sleep_for(std::chrono::microseconds(sleepBetweenPackets));
 			else if (packetPerSec == 1)
 				std::this_thread::sleep_for(std::chrono::seconds(1));
 
--- a/Pcap++/src/DpdkDevice.cpp
+++ b/Pcap++/src/DpdkDevice.cpp
@@ -20,6 +20,8 @@
 #include "rte_cycles.h"
 #include <string>
 #include <unistd.h>
+#include <chrono>
+#include <thread>
 
 #define MAX_BURST_SIZE 64
 
@@ -1115,7 +1117,7 @@ namespace pcpp
 					{
 						PCPP_LOG_DEBUG(
 						    "Since NIC couldn't send all packet in this iteration, waiting for 0.2 second for H/W descriptors to get free");
-						usleep(200000);
+						std::this_thread::sleep_for(std::chrono::microseconds(200000));
 						lastSleep = packetsSent;
 					}
 				}
--- a/Pcap++/src/KniDevice.cpp
+++ b/Pcap++/src/KniDevice.cpp
@@ -7,7 +7,6 @@
 #include "SystemUtils.h"
 
 #include <unistd.h>
-#include <time.h>
 #include <thread>
 #include <sys/ioctl.h>
 #include <net/if.h>
@@ -410,7 +409,8 @@ namespace pcpp
 		struct rte_kni* kni_dev = device->m_Device;
 		for (;;)
 		{
-			nanosleep(&sleepTime, nullptr);
+			std::this_thread::sleep_for(std::chrono::seconds(sleepTime.tv_sec) +
+			                            std::chrono::nanoseconds(sleepTime.tv_nsec));
 			rte_kni_handle_request(kni_dev);
 			if (stopThread)
 			{
--- a/Pcap++/src/NetworkUtils.cpp
+++ b/Pcap++/src/NetworkUtils.cpp
@@ -31,7 +31,7 @@ namespace pcpp
 		std::mutex& mutex;
 		std::condition_variable& cond;
 		IPv4Address ipAddr;
-		clock_t start;
+		std::chrono::steady_clock::time_point start;
 		MacAddress result;
 		double arpResponseTime;
 	};
@@ -39,7 +39,7 @@ namespace pcpp
 	static void arpPacketReceived(RawPacket* rawPacket, PcapLiveDevice*, void* userCookie)
 	{
 		// extract timestamp of packet
-		clock_t receiveTime = clock();
+		auto receiveTime = std::chrono::steady_clock::now();
 
 		// get the data from the main thread
 		ArpingReceivedData* data = (ArpingReceivedData*)userCookie;
@@ -66,8 +66,8 @@ namespace pcpp
 			return;
 
 		// measure response time
-		double diffticks = receiveTime - data->start;
-		double diffms = (diffticks * 1000) / CLOCKS_PER_SEC;
+		auto duration = receiveTime - data->start;
+		auto diffms = std::chrono::duration<double, std::milli>(duration).count();
 
 		data->arpResponseTime = diffms;
 		data->result = arpReplyLayer->getSenderMacAddress();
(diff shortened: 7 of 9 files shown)


Common++/header/IpAddress.h
```diff
@@ -39,8 +41,21 @@ namespace pcpp
 
 		/// A constructor that creates an instance of the class out of 4-byte array.
 		/// @param[in] bytes The address as 4-byte array in network byte order
-		IPv4Address(const uint8_t bytes[4])
+		/// @remarks This constructor assumes that the provided array is exactly 4 bytes long.
+		/// Prefer using the constructor with size parameter if the array length is not guaranteed to be 4 bytes.
+		IPv4Address(const uint8_t bytes[4]) : IPv4Address(bytes, 4)
```
Dimi1010, line 46: Unsure if we want to deprecate these? Generally I dislike API that just assumes the buffer size is enough. 🤔

Common++/header/IpAddress.h
```diff
@@ -160,9 +175,22 @@ namespace pcpp
 
 		/// A constructor that creates an instance of the class out of 16-byte array.
 		/// @param[in] bytes The address as 16-byte array in network byte order
-		IPv6Address(const uint8_t bytes[16])
+		/// remarks This constructor assumes that the provided array is exactly 16 bytes long.
+		/// Prefer using the constructor with size parameter if the array length is not guaranteed to be 16 bytes.
+		IPv6Address(const uint8_t bytes[16]) : IPv6Address(bytes, 16)
```
Dimi1010, line 174: ditto

Common++/header/MacAddress.h
```diff
@@ -28,8 +31,20 @@ namespace pcpp
 		/// The byte array length should be 6 (as MAC address is 6-byte long), and the remaining bytes are ignored.
 		/// If the byte array is invalid, the constructor throws an exception.
 		/// @param[in] addr A pointer to the byte array containing 6 bytes representing the MAC address
-		explicit MacAddress(const uint8_t addr[6])
+		/// @deprecated Deprecated in favor of sized MacAddress(const uint8_t addr[6], size_t size) constructor.
+		explicit MacAddress(const uint8_t addr[6]) : MacAddress(addr, 6)
```
Dimi1010, line 36: ditto

Common++/header/MacAddress.h
```diff
@@ -127,18 +143,47 @@ namespace pcpp
...
+			*arr = copyToNewBuffer(size).release();
 		}
 
 		/// Gets a pointer to an already allocated byte array and copies the address value to it.
 		/// This method assumes array allocated size is at least 6 (the size of a MAC address)
 		/// @param[in] arr A pointer to the array which address will be copied to
+		/// @remarks This method assumes that the provided array is at least 6 bytes long.
+		/// Prefer using the copyTo(uint8_t* buffer, size_t size) method if the array length is not guaranteed to be 6
+		/// bytes.
 		void copyTo(uint8_t arr[6]) const
```
Dimi1010, line 155: ditto

Common++/header/IpAddress.h
```diff
@@ -224,16 +252,39 @@ namespace pcpp
...
+		PCPP_DEPRECATED("Use copyToNewBuffer instead.")
 		void copyTo(uint8_t** arr, size_t& length) const;
 
 		/// Gets a pointer to an already allocated byte array and copies the address value to it.
 		/// This method assumes array allocated size is at least 16 (the size of an IPv6 address)
 		/// @param[in] arr A pointer to the array which address will be copied to
+		/// @remarks This method assumes that the provided array is at least 16 bytes long.
+		/// Prefer using the copyTo(uint8_t* buffer, size_t size) method if the array length is not guaranteed to be 16
+		/// bytes.
 		void copyTo(uint8_t* arr) const
```
Dimi1010, line 254: ditto

Common++/header/IpAddress.h
```diff
@@ -224,16 +252,39 @@ namespace pcpp
...
+			{
+				throw std::out_of_range("Buffer size is smaller than IPv6 address size");
+			}
+			memcpy(buffer, m_Bytes.data(), m_Bytes.size() * sizeof(uint8_t));
 		}
 
+		/// @brief Allocates a byte array and copies address value into it.
+		/// @param[out] size Returns the size in bytes of the allocated array. Usually 16.
+		/// @return A unique pointer to the allocated byte array containing the address value
+		std::unique_ptr<uint8_t[]> copyToNewBuffer(size_t& size) const;
```
Dimi1010: Debating if its better to return `unique_ptr` with size as an out param, or return `pair<unique_ptr<...>, size_t>`

PS: Kinda starting to lean to pair as it will keep the buffer and size logically grouped? 🤔

Common++/header/MacAddress.h
```diff
@@ -127,18 +143,47 @@ namespace pcpp
...
+			{
+				throw std::out_of_range("Buffer size is smaller than MAC address size");
+			}
+			std::copy(m_Address.begin(), m_Address.end(), buffer);
+		}
+
+		/// @brief Allocates a byte array and copies address value into it.
+		/// @param[out] size The size of the allocated array in bytes. Usually 6.
+		/// @return A unique pointer to the allocated byte array containing the MAC address.
+		std::unique_ptr<uint8_t[]> copyToNewBuffer(size_t& size) const
```
Dimi1010: ditto

Common++/header/IpAddress.h
tigercosmos: Instead, we can provide `IPv4Address(const std::array<uint8_t, 4> bytes)`

Common++/header/IpAddress.h
tigercosmos: We can consider to use `std::array` here as well?

Common++/header/IpAddress.h
Dimi1010: We already do. It's the next overload down.

Common++/header/IpAddress.h
Dimi1010: There is `toByteArray` that returns an `std::array` type.

This one mostly covers existing functionality that was in `copyTo` that allocates heap buffer. Which I suppose can be cheaper if it will be changing ownership around a lot?

Common++/header/IpAddress.h
Dimi1010: Some IP objects are constructed from raw buffers in the library, tho. Like when parsing IP from packet structures.

tigercosmos: I see. I would still vote for removing this to make it more C++.

Common++/header/IpAddress.h
tigercosmos: I am okay with having `std::unique_ptr` API, but I wonder if people need it. IMO, most of the time, people may only need the reference to the bytes directly from the `IpAddress` object. In rare cases, they can just copy the bytes. Then, we can keep the interface as simple as possible.

Dimi1010: That is true. I can't find a lot of usages where a new buffer is required. Most of the time it is copied to an already existing buffer, if the copyTo api is used.

Common++/header/MacAddress.h
```diff
@@ -127,18 +143,47 @@ namespace pcpp
 		/// Allocates a byte array of length 6 and copies address value into it. Array deallocation is user
 		/// responsibility
 		/// @param[in] arr A pointer to where array will be allocated
+		/// @deprecated Use copyToNewBuffer instead as it returns a unique pointer to the allocated array.
+		PCPP_DEPRECATED("Use copyToNewBuffer instead as it returns a unique pointer to the allocated array.")
 		void copyTo(uint8_t** arr) const
 		{
-			*arr = new uint8_t[m_Address.size()];
-			std::copy(m_Address.begin(), m_Address.end(), *arr);
+			size_t size;
```
tigercosmos: It seems incorrect here as the `size` is unassigned. I guess you want `size = m_Address.size()`?

Common++/header/MacAddress.h
```diff
@@ -127,18 +143,47 @@ namespace pcpp
...
+			return copyTo(arr, 6);
+		}
+
+		/// @brief Copies the address value to a user-provided buffer.
+		/// @param[in] buffer A pointer to the buffer where the address will be copied
+		/// @param[in] size The size of the buffer in bytes
+		/// @throws std::out_of_range If the provided size is smaller than 6 bytes.
+		void copyTo(uint8_t* buffer, size_t size) const
+		{
+			if (size < m_Address.size())
```
tigercosmos, line 171: I think we should compare if the size is 6?(then maybe we don't need to compare it?) It doesn't make sense to me to provide either a bigger or a smaller size.

Common++/header/MacAddress.h
Dimi1010: It actually doesn't matter. `copyToNewBuffer` reports the size of the new buffer via an out param so it gets assigned there. In this instance we don't actually use it anywhere, but it is a required param for method signature.

Maybe it can be better to have the return be `pair<u_ptr, size_t>`?

Tests/Pcap++Test/Tests/IpMacTests.cpp
```diff
@@ -270,15 +270,15 @@ PTF_TEST_CASE(TestMacAddress)
 	oss << macAddr1;
 	PTF_ASSERT_EQUAL(oss.str(), "11:02:33:04:55:06");
 
-	uint8_t* arrToCopyTo = nullptr;
-	macAddr3.copyTo(&arrToCopyTo);
+	size_t arrLen = 0;
+	auto arrToCopyTo = macAddr3.copyToNewBuffer(arrLen);
```
tigercosmos: Can we add more test cases for the newly added APIs?

Common++/header/MacAddress.h
tigercosmos: It sounds weird to me that there is an unused parameter... It means that the API design can be improved, though.

Common++/header/IpAddress.h
```diff
@@ -39,8 +41,21 @@ namespace pcpp
...
+		/// @remarks This constructor assumes that the provided array is exactly 4 bytes long.
+		/// Prefer using the constructor with size parameter if the array length is not guaranteed to be 4 bytes.
+		IPv4Address(const uint8_t bytes[4]) : IPv4Address(bytes, 4)
+		{}
+
+		/// A constructor that creates an instance of the class out of a 4-byte array.
+		/// @param[in] bytes The address as 4-byte array in network byte order
+		/// @param[in] size The size of the array in bytes
+		/// @throws std::out_of_range If the provided size is smaller than 4 bytes.
+		IPv4Address(const uint8_t bytes[4], size_t size)
```
seladb: Maybe we want this c'tor to be:
```cpp
IPv4Address(const uint8_t* bytes, size_t size)
```

Common++/header/IpAddress.h
```diff
@@ -160,9 +175,22 @@ namespace pcpp
...
+		/// @remarks This constructor assumes that the provided array is exactly 16 bytes long.
+		/// Prefer using the constructor with size parameter if the array length is not guaranteed to be 16 bytes.
+		IPv6Address(const uint8_t bytes[16]) : IPv6Address(bytes, 16)
+		{}
+
+		/// @brief A constructor that creates an instance of the class out of a 16-byte array.
+		/// @param bytes The address as 16-byte array in network byte order
+		/// @param size The size of the array in bytes
+		/// @throws std::out_of_range If the provided size is smaller than 16 bytes.
+		IPv6Address(const uint8_t bytes[16], size_t size)
```
seladb: ditto

Common++/header/IpAddress.h
Dimi1010: Ended up deprecating the allocating `copyTo` API altogether. It is not commonly used and it is clunky IMO.
The use case can be handled by the new `size_t copyTo(uint8_t* buffer, size_t size)` function that returns the written bytes or the required size. It also supports querying the function.

Example usage:
```cpp
MacAddress mac;
size_t addrLen = mac.copyTo(nullptr, 0); // Query the function for a required buffer.
uint8_t* addr = new uint8_t[addrLen];
if(mac.copyTo(addr, addrLen) != addrLen) { /* Error check */ }
// Use addr...
```

Common++/header/MacAddress.h
Dimi1010: Resolving this as the changes to `copyTo(uint8_t** arr)` implementation were reverted.

Tests/Pcap++Test/Tests/IpMacTests.cpp
Dimi1010: The new functions should have full coverage now.

Common++/header/IpAddress.h
```diff
@@ -39,8 +41,27 @@ namespace pcpp
...
+			if (bytes == nullptr)
+			{
+				throw std::invalid_argument("Buffer pointer is null");
+			}
+
+			if (size < 4)
+			{
+				throw std::out_of_range("Buffer size is smaller than IPv4 address size");
+			}
 			memcpy(m_Bytes.data(), bytes, 4 * sizeof(uint8_t));
```
seladb: nit: maybe move the implementation to `IpAddress.cpp`?

Same for the rest of the methods here and in `MacAddress.h`

Common++/header/IpAddress.h
```diff
@@ -224,14 +264,50 @@ namespace pcpp
 		/// Allocates a byte array and copies address value into it. Array deallocation is user responsibility
 		/// @param[in] arr A pointer to where array will be allocated
 		/// @param[out] length Returns the length in bytes of the array that was allocated
+		/// @deprecated Allocating copyTo API is deprecated.
+		PCPP_DEPRECATED("Allocating copyTo API is deprecated.")
```
seladb: Why do we want to deprecate this API?
If someone wants the library to allocate the buffer for them, what's wrong with it? 🤔

Common++/header/IpAddress.h
```diff
@@ -224,14 +264,50 @@ namespace pcpp
...
+			{
+				if (size != 0)
+				{
+					throw std::invalid_argument("Buffer is null but size is not zero");
+				}
+
+				return requiredSize;
+			}
+
+			if (size < requiredSize)
```
seladb: do we want to support an array size of less than 16 bytes? I'm not sure if there a need to copy just a portion of an IPv6 address

Common++/header/MacAddress.h
```diff
@@ -127,6 +150,8 @@ namespace pcpp
 		/// Allocates a byte array of length 6 and copies address value into it. Array deallocation is user
 		/// responsibility
 		/// @param[in] arr A pointer to where array will be allocated
+		/// @deprecated Allocating copyTo API is deprecated.
+		PCPP_DEPRECATED("Allocating copyTo API is deprecated.")
```
seladb: ditto: why deprecating?

Common++/header/MacAddress.h
```diff
@@ -136,9 +161,42 @@ namespace pcpp
...
+			{
+				if (size != 0)
+				{
+					throw std::invalid_argument("Buffer is null but size is not zero");
+				}
+
+				return requiredSize;
+			}
+
+			if (size < m_Address.size())
```
seladb: ditto: do we need to support copying to an array smaller than 6 bytes?

Tests/Pcap++Test/Tests/IpMacTests.cpp
```diff
@@ -113,15 +129,38 @@ PTF_TEST_CASE(TestIPAddress)
...
-		                              0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x04 };
+
+	std::array<uint8_t, 16> addrAsByteArray;
+	ip6AddrFromIpAddr.copyTo(addrAsByteArray.data());
+	std::array<uint8_t, 16> expectedByteArray = { 0x26, 0x07, 0xF0, 0xD0, 0x10, 0x02, 0x00, 0x51,
+		                                          0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x04 };
+	for (int i = 0; i < 16; i++)
+	{
+		PTF_ASSERT_EQUAL(addrAsByteArray[i], expectedByteArray[i]);
+	}
```
seladb: you can use either `PTF_ASSERT_VECTORS_EQUAL` or `PTF_ASSERT_BUF_COMPARE`.

Ditto the asserts below

Common++/header/IpAddress.h
Dimi1010: It doesn't support copying to an array of less than 16 bytes. If the buffer size is less, no copy is made, and the required buffer size is returned.

If the array is larger, the size of the array that has been written to is returned.

Essentially:
- If the return value is greater than the size input of the buffer, a buffer of at least the size of the return value is needed.
- If the return value is less or equal to the input size param, that is the number of bytes that are updated.

That patern allows runtime querying of the required buffer size without knowing and hardcoding it.

Common++/header/IpAddress.h
Dimi1010: It is very similar to the new `size_t copyTo(uint8_t*, size_t)` function that performs a pure copy to the input buffer. The only difference is that one is a pointer and the other a pointer to pointer param.

I also don't think it should be the `copyTo` function's job to allocate a new buffer. IMO, it breaks the SRP. I understand that is was needed because of how the previous `copyTo` function worked, it did not provide a way to check the required size of a buffer without the user *knowing* its N bytes.

The new function allows reporting of the required buffer size (query mode), so it should be left to the user to handle the memory management of how he wants to provide the required buffer. It allows for simpler, more consistent API of the class and less potential for errors due to the class allocating a buffer the user did not expect.

Common++/header/IpAddress.h
Dimi1010: It should already be that way in the latest commit?

Common++/header/IpAddress.h
seladb: > The only difference is that one is a pointer and the other a pointer to pointer param.

It also allocates a new buffer before copying the data to it, which is the main difference between the 2 overloads.



> I also don't think it should be the `copyTo` function's job to allocate a new buffer. IMO, it breaks the SRP. I understand that is was needed because of how the previous `copyTo` function worked, it did not provide a way to check the required size of a buffer without the user _knowing_ its N bytes.

The user still needs to know the required size to allocate enough bytes before calling this method. Maybe in some cases, users want the library to do that for them.

Common++/header/IpAddress.h
seladb: I'm sorry, I misread the code. I find this API a bit confusing... as a user, I allocate a byte array of a certain size, then call this method. Now, in order to know if the copy succeeded or not, I need to check the return size and compare it to the size I provided to the function. I guess throwing an exception is better in this case...

Common++/header/IpAddress.h
Dimi1010: > It also allocates a new buffer before copying the data to it, which is the main difference between the 2 overloads.

This was meant as the only difference between the two signatures is a degree of indirection, so it will be easy to mistake one for the other.


> > I also don't think it should be the `copyTo` function's job to allocate a new buffer. IMO, it breaks the SRP. I understand that was needed because of how the previous `copyTo` function worked, it did not provide a way to check the required size of a buffer without the user _knowing_ its N bytes.
> 
> The user still needs to know the required size to allocate enough bytes before calling this method. Maybe in some cases, users want the library to do that for them.

Well, an earlier commit had the functionality remade to `copyToNewBuffer` that returned an `unique ptr<uint8_t[]>`, but the API for it ended up kinda clunky. I could re-add that, if needed.

Yes, they still need to know it, but they can control the method of allocation. It follows the somewhat standard pattern of `query` -> `allocate` -> `populate`. I personally dislike black box allocations that then rely on the user knowing the precice method the library allocated the memory inside the function. 

Usually, a black box allocation function is accompanied by a black box deallocation function. In this case, there is only an allocation function.

Common++/header/IpAddress.h
Dimi1010: Well, it is a pattern that allows the user to query the precice required size and not have to guestimate / handcode the allocation size.

Common use case is:
1. Call the method with nullptr + 0 size -> get required bytes
2. Allocate buffer
3. Call method with (buffer, bufSize)
4. Check if the return value is less or equal to bufferSize.

It also allows the user to guestimate the buffer if there is a common case. But if for some reason, the initial buffer is insufficient, the pattern allows for hot retry with the precise amount without triggering the exception mechanism. Here, its value is kinda diminished because the buffer is always 16, but it shines when the required buffer is dynamic in range.

seladb: I think IP addresses are well known concepts, and even more so for people who use these classes. They should be aware that IPv4 is 4 bytes and IPv6 is 16 bytes. IMHO throwing an exception here should be good enough

Common++/header/IpAddress.h
seladb: I'm pretty sure I saw this pattern somewhere where `uint8_t**` means that the buffer is allocated by the library. However I'm not sure where I saw it (I wrote these methods quite a few year ago).

All I'm saying is that I don't think we should deprecate (and later remove) these methods. We can just keep them as they don't break anything and might be useful for some people.

Dimi1010: I see. I have seen that pattern, yes.
My main concern is that `copyTo(uint8_t** ptr, size_t& size)` and `copyTo(uint8_t* ptr, size_t size)` can be easily mistakenly called.

For example, a user wants to copy to a new buffer but forgets to call with `&ptr`, and instead calls with `ptr`. The call will be resolved to the non allocating overload that will either do a no-op on nullptr or explode if the ptr is uninitialized.

Perhaps we can keep the operation but rename the method to something like `copyToNewBuffer`?

tigercosmos: I agree that we should just keep one. IMO, `**ptr` looks more like C-style.

Common++/header/IpAddress.h
tigercosmos: In this case, why not check if the size is either 4 or 16?

tigercosmos (approved):

Common++/header/IpAddress.h
Dimi1010: I agree that they are well known concepts. If the size is known, step 1 can be skipped and the buffer can be directly allocated / used. I think its a good idea to still have a way to query the size at runtime, even if it should be well known.

Additionally, I counted `copyTo` as a low-level operation, so I tried to minimize the exception throwing for the most common error case (insufficient buffer). Ideally, I made the API that way to keep the signature independent of the actual type that is being copied, so we can have a unified `copyTo` API signature for both dynamic and static length objects, thus allowing more flexibility down the line if needed. The initial signature comes from `size_t writeToBuffer(uint8_t* buffer, size_t bufferLen) const;` in #1820, where it was used to serialize helper structs.

> In this case, why not check if the size is either 4 or 16?

The checks are not precise 4 or 16 so it can potentially write to overprovisioned buffers. That is one of primary reasons the function returns the num of written bytes, instead of just true/false on success.

Tests/Pcap++Test/Tests/IpMacTests.cpp
Dimi1010: fixed 4143274

Common++/header/IpAddress.h
Dimi1010: Update 019b376: Re-added allocating `copyTo` with the current signature under the name `copyToNewBuffer(uint8_t** buffer, size_t& size)` and updated deprecations.

Common++/header/IpAddress.h
Dimi1010: Moved most of the `copyTo` methods to cpp. I think the constructors can stay here? They are kinda small and can be inlined better that way?

Dimi1010: Fixed fb3f8f3

Tests/Pcap++Test/Tests/IpMacTests.cpp
```diff
@@ -21,6 +21,22 @@ extern PcapTestArgs PcapTestGlobalArgs;
 
 PTF_TEST_CASE(TestIPAddress)
 {
+	using pcpp::IPAddress;
+	using pcpp::IPv4Address;
+	using pcpp::IPv6Address;
```
seladb: nit: we can probably remove that to be consistent with the rest of this test and the other tests

Tests/Pcap++Test/Tests/IpMacTests.cpp
```diff
@@ -249,6 +300,8 @@ PTF_TEST_CASE(TestIPAddress)
 
 PTF_TEST_CASE(TestMacAddress)
 {
+	using pcpp::MacAddress;
```
seladb: ditto

seladb (approved):

Tests/Pcap++Test/Tests/IpMacTests.cpp
Dimi1010: Fixed 3803685

Tests/Pcap++Test/Tests/IpMacTests.cpp
Dimi1010: Fixed 3803685

PR #1850, opened 2025-06-19
Add decoding and encoding of `BitString` ASN.1 records

seladb:
As part of https://github.com/seladb/PcapPlusPlus/issues/1835, X509 certificates require support for ASN.1 records of type `BitString`

### DIFF
--- a/Packet++/src/Asn1Codec.cpp
+++ b/Packet++/src/Asn1Codec.cpp
@@ -336,6 +336,11 @@ namespace pcpp
 					newRecord = new Asn1BooleanRecord();
 					break;
 				}
+				case Asn1UniversalTagType::BitString:
+				{
+					newRecord = new Asn1BitStringRecord();
+					break;
+				}
 				case Asn1UniversalTagType::Null:
 				{
 					newRecord = new Asn1NullRecord();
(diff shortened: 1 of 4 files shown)


Packet++/src/Asn1Codec.cpp
```diff
@@ -958,4 +963,104 @@ namespace pcpp
...
+		    .substr(0, m_NumBits);
+	}
+
+	std::vector<uint8_t> Asn1BitStringRecord::BitSet::toBytes() const
+	{
+		std::vector<uint8_t> result;
+		for (const auto bs : m_Data)
+		{
+			result.push_back(static_cast<uint8_t>(bs.to_ulong()));
+		}
```
Dimi1010: nit: Vector reserve
It is possible to reserve the result vector size in 1 call, instead of relying on push_back auto reallocation, as we know the output will be `m_data.size()`

Packet++/src/Asn1Codec.cpp
```diff
@@ -1203,4 +1208,104 @@ namespace pcpp
...
+	{
+		return std::accumulate(m_Data.begin(), m_Data.end(), std::string(),
+		                       [](const std::string& acc, const auto& bs) { return acc + bs.to_string(); })
+		    .substr(0, m_NumBits);
+	}
+
+	std::vector<uint8_t> Asn1BitStringRecord::BitSet::toBytes() const
+	{
+		std::vector<uint8_t> result;
+		for (const auto bs : m_Data)
```
Dimi1010: References are not deduced by auto. A copy is being made here for each element.
```suggestion
		for (const auto& bs : m_Data)
```

Packet++/src/Asn1Codec.cpp
```diff
@@ -1203,4 +1208,104 @@ namespace pcpp
...
+	{
+		initFromString(value);
+		return *this;
+	}
+
+	std::string Asn1BitStringRecord::BitSet::toString() const
+	{
+		return std::accumulate(m_Data.begin(), m_Data.end(), std::string(),
+		                       [](const std::string& acc, const auto& bs) { return acc + bs.to_string(); })
+		    .substr(0, m_NumBits);
```
Dimi1010: The previous implementation was actually slightly better, due to less temporary strings.

Accumulate here creates two temporary strings per element - one for the `bs.to_string()` return value and another for the `acc + bs.to_string()` result.

The old implementation only had 1 temporary string per element that is the result of `bs.to_string()`.
Reallocations of the result string can also be eliminated by a reserve call prior to processing, as the return size can be estimated since the bitsets are constant.
```suggestion
		std::string result;
		result.reserve(m_data.size() * 8); // 1 char per bit
		
		for (const auto& bs : m_Data)
		{			   
			result += bs.to_string();	
		}	
		return result.substr(0, m_NumBits);
```

Dimi1010: nit:
`return result.substr(0, m_NumBits)` technically returns a copy of the substring. Since the only operation is cutting characters from the back, a `resize(m_NumBits)` operation can do the same job without a copy at the expense of keeping the previously allocated capacity of the result string. (`resize` does not reallocate if size is reduced).

I don't expect a huge performance difference from either unless the function is called in a hot loop, so I leave it up to you which one you prefer.

seladb: The previous implementation had a cppcheck warning: https://github.com/seladb/PcapPlusPlus/actions/runs/15753205817/job/44402647810 🤔 

```

Packet++/src/Asn1Codec.cpp:834:11: style: Consider using std::accumulate algorithm instead of a raw loop. [useStlAlgorithm]
   result += bs.to_string();
          ^
```

Packet++/src/Asn1Codec.cpp
seladb: Fixed in https://github.com/seladb/PcapPlusPlus/pull/1850/commits/f707262a30c5c412a106998b5c1013d74ee3f1a2

Packet++/src/Asn1Codec.cpp
seladb: Fixed in https://github.com/seladb/PcapPlusPlus/pull/1850/commits/f707262a30c5c412a106998b5c1013d74ee3f1a2

Packet++/src/Asn1Codec.cpp
Dimi1010: Hmm, I would recommend suppressing the warning here if it shows again, as it is suboptimal in the current case.

C++20 added some mechanics to make `accumulate` more performant, but they aren't available in 14.

seladb: Addressed in this commit: https://github.com/seladb/PcapPlusPlus/pull/1850/commits/f07a9006a12414a72c532795f4935a1054b9a851

PR #1851, opened 2025-06-21
Add decoding and encoding of ASN.1 string records 

seladb:
As part of https://github.com/seladb/PcapPlusPlus/issues/1835, X509 certificates require support for ASN.1 records of types `UTF8String`, `PrintableString` and  `IA5String`

- Create an abstract template base class `Asn1StringRecord`
- Refactor `Asn1OctetStringRecord` to inherit from `Asn1StringRecord`
- Create the new record classes: `Asn1UTF8StringRecord`, `Asn1PrintableStringRecord`, `Asn1IA5StringRecord`

### DIFF
--- a/Packet++/header/Asn1Codec.h
+++ b/Packet++/header/Asn1Codec.h
@@ -8,6 +8,8 @@
 #include <chrono>
 #include "PointerVector.h"
 
+#include <iostream>
+
 /// @file
 
 /// @namespace pcpp
@@ -491,9 +493,35 @@ namespace pcpp
 		Asn1EnumeratedRecord() = default;
 	};
 
+	/// @class Asn1StringRecord
+	/// An abstract class for representing ASN.1 string records.
+	/// This class is not instantiable, users should use the derived classes
+	class Asn1StringRecord : public Asn1PrimitiveRecord
+	{
+	public:
+		/// @return The string value of this record
+		std::string getValue()
+		{
+			decodeValueIfNeeded();
+			return m_Value;
+		};
+
+	protected:
+		Asn1StringRecord() = default;
+		Asn1StringRecord(const std::string& value, Asn1UniversalTagType tagType);
+		Asn1StringRecord(const uint8_t* value, size_t valueLength, Asn1UniversalTagType tagType);
+
+		void decodeValue(uint8_t* data, bool lazy) override;
+		std::vector<uint8_t> encodeValue() const override;
+
+		std::vector<std::string> toStringList() override;
+
+		std::string m_Value;
+	};
+
 	/// @class Asn1OctetStringRecord
 	/// Represents an ASN.1 record with a value of type Octet String
-	class Asn1OctetStringRecord : public Asn1PrimitiveRecord
+	class Asn1OctetStringRecord : public Asn1StringRecord
 	{
 		friend class Asn1Record;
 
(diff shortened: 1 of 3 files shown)


Packet++/header/Asn1Codec.h
```diff
@@ -492,21 +492,21 @@ namespace pcpp
...
 	};
 
-	/// @class Asn1OctetStringRecord
-	/// Represents an ASN.1 record with a value of type Octet String
-	class Asn1OctetStringRecord : public Asn1PrimitiveRecord
+	/// @class Asn1StringRecord
+	/// An abstract class for representing ASN.1 string records.
+	/// This class is not instantiable, users should use the derived classes
+	template <Asn1UniversalTagType TagType,
+	          typename std::enable_if<std::is_same<decltype(TagType), Asn1UniversalTagType>::value, bool>::type = false>
```
Dimi1010: The enable if clause might be redundant here, as the first param is already constrained to the specific type (`Asn1UniversalTagType`).

Packet++/header/Asn1Codec.h
```diff
@@ -492,40 +492,94 @@ namespace pcpp
...
 
+	/// @class Asn1StringRecord
+	/// An abstract class for representing ASN.1 string records.
+	/// This class is not instantiable, users should use the derived classes
+	template <Asn1UniversalTagType TagType> class Asn1StringRecord : public Asn1PrimitiveRecord
+	{
+	public:
+		/// A constructor to create a record from a printable string value
+		/// @param value A string to set as the record value
+		explicit Asn1StringRecord(const std::string& value) : Asn1PrimitiveRecord(TagType), m_Value(value)
```
Dimi1010: Should this not be protected? It allows creating string records with arbitrary universal tag values?

Packet++/header/Asn1Codec.h
```diff
@@ -492,40 +492,94 @@ namespace pcpp
...
-		std::string m_Value;
+		Asn1OctetStringRecord() = default;
+
 		bool m_IsPrintable = true;
+	};
 
-		Asn1OctetStringRecord() = default;
+	/// @class Asn1UTF8StringRecord
+	/// Represents an ASN.1 record with a value of type UTF8 String
+	class Asn1UTF8StringRecord : public Asn1StringRecord<Asn1UniversalTagType::UTF8String>
```
Dimi1010, line 566: Nit: If string record keeps its public constructor, this can be an alias instead of a discrete type.

Packet++/header/Asn1Codec.h
```diff
@@ -492,40 +492,94 @@ namespace pcpp
...
+	{
+	public:
+		using Asn1StringRecord::Asn1StringRecord;
+	};
+
+	/// @class Asn1PrintableStringRecord
+	/// Represents an ASN.1 record with a value of type Printable String
+	class Asn1PrintableStringRecord : public Asn1StringRecord<Asn1UniversalTagType::PrintableString>
```
Dimi1010, line 582: Nit, Ditto: alias

Packet++/header/Asn1Codec.h
```diff
@@ -492,40 +492,94 @@ namespace pcpp
...
+	{
+	public:
+		using Asn1StringRecord::Asn1StringRecord;
+	};
+
+	/// @class Asn1IA5StringRecord
+	/// Represents an ASN.1 record with a value of type IA5 String
+	class Asn1IA5StringRecord : public Asn1StringRecord<Asn1UniversalTagType::IA5String>
```
Dimi1010, line 598: Nit, Ditto: alias

Packet++/header/Asn1Codec.h
seladb: You're right, but this means we must define explicit constructors for each subclass. I'll make this change

Dimi1010: No need for that, the using declaration should work as publicizing the protected base ctor in the derived classes.

Dimi1010: On second thought, not 100% sure that will work, but it might be worth a try.

Dimi1010: An alternative might be to keep the public constructor but add static asserts to the class so it raises a compile error if instantiated with invalid tag value?

seladb: > No need for that, the using declaration should work as publicizing the protected base ctor in the derived classes.

Yeah, this doesn't work...

> An alternative might be to keep the public constructor but add static asserts to the class so it raises a compile error if instantiated with invalid tag value?

We could but it defeats the purpose of the abstract class concept. I think exposing an explicit constructor in the child classes is good enough and doesn't involve too much code

PR #1854, opened 2025-06-30
Refactor: Use std::make_unique where applicable

tigercosmos:
Replaced direct usage of new with std::unique_ptr with std::make_unique for improved memory safety and conciseness in C++14 and later.

Introduced static factory methods for PcapLiveDevice and WinPcapLiveDevice to allow std::unique_ptr creation while keeping their constructors protected.

### DIFF
--- a/Pcap++/src/DpdkDeviceList.cpp
+++ b/Pcap++/src/DpdkDeviceList.cpp
@@ -198,7 +198,7 @@ namespace pcpp
 		// Initialize a DpdkDevice per port
 		for (int i = 0; i < numOfPorts; i++)
 		{
-			auto newDevice = std::unique_ptr<DpdkDevice>(new DpdkDevice(i, mBufPoolSizePerDevice, mBufDataSize));
+			auto newDevice = std::make_unique<DpdkDevice>(i, mBufPoolSizePerDevice, mBufDataSize);
 			PCPP_LOG_DEBUG("DpdkDevice #" << i << ": Name='" << newDevice->getDeviceName() << "', PCI-slot='"
 			                              << newDevice->getPciAddress() << "', PMD='" << newDevice->getPMDName()
 			                              << "', MAC Addr='" << newDevice->getMacAddress() << "'");
--- a/Pcap++/src/DpdkDeviceList.cpp
+++ b/Pcap++/src/DpdkDeviceList.cpp
@@ -198,7 +198,8 @@ namespace pcpp
 		// Initialize a DpdkDevice per port
 		for (int i = 0; i < numOfPorts; i++)
 		{
-			auto newDevice = std::make_unique<DpdkDevice>(i, mBufPoolSizePerDevice, mBufDataSize);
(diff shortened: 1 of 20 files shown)


Dimi1010 (approved):

Pcap++/src/DpdkDeviceList.cpp
```diff
@@ -198,7 +198,7 @@ namespace pcpp
 		// Initialize a DpdkDevice per port
 		for (int i = 0; i < numOfPorts; i++)
 		{
-			auto newDevice = std::unique_ptr<DpdkDevice>(new DpdkDevice(i, mBufPoolSizePerDevice, mBufDataSize));
+			auto newDevice = std::make_unique<DpdkDevice>(i, mBufPoolSizePerDevice, mBufDataSize);
```
Dimi1010: DpdkDevice ctor is not public.

Pcap++/src/PfRingDeviceList.cpp
```diff
@@ -92,7 +92,7 @@ namespace pcpp
 						PCPP_LOG_DEBUG("PF_RING version is: " << m_PfRingVersion);
 					}
 
-					auto newDev = std::unique_ptr<PfRingDevice>(new PfRingDevice(currInterface->name));
+					auto newDev = std::make_unique<PfRingDevice>(currInterface->name);
```
Dimi1010: PfRingDevice ctor is not public.

Pcap++/src/DpdkDeviceList.cpp
tigercosmos: Reverted.

Pcap++/src/PfRingDeviceList.cpp
tigercosmos: Reverted.

Pcap++/src/PcapLiveDeviceList.cpp
```diff
@@ -60,9 +60,9 @@ namespace pcpp
 		     currInterface = currInterface->next)
 		{
 #if defined(_WIN32)
-			auto dev = std::unique_ptr<PcapLiveDevice>(new WinPcapLiveDevice(currInterface, true, true, true));
+			auto dev = WinPcapLiveDevice::createWinPcapLiveDevice(currInterface, true, true, true);
 #else  //__linux__, __APPLE__, __FreeBSD__
-			auto dev = std::unique_ptr<PcapLiveDevice>(new PcapLiveDevice(currInterface, true, true, true));
+			auto dev = PcapLiveDevice::createPcapLiveDevice(currInterface, true, true, true);
```
seladb: Why do we need  to expose the `WinPcapLiveDevice::createWinPcapLiveDevice()` method as public?
It makes sense to keep the existing protected constructor and use it in the same way as we do today

tigercosmos: Sure. Reverted.

PR #1857, opened 2025-07-01
CI: windows, remove 2019, add 2025

tigercosmos:
The windows-2019 runner image is being deprecated, consider switching to windows-2022(windows-latest) or windows-2025 instead. For more details see https://github.com/actions/runner-images/issues/12045.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -579,11 +579,11 @@ jobs:
       matrix:
         include:
           - os: "windows-2025"
-            platform: "Visual Studio 16 2019"
+            platform: "Visual Studio 17 2022"
             arch: "x64"
             pcap_lib: "npcap"
           - os: windows-2025
-            platform: "Visual Studio 16 2019"
+            platform: "Visual Studio 17 2022"
             arch: Win32
             pcap_lib: "winpcap"
           - os: windows-2022
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -581,12 +581,12 @@ jobs:
           - os: windows-2025
             platform: "Visual Studio 17 2022"
             arch: Win32
-            pcap_lib: "winpcap"
+            pcap_lib: "npcap"
           - os: windows-2025
             platform: "Visual Studio 17 2022"
             arch: "x64"
             pcap_lib: "winpcap"
-          - os: windows-2025
+          - os: windows-2022
             platform: "Visual Studio 17 2022"
             arch: "x64"
             pcap_lib: "npcap"
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -578,11 +578,11 @@ jobs:
     strategy:
       matrix:
         include:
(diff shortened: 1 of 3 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -578,12 +578,12 @@ jobs:
...
-          - os: "windows-2019"
-            platform: "Visual Studio 16 2019"
+          - os: "windows-2025"
+            platform: "Visual Studio 17 2022"
             arch: "x64"
             pcap_lib: "npcap"
-          - os: windows-2019
-            platform: "Visual Studio 16 2019"
+          - os: windows-2025
+            platform: "Visual Studio 17 2022"
```
seladb: I don't think we need to run VS2022 on both versions of Windows. Since VS2019 is not supported in the newer images, we unfortunately need to drop support for VS2019.

We can keep VS2022 running on `windows-2025`

.github/workflows/package.yml
```diff
@@ -266,7 +266,7 @@ jobs:
   visual-studio:
     strategy:
       matrix:
-        os: [ windows-2019, windows-2022 ]
+        os: [ windows-2025, windows-2022 ]
```
seladb: ditto: no need to run on both versions of Windows

tigercosmos: But why do we run both before?

.github/workflows/build_and_test.yml
tigercosmos: @seladb  I don't get your point here. We now have Windows 2022 and Windows 2025 and both running on VS 2022 since there is no VS 2025.

seladb: yes, that's why I think we can only test VS2022 on Windows 2025 and have less variants. We can probably have only 3:

```yaml
      matrix:
        include:
          - os: "windows-2025"
            arch: "x64"
            pcap_lib: "npcap"
          - os: windows-2025
            arch: Win32
            pcap_lib: "npcap"
          - os: windows-2025
            arch: "x64"
            pcap_lib: "winpcap"
```

.github/workflows/package.yml
seladb: Until the latest version we published both VS2019 and VS2022 packages. Now I guess we can't publish VS2019 anymore...

tigercosmos:
@seladb  It's ready again.

.github/workflows/build_and_test.yml
```diff
@@ -578,19 +578,15 @@ jobs:
...
-          - os: "windows-2019"
-            platform: "Visual Studio 16 2019"
-            arch: "x64"
-            pcap_lib: "npcap"
-          - os: windows-2019
-            platform: "Visual Studio 16 2019"
+          - os: windows-2025
+            platform: "Visual Studio 17 2022"
             arch: Win32
             pcap_lib: "winpcap"
```
seladb: Since WinPcap is deprecated I think we should have 2 build running with Npcap and one with WinPcap.
I'd suggest having this build running with Npcap

seladb (approved):
See just one comment, otherwise LGTM

PR #1858, opened 2025-07-01
Refactor: Use std::chrono for sleep and SystemUtils #1854

adityaxa:
Replaced time handling to use C++14 std::chrono for improved  portability and type safety.

### Changes:
- Replaced `gettimeofday()` and `clockGetTime()` with `std::chrono::system_clock` in SystemUtils.cpp for Consistent time handling across Windows, Linux, and macOS
- Replaced `clock()` with `std::chrono::steady_clock` for time measurements in NetworkUtils.cpp  
- Replaced `usleep()` and `nanosleep()` with `std::this_thread::sleep_for()`
- Removed platform-specific time code (~60 lines) while maintaining the same API

Note: This is the first in a series of incremental updates for the time-handling code.


Common++/src/SystemUtils.cpp
```diff
@@ -211,63 +207,18 @@ namespace pcpp
...
-		nsec = ((count.QuadPart % clock_gettime_counts_per_sec.QuadPart) * CLOCK_GETTIME_BILLION) /
-		       clock_gettime_counts_per_sec.QuadPart;
+		auto now = system_clock::now();
+		auto duration = now.time_since_epoch();
 
-		return 0;
-
-#elif defined(__APPLE__)
+		auto seconds = duration_cast<std::chrono::seconds>(duration);
+		auto nanoseconds = duration_cast<std::chrono::nanoseconds>(duration - seconds);
```
Dimi1010: nit: If we are using `using namespace std::chrono` in the function call scope. Maybe we can shorten this to `seconds` and `nanoseconds` here too?

Dimi1010 (approved):
LGTM

Common++/src/SystemUtils.cpp
adityaxa: Thanks for the suggestion!
I've updated the code to drop std::chrono:: qualifiers. Let me know if anything else needs tweaking.

seladb:
Thank you @adityaxa for working on this refactoring, much appreciated! 🙏

PR #1872, opened 2025-07-11
Added a warning log level.

Dimi1010:
Split of #1845

This PR adds a Warn log level that sits between ERROR and INFO and can be used to signalize non-critical issues in the runtime.


tigercosmos (approved):
LGTM

Issue #1870, opened 2025-07-08
PcapSplitter Error


ferhs343:
### Question

Hello PcapPlusPlus team, the reason for my message is because I'm experiencing an error in the PcapSplitter tool. When trying to trim a pcap by "connection," an error persists. The error is as follows:

[ERROR: PcapFileDevice.cpp: openAppend:754 ] Cannot open './Pcaps/Trims/download_2025235814-connection-tcp_192-168-141-82_7680-187-188-32-129_7345.pcap' for reading and writing
Finished. Read and written 589,633 packets to 9,402 files

It appears that the file "download_2025235814-connection-tcp_192-168-141-82_7680-187-188-32-129_7345.pcap" is not being created by the tool, which results in it not being able to open in append mode.

I've attached evidence of the error below:

<img width="361" height="14" alt="Image" src="https://github.com/user-attachments/assets/7d0e9682-cd7f-4582-b538-0266e64d06a8" />

<img width="854" height="29" alt="Image" src="https://github.com/user-attachments/assets/2842c280-f81f-43d3-8ae3-35778ce455fa" />

I would greatly appreciate it if you could help me resolve this error. Thank you and excellent work you're doing :)

### Operating systems

Linux

seladb:
> It appears that the file "download_2025235814-connection-tcp_192-168-141-82_7680-187-188-32-129_7345.pcap" is not being created by the tool, which results in it not being able to open in append mode.

@ferhs343 I'm not sure I understand... PcapSplitter generates files while trying to split the pcap file. It opens them in "append" mode because packets in the same connection don't necessarily one after the other and the tool tries to cap the number of concurrent open files.

The error you mentioned suggests that `fopen(filename, "rb+")` failed, which might be because of permission issues? 🤔

ferhs343:
Yes, from what I understand, this error occurs because the fopen() method expects the file to already physically exist on the system. However, when I try to cut a PCAP file, this error persists, indicating that the file isn't being created, and consequently, fopen() fails.

I've been running several tests with different PCAP files, and I've noticed that this doesn't happen with some.

ferhs343:
> > It appears that the file "download_2025235814-connection-tcp_192-168-141-82_7680-187-188-32-129_7345.pcap" is not being created by the tool, which results in it not being able to open in append mode.
> 
[...]

PR #1879, opened 2025-07-17
PcapSplitter - fix a bug when splitting by UDP connections

seladb:
Fixes https://github.com/seladb/PcapPlusPlus/issues/1870, the issue is:
- PcapSplitter only keeps up to 250 files open to avoid maintaining to many file handlers
- When it reaches the limit it closes the LRU file
- If a packet arrives on a connection whose file is already closed, it will reopen the file in append mode
- In this example, it was a UDP flow, so there are no clear src and dst ports. The file was initially opened with the name of `udp_src_ip_src_port_dst_ip_dst_port.pcap`, then closed, and when a new packet came on this flow, it had the opposite IP/ports, so it tried to reopen a file `udp_dst_ip_dst_port_src_ip_src_port.pcap` which didn't exist
- The fix is to always open/reopen the files with a name that has the highest port first and the lowest port later

### DIFF
--- a/Examples/PcapSplitter/ConnectionSplitters.h
+++ b/Examples/PcapSplitter/ConnectionSplitters.h
@@ -199,8 +199,17 @@ public:
 			if (udpLayer != nullptr)
 			{
 				sstream << "udp_";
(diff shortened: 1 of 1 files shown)


Dimi1010 (approved):
lgtm

tigercosmos (approved):

PR #1880, opened 2025-07-17
Improvements on lazy decoding of Asn1Codec.

Dimi1010:
This pull request streamlines the Asn1Record lazy decoding of value data.

### Logical `const`-ness of an object
The implementation of lazy evaluation assumes that a `const` qualified object should have an constant logical state, but not necessarily a constant binary representation in memory.

This assumption allows inclusion of `mutable` qualified cache fields that operate on the following rules:
- On first access, the cache field is atomically populated from a data source in a thread-safe manner.
- On subsequent access, the data is fetched directly from the cache field. After the first access, the cached field operates as a normal field and can be treated as such, following `const`-correctness as if the field is not `mutable` qualified.

### Integration with `Asn1Codec`:
* **Field decoding changes**: Replaced the `decodeValue` method to remove the `lazy` parameter, relying on `decodeValueIfNeeded` for lazy decoding.
* **Mutable fields**: Marked fields such as `m_Value` and `m_IsPrintable` as `mutable` to allow modification during lazy evaluation in const methods.

### API Enhancements:
* **Const-correctness**: Updated methods like `getValue` and `toStringList` to be `const`, ensuring they can be called on const instances while maintaining lazy decoding functionality.
* **New `encodeValueSafe` method**: Added to ensure the value is decoded before encoding, improving safety and usability.

### Other Changes:
* **Lazy sub-record decoding**: Introduced a static flag `LazySubRecordDecoding`, defaulting to `false` due to known issues.

### Limitations:
- The current lazy decoding implementation is not thread safe. Thread safety will be added in a future PR.


Common++/header/Lazy.h
```diff
@@ -0,0 +1,133 @@
...
+			}
+
+		protected:
+			/// @brief Initializes the class without a source object and sets the state to Evaluated.
+			/// This is useful for derived classes that set their fields manually.
+			LazyFieldEvaluationBase() = default;
+
+			/// @brief Initializes the class with a source object.
+			/// @param source The source object from which the fields are evaluated.
+			explicit LazyFieldEvaluationBase(T source) : m_Source(std::move(source))
```
Dimi1010: Needs to set state to not evaluated.

seladb:
@Dimi1010 I like this solution! It adds the missing const correctness which is really needed in ASN.1.
However, the `LazyFieldEvaluationBase` class, even though it's nicely implemented, seems to be somewhat complex and over-engineered. We don't need this pattern anywhere else (yet) so we can keep things simple within the ASN.1 codec and extract it to a base class at a later point in time if we need to.
I'd also break it down into multiple PRs:
- First PR - use `mutable` and fix const correctness - this is the most important thing we want to fix with a hopefully minimal amount of changes
- Second PR - add thread-safety
- Third PR - more refactoring and cleanup as needed

Dimi1010:
@seladb The reason I extracted it to a separate class is that I found it easier to reason about lazy loading values in isolation, without the rest of the Asn1Codec infrastructure.

The thread safety is there because I have seen usages where const values are used as read-only across different threads, so I preferred to add that from the get go. I suppose that can be added as a different PR.
Removing it mostly changes `std::atomic<LazyState>` back to `LazyState` and changes `ensureEvaluated` to:
```
if(state == LazyState::NotEvaluated)
  evaluate();
  state = ... // success or fail
  
if (state == Success)
  return value;
if (state == Fail)
  throw;
throw "we should not be here"
```

seladb:
Separating into smaller PRs is always preferred IMO (where it's possible), especially for changes in existing code, which is harder to follow.

I'd keep things simple inside the ASN.1 codec without extracting to a separate class. I agree that thread safety is a nice feature to add

Packet++/src/Asn1Codec.cpp
```diff
@@ -421,17 +411,43 @@ namespace pcpp
...
-	void Asn1Record::decodeValueIfNeeded()
+	void Asn1Record::decodeValueIfNeeded() const
 	{
-		if (m_EncodedValue != nullptr)
+		// TODO: This is not thread-safe and can cause issues if multiple threads access the same record
+		if (m_LazyDecodeState == internal::LazyState::NotEvaluated)
 		{
-			decodeValue(m_EncodedValue, true);
-			m_EncodedValue = nullptr;
+			if (m_EncodedValue == nullptr /* || m_ValueLength == 0 */)
```
Dimi1010: @seladb `m_ValueLength` isn't actually decoded sometimes when `decodeValue` is called, even though `decodeLength` has already been called in constructor?

seladb: @Dimi1010 I'm not sure what you mean, can you please elaborate?

Dimi1010: Found the issue. `Asn.1NullRecord` triggered the `|| m_ValueLength == 0` because it technically is 0.
I suppose we can just throw on `m_EncodedValue == nullptr` as it was originally.

Packet++/header/Asn1Codec.h
```diff
@@ -109,6 +109,36 @@ namespace pcpp
...
+			Lazy,
+			/// The value is evaluated immediately on construction (eager decoding).
+			Eager
+		};
+
+		/// @enum Asn1LoadState
+		/// @brief Represents the state of lazy evaluation for ASN.1 record decoding.
+		/// Used internally to track whether a record's value has been decoded, is being decoded, or if an error
+		/// occurred.
+		enum class Asn1LoadState : int
```
seladb: I'm not sure why we need the state. Previously we'd use the `m_EncodedValue` as the "state" - if it's not null, we know we need to decode, otherwise we can skip. 
Now we have both the state and  `m_EncodedValue`, which seems a bit redundant since there can only be 2 states - "need to evaluate" or "no need to evaluate". Why adding the variable and the additional 2 states (`Evaluating` is not even being used)

Dimi1010: Evaluating was needed for the thread safe version in the duration the value was computed, it isn't needed currently, but I plan on adding the thread safety in another PR after this is merged.

Technically there are 3 states. If the evaluation fails the first time (exception is thrown), you probably want the second time to also throw for consistency.

Using `m_EncodedValue` as the state too, would require that to also be `mutable`, as it needs to be modified after decoding. Also it would make the work of getting the version thread safe harder, as that would require a separate state again.

seladb: Why do we care about thread safety? 🤔 
The other protocol parsers aren't thread-safe, and the reason is that in most applications, it doesn't make sense for multiple threads to access the same packet or message, because that'd involve locks, which slows things down.

If the evaluation fails the first time, we can probably set `m_EncodedValue` back to `nullptr` so the evaluation will retry and probably fail again. We can ke it `mutable` like we make the other private members. Is there a reason not to do it?

Dimi1010: Because the methods are marked as `const`. IMO, most users expect a const method to allow multiple readers, no writers scenario. Which the current implementation does not allow, as that would be a data race.

Thread safety does not always involve locks. In the current case, I had it done with an atomic integer sync, which is almost always lock free for most modern cpus.

seladb: I wouldn't worry too much about thread safety, because we don't guarantee that on any of our protocols. Most applications (if not all) should process each packet in a single thread, unless there's a good reason not to do so and then they should check the internal implementation to make sure they don't miss anything. 

I'd prioritize simplicity over thread safety in this case. We can always add it later if someone asks, but I doubt if that'll ever happen

Dimi1010: Eh, we can keep without thread safety in this PR.

Generally I would prefer to have it thread safe, as it is can avoid the users being surprised by a `const` method being unsafe for something that is commonly assumed. It is an issue that is hard to debug while not precisely searching for it.
Since the load mechanism is hidden, IMO the users shouldn't have to dig into the internals to figure it out.

I can add it in another PR after this, but the atomic implementation should have minimal performance impact.

seladb: The discussion started with the additional state instead of just using `m_EncodedValue` as the state. I still think using `m_EncodedValue` is simpler and better. The purpose of this PR is to improve the lazy decoding, so let's focus on that. We can discuss thread safety later

Dimi1010: Reverted the implementation back to using `m_EncodedValue`. 2fef6c70172c0668e6437cf5dc7852d5b8f3e874 4d34d426c7c5164be86adbad53351a4094891230

Packet++/header/Asn1Codec.h
```diff
@@ -516,21 +547,22 @@ namespace pcpp
...
 		}
 
-		void decodeValue(uint8_t const* data, bool lazy) override
+		void decodeValue(uint8_t const* data) const override
 		{
 			m_Value = std::string(reinterpret_cast<char const*>(data), m_ValueLength);
 		}
 		std::vector<uint8_t> encodeValue() const override
 		{
+			decodeValueIfNeeded();
```
seladb: Is this needed in `encodeValue()`? 🤔

Dimi1010: Technically, yes. But it is probably better to have it in `encodeValueSafe`, so I will remove it from here.

Issue #1876, opened 2025-07-16
When the SSL protocol is determined, check whether the port can be changed to the configuration


yalanmao:
### Question

/// A static methods that gets raw data of a layer and checks whether this data is a SSL/TLS record or not. This
/// check is done using the source/dest port and matching of a legal record type in the raw data. The list of
/// ports identified as SSL/TLS is hard-coded and includes the following ports:
/// - Port 443 [HTTPS]
/// - Port 261 [NSIIOPS]
/// - Port 448 [DDM-SSL]
/// - Port 563 [NNTPS]
/// - Port 614 [SSHELL]
/// - Port 465 [SMTPS]
/// - Port 636 [LDAPS]
/// - Port 989 [FTPS - data]
/// - Port 990 [FTPS - control]
/// - Port 992 [Telnet over TLS/SSL]
/// - Port 993 [IMAPS]
/// - Port 994 [IRCS]
/// - Port 995 [POP3S]
/// @param[in] srcPort The source port of the packet that contains the raw data. Source port (or dest port) are
/// a criteria to identify SSL/TLS packets
/// @param[in] dstPort The dest port of the packet that contains the raw data. Dest port (or source port) are a
/// criteria to identify SSL/TLS packets
/// @param[in] data The data to check
/// @param[in] dataLen Length (in bytes) of the data
/// @param[in] ignorePorts SSL/TLS ports are only relevant for parsing the first SSL/TLS message, but are not
/// relevant for parsing subsequent messages. This parameter can be set to "true" to skip SSL/TLS ports check.
/// This is an optional parameter and its default is "false"
static bool IsSSLMessage(uint16_t srcPort, uint16_t dstPort, uint8_t* data, size_t dataLen,
                         bool ignorePorts = false);

<img width="803" height="395" alt="Image" src="https://github.com/user-attachments/assets/c03fba0d-657a-4568-a488-412cff95a971" />
The ignorePorts parameter is not passed on line 380, resulting in the TLS protocol not being resolved outside the defined port

<img width="1321" height="868" alt="Image" src="https://github.com/user-attachments/assets/2b111d7b-5ba0-4644-bfa9-b5532331681c" />

### Operating systems

Windows

yalanmao:
<img width="1612" height="337" alt="Image" src="https://github.com/user-attachments/assets/24f9f844-e3fc-4946-8369-bbb75b251e04" />
in this function ignorePorts=true，Is this a bug?

[...]

Issue #671, opened 2021-06-19
Can PcapPlusPlus support custom resolution to a specific protocol type based on the port number? Rather than according to the default RFC port definition. 


angelyouyou:
In many cases, the protocol can customize or specify the port number instead of using the fixed allocation on the RFC. After I read the code of the TCP and UDP layer of PcapPlusPlus, I found that PcapPlusPlus is always recognized based on the default port number on the RFC. This is not flexible enough and scalability is not strong in many application scenarios. Do we have plans to add this kind of custom parsing in the future? If a custom parsing method is specified in the configuration file, the custom method will be used first, otherwise The default RFC-based port is still used for resolution.

seladb:
yes, this request has been discussed before here: https://github.com/seladb/PcapPlusPlus/issues/644. Right now it's not supported but if you want you're more than welcome to implement this feature. We can discuss what's the best way to do it.

angelyouyou:
> yes, this request has been discussed before here: #644. Right now it's not supported but if you want you're more than welcome to implement this feature. We can discuss what's the best way to do it.

I think we can define a configuration file to re-define the protocol recognization.Like the wireshark, it can specify the port of the protocol to recognize.
It is not difficult to implement, but very useful.I'd like to implement it.

seladb:
This can be a good addition to this project. This is just for configuring ports, is that correct?
I think we can let the user choose the location of the file. It can be in json format as you mentioned.
Please note that this will require a change in the existing layers to be able to read the data from the file as well, for example:
- [HTTP ports](https://github.com/seladb/PcapPlusPlus/blob/f94bbf8133a32033fcb80b765d543ce147f57a46/Packet%2B%2B/header/HttpLayer.h#L80)
- [SIP ports](https://github.com/seladb/PcapPlusPlus/blob/f94bbf8133a32033fcb80b765d543ce147f57a46/Packet%2B%2B/header/SipLayer.h#L119)
- [TLS/SSL ports](https://github.com/seladb/PcapPlusPlus/blob/f94bbf8133a32033fcb80b765d543ce147f57a46/Packet%2B%2B/header/SSLLayer.h#L179)
- And more...

You basically need to:
- Read the file if exists and load it to some class in memory - we can probably keep a map between the protocol type and the custom ports
- Modify each layer to look at this map and return all the relevant ports - both the default ones and the ports in this map

PR #1881, opened 2025-07-17
Customizable parser configuration

Dimi1010:
This PR adda a new class `ParserConfiguration` with the intent to enable user customization of parsing logic.
The initial experimental implementation includes `PortMapper` that can be used to define custom port rule mappings.

### The default parser configuration
A default configuration singleton is provided by the runtime via `ParserConfiguration::getDefault()` containing all default parsing logic. This configuration is to be used in all instances when no custom configuration is provided.

#### Modifying the default configuration
The reference returned by `ParserConfiguration::getDefault()` allows modification of the underlying configuration. This can be used to modify the default parsing behaviour or even override it completely by assigning a custom configuration.

A modified default configuration can be reset to its initial state by calling `ParserConfiguration::resetDefault`.

**NB: The configuration object is currently not thread safe for modification. Do not modify the config if an active parsing is being performed.** 

### Parser extension
- Added new overload `Layer::parseNextLayer(ParserConfiguration const& config)`: The new overload is to be considered the main signature of the parsing system. The parsed configuration is used to determine parser behaviour.
- Changed `Layer::parseNextLayer()` to forward to the new overload: The parameter-less version of `parseNextLayer` has been converted to a non-virtual function that calls the new overload with the default parsing configuration.

#### Limitations
- Currently the `Packet` class only utilizes the default parser during parsing.


Dimi1010:
@seladb what do you think of this implementation for dynamic port mapping as a potential solution to #1876 and #671?

seladb:
@ Dimi1010 overall, it's a good solution that could work! Here are a few comments:
- We need to figure out how the configuration is loaded. I think it should be loaded once when the app starts and possibly kept in a static variable. The packet parser can access it to get the configuration
- `parseNextLayer(ParserConfiguration const& config)` could be made protected to make sure this overload is never called by the user, ensuring that the configuration is always static and loaded once
- It'd be nice to tie the default configuration of a protocol to the layer representing it, like we do today with `isHttpPort()` etc. I'm not exactly sure how to do it though, maybe we can think about it further
- It'd be nice to be able to load and save the configuration from/to json
- Something to think about: should the loaded configuration be on top of the default or replace it completely? For example, the user might define additional ports for HTTP or may want to replace the existing ports altogether. Another example: the user wants to replace HTTP ports, but keep the default SSL/TLS ports
- I'm not sure if we need a `PortPair` or just a port that can be either src or dst. Usually we look at a single port and not a combination of them. Even if we want to keep  `PortPair` we should probably give an option to define only one port and translate it behind the scenes to a `PortPair` with similar src and dst ports

Dimi1010:
> * We need to figure out how the configuration is loaded. I think it should be loaded once when the app starts and possibly kept in a static variable. The packet parser can access it to get the configuration

That is the default behaviour already. `ParserConfiguration::getDefault()` returns a reference to a static variable.
It is only constructed on first access and then reused on following calls.
```cpp
static inline ParserConfiguration& getDefault()
{
	static ParserConfiguration defaultConfig = makeDefaultConfiguration();
	return defaultConfig;
}
```

> * `parseNextLayer(ParserConfiguration const& config)` could be made protected to make sure this overload is never called by the user, ensuring that the configuration is always static and loaded once

I don't agree with that. IMO, both should be available to the users. In the general use case, the user will configure / modify the default parser and use the parameter-less overload as is (`parseNextLayer()`), but some might want to have several configurations for parsing. Keeping the overload `public` allows that behaviour while it doesn't sacrifice the general use case performance.

> * It'd be nice to tie the default configuration of a protocol to the layer representing it, like we do today with `isHttpPort()` etc. I'm not exactly sure how to do it though, maybe we can think about it further

I suppose that might be possible but the default configuration mappings are created in `ParserConfig.cpp`. That would imply that TU including all the layers?
I think it would be better to have all the mappings in a single location so it is easier to spot conflicting mappings?
I am actually unsure if we have any port mappings that map to different next layers depending on the parsing layer.

> * It'd be nice to be able to load and save the configuration from/to json

Can be done, but it wasn't a priority for the initial draft.

> * Something to think about: should the loaded configuration be on top of the default or replace it completely? For example, the user might define additional ports for HTTP or may want to replace the existing ports altogether. Another example: the user wants to replace HTTP ports, but keep the default SSL/TLS ports

The way I imagine the system working is thus:
1. the connection has `portSrc` and `portDst`
2. The mapper is queried with (`portSrc`, `portDst`) pair:
3. The mapper returns a matrix of the possible hits:
3.1. full match - there is an entry mapping the entire pair to a protocol type
3.2. partial match (src) - there is an entry mapping `portSrc` and any `portDst` value to a protocol type
3.3. partial match (dst) - there is an entry mapping `portDst` and any `portSrc` value to a protocol type

Example:
For example pair `(src: 80, dst: 65444)` would match:
- `(src: 80, dst: 65444)` -> `HttpResponse`
- `(src: 80, dst: *)` -> `HttpResponse`
- `(src: *, dst: 65444)` -> `UnknownProtocol`

If wildcards are used in the input pair the exact match will be the same as the partial match of the non-wildcard port.
The match `(src: *, dst: *)` always results `UnknownPort`.

> * I'm not sure if we need a `PortPair` or just a port that can be either src or dst. Usually we look at a single port and not a combination of them. Even if we want to keep  `PortPair` we should probably give an option to define only one port and translate it behind the scenes to a `PortPair` with similar src and dst ports

`PortPair` provides factory methods `fromSrc` and `fromDst` that create a pair that uses a wildcard for the opposite port, allowing easy creation from only one port.

Generally symmetrical bindings are handled on the mapping level, when the rule is being registered. In `PortMapper::addPortMapping(PortPair port, ProtocolType protocol, bool symmetrical = false)`, the flag `symmetrical` registers `(src: A, dst: B)` and `(src: B, dst: A)` to map to the same `protocol` if set to `true`.


@seladb Hope that answers the questions!

seladb:
> That is the default behaviour already. `ParserConfiguration::getDefault()` returns a reference to a static variable. It is only constructed on first access and then reused on following calls.
> 
> ```c++
> static inline ParserConfiguration& getDefault()
> {
> 	static ParserConfiguration defaultConfig = makeDefaultConfiguration();
> 	return defaultConfig;
> }
> ```

Yes, I was referring to the configuration loaded by the user, not only the default one


> I don't agree with that. IMO, both should be available to the users. In the general use case, the user will configure / modify the default parser and use the parameter-less overload as is (`parseNextLayer()`), but some might want to have several configurations for parsing. Keeping the overload `public` allows that behaviour while it doesn't sacrifice the general use case performance.

`parseNextLayer()`, although it's a public method, shouldn't be called by users. Users should create a `Packet` instance and let PcapPlusPlus parse it. We need to remember that packet parsing is on the critical path, and we must ensure that any changes we make don't affect performance. That's the reason I don't want to encourage users to change the configuration at runtime. If we get feedback from users about it we can consider it in the future



> The way I imagine the system working is thus:
> 
>     1. the connection has `portSrc` and `portDst`
> 
>     2. The mapper is queried with (`portSrc`, `portDst`) pair:
> 
>     3. The mapper returns a matrix of the possible hits:
>        3.1. full match - there is an entry mapping the entire pair to a protocol type
>        3.2. partial match (src) - there is an entry mapping `portSrc` and any `portDst` value to a protocol type
>        3.3. partial match (dst) - there is an entry mapping `portDst` and any `portSrc` value to a protocol type
> 
> 
> Example: For example pair `(src: 80, dst: 65444)` would match:
> 
>     * `(src: 80, dst: 65444)` -> `HttpResponse`
> 
>     * `(src: 80, dst: *)` -> `HttpResponse`
> 
>     * `(src: *, dst: 65444)` -> `UnknownProtocol`
> 
> 
> If wildcards are used in the input pair the exact match will be the same as the partial match of the non-wildcard port. The match `(src: *, dst: *)` always results `UnknownPort`.

Will the user configuration be on top of the default one or replace it? For example: HTTP default are port 80 and 8080. The user might want to replace that with port 1234, or add an additional port 1234. How can we support both?
Another example: the user wants to replace ports 80 and 8080 with port 1234 for HTTP, but keep the default ports for SSL/TLS

> `PortPair` provides factory methods `fromSrc` and `fromDst` that create a pair that uses a wildcard for the opposite port, allowing easy creation from only one port.
> 
> Generally symmetrical bindings are handled on the mapping level, when the rule is being registered. In `PortMapper::addPortMapping(PortPair port, ProtocolType protocol, bool symmetrical = false)`, the flag `symmetrical` registers `(src: A, dst: B)` and `(src: B, dst: A)` to map to the same `protocol` if set to `true`.

I'd create a simpler interface for the users that can define a single port and have the port mapper create 2 rules behind the scenes

Dimi1010:
> Yes, I was referring to the configuration loaded by the user, not only the default one

The idea is that the user can either modify the default configuration, or make his own configuration instance.
The user can make changes to the default one by: `getDefault().addPortMapping(...);`

In the case of modifying the default, the user's changes essentially change the global default.
Reset of the default config is also possible via `resetDefault()` method that restores the "factory defaults".

If they want to have a separate instance, the can either make one from scratch, or make a copy of the default and store it at their convenience. A manually constructed config can even be assigned to the default via `*::getDefault() = otherConfig;`

Ideally the user would be able to inject a configuration to the parsing logic of `Packet`, but currently that isn't possible, so the only reason to keep separate instances of the configurations is for backup purposes.

> `parseNextLayer()`, although it's a public method, shouldn't be called by users. Users should create a `Packet` instance and let PcapPlusPlus parse it. We need to remember that packet parsing is on the critical path, and we must ensure that any changes we make don't affect performance. That's the reason I don't want to encourage users to change the configuration at runtime. If we get feedback from users about it we can consider it in the future

Yes, it is mainly called internally. There was an idea to eventually allow Packets to be parsed with different configurations, specified at the packet parse time. So there is a potential use case for the `Packet` class itself to inject a parser configuration to the `Layer` classes. Haven't really figured out how to do that yet tho.

This will be an additive change, so in the base use case, the packet fetches the default parser config static.
In any case, the access to the configuration is via pointer indirection in both cases, so there isn't a comparative performance difference.

I haven't profiled the solution yet, but the most likely performance factor would be the `PortMap` implementation itself, not how the configuration dependency is actually injected into the parsing logic.

> Will the user configuration be on top of the default one or replace it? For example: HTTP default are port 80 and 8080. The user might want to replace that with port 1234, or add an additional port 1234. How can we support both? Another example: the user wants to replace ports 80 and 8080 with port 1234 for HTTP, but keep the default ports for SSL/TLS

For port mapping, each configuration instance is essentially a global port table. The user can add entries, remove entries or completely replace the entire table.

For the example, if the user wants port 1234 to be HTTP, the user adds a rule (1234, *) -> HTTP. If the user wants to have 80 not be HTTP, he removes the rule from the map that is used for parsing.

I am unsure by what you mean by keeping the default ports for SSL/TLS. Each port rule is independent of each other.
Modifying the `portPair` -> HTTP rule would not affect the rule `portPair` -> SSL. (Assuming I have understood everything)

> I'd create a simpler interface for the users that can define a single port and have the port mapper create 2 rules behind the scenes

Wdym? Like a convenience method that takes `uint16_t` and creates `PortPair(x, Any)` and `PortPair(Any, x)`? I suppose that is possible but it is a bit ambiguous as that can easily be inferred to be `PortPair(x, x)`.

I think it is better to only allow explicit port pairs for clarity when users are defining their own rules. `symmetrical` already handles the use case of defining the reverse rule to their provided pair for them.

```cpp
void PortMapper::addPortMapping(PortPair port, ProtocolType protocol, bool symmetrical)
{
	if (port == PortPair())
	{
		throw std::invalid_argument("PortPair cannot be empty (both src and dst ports are 0)");
	}

	auto insertResult = m_PortToProtocolMap.insert({ port, protocol });
	insertResult.first->second = protocol;  // Update the protocol if it already exists
	if (!insertResult.second)
	{
		PCPP_LOG_WARN("Port " << port << " is already mapped to protocol " << insertResult.first->second
		                      << ", updating to " << protocol);
	}

	if (symmetrical && port.portSrc != port.portDst)
	{
		// Add the symmetrical mapping
		PortPair symmetricalPort = { port.portDst, port.portSrc };
		addPortMapping(symmetricalPort, protocol, false);
	}
}
```

Dimi1010:
@seladb Another case I found. FTP has both FTP default port (20) and FTP Data (port 21), but both map to ProtocolType::FTP. 
Can we add a separate FTP Data protocol type? Having both being mapped to FTP would make it hard for the layers to disambiguate, as the mapper would return FTP in both cases. (or would require an additional hard coded check which defeats the purpose)

seladb:
> The idea is that the user can either modify the default configuration, or make his own configuration instance. The user can make changes to the default one by: `getDefault().addPortMapping(...);`
> 
> In the case of modifying the default, the user's changes essentially change the global default. Reset of the default config is also possible via `resetDefault()` method that restores the "factory defaults".
> 
> If they want to have a separate instance, the can either make one from scratch, or make a copy of the default and store it at their convenience. A manually constructed config can even be assigned to the default via `*::getDefault() = otherConfig;`
> 
> Ideally the user would be able to inject a configuration to the parsing logic of `Packet`, but currently that isn't possible, so the only reason to keep separate instances of the configurations is for backup purposes.

I think that's a very reasonable solution, offering a lot of flexibility to users. They can start with the default and modify it as needed, or replace it with their configuration 👍 


> Yes, it is mainly called internally. There was an idea to eventually allow Packets to be parsed with different configurations, specified at the packet parse time. So there is a potential use case for the `Packet` class itself to inject a parser configuration to the `Layer` classes. Haven't really figured out how to do that yet tho.
> 
> This will be an additive change, so in the base use case, the packet fetches the default parser config static. In any case, the access to the configuration is via pointer indirection in both cases, so there isn't a comparative performance difference.
> 
> I haven't profiled the solution yet, but the most likely performance factor would be the `PortMap` implementation itself, not how the configuration dependency is actually injected into the parsing logic.

I think we agree on that 👍 
The use case of a `Packet` instance to modify the parser configuration is interesting, but as you mentioned we can discuss it later as such use cases arise

> Wdym? Like a convenience method that takes `uint16_t` and creates `PortPair(x, Any)` and `PortPair(Any, x)`? I suppose that is possible but it is a bit ambiguous as that can easily be inferred to be `PortPair(x, x)`.
> 
> I think it is better to only allow explicit port pairs for clarity when users are defining their own rules. `symmetrical` already handles the use case of defining the reverse rule to their provided pair for them.
> 
> ```c++
> void PortMapper::addPortMapping(PortPair port, ProtocolType protocol, bool symmetrical)
> {
> 	if (port == PortPair())
> 	{
> 		throw std::invalid_argument("PortPair cannot be empty (both src and dst ports are 0)");
> 	}
> 
> 	auto insertResult = m_PortToProtocolMap.insert({ port, protocol });
> 	insertResult.first->second = protocol;  // Update the protocol if it already exists
> 	if (!insertResult.second)
> 	{
> 		PCPP_LOG_WARN("Port " << port << " is already mapped to protocol " << insertResult.first->second
> 		                      << ", updating to " << protocol);
> 	}
> 
> 	if (symmetrical && port.portSrc != port.portDst)
> 	{
> 		// Add the symmetrical mapping
> 		PortPair symmetricalPort = { port.portDst, port.portSrc };
> 		addPortMapping(symmetricalPort, protocol, false);
> 	}
> }
> ```

I'd provide a simple interface, something along the lines of:
I think this will be simpler for most users to understand: "use port 80 for HTTP; use port 443 for TLS/SSL"
```cpp
void PortMapper::addPortMapping(uint16_t port, ProtocolType protocol)
{
    addPortMapping({ port, port }, protocol, true);
}
```

Dimi1010:
> I'd provide a simple interface, something along the lines of:
> I think this will be simpler for most users to understand: "use port 80 for HTTP; use port 443 for TLS/SSL"
> ```cpp
> void PortMapper::addPortMapping(uint16_t port, ProtocolType protocol)
> {
>     addPortMapping({ port, port }, protocol, true);
> }
> ```

Sure, we can do that.

Although I think the more general case is if the mapping was `{ port : AnyPort }` and `{ AnyPort : port }`, rather than `{ port, port }`, no?

seladb:
> > I'd provide a simple interface, something along the lines of:
> > I think this will be simpler for most users to understand: "use port 80 for HTTP; use port 443 for TLS/SSL"
> > ```c++
> > void PortMapper::addPortMapping(uint16_t port, ProtocolType protocol)
> > {
> >     addPortMapping({ port, port }, protocol, true);
> > }
> > ```
> 
> Sure, we can do that.
> 
> Although I think the more general case is if the mapping was `{ port : AnyPort }` and `{ AnyPort : port }`, rather than `{ port, port }`, no?

Ah yes, I think you're right 👍

Packet++/src/ParserConfig.cpp
```diff
@@ -0,0 +1,222 @@
...
+		mapper.addPortMapping(PortPair::fromDst(3799), Radius, true);  // RADIUS over TCP
+
+		// NTP port mappings
+		mapper.addPortMapping(PortPair::fromDst(123), NTP, true);  // NTP over UDP
+
+		// Wake-on-LAN port mappings
+		mapper.addPortMapping(PortPair::fromDst(9), WakeOnLan, false);  // Wake-on-LAN over UDP
+		mapper.addPortMapping(PortPair::fromDst(7), WakeOnLan, false);  // Wake-on-LAN over UDP
+		// Would result in Pair (0, 0) which is invalid for the mapper.
+		// mapper.addPortMapping(PortPair::fromDst(0), WakeOnLan, false);  // Wake-on-LAN over UDP (broadcast)
```
Dimi1010: @seladb Due to the way the mapper works, we can't actually do mapping `(AnyPort, 0)`, because 0 is used as a magic value for the `AnyPort`. That would result in `(AnyPort, AnyPort)` pair.

I can either try to make a workaround (somehow), or we can drop the mapping.
From what I can read usage of Port 0 is kinda rare?

seladb: @egecetin write this parser so maybe he knows if it's a rate case?

If we want to support port 0 maybe we can make `PortPair` a pair of 2 `uint32_t` values, but only allow assigning it `uint16_t` values - and make `AnyPort` of value that is larger than `uint16_t`...

seladb:
@Dimi1010 I think I remember why I didn't implement this until now: I wasn't sure what the performance implications might be. Packet parsing is the fast path of PcapPlusPlus, and changing it from hard-coded values to using a map might have performance implications, especially in high throughput network traffic (for example: when using DPDK, or even when using libpcap with high traffic). Is it possible to measure it somehow? 🤔

Dimi1010:
> @Dimi1010 I think I remember why I didn't implement this until now: I wasn't sure what the performance implications might be. Packet parsing is the fast path of PcapPlusPlus, and changing it from hard-coded values to using a map might have performance implications, especially in high throughput network traffic (for example: when using DPDK, or even when using libpcap with high traffic). Is it possible to measure it somehow? 🤔 

It should be able to be measured. A large Pcap file that is loaded and then timed as it is fed into Packet parsing should suffice since we only need the comparative benchmarks for the parsing module.

Dimi1010:
@seladb Check #1887 for a benchmark that evaluates only the pure parsing performance.

And the current PR head does drop performance somewhat, although that was always gonna be the case. 🤔 

Might test it with 3rd party map implementations, to see if the performance can get better.
`std::unordered map` is notorious for being kinda bad for very high performance situations compared to 3rd party maps, due to API constraints set by the standard (iterator invalidation limitations iirc).

```
2025-07-21T20:56:55+03:00
Running ...\PcapPlusPlus\out\build\win-msvc-x64-npcap\examples_bin\Release\BenchmarkExampleGoogle.exe
Run on (16 X 4700 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 1024 KiB (x8)
  L3 Unified 98304 KiB (x1)
Build info: Jul 21 2025 20:55:43
Git info: Git branch 'dev', commit '40d8de02'
Pcap file: ...\PcapPlusPlus\Examples\PcapPlusPlus-benchmark\traffic-sample.pcapng
PcapPlusPlus version: v25.05+ (non-official release)
-------------------------------------------------------------------------------
Benchmark                     Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------
BM_PcapFileRead             366 ns          361 ns      2036364 bytes_per_second=1.73153Gi/s items_per_second=2.77247M/s
BM_PcapFileWrite           3515 ns         3453 ns       203636 bytes_per_second=14.9147Mi/s items_per_second=289.616k/s
BM_PacketParsing            497 ns          497 ns      1445161 bytes_per_second=1.25648Gi/s items_per_second=2.01033M/s
BM_PacketPureParsing        125 ns          125 ns      6400000 bytes_per_second=5.02254Gi/s items_per_second=8.03137M/s
BM_PacketCrafting           360 ns          361 ns      1947826 bytes_per_second=153.231Mi/s items_per_second=2.77024M/s


2025-07-21T20:57:49+03:00
Running ...\PcapPlusPlus\out\build\win-msvc-x64-npcap\examples_bin\Release\BenchmarkExampleGoogle.exe
Run on (16 X 4700 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 1024 KiB (x8)
  L3 Unified 98304 KiB (x1)
Build info: Jul 21 2025 20:57:47
Git info: Git branch 'feature/parser-config', commit '27e79425'
Pcap file: ...\PcapPlusPlus\Examples\PcapPlusPlus-benchmark\traffic-sample.pcapng
PcapPlusPlus version: v25.05+ (non-official release)
-------------------------------------------------------------------------------
Benchmark                     Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------
BM_PcapFileRead             366 ns          377 ns      1866667 bytes_per_second=1.65863Gi/s items_per_second=2.65438M/s
BM_PcapFileWrite           3552 ns         3453 ns       203636 bytes_per_second=14.9147Mi/s items_per_second=289.616k/s
BM_PacketParsing            507 ns          500 ns      1000000 bytes_per_second=1.24968Gi/s items_per_second=1.99967M/s
BM_PacketPureParsing        132 ns          129 ns      4977778 bytes_per_second=4.85874Gi/s items_per_second=7.77019M/s
BM_PacketCrafting           363 ns          361 ns      2036364 bytes_per_second=153.381Mi/s items_per_second=2.77292M/s

...\PcapPlusPlus\out\build\win-msvc-x64-npcap\examples_bin\Release\BenchmarkExampleGoogle.exe (process 68964) exited with code 0 (0x0).
Press any key to close this window . . .
```

seladb:
Hmm it's a drop of ~4-6%. We can probably run it on different platforms to see if we get similar results (especially on Linux, which many of the high throughput applications run on). I guess such a drop is not ideal. If there was a way to "configure" what to use it'd be nice, but I'm not sure there's a clean way to do that 🤔

Dimi1010:
It probably can be done, with some compile time magic, where the `PortMapper`'s internals are swapped-in to a hard coded switch case or something depending on a compile time flag.

seladb:
> It probably can be done, with some compile time magic, where the `PortMapper`'s internals are swapped-in to a hard coded switch case or something depending on a compile time flag.

I'm not sure if we want to get into it, it might make things even more complex than they are now...

Packet++/src/ParserConfig.cpp
egecetin: @seladb @Dimi1010 I'm not sure fully understand since not followed up lately but in WoL there are two cases,

- For UDP we can just assume known ports (7 and 9) are WoL packets
- For directly Ether2 packets there is a type indicator in ethernet header (0x0842) and there is no other encapsulation (like UDP or anything) just raw ethernet packet, so no ports

Dimi1010: @egecetin 

In the current version `WakeOnLanLayer::isWakeOnLanPort` is implemented as [[1]](https://github.com/seladb/PcapPlusPlus/blob/a667b2af55b91f69a88fbe05fc0f6e719c5d9e7c/Packet%2B%2B/header/WakeOnLanLayer.h#L105-L108):
```cpp
static bool isWakeOnLanPort(uint16_t port)
{
	return (port == 0) || (port == 7) || (port == 9);
}
```

UDP Layer considers a packet WoL under the following conditions [[2]](https://github.com/seladb/PcapPlusPlus/blob/a667b2af55b91f69a88fbe05fc0f6e719c5d9e7c/Packet%2B%2B/src/UdpLayer.cpp#L138-L139):
```cpp
if ((WakeOnLanLayer::isWakeOnLanPort(portDst) && WakeOnLanLayer::isDataValid(udpData, udpDataLen)))
	m_NextLayer = new WakeOnLanLayer(udpData, udpDataLen, this, m_Packet);
```

Is it safe to no longer consider a udp packet with destination port 0 as a WoL packet?

egecetin: I found why, from Wikipedia,

> Since the magic packet is only scanned for the string above, and not actually parsed by a full protocol stack, it could be sent as payload of any network- and transport-layer protocol, although it is typically sent as a [UDP](https://en.wikipedia.org/wiki/User_Datagram_Protocol) [datagram](https://en.wikipedia.org/wiki/Datagram) to [port](https://en.wikipedia.org/wiki/TCP_and_UDP_port) 0 (reserved port number),[[7]](https://en.wikipedia.org/wiki/Wake-on-LAN#cite_note-8) 7 ([Echo Protocol](https://en.wikipedia.org/wiki/Echo_Protocol)) or 9 ([Discard Protocol](https://en.wikipedia.org/wiki/Discard_Protocol)),[[8]](https://en.wikipedia.org/wiki/Wake-on-LAN#cite_note-9) or directly over Ethernet using [EtherType](https://en.wikipedia.org/wiki/EtherType) 0x0842.[[9]](https://en.wikipedia.org/wiki/Wake-on-LAN#cite_note-10) A [connection-oriented](https://en.wikipedia.org/wiki/Connection-oriented) transport-layer protocol like [TCP](https://en.wikipedia.org/wiki/Transmission_Control_Protocol) is less suited for this task as it requires establishing an active connection before sending user data.

It is also possible port 0. Probably I saw it and added also port 0 but also examples from internet uses a raw ethernet or other known ports (7 and 9) so why there is no example in Tests

Dimi1010: So, we should keep the port 0 available for mapping?

egecetin: Unfortunately yes

Dimi1010: Refactored the `PortPair`. We should be able to support port 0 now.

Packet++/header/Layer.h
```diff
@@ -197,6 +206,8 @@ namespace pcpp
 			return m_NextLayer != nullptr;
 		}
 
+		virtual void doParseNextLayer(ParserConfiguration const& config) = 0;
```
Dimi1010, line 209: Changed parsing implementation to be done in `doParseNextLayer` with `parseNextLayer` being converted to compile-time dispatch function. The short version is that C++ function hiding mechanics make the parameter-less overload of `parseNextLayer` inaccessible in derived classes, if they implement an override of the `parseNextLayer(ParserConfiguration const& config)`.

Two ways to solve the problem were considered:
1. Move the virtual dispatch implementation to a protected function and keep the API function overloads as non-virtual functions that forward to the implementation.
2. Add `using Layer::parseNextLayer;` to all derived classes.

Option 1 was selected as it provides compiler enforcement. Option 2 can't be enforced by the compiler, and therefor less desirable to be applied over a large collection of derived classes.

See more details in 15ed36f20fa91b754f34228db22b8ec56c0cbfba commit message.

seladb:
@Dimi1010 I think we can close this PR?

Dimi1010:
I guess. I somewhat hit a roadblock with the map implementation. Might reopen it later if I decide to work on in again.

Issue #1675, opened 2025-01-01
Enable more clang-tidy checks


egecetin:
Currently a simple clang-tidy check is running for the repo. More checks should be enabled and raised issues should be fixed for,

- [ ] Common++ (#1639)
- [ ] Packet++
- [ ] Pcap++
- [ ] Examples
- [ ] Tests
- [ ] 3rdParty

OxotaKpenkoe:
Looks like .clang-tidy-new / clang-tidy-all-new.sh haven't been updated in about a year and aren't wired into CI or pre-commit at all. Is this approach still relevant? I'd like to pick this up for the Packet++ module if so.

egecetin:
@OxotaKpenkoe Sure. I didn't have enough time to complete work, so help is appreciated

PR #1639, opened 2024-11-07
Clang-tidy fixes for Common++

egecetin:
Fix various clang-tidy warnings for Common++

Requires #1677

PR #1884, opened 2025-07-21
Clang tidy fixes for Common++

egecetin:
Because of the slow transition and fast changes in Common++ new warnings are generated by clang-tidy in Common++. This PR fixes new issues.

Part of #1675


Dimi1010 (approved):
LGTM

Common++/src/Logger.cpp
```diff
@@ -103,8 +103,8 @@ namespace pcpp
 		std::ostringstream sstream;
 		sstream << file << ": " << method << ":" << line;
 
-		std::unique_lock<std::mutex> lock(logMutex);
+		std::unique_lock<std::mutex> const lock(logMutex);
 		std::cerr << std::left << "[" << std::setw(5) << Logger::logLevelAsString(logLevel) << ": " << std::setw(45)
-		          << sstream.str() << "] " << logMessage << std::endl;
+		          << sstream.str() << "] " << logMessage << '\n';
```
seladb: We always use `std::endl`, not sure why `\n` is better, but maybe we should stay consistent and always use `std::endl`

Dimi1010: Because '\n' does not flush the stream buffer. `std::endl` flushes it every time it is encountered.
For `std::cerr` that doesn't do much as that stream is unbuffered, but it slows down `std::cout` a lot if many flushes are done unnecessarily, so it is generally more performant to use '\n'

egecetin: It is a performance warning from clang-tidy. I'm not sure it impacts heavily. `std::endl` forces a flush but `\n` doesn't so it might optimize the performance of stream output slightly

seladb: Oh I didn't know that. But for logs, don't we want to flush the buffer to make sure the log is written?

egecetin: Actually it will flush but after some time, just not immediately so it can write multiple lines at once.

seladb: yes, but for logs you'd like the buffer to get flushed after every message so you don't lose messages or get them later than you expect. For example, assume you have this code:
```cpp
LOG_ERROR("some error");
std::cout << "some output << std::endl;
```

You might get "some output" before "some error" which is not what you expect...

Dimi1010: The output is still ordered, even if buffered.

From what I know `std::cout` flushes when:
- encounters `std::endl` or `std::flush`
- `std::cout.flush()` is called
- input is queried from `std::cin`
- output is written to `std::cerr`
- (possibly) the internal buffer is full
- if `std::ios_base::sync_with_stdio(true)` is set (default behavour), usage of C stream functions (`printf`, `scanf`) also flushes accordingly if needed.

seladb: got it, so my example is not good.
However I remember wasting a lot of time in the past when I couldn't figure out why logs weren't printed, just to discover the buffer wasn't flushed 😕

Dimi1010: yeah... it can happen. Generally it is fine to use `std::endl` at the end of a output operation, but it is counter productive to have multiple `std::endl`s in the middle of output preparation.

Good example - Single flush at the end of the message to display it.
```cpp
std::cout << "This is some text! \n This is new line text." << std::endl;
```

Bad example - Flush in the middle of the message, without need.
```cpp
std::cout << "This is some text!" << std::endl << "This is new line text." << std::endl;
```

seladb: Yes I agree. In this case, it's a single flush per line, isn't it?

Dimi1010: for the bad example, yes. It is a single flush per line.

seladb: so do we agree to move it back to `std::endl`?

Dimi1010: The stream is `std::cerr`, which is unbuffered, so it kinda doesn't matter which one is used here. I suppose '\n` avoids a virtual call to flush (no op function in this case).

Im fine with either.

seladb: @egecetin let's revert to `std::endl` to follow the same conventions we use everywhere?

egecetin: @seladb Sure but all of them including `std::endl` changes in #1639 or just this one?

Dimi1010: Doesn't #1639 only change the std::endl in the logger? If so, that is the same function as now. The change probably got lost when I refactored the logger.

I do think that the `std::endl`s sprinkled in examples (mainly the help strings) need to be changed to `'\n'` tho.

egecetin: Yes you are right, all of them were `Logger` or `TablePrinter`. But my question is should I suppress the warning completely from all files in clang-tidy or just very simple cases like this? 

In examples there are some cases just multiline strings are printed with `std::endl` in `printUsage()` or `EXIT_WITH_ERROR()`. For `printUsage` I think it should be definitely `\n` but what should be done for error macros? One thing in my mind use `std::endl` for just error strings and in last line of complex multiline strings to ensure it is flushed in last line. But it will require many inline suppressions

Dimi1010: > In examples there are some cases just multiline strings are printed with `std::endl` in `printUsage()` or `EXIT_WITH_ERROR()`. For `printUsage` I think it should be definitely `\n` but what should be done for error macros? One thing in my mind use `std::endl` for just error strings and in last line of complex multiline strings to ensure it is flushed in last line. But it will require many inline suppressions

I think using std::endl at the end of a conceptual message is a good solution.

seladb: Maybe it's better to just suppress this error and fix whatever we think needs fixing?

egecetin: @seladb Ok agreed I'll add global suppression for it. We can discuss it for Examples more precisely (which should keep or replaced) in its PR. Anyway, I already reverted in this PR, so it is ready to merge if there is no anything different than this

seladb (approved):

PR #1892, opened 2025-07-24
Updated Logger to use `std::function`

Dimi1010:
This PR updates `pcpp::Logger::LogPrinter` to alias to `std::function<void(...)>` from `std::add_pointer_t<void(...)>`.

Micro benchmarking found no noticeable performance drop from the change.

In the benchmarks the print function was replaced with an int accumulator function to avoid variance due to console IO.
```cpp
volatile int dummy = 0;

static void printer(pcpp::LogLevel logLevel, const std::string& logMessage, const std::string& file,
                    const std::string& method, int line)
{
	benchmark::DoNotOptimize(logLevel);
	dummy++;
}
```

Run on (16 X 4700 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 1024 KiB (x8)
  L3 Unified 98304 KiB (x1)
Windows 11 x64, MSVC Release Build

| Test Run | Time | CPU Time | Iterations |
| ------------- | ------------- | ------------- | ------------- |
| Func Pointer N1 | 223 ns | 220 ns | 2986667 |
| Func Pointer N2 | 225 ns | 225 ns | 3200000 |
| Func Pointer N3 | 226 ns | 220 ns | 3200000 |
| Func Pointer N4 | 223 ns | 225 ns | 3200000 |
| std::function N1 | 225 ns | 225 ns | 3200000 |
| std::function N2 | 224 ns | 225 ns | 2986667 |
| std::function N3 | 224 ns | 220 ns | 3200000 |
| std::function N4 | 226 ns | 225 ns | 2986667 |

### DIFF
--- a/Common++/header/Logger.h
+++ b/Common++/header/Logger.h
@@ -246,9 +246,8 @@ namespace pcpp
 		/// @param[in] method The method in PcapPlusPlus code the log message is coming from
 		/// @param[in] line The line in PcapPlusPlus code the log message is coming from
 		/// @remarks The printer callback should support being called from multiple threads simultaneously.
-		using LogPrinter =
-		    std::function<void(LogLevel logLevel, const std::string& logMessage, const std::string& file,
-		                       const std::string& method, const int line)>;
+		using LogPrinter = std::function<void(LogLevel logLevel, const std::string& logMessage, const std::string& file,
+		                                      const std::string& method, const int line)>;
 
 		/// A static method for converting the log level enum to a string.
 		/// @param[in] logLevel A log level enum
--- a/Common++/header/Logger.h
+++ b/Common++/header/Logger.h
@@ -7,6 +7,7 @@
 #include <mutex>
 #include <ostream>
 #include <sstream>
+#include <functional>
 #include "DeprecationUtils.h"
 #include "ObjectPool.h"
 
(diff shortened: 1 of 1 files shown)


tigercosmos (approved):
Cool, I am surprised by this result. I guess the compiler is smart.

Dimi1010:
Apparently, most major implementations have SSO on `std::function` where if the callable is small enough, it is stored inline, instead of on the heap.

Common++/header/Logger.h
```diff
@@ -417,10 +421,10 @@ namespace pcpp
 	do                                                                                                                 \
 	{                                                                                                                  \
 		auto& logger = pcpp::Logger::getInstance();                                                                    \
-		if (logger.shouldLog(level, LOG_MODULE))                                                                       \
+		if (logger.shouldLog(level, PCPP_FULLY_QUALIFIED_LOG_MODULE))                                                  \
 		{                                                                                                              \
-			auto ctx =                                                                                                 \
-			    logger.createLogContext(level, pcpp::LogSource(LOG_MODULE, PCAPPP_FILENAME, __FUNCTION__, __LINE__));  \
+			auto ctx = logger.createLogContext(                                                                        \
+			    level, pcpp::LogSource(PCPP_FULLY_QUALIFIED_LOG_MODULE, PCAPPP_FILENAME, __FUNCTION__, __LINE__));     \
```
seladb: Why is this new macro `PCPP_FULLY_QUALIFIED_LOG_MODULE` needed?

Dimi1010: I tried using the log macro in benchmarks to test, and it broke because it was not inside the pcpp namespace, therefor the enum could not be found.

seladb: This is because we change `std::add_pointer` to `std::function`?

Dimi1010: No. It's unrelated to that change directly.

It was due to the perf testing I did. I can separate it if you want.

seladb: Yes, maybe we can separate it?

Dimi1010: Should be fixed now

seladb (approved):

PR #1894, opened 2025-07-25
Add Base64 encoding and decoding

seladb:
Part of https://github.com/seladb/PcapPlusPlus/issues/1835

This PR introduces Base64 encoding and decoding functionality to PcapPlusPlus, based on the implementation from https://github.com/tobiaslocker/base64

The primary motivation is to support encoding and decoding of [PEM](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail) formats, which are commonly used for encoding cryptographic keys and certificates.


Common++/header/GeneralUtils.h
```diff
@@ -55,6 +55,60 @@ namespace pcpp
 		return (number + mask) & ~mask;
 	}
 
+	/// @class Base64
```
seladb, line 59: Maybe we can add this class in its own file, but I think adding it in `GeneralUtils.h` is reasonable...

Common++/src/GeneralUtils.cpp
```diff
@@ -108,4 +111,511 @@ namespace pcpp
...
+		if (numPadding > 2)
+		{
+			throw std::invalid_argument("Invalid base64 encoded data - Found more than 2 padding characters");
+		}
+
+		size_t decodedSize = (input.size() * 3 / 4) - numPadding;
+		if (decodedSize > resultByteArrSize)
+		{
+			throw std::invalid_argument("Not enough space in result buffer for decoded data");
+		}
```
Dimi1010: IMO, the function should return the requested size here, instead of throwing.

Currently the thrown error does not give any indication on how much space is required, just a general "Need more space". How much? 1 byte, 500? ¯\\\_(ツ)\_/¯ This can lead to situations where the caller has to call / catch / reallocate multiple times, if the buffer increment step is not enough.

Returning the required size would allow the caller to immediately identify how much is needed and reallocate accordingly. The difference in API would be that after execution a check needs to be made if `returnValue <= resultByteArrSize` which would be `true` on success or `false` if more memory is required.

Given that this is low-level API, (as opposed to the API that returns value types), avoiding an exception throw on a common error condition that is easily rectifiable is a benefit.

Common++/header/GeneralUtils.h
Dimi1010: A different file would be beneficial due to the size of the cpp file implementation, but it is fine for now.

Common++/header/GeneralUtils.h
```diff
@@ -55,6 +56,60 @@ namespace pcpp
...
+		/// Decode a Base64 string to a byte array
+		/// @param[in] input The Base64 string to be decoded
+		/// @param[out] resultByteArr A pre-allocated byte array where the result will be written to
+		/// @param[in] resultByteArrSize The size of the pre-allocated byte array
+		/// @return The size of the decoded data
+		static size_t decodeToByteArray(const std::string& input, uint8_t* resultByteArr, size_t resultByteArrSize);
+
+	private:
+		static constexpr uint32_t badChar = 0x01ffffff;
+		static constexpr char paddingChar = '=';
```
Dimi1010, line 115: Do we need these to be in the header?

Common++/src/GeneralUtils.cpp
```diff
@@ -108,4 +111,511 @@ namespace pcpp
...
+	}
+
+	std::string Base64::encode(const std::string& input)
+	{
+		return encode(reinterpret_cast<const uint8_t*>(input.c_str()), input.length());
+	}
+
+	std::string Base64::encodeHexString(const std::string& hexStringInput)
+	{
+		size_t inputLen = hexStringInput.length() / 2;
```
Dimi1010, line 473: nit: `std::string`'s size is sometimes queried with `length()` and sometimes with `size()`.

IMO, `size` is preferrable as it matches the other generic containers.

Common++/src/GeneralUtils.cpp
seladb: I think we had this discussion before, and I don't agree with this API. I believe we should treat this scenario as any other invalid input. I can add the required size to the exception we throw so the user knows how much more buffer they should allocate

Common++/header/GeneralUtils.h
seladb: We don't have to, I just thought it'd be nicer to include them in the class because the file contains more than base64 decoding/encoding logic. Ideally I'd also include `base64EncodingTable*` and `base64DecodingTable*` in the class as well but `constexpr` has to be included in the header file which will make it much less readable

Common++/src/GeneralUtils.cpp
seladb: Sure, should we change all occurrences of `length()` to `size()`?

Dimi1010: On second thoughts... Did a check and we already appear to use `length` in a lot of places, so it would be quite a lot of work for dubious benefit. I guess it is fine.

Will mark this as resolved. :)

Common++/src/GeneralUtils.cpp
Dimi1010: Sure, I guess it can throw an exception with the required bytes... although I think it would be beneficial to have a way to query the required length, avoiding the exception mechanism altogether, since it is a common error case.

Perhaps having a second function that just calculates the decoded string's length instead, so it can be queried prior to the actual decode call?

seladb: I added this method in this commit: https://github.com/seladb/PcapPlusPlus/pull/1894/commits/0c1b2a7cf682510167c22253160f904937374d40

I couldn't find a good way to avoid code duplication between `decodeToByteArray()` and `getDecodedSize()`, because `numPadding` is used later in `decodeToByteArray()`. Please let me know if you have any idea there

Dimi1010: It can possibly be extracted into a helper function with internal linkage that returns `numPadding` and the length, but it seems fine as is.

Dimi1010 (approved):
Clang format complains about formatting the brace initializations. But other than that LGTM.

Issue #1890, opened 2025-07-23
Wrong offset calculation in SOME/IP-TP


simon-germany:
### Bug description

**Describe the bug**

The methods getOffset() and setOffset() in SomeIpTpLayer are not reading and writing the offset field of the transport protocol layer correctly. It is wrong by factor 16. The reason is a wrong bit shift (>> 4 and << 4):

 ``` 
uint32_t SomeIpTpLayer::getOffset() const
{
	return (be32toh(getSomeIpTpHeader()->offsetAndFlag) & SOMEIP_TP_OFFSET_MASK) >> 4;
}

void SomeIpTpLayer::setOffset(uint32_t offset)
{
	uint32_t val = (offset << 4) | (be32toh(getSomeIpTpHeader()->offsetAndFlag) & ~SOMEIP_TP_OFFSET_MASK);
		getSomeIpTpHeader()->offsetAndFlag = htobe32(val);
}
 ``` 

The author simply read the 28 offset bits and mapped it to an uint32_t skipping the 4 most significant bits.

However, the Autosar Standard (https://www.autosar.org/fileadmin/standards/R22-11/FO/AUTOSAR_PRS_SOMEIPProtocol.pdf , page 51) states:

> **[PRS_SOMEIP_00724]** 
> The Offset field shall transport the upper 28 bits of a uint32. The lower 4 bits shall be always interpreted as 0.
> **Note:**
> This means that the offset field can only transport offset values that are multiples of 16 bytes.

**Expected behavior**
Correct offset calculation

### PcapPlusPlus versions tested on

v23.09

### Operating systems tested on

Windows

### Compiler version

MS VS2019

seladb:
Thanks for reporting this issue @simon-germany ! To be honest, I'm not familiar with the SOME/IP protocol. @kolbex is the one who added it, so maybe he can provide more insights.
Anyway - maybe you can open a PR with a fix?

simon-germany:
Unfortunately, in my organization there are very strict IT rules and barriers (e. g. firewall blocks GIT ports).

But I can provide my small fix here. Only 2 lines of code are effected:


```
uint32_t SomeIpTpLayer::getOffset() const
{
	return (be32toh(getSomeIpTpHeader()->offsetAndFlag) & SOMEIP_TP_OFFSET_MASK);
}

void SomeIpTpLayer::setOffset(uint32_t offset)
{
[...]

PR #1898, opened 2025-07-26
Fix wrong offset calculation in SOME/IP-TP

seladb:
Fixes https://github.com/seladb/PcapPlusPlus/issues/1890


simon-germany (approved):
Hello @seladb,

sorry, I am in vacations and missed your first message. Thank you very much for the fix, it looks correct. I like that you added a check in the setter!

tigercosmos (approved):

Issue #1899, opened 2025-07-26
Separate `RawPacket` and `MBufRawPacket`


Dimi1010:
### Preface
Currently there are 2 main "raw" packet container types: `RawPacket` and `MBufRawPacket`. Both are specialized in different use cases:
- **RawPacket:** Specialized to hold a buffer on the free store or with external storage duration (user must ensure lifetime).
- **MBufRawPacket:** Specialized to utilize DPDK's MBuf infrastructure and read/write data to it.

### The problem - `RawPacket` is currently works as both "concrete class" and "base class"
In the majority of the library, a pointer to `RawPacket` is used where a packet data container is needed. This leads to the current implementation of `MBufRawPacket` subclassing from `RawPacket` to be interchangeable.

The primary issue with this design choice is that `RawPacket` now has double duty of both being a final implementation class and an abstract base class. The need to both have it's own functionality and the need to accommodate subclassing leads to maintainability issues and hard to read code, especially in the copy operations.

### Proposed solution
A potential solution would be to extract a base class (or interface) over `RawPacket` and `MBufRawPacket` that defines the common interface or functionality between the packets. This disentangles the need for both classes to work around each other, simplifying the code.

seladb:
I remember I had these thoughts when I created `MBufRawPacket`, but I'm not sure creating an interface or a base class is really needed. `MBufRawPacket` overrides the necessary functions of `RawPacket` and is more of a niche use case needed for DPDK. 


> This leads to the current implementation of `MBufRawPacket` subclassing from `RawPacket` to be interchangeable.

Where do you see it as a problem or a blocker to a feature we want to implement but can't?

Dimi1010:
The primary issues I have in mind would be from the internal capacity tracking and buffer policies described in #1888. It can possibly be worked around that but it would be messy and probably less maintainable in the long run as every change in `RawPacket` must be checked so it does not break something in `MBufRawPacket`.

seladb:
Since we only have 2 classes `RawPacket` and `MBufRawPacket` and most of the code uses `RawPacket`, I don't think it'd be very hard to maintain both. If we add more types of raw packets in the future it may be more difficult to maintain, but I don't currently see a feature that would need that. If we change `RawPacket`'s behavior we have to accomodate these changes in both classes anyway in order for everything to work correctly

Dimi1010:
> If we change RawPacket's behavior we have to accomodate these changes in both classes anyway in order for everything to work correctly

[...]

Issue #1888, opened 2025-07-22
Enhancements on `RawPacket` infrastructure.


Dimi1010:
This issue is a general overview of infrastructure and API enhancements related to `RawPacket` mostly explored in #1845 .

### Separation of `RawPacket` and `MBufRawPacket`

Currently `MBufRawPacket` inherits from `RawPacket` mainly to be interchangeable with it in the `Layer` infrastructure. The two packet classes however, differ significantly in their internal mechanics.

- `RawPacket` focuses on heap allocated or externally managed memory.
- `MBufRawPacket` focuses on packets in DPDK MBuf managed memory.

It is my opinion that need for `MBufRawPacket` to inherit from `RawPacket` causes issues with the classes' maintenance. Notable cases are `RawPacket`'s copy operations, which need to work both for `RawPacket` and for `MBufRawPacket`.

My proposal for resolution of the issue is the addition of a pure interface class `IRawPacket` that provides a common interface and that both classes inherit it. `RawPacket` can keep its existing functionality intact without the need to handle cases where it is derived from, while `MBufRawPacket` is detached from the base implementation allowing greater freedom of implementation.

The existing `Packet` infrastructure can be refactored to utilize `IRawPacket*` where `RawPacket*` was previously used, with minimal changes to the internals.

### Improvements of const correctness of `RawPacket` constructor parameters.

In the current implementation the `RawPacket` constructor accepts a `uint8_t const*` data buffer pointer. This is incorrect as the pointer is immediately stripped of its `const` qualifier. It is my opinion that that constructor should be deprecated and an overload that accepts a non-const qualified data buffer pointer should be provided instead.

This change forces the `const` stripping to be done outside of the class, ensuring that the user makes a conscious decision to ignore const-correctness and pass a buffer that should not be modified.

### `RawPacket` as a fully functional data container.

In the current implementation `RawPacket` does not track the capacity of the buffer it owns / references and relies on external users to track its capacity for it. This is mentioned in the descriptions of `RawPacket::appendData` and `RawPacket::insertData`, warning of memory corruption if invalid input is provided.

I propose that the buffer capacity is tracked internally, as that would allow better memory safety during append or insert operations. It would also allow better control over the buffer depending on its ownership.

### Improving clarity on buffer ownership and expanded behaviour

[...]

PR #1900, opened 2025-07-26
Add `RawPacketBase` as common base class between raw packet implementations.

Dimi1010:
Part of #1899

This PR adds a class `RawPacketBase` that is to work as a common base class between `RawPacket` and `MBufRawPacket`.
The current iteration only takes over the management for `timestamp` and `linkLayer` as a minimal example, without defining common API yet.

The class is defined in the `internal` namespace while in development, with the eventual goal to be "published" to the main namespace when complete.

### DIFF
--- a/Packet++/header/RawPacket.h
+++ b/Packet++/header/RawPacket.h
@@ -251,20 +251,82 @@ namespace pcpp
 	/// Max packet size supported
 #define PCPP_MAX_PACKET_SIZE 65536
 
+	namespace internal
+	{
+		// TODO: Class is experimental and should not be used in user code yet. Eventually promote it to public API
+		/// @brief A base class for raw packet classes, containing common attributes and methods.
+		class RawPacketBase
+		{
+		protected:
+			/// @brief A default constructor that zero initializes the timestamp and sets the link layer type to
+			/// Ethernet.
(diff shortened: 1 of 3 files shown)


seladb:
@Dimi1010 I think we discussed it somewhere - I don't think this refactor is needed. Can we close this PR?

Dimi1010:
@seladb I suppose we can close it for now. We did discuss it. I think we decided to defer the refactor, not discard it all together.

IIRC, the main cause for this was that the entanglement of `RawPacket` and `MBufRawPacket` was making it difficult for `RawPacket` to track its own capacity. Currently it only tracks the used buffer size, and capacity is tracked in `Packet`, which makes the `RawPacket` methods vulnerable to buffer overflows.

seladb:
> think we decided to defer the refactor, not discard it all together.

If I remember the discussion, we decided it isn't worth the effort

Dimi1010:
> If I remember the discussion, we decided it isn't worth the effort

This was how it ended:
https://github.com/seladb/PcapPlusPlus/issues/1888#issuecomment-3154119804

Dimi1010:
Anyway, I can close then PR for now.

PR #1916, opened 2025-08-10
Add discrete FTPControl and FTPData protocol types.

Dimi1010:
This PR adds discrete `FTPControl` and `FTPData` protocol types, while changing `FTP` to protocol family.

This is done to be able to better distinguish between FTP data channel and FTP control channel as they are normally on different ports. (relevant to #1881)

Potential breaking change:
- the protocol type value `42` now refers to `FTPControl` instead of `FTP` and will only match control payloads.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,12 +93,18 @@ set(
   "1"
   "2"
   "3"
+  "4"
+  "Off"
+  "Error"
+  "Warn"
+  "Info"
+  "Debug"
 )
 set(
   PCAPPP_LOG_LEVEL
   ""
   CACHE STRING
-  "Compile time log level (0 - Off, 1 - Error, 2 - Info, 3 - Debug) Default(empty): Debug"
+  "Compile time log level (0 - Off, 1 - Error, 2 - Warn, 3 - Info, 4 - Debug) Default(empty): Debug"
 )
 set_property(CACHE PCAPPP_LOG_LEVEL PROPERTY STRINGS ${PCAPP_ALLOWED_LOG_LEVELS})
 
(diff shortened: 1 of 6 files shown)


egecetin (approved):

Packet++/header/ProtocolType.h
```diff
@@ -223,6 +223,12 @@ namespace pcpp
 	/// Diagnostic over IP protocol (DOIP)
 	const ProtocolType DOIP = 59;
 
+	/// File Transfer Protocol (FTP) Data channel
+	const ProtocolType FTPData = 60;
+
+	/// FTP protocol family (FTPControl and FtpData protocols)
+	const ProtocolTypeFamily FTP = 0x3c29;
```
tigercosmos, line 230: I wonder why FTP value is 0x3c29?

Dimi1010: It follows the convention for other protocol family packs.

Hex 3c = Dec 60 = FTPData
Hex 29 = Dec 41 = FTPControl

A single protocol value is 8 bits, while a family is 32 bits packed 4x 8 bit protocol type values. Essentially being `(protocol1, protocol2, protocol3, protocol4)`.

The public API takes 32-bit family packs, so if a single protocol is passed, the value gets zero extended to `(unknown protocol, unknown protocol, unknown protocol, protocol value)`.

Ideally, I would have this done with structs and constepxr construction, but we need cpp17 for that to work alongside the current usages inside switch cases. Until then, hard coded magic values it is.

tigercosmos: Wow... TBH it's a little hard to get the idea just from the source code. Could you add the comment in the code, thanks!

tigercosmos (approved):

Packet++/header/ProtocolType.h
Dimi1010: 👍 Sure I can try to add documentation on how it works. Will be in a different PR tho.

seladb (approved):

PR #1917, opened 2025-08-10
Add X509Toolkit - X509 example

seladb:
Part of https://github.com/seladb/PcapPlusPlus/issues/1835

This PR adds `X509Toolkit` - an example application that shows the APIs for parsing X509 certificates


Examples/X509Toolkit/PcapExtract.h
```diff
@@ -0,0 +1,318 @@
...
+#include "Logger.h"
+#include <iostream>
+#include <vector>
+#include <unordered_map>
+
+#if defined(_WIN32)
+#	define DIR_SEPARATOR '\\'
+#else
+#	define DIR_SEPARATOR '/'
+#endif
```
Dimi1010: Can we make this a constexpr function?

```suggestion
constexpr char getOsPathSeparator() {
#if defined(_WIN32)
  return '\\';
#else
  return '/';
#endif
}
```

Dimi1010: nit: `ifdef _WIN32`

Examples/X509Toolkit/main.cpp
```diff
@@ -0,0 +1,347 @@
...
+	          << "               Use -s to display extraction statistics after processing." << std::endl
+	          << "               If -o is not specified, the certificates are written to stdout." << std::endl
+	          << std::endl
+	          << "Examples:" << std::endl
+	          << "  " << pcpp::AppName::get() << " convert -i cert.der -o cert.pem -f PEM" << std::endl
+	          << "  " << pcpp::AppName::get() << " info -i cert.pem" << std::endl
+	          << "  " << pcpp::AppName::get() << " json -i cert.pem -o cert.json" << std::endl
+	          << "  " << pcpp::AppName::get() << " expire -i cert.pem" << std::endl
+	          << "  " << pcpp::AppName::get() << " pcap-extract -i tls.pcap -o MyCertDir -f PEM -s" << std::endl
+	          << std::endl;
```
Dimi1010: nit: Can we move to using '\n' instead of `std::endl` in the middle of string streams?
The [core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slio50-avoid-endl) recommend against its usage due to extra flushing and having it in the examples promotes bad coding practices imo.

Examples/X509Toolkit/main.cpp
```diff
@@ -0,0 +1,347 @@
...
+		exit(1);                                                                                                       \
+	} while (0)
+
+#define EXIT_WITH_ERROR_AND_PRINT_USAGE(reason)                                                                        \
+	do                                                                                                                 \
+	{                                                                                                                  \
+		printUsage();                                                                                                  \
+		std::cout << std::endl << "ERROR: " << reason << std::endl << std::endl;                                       \
+		exit(1);                                                                                                       \
+	} while (0)
```
Dimi1010, line 25: Curious, do those 2 have to be macros? Can we get away with using functions?

Examples/X509Toolkit/main.cpp
Dimi1010, line 1: Some functions in this file have `static` marker and some don't (excluding `main(...)`).

Since this is `main.cpp` this I personally don't see the need to have the functions with internal linkage, but if we want them to, perhaps we can put them under an unnamed namespace instead?

Examples/X509Toolkit/main.cpp
seladb: I prefer to leave it as is to make it consistent with the other examples

Examples/X509Toolkit/PcapExtract.h
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/1917/commits/108969fc6fe2a04cebf8ef28e87f88221dba9437

Examples/X509Toolkit/main.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/1917/commits/108969fc6fe2a04cebf8ef28e87f88221dba9437

Examples/X509Toolkit/main.cpp
seladb: I added `static` where it was missing in https://github.com/seladb/PcapPlusPlus/pull/1917/commits/108969fc6fe2a04cebf8ef28e87f88221dba9437. I don't think an unnamed namespace is needed here, please let me know what you think

Dimi1010: They do practically the same thing for functions and variables. The unnamed namespace has the advantage of being able to mark classes as internal linkage too, but it doesn't matter in this case.

looks good 👍

Examples/X509Toolkit/main.cpp
```diff
@@ -0,0 +1,345 @@
...
+	    << "               Use -s to display extraction statistics after processing.\n"
+	    << "               If -o is not specified, the certificates are written to stdout.\n"
+	    << "\n"
+	    << "Examples:\n"
+	    << "  " << pcpp::AppName::get() << " convert -i cert.der -o cert.pem -f PEM\n"
+	    << "  " << pcpp::AppName::get() << " info -i cert.pem\n"
+	    << "  " << pcpp::AppName::get() << " json -i cert.pem -o cert.json\n"
+	    << "  " << pcpp::AppName::get() << " expire -i cert.pem\n"
+	    << "  " << pcpp::AppName::get() << " pcap-extract -i tls.pcap -o MyCertDir -f PEM -s\n"
+	    << std::endl;
```
Dimi1010, line 76: Note: C/Cpp compilers have the feature where consecutive string literals are merged under the hood.

```cpp
std::cout << 
  "First Line \n"
  "Second Line";
```
In the above example, the string is concatenated into "First Line \nSecond Line" during compilation, allowing some flexibility in formatting in source while allowing the string to be handled as a single literal.

It is fine either way, I just thought it was a neat feature I should mention.

seladb: Thanks for mentioning it! I guess we can keep it as is to be consistent with the rest of the code, but nice feature indeed!

PR #1919, opened 2025-08-13
Added documentation comment for protocol type and protocol type family.

Dimi1010:
In relation to https://github.com/seladb/PcapPlusPlus/pull/1916#discussion_r2270700810.

This PR adds a short description on `ProtocolType` and `ProtocolTypeFamily` usage in a separate doxygen "topic" page.

Note: The paragraphs enclosed inside the `@internal` and `@endinternal` scope, will not generate any documentation unless doxygen is ran with `INTERNAL_DOCS=YES`. As the section there relates to the internal mechanics and can be subject to change, I decided it is best not to specify it directly in public documentation, to avoid users of the library relying on the information.

### DIFF
--- a/Packet++/header/ProtocolType.h
+++ b/Packet++/header/ProtocolType.h
@@ -8,6 +8,28 @@
 /// @brief The main namespace for the PcapPlusPlus lib
 namespace pcpp
 {
+	/// @defgroup ProtocolTypes ProtocolType and ProtocolTypeFamily
+	/// In the PcapPlusPlus library, specific protocols are identified by a `ProtocolType` value, which is an 8-bit
+	/// unsigned integer. Each Layer class has a ProtocolType that can be used to identify the protocol of that layer.
+	///
+	/// As there are some situations where multiple protocols can be grouped together (e.g. IPv4 and IPv6 into IP),
+	/// PcapPlusPlus also defines ProtocolTypeFamily, which is a group of multiple protocol types. In functions where a
+	/// ProtocolTypeFamily parameter is expected, a single ProtocolType value can be passed instead. The library will
+	/// treat it as a family containing only that single protocol.
+	///
+	/// @internal
+	/// 
+	/// A ProtocolTypeFamily pack is represented as a 32-bit unsigned integer. Each octet in the 32-bit value is a
+	/// ProtocolType, allowing for up to 4 protocols to be packed into a single ProtocolTypeFamily value. For example,
+	/// the ProtocolTypeFamily for IP is represented as 0x203, which contains both IPv4 (0x02) and IPv6 (0x03)
+	/// protocols. A single ProtocolType casted to ProtocolTypeFamily will result the value '0x000000xx' where 'xx' is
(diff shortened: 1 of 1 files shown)


Dimi1010:
Do tell if there are any suggestions on expanding the section.

Packet++/header/ProtocolType.h
```diff
@@ -8,6 +8,28 @@
...
+	/// @defgroup ProtocolTypes ProtocolType and ProtocolTypeFamily
+	/// In the PcapPlusPlus library, specific protocols are identified by a `ProtocolType` value, which is an 8-bit
+	/// unsigned integer. Each Layer class has a ProtocolType that can be used to identify the protocol of that layer.
+	///
+	/// As there are some situations where multiple protocols can be grouped together (e.g. IPv4 and IPv6 into IP),
+	/// PcapPlusPlus also defines ProtocolTypeFamily, which is a group of multiple protocol types. In functions where a
+	/// ProtocolTypeFamily parameter is expected, a single ProtocolType value can be passed instead. The library will
+	/// treat it as a family containing only that single protocol.
+	///
+	/// @internal
```
seladb, line 20: nit: I'm not sure this section should be defined as `@internal`

Dimi1010: I marked it as such, as I think it would be beneficial for those 2 typedefs to be converted to actual structs for type safety eventually.

The current issue with that is their usages in switch statemets. To use a structure in a `case XXX: stmt` requires the structure to be castable to integral type at compile time. To do this would require all protocol definitions to be `constexpr` variables, which has some issues when in headers pre-cpp17.

This is why I put that section as internal documentation.

Packet++/header/ProtocolType.h
```diff
@@ -8,6 +8,28 @@
 /// @brief The main namespace for the PcapPlusPlus lib
 namespace pcpp
 {
+	/// @defgroup ProtocolTypes ProtocolType and ProtocolTypeFamily
```
tigercosmos, line 11: It would be better if we could do the check using the compiler, but if not, we reviewers need to read the code carefully.

Packet++/header/ProtocolType.h
seladb: I don't have a strong opinion, just thought this information could be useful for everyone using this library, not just its developers. If you feel strongly about keeping it, I'm ok with it too 🙂 

The PR is ready to merge, so whatever you think...

Dimi1010: I'm gonna keep it internal for now. We can publish more documentation on it if we move to discrete types instead of aliases. As is, the users shouldn't really depend on any internal relations on how it works, as the actual public use cases for `ProtocolType` and `ProtocolTypeFamily` are as a more complicated Enum tbh.

Packet++/header/ProtocolType.h
Dimi1010: It would be, but we need cpp17 for that.

PR #1926, opened 2025-08-16
Fix tests to work with out of source builds.

Dimi1010:


seladb:
@Dimi1010 should we close this PR?

Dimi1010:
Sure, we can do it for now.

ci/run_tests/run_tests.py
seladb, line 1: Are the changes in this file related to this PR?

I'm not sure we need the `Runner` class 🤔

ci/run_tests/run_tests_windows.py
seladb, line 1: Ditto: are the changes related to this PR?

ci/run_tests/run_tests.py
Dimi1010: Yes, they are. Otherwise, the CI doesn't run correctly as it needs the dynamic build directory to run the executables.

seladb: Ah you're right! In order to get this PR merged quickly, I suggest making the minimum changes needed in this file and in `run_tests_windows.py`. We can discuss larger refactor later (although I claim it's not needed)

seladb:
@Dimi1010 the reason the coverage tests are failing is a new gcovr version that was [released recently](https://github.com/gcovr/gcovr/releases/tag/8.5). I'm not entirely sure how to fix it, but for now we can pin to the latest version supported by all platforms as I did [here](https://github.com/seladb/PcapPlusPlus/pull/2034)

ci/run_tests/run_tests.py
Dimi1010: My concern is that the original run script's main function is getting very bloated, and adding this PR's changes won't help. I split it because I had issues with reading through it due to almost everything being dumped there.

I can split the changes of this PR from the run refactor, but I would prefer to do the refactor of the run script first since it is difficult to read and the added changes from this PR would make it even more so.

seladb: I don't think it's difficult to read. In Python you don't have to put everything under classes. That's why I think it's a discussion for a different PR, it'd be best to make the necessary changes here to get this PR merged and worry about the refactoring later

Dimi1010: > I don't think it's difficult to read.

The `run_tests.py` isn't that difficult, but the `run_tests_windows` is, due to all the code being in `main`. I would prefer both scripts be mirrors of each other where possible.

> In Python you don't have to put everything under classes.

Indeed, you don't have to. In this case I did it to save on common parameters. Otherwise the `build_dir`, `packet_test_path` and `pcap_test_path` would have had to be function parameters that would be duplicates.

> it'd be best to make the necessary changes here to get this PR merged and worry about the refactoring later

The necessary changes would make the code even more messy (referring to the windows script), which is why I want to do the refactor first (even as a separate PR). Otherwise, we are doing double the work, and I personally dislike doing double the work when I can avoid it...

seladb: This 

> The `run_tests.py` isn't that difficult, but the `run_tests_windows` is, due to all the code being in `main`. I would prefer both scripts be mirrors of each other where possible.

I don't think that `run_tests_windows.py` is hard to read



> The necessary changes would make the code even more messy (referring to the windows script), which is why I want to do the refactor first (even as a separate PR). Otherwise, we are doing double the work, and I personally dislike doing double the work when I can avoid it...

Again, I don't think the Windows script is hard to read, and the changes won't make it much more messy. I think doing the necessary changes in order to get this PR merged it the right thing to do...

Dimi1010: Minimized the changes to the windows run script.

ci/run_tests/run_tests_windows.py
```diff
@@ -101,57 +117,95 @@ def main():
...
+    )
+    parser.add_argument(
+        "--pcap-test-exe",
+        type=str,
+        default=os.path.join("Tests", "Pcap++Test", "Pcap++Test.exe"),
+        help="Custom path to Pcap++ test executable. Can be relative to the build directory.",
+    )
+    parser.add_argument(
+        "-v", "--verbose", action="store_true", help="Enable verbose logging output"
+    )
```
seladb: Is there any reason why we need this flag? 🤔 
We can always enable logging, it's just a script that runs in CI...

Dimi1010: It's not critical. We can run always in debug or always in info, if its simpler.
I added that mostly as a switch so it doesn't crowd the test results output when not needed.

seladb: I'd consider removing it, we can always add it later if we want...

I'll leave it up to you to decide. Anyway the PR is LGTM

Dimi1010: Set the logging to always use INFO. If more is needed it can be changed to debug later.

PR #1927, opened 2025-08-18
Bugfixes in X509 decoder

seladb:
This PR fixes a few bugs:
- Support cases where the "version" field doesn't exist (the version in this case should be v1)
- Support cases where neither "extensions", "subject unique ID" and "issuer unique ID" exists
- Add test for these cases


Packet++/src/X509Decoder.cpp
```diff
@@ -788,25 +791,25 @@ namespace pcpp
 
 		X509Name X509TBSCertificate::getIssuer() const
 		{
-			auto root = getSubRecordAndCast<Asn1SequenceRecord>(m_Root, getIndex(m_IssuerOffset), "Issuer");
```
Dimi1010, line 791: How does this work? `m_IssuerOffset` appears to be the same value as before, but the version offset was removed?

seladb: In all tests until this PR the version field always existed and its offset was 0, so `getIndex(m_IssuerOffset) == m_IssuerOffset`

The `getIndex()` method is actually redundant because all offsets are already calculated correctly, whether the version field exists or not

Tests/Packet++Test/Tests/X509Tests.cpp
```diff
@@ -329,6 +329,14 @@ PTF_TEST_CASE(X509VariantsParsingTest)
 		auto x509Cert = pcpp::X509Certificate::fromDERFile("PacketExamples/x509_cert_serial_lead_zeros.der");
 		PTF_ASSERT_EQUAL(x509Cert->getSerialNumber().toString(), "80");
 	}
+
+	// No version and extensions fields
+	{
+		auto x509Cert = pcpp::X509Certificate::fromPEMFile("PacketExamples/x509_cert_no_version_extension_fields.pem");
+		PTF_ASSERT_EQUAL(x509Cert->getVersion(), pcpp::X509Version::V1, enumclass);
+		PTF_ASSERT_EQUAL(x509Cert->getNotAfter().toString(), "2125-07-24 08:27:07");
```
Dimi1010: This seems to fail on mingw.  the date is set to `1970-01-01 00:00:`

seladb: Yes, the 32-bit version fails because `time_t` is of size 32-bit, so the max data is `1970 + (2^32 seconds) = ~2038`.
This is probably something to address in a separate PR, but for this PR, it doesn't matter so I changed it to `getNotBefore()` which is in 2025

Dimi1010 (approved):
LGTM

PR #1928, opened 2025-08-18
Replace packet creation macros in tests with cpp functions.

Dimi1010:
This PR adds machinery to replace `READ_FILE_AND_CREATE_PACKET` and `READ_FILE_AND_CREATE_PACKET_LINKTYPE` macro usages with cpp function `pcpp_tests::utils::createPacketFromHexResource`.

Notable classes and functions:
- **ResourceProvider:** a resource repository class that handles finding and loading resource files. (packet dumps in this case)
- **PacketFactory:** a standardized factory to create `RawPacket` instances from buffers, using set time and link type.
This class mostly removes the need for `timeval time; gettimeofday(&time, nullptr);` at the start of each test case.
NB: The factory returns `unique_ptr<RawPacket>` because `RawPacket` does not support move semantics.
- **createPacketFromHexResource:** the main function of the PR. It encapsulates the work previously done in the `READ_FILE_AND_CREATE_PACKET` macro. It takes a path to the resource and optional PacketFactory and/or ResourceProvider instances to specify behaviour of the procedure.


Tests/PcppTestUtilities/PacketFactory.hpp
seladb, line 1: Why `.hpp` and not `.h`? 🤔

Tests/PcppTestUtilities/Resources.hpp
seladb, line 1: ditto

Tests/PcppTestUtilities/Resources.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+		Resource ResourceProvider::loadResource(const char* filename, ResourceType resourceType) const
+		{
+			// Somewhat inefficient as it copies the data into a vector first, but it should work for testing,
+			// as it saves on code duplication.
+			auto vecBuffer = loadResourceToVector(filename, resourceType);
+
+			Resource resource;
+			resource.length = vecBuffer.size();
+			resource.data = std::make_unique<uint8_t[]>(vecBuffer.size());
```
seladb, line 65: Shouldn't this use `vecBuffer.data()`?

Tests/PcppTestUtilities/Resources.cpp
```diff
@@ -0,0 +1,127 @@
...
+			return resource;
+		}
+
+		std::vector<uint8_t> ResourceProvider::loadResourceToVector(const char* filename,
+		                                                            ResourceType resourceType) const
+		{
+			std::string fullPath;
+			if (!m_DataRoot.empty())
+			{
+				fullPath = m_DataRoot + '/' + filename;
```
seladb: Does this work well cross-platform? Should we use `'/'` for unix-based and `'\\'` for Windows?

Tests/PcppTestUtilities/Resources.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include <cstdint>
+
+namespace pcpp_tests
+{
+	namespace utils
+	{
+		enum class ResourceType
+		{
+			BinaryData,  ///< Resource is a file containing binary data
```
seladb: Do we ever load binary data anywhere?
If not we can remove it for now, and add back format as we need them

Tests/PcppTestUtilities/Resources.hpp
```diff
@@ -0,0 +1,54 @@
...
+			}
+
+			Resource loadResource(const char* filename, ResourceType resourceType) const;
+
+			std::vector<uint8_t> loadResourceToVector(std::string const& filename, ResourceType resourceType) const
+			{
+				return loadResourceToVector(filename.c_str(), resourceType);
+			}
+
+			std::vector<uint8_t> loadResourceToVector(const char* filename, ResourceType resourceType) const;
```
seladb: I don't think we need both overloads for each method (one that takes `const char*` and another that takes `std::string`. We anyway only use the `const char*` in all of our tests

Tests/PcppTestUtilities/Resources.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+	void setDefaultResourceProvider(utils::ResourceProvider* resourceProvider)
+	{
+		provider = resourceProvider;
+	}
+
+	utils::ResourceProvider* getDefaultResourceProvider()
+	{
+		if (provider == nullptr)
+			throw std::runtime_error("Resource provider is not set");
```
seladb: nit: add curly brackets around it

Tests/PcppTestUtilities/Resources.cpp
```diff
@@ -0,0 +1,127 @@
...
+			{
+				std::vector<uint8_t> buffer;
+
+				char hexPair[2];  // 0 - high, 1 - low
+				while (stream.read(hexPair, 2))
+				{
+					buffer.push_back(hexPairToByte(hexPair));
+				}
+				return buffer;
+			}
```
seladb: I think that in C++14 there should be a nicer way to read a hex stream file into a vector. I asked ChatGPT and here is what it provided (I didn't test this code):
```cpp
std::vector<uint8_t> readHexFile(const std::string& filename) {
    std::ifstream file(filename);
    if (!file)
        throw std::runtime_error("Failed to open file: " + filename);

    // Read the entire file into a string
    std::ostringstream oss;
    oss << file.rdbuf();
    std::string hexStr = oss.str();

    // Check for even length
    if (hexStr.length() % 2 != 0)
        throw std::runtime_error("Hex string has an odd length.");

    std::vector<uint8_t> result;
    result.reserve(hexStr.length() / 2);

    for (std::size_t i = 0; i < hexStr.length(); i += 2) {
        std::string byteStr = hexStr.substr(i, 2);
        uint8_t byte = static_cast<uint8_t>(std::stoul(byteStr, nullptr, 16));
        result.push_back(byte);
    }

    return result;
}
```

Tests/PcppTestUtilities/PacketFactory.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+		/// @brief Creates a RawPacket from a resource file.
+		/// @param resourceName The name of the resource file to read the packet data from.
+		/// @param factory The PacketFactory to use for creating the RawPacket.
+		/// @param resourceProvider An optional ResourceProvider to use for loading the resource file.
+		///   Uses the default resource provider if not provided.
+		/// @return A RawPacket object created from the resource file.
+		std::unique_ptr<pcpp::RawPacket> createPacketFromHexResource(
+		    const std::string& resourceName, const PacketFactory& factory = PacketFactory(),
+		    ResourceProvider const* resourceProvider = nullptr);
```
seladb: Is this param necessary? It'll always be set in `main.cpp`

Tests/PcppTestUtilities/PacketFactory.hpp
```diff
@@ -0,0 +1,62 @@
...
+		public:
+			/// @brief The time used for creating packets in the factory.
+			timespec factoryTime;
+			/// @brief The default link layer type for packets created by this factory.
+			pcpp::LinkLayerType defaultLinkType = pcpp::LinkLayerType::LINKTYPE_ETHERNET;
+
+			/// @brief Creates a new PacketFactory instance with the current time as factoryTime.
+			PacketFactory();
+			explicit PacketFactory(timespec time);
+			explicit PacketFactory(timeval time);
```
seladb: I wonder if we need these c'tors. Shouldn't we take the current time `std::chrono::steady_clock::now()` and convert it to a `timespec` when creating the `RawPacket`?

Tests/Packet++Test/CMakeLists.txt
```diff
@@ -51,7 +51,15 @@ add_executable(
...
-target_link_libraries(Packet++Test PUBLIC memplumber Packet++ PcppTestFramework EndianPortable)
+target_link_libraries(
+  Packet++Test
+  PUBLIC memplumber Packet++ PcppTestUtilities PcppTestFramework EndianPortable
+)
+
+if(WIN32)
+  # Otherwise winsock2.h will get conflict definitions with windows.h
+  target_compile_definitions(Packet++Test PUBLIC WIN32_LEAN_AND_MEAN)
+endif()
```
seladb, line 62: Why do we have these conflicts now? 🤔

Tests/PcppTestUtilities/Resources.cpp
Dimi1010: `make_unique`'s specialization for unique pointers holding array types takes size param, instead of a value.

It is equivalent to `unique_ptr<uint8_t[]>(new uint8_t[sizeParam]);`.

seladb: you're right, my bad

Tests/PcppTestUtilities/PacketFactory.hpp
Dimi1010: Probably, but at the moment, most of the tests that require a set time already generate `timespec` or `timeval`. We could add additional constructors later.

Tests/PcppTestUtilities/PacketFactory.hpp
Dimi1010: I have found that it helps to disambiguate between C++ only and C/C++ headers. It is mostly a convention used by some cpp libraries, most notably boost.

Can change it back to `.h` to keep it consistent tho.

Tests/PcppTestUtilities/PacketFactory.hpp
Dimi1010: I would prefer to keep it as it makes reusing the function with a different provider somewhat easier if needed without having to modify the signature. It is a dependency injection pattern and I don't think it adds much complexity, given that the param is optional.

Tests/PcppTestUtilities/Resources.cpp
Dimi1010: Fixed 93f267f76983910d286bbb216cdee7ce1b842b4b

Tests/PcppTestUtilities/Resources.hpp
Dimi1010: Unsure if we will use it. I added it as a default when I initially designed the class, but I haven't checked how all the example files are loaded. So far, I haven't found a use.

Tests/PcppTestUtilities/Resources.hpp
Dimi1010: Sure. I added those mostly as a convenience, so there is no need to write `string.c_str()`. They can be added later if an actual need arises.

Dimi1010: Nevermind... `createPacketFromHexResource` actually uses the `std::string` overload, so I guess we do use them.

Tests/Packet++Test/CMakeLists.txt
Dimi1010: It is an issue with the headers and the include order.

`winsock2.h` redefines a number of symbols that are defined in `winsock.h` (included in `windows.h`). Generally `winsock2.h` has the following line `#define _WINSOCKAPI_   /* Prevent inclusion of winsock.h in windows.h */` which should disable the `winsock.h` header from being defined, **BUT** that has to be processed **before** the first `windows.h` header is encountered.

In Packet++ and dependencies, we use `winsock2.h` in numerous files, which is fine because we don't use `windows.h` in the library.

The issue arises in `Packet++Test`:
In `main.cpp` the `getopt.h` header includes `windows.h` internally. It also includes `TestUtils.h` which previously did not have any connection to `Packet++` (w/`PCPP_TESTS_DEBUG` also has the same issues).

However, the change adds `PacketFactory.hpp` to `TestUtils.h` which includes `RawPacket.h` which includes `winsock2.h`. This is processed after `#include <getopt.h>`, so the guard on `winsock2,h` to disable `winsock.h` does nothing as that header is already processed. Next thing you know, the compilation explodes due to redefinitions.

The options to deal with it are:
- reorder the headers to have `getopt.h` after `TestUtils.h`, and keep track of all such changes in the future.
- add `WIN32_LEAN_AND_MEAN` which disables a bunch of rarely used includes from `windows.h`. `winsock.h` is one of the headers that are disabled by that definition and we save ourselves the entire headache in our test executables.
- There is possibly the way to define `_WINSOCKAPI_` manually instead, but that raises warnings in `winsock2.h` about it already being defined. (but doesn't crash)

Isn't Windows fun? :)

Tests/PcppTestUtilities/PacketFactory.hpp
seladb: I think we can change it to `.h` to make it consistent with other files in this project

Tests/PcppTestUtilities/Resources.hpp
seladb: Maybe we can remove it for now, and add it in the future if we need it?

Tests/PcppTestUtilities/Resources.hpp
seladb: it's only one usage so we can do `resourceName.c_atr()`

Tests/PcppTestUtilities/Resources.cpp
Dimi1010: I am unsure if I like that code snippet more than the current implementation.

It has 2 extra copies of the full sequence of data that the current implementation avoids.
- `fstream` -> `ostringstream` - copy full sequence (extra)
- `ostringstream.str()` - copy full sequence (extra)
- `hexStr.substr()` - copy (exists in both impl)
- `result.push_back` - copy (exists in both impl)

The only benefits I see in the code snippet are:
- The string is checked to be a multiple of 2 prior to processing.
- The vector is pre-allocated with reserve.
Both can be achieved by running `getFileLength` inside `readHexResource`.

Tests/PcppTestUtilities/PacketFactory.hpp
seladb: I guess it's ok, however, in general I prefer to keep things simple unless we can't...

Tests/PcppTestUtilities/Resources.hpp
Dimi1010: I have another `createPacketAndBufferFromHexResource` in a side branch, for the test cases that require the buffer and packet that also uses it, but I get your point.

Sure we can remove it.

Tests/PcppTestUtilities/PacketFactory.hpp
seladb: Most tests need the current time. What I'm suggesting is to only keep the default constructor that generates the current time

Tests/Packet++Test/CMakeLists.txt
seladb: Hehe, Windows is indeed fun 🙂

Tests/PcppTestUtilities/Resources.cpp
seladb: Performance is not an issue here because it's only used in tests. We have an opportunity to write a more modern code without performance constraints so I think we should use it :)

Tests/PcppTestUtilities/PacketFactory.hpp
Dimi1010: True, most tests don't require it, and they use the default factory provided from `createPacketFromHexResource`

There are still a couple that do. [PrintPacketAndLayersTest [1]](https://github.com/Dimi1010/PcapPlusPlus/blob/50d7ae023968b5992f525553021254b3cadf4cbd/Tests/Packet%2B%2BTest/Tests/PacketTests.cpp#L992-L995) does. It already generates `timeval time` via C code that is then used in the test too. In that case, I prefer adding a constructor than having to modify the test extensively. I am trying to keep the changes to the individual tests small, so it will be easier to review.

The lack of usage in the current PR for some features is mostly due to the fact that I split the changes into 3 branches and the first mostly focuses on the utility framework, rather than test conversion.

[1] - That is a link to a different branch that builds off this one.

Tests/PcppTestUtilities/Resources.cpp
Dimi1010: Sure. Updated 43247673684b596bb0068a1eaf835b0fd6da83e2 + a3ffc8a839ef5450b8d3beb5c0be3ba40bf2dffb

Didn't use the entire fragment, but it is easier to read IMO.

Tests/PcppTestUtilities/Resources.hpp
Dimi1010: Sure, removed 520226f43132a4e3a9fbd6c9dcc051a6678f07a8

Tests/PcppTestUtilities/Resources.cpp
Dimi1010: Well, seemed to work fine on my windows machine. (famous last words).

Updated it to be OS specific in any case. 913bb4ea6b483fcd93354ac86637e9fd3f9e3a06

Tests/PcppTestUtilities/Resources.hpp
Dimi1010: Removed cba8e56c502ce76ef1c35cbe28cc69df25286edc

Tests/PcppTestUtilities/PacketFactory.hpp
Dimi1010: Fixed fb5515de580f3a0f47753649566b9ef84bb88a75

PS: That kinda broke all other PR comments by marking them outdated.

Tests/PcppTestUtilities/Resources.hpp
Dimi1010: Fixed fb5515de580f3a0f47753649566b9ef84bb88a75

Tests/PcppTestUtilities/PacketFactory.h
```diff
@@ -0,0 +1,62 @@
...
+
+			/// @brief Creates a new PacketFactory instance with the current time as factoryTime.
+			PacketFactory();
+			explicit PacketFactory(timespec time);
+			explicit PacketFactory(timeval time);
+			explicit PacketFactory(pcpp::LinkLayerType linkType);
+
+			PacketFactory withTime(timespec time);
+			PacketFactory withTime(timeval time);
+			PacketFactory withLinkType(pcpp::LinkLayerType linkType);
```
seladb: Some public methods, classes and enums in this PR have doxygen documentation and some don't. I'm ok with not having any doxygen documentation because these are internal entities, but let's be consistent?

seladb: Where are we planning to use `withTime` and `withLinkType` vs. use the appropriate c'tor?

Dimi1010: Currently, nowhere. I initially made those when I thought of using a different API, but eventually settled on a constructor.
Will remove them, if you want to keep the API clean.

Dimi1010: Fixed 062e8f2224d1816e158d6d73bf281bd64ce2614a

Dimi1010: Removed 8d7b7e3680c353553d0532c62e88323100ad2d6a

seladb (approved):

PR #1929, opened 2025-08-20
Partial update tests to use new resource system.

Dimi1010:
This PR is a partial update of tests to utilize `pcpp_tests::utils::createPacketFromHexResource`.
This PR in focuses on simple cases, where regex replacement is sufficient:

Replacement 1:
original line: `READ_FILE_AND_CREATE_PACKET(2, "PacketExamples/Dhcp2.dat");`
new line: `auto rawPacket2 = createPacketFromHexResource("PacketExamples/Dhcp2.dat");`

Replacement 2:
original line: `pcpp::Packet dhcpPacket2(&rawPacket2);`
new line: `pcpp::Packet dhcpPacket2(rawPacket2.get());`

Removed section from start of test bodies, where applicable:
```cpp
timeval time;
gettimeofday(&time, nullptr);
```

Added `using pcpp_tests::utils::createPacketFromHexResource;` to test translation units.

Fixed malformed .dat hex files that contained a new line at the end of the file.

PR #1930, opened 2025-08-22
Fix telnet off by one error.

Dimi1010:
Relevant to the CI failure [[1]](https://github.com/seladb/PcapPlusPlus/actions/runs/17156780437/job/48675456270) in #1929 

The old implementation appears to have an off by one error.
`idx` would have advanced in the interval `[0, m_DataLen - advanceOffset]` instead of in the `[0, m_DataLen - advanceOffset)`.
Assuming `advance offset = 0`, the index advances one past the end of the buffer.

The new implementation instead only calculates the begin and end position and uses `std::copy_if` to perform the operation.


Dimi1010:
Can someone tell me why the fuzzers are unhappy?

seladb:
@Dimi1010 Fuzz CI fails 🤔

Dimi1010:
> @Dimi1010 Fuzz CI fails 🤔

Ik, dunno why tho, since I haven't really interacted with it much.

seladb:
Maybe there is no bug after all?

Fuzz CI says there is a buffer overflow:

```
==40==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x50b00000bdbf at pc 0x561ad77ca34f bp 0x7ffe100f9050 sp 0x7ffe100f9048
READ of size 1 at 0x50b00000bdbf thread T0
SCARINESS: 12 (1-byte-read-heap-buffer-overflow)
    #0 0x561ad77ca34e in copy_if<const unsigned char *, std::__1::back_insert_iterator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, (lambda at /src/PcapPlusPlus/Packet++/src/TelnetLayer.cpp:142:61)> /usr/local/bin/../include/c++/v1/__algorithm/copy_if.h:24:16
    #1 0x561ad77ca34e in pcpp::TelnetLayer::getDataAsString(bool) /src/PcapPlusPlus/Packet++/src/TelnetLayer.cpp:142:4
```

Dimi1010:
Which makes no sense since we should be  iterating over a reduced range of elements.

Unless for some reason `std::distance(m_Data, dataPos)` ends up negative.

seladb:
> Which makes no sense since we should be iterating over a reduced range of elements.
> 
> Unless for some reason `std::distance(m_Data, dataPos)` ends up negative.

This could be... the generated packets in Fuzz CI are usually malformed, the goal is to test the library's resistance

Dimi1010:
> > Which makes no sense since we should be iterating over a reduced range of elements.
> > 
> > Unless for some reason `std::distance(m_Data, dataPos)` ends up negative.
> 
> This could be... the generated packets in Fuzz CI are usually malformed, the goal is to test the library's resistance

The only way `distance` can end up negative is if the pointer is set before `m_Data` as the func resolves to `dataPos - m_Data` :/

Packet++/src/TelnetLayer.cpp
```diff
@@ -132,13 +134,15 @@ namespace pcpp
...
-					ss << dataPos[idx];
-			}
-			return ss.str();
+			// End of range is corrected by the advance offset.
+			auto const* beginIt = dataPos;
+			auto const* endIt = dataPos + m_DataLen - std::distance(m_Data, dataPos);
+
+			std::string result;
+			std::copy_if(beginIt, endIt, std::back_inserter(result), [](char ch) -> bool {
+				return ch > 31 && ch < 127;  // From SPACE to ~
```
clementperon, line 241: Maybe using isprint() ?

Packet++/src/TelnetLayer.cpp
```diff
@@ -132,13 +134,15 @@ namespace pcpp
...
-			std::stringstream ss;
-			for (size_t idx = 0; idx < m_DataLen - (dataPos - m_Data) + 1; ++idx)
-			{
-				if (int(dataPos[idx]) < 127 && int(dataPos[idx]) > 31)  // From SPACE to ~
-					ss << dataPos[idx];
-			}
-			return ss.str();
+			// End of range is corrected by the advance offset.
+			auto const* beginIt = dataPos;
+			auto const* endIt = dataPos + m_DataLen - std::distance(m_Data, dataPos);
```
clementperon, line 237: Add a check if std::distance is > 0 ?

Dimi1010:
I think I found the issue. There is additional bug in `getNextDataField`. Due to lack of bounds checking it runs off the buffer when checking for data fields and checks the element past the end. Thus `dataPos` ends outside the range [m_Data, m_Data + m_DataLen), equaling `m_Data + m_DataLen`.

This happens when the telenet packet does not contain any data. I suppose it might also be triggered by junk data in fuzz too.

Packet++/src/TelnetLayer.cpp
Dimi1010: Added this [section](https://github.com/seladb/PcapPlusPlus/pull/1930/files#diff-affa9e6dca5857618a35932cfc6e1c1e462a889aaab41c22c0f980cd11128848R256-R263) which should make `std::distance(m_Data, datapos)` always in the range `[m_Data, m_Data + m_DataLen)`.

Packet++/src/TelnetLayer.cpp
```diff
@@ -4,21 +4,55 @@
...
-		return pos[0] != static_cast<int>(TelnetCommand::InterpretAsCommand) ||
-		       pos[1] == static_cast<int>(TelnetCommand::InterpretAsCommand);
-	}
+		/// @brief Checks if a given sequence matches Telnet data pattern.
+		/// @param first Start of the sequence to check
+		/// @param maxCount Maximum number of bytes to check
+		/// @return True if the buffer matches Telnet data pattern, false otherwise
+		/// @throw std::runtime_error if an IAC symbol is found at the end of the buffer without a
+		/// following byte
+		bool isTelnetData(uint8_t const* first, size_t maxCount)
```
clementperon: static

Packet++/src/TelnetLayer.cpp
```diff
@@ -4,21 +4,55 @@
...
+			return first[1] == static_cast<int>(TelnetLayer::TelnetCommand::InterpretAsCommand);
+		}
+
+		/// @brief Checks if a given sequence matches Telnet command pattern.
+		/// @param first Start of the sequence to check
+		/// @param maxCount Maximum number of bytes to check
+		/// @return True if the buffer matches Telnet command pattern, false otherwise
+		/// @throw std::runtime_error if an IAC symbol is found at the end of the buffer without
+		/// a following byte
+		bool isTelnetCommand(uint8_t const* first, size_t maxCount)
```
clementperon, line 67: ditto

Packet++/src/TelnetLayer.cpp
Dimi1010: anonymous namespace does the same thing. both mean internal linkage. [[cppReference]](https://en.cppreference.com/w/cpp/language/namespace.html#:~:text=Unnamed%20namespaces%20as%20well%20as%20all%20namespaces%20declared%20directly%20or%20indirectly%20within%20an%20unnamed%20namespace%20have%20internal%20linkage%2C%20which%20means%20that%20any%20name%20that%20is%20declared%20within%20an%20unnamed%20namespace%20has%20internal%20linkage.)

Packet++/src/TelnetLayer.cpp
```diff
@@ -64,36 +98,67 @@ namespace pcpp
...
+				}
+
+				// If not data, move to next field
+				pos += getFieldLen(pos, std::distance(pos, endIt));
+			}
+		}
+		catch (std::runtime_error const& e)
+		{
+			// Most likely an IAC at the end of the buffer
+			PCPP_LOG_ERROR("Telnet Parse Error: " << e.what());
```
Dimi1010: Do we want these downgraded to `debug` level logs?
Telnet packets can theoretically be incomplete due to them being a tcp stream, so that can be somewhat common?

Packet++/src/TelnetLayer.cpp
Dimi1010: Maybe. Will open it in another PR afterwards, tho. This one got a bit heavy trying to fix the things that broke when fixing the bug.

Dimi1010:
@seladb fuzz passes now.

Packet++/src/TelnetLayer.cpp
```diff
@@ -4,21 +4,94 @@
...
+		bool isSequenceType(uint8_t const* first, size_t maxCount, TelnetSequenceType seqType)
+		{
+			return getTelnetSequenceType(first, maxCount) == seqType;
+		}
+
+		/// @brief Checks if a given sequence matches Telnet data pattern.
+		/// @param first Start of the sequence to check
+		/// @param maxCount Maximum number of bytes to check
+		/// @return True if the buffer matches Telnet data pattern, false otherwise
+		PCPP_MAYBE_UNUSED
```
seladb: Why do we need this macro? This is not part of the library API, so if it's not used, we can remove it

Packet++/src/TelnetLayer.cpp
```diff
@@ -4,21 +4,94 @@
...
+				PCPP_LOG_DEBUG("Telnet Parse Error: IAC (FF) must always be followed by another octet");
+				return TelnetSequenceType::Unknown;
+			}
+
+			if (first[1] == static_cast<int>(TelnetLayer::TelnetCommand::InterpretAsCommand))
+			{
+				// "FF FF" means data continue
+				return TelnetSequenceType::UserData;
+			}
+			else
```
seladb: nit: the `else` is not needed. We can do:
```cpp
if (first[1] == static_cast<int>(TelnetLayer::TelnetCommand::InterpretAsCommand))
{
	// "FF FF" means data continue
	return TelnetSequenceType::UserData;
}

// "FF X" where X != "FF" means command
return TelnetSequenceType::Command;
```

Packet++/src/TelnetLayer.cpp
```diff
@@ -4,21 +4,94 @@
...
+
+		/// @brief Checks if a given sequence matches a specific Telnet sequence pattern.
+		/// @param first Start of the sequence to check
+		/// @param maxCount Maximum number of bytes to check
+		/// @param seqType The desired Telnet sequence type to check for
+		/// @return True if the buffer matches the desired Telnet sequence pattern, false otherwise
+		bool isSequenceType(uint8_t const* first, size_t maxCount, TelnetSequenceType seqType)
+		{
+			return getTelnetSequenceType(first, maxCount) == seqType;
+		}
```
seladb: I think this method is redundant. We can implement `isTelnetCommand` like this:
```cpp
bool isTelnetCommand(uint8_t const* first, size_t maxCount)
{
	return getTelnetSequenceType(first, maxCount) == TelnetSequenceType::Command);
}
```

Dimi1010: Possibly, although if we are removing `isTelnetData` due to being unused, it kinda makes for asymmetric utility functions, where we have one for command sequence but not for data sequence.

Packet++/src/TelnetLayer.cpp
Dimi1010: Removed a70ed399845b496f1f47eacbeda2bb1a069e2388 & aaf251c6aeeb2070eb3c6b16c5c64998232c67cf

Packet++/src/TelnetLayer.cpp
Dimi1010: Fixed dcf4d92fc7006843ef73a3ee4383cf3f186d3589

Packet++/src/TelnetLayer.cpp
seladb: As you said, these are utility functions and not a public API, so we can implement only functions that we need. In the future, we can add `isTelnetData` if we need it

Packet++/src/TelnetLayer.cpp
```diff
@@ -117,36 +206,59 @@ namespace pcpp
...
+			if (!dataPos)
 			{
-				if (int(dataPos[idx]) < 127 && int(dataPos[idx]) > 31)  // From SPACE to ~
-					ss << dataPos[idx];
+				PCPP_LOG_DEBUG("Packet does not have a data field");
+				return std::string();
 			}
-			return ss.str();
+
+			PCPP_ASSERT(dataPos >= m_Data && dataPos < (m_Data + m_DataLen),
```
seladb, line 232: I think this `PCPP_ASSERT` macro and `AssertionUtils` are redundant. Let's just throw an exception or print an error log + return an empty string if that happens?

Dimi1010: Its not. The assert macro is so it is only added to the debug build. In release mode this is a noop.

This is done as the check should **always** be true in production code. Since the value of `dataPos` is always fetched from internal functions, therefor should have expected values at all times, in production code checking it is pure overhead that can be skipped / pruned.

seladb: If we're sure it'll never happen, we shouldn't check it. If there is a chance it'll happen, we should throw an exception. 
I don't like using these asserts that are only relevant for debug builds...

Dimi1010: This entire bug happened because we didn't check it and we had another bug cover for it so it worked semi correctly.
These always true asserts are sanity checks that help prevent situations precisely like these.

The reason they are skipped in Release is that they usually add overhead that is unnecessary in production after the code has already been verified. An example for their use is the VC++ standard library that has boundary checks on calls like `std::vector<T>::operator[]` in debug, and no checks in release. It helps catch bugs early and also maintain performance when it matters.

seladb: I'm not sure I understand. The bug happened because we didn't have tests for it, right? 


> The reason they are skipped in Release is that they usually add overhead that is unnecessary in production after the code has already been verified

I understand that asserts are removed to prevent overhead, but what I don't understand is what does "code has already been verified" mean? The only way to verify this code is by adding tests, if we don't add tests this code is never called in debug or in release...

Dimi1010: > The bug happened because we didn't have tests for it, right?

We did have tests for the method `getDataAsString`. And under normal conditions it should have returned an empty string, which it usually did and passed the tests.

The primary bug was the loop `for (size_t idx = 0; idx < m_DataLen - (dataPos - m_Data) + 1; ++idx)` which ran one over the buffer because of the `+1`. Fixing that caused the fuzz to start failing.

That leads us to the second bug:
Normally the code expected that if there is no data field `getNextDataField` to return `nullptr`, and get caught by the `if(!dataPos)` branch. That did not happen. 

`getNextDataField` had another bug where on the final loop iteration, it can call `isDataField` on the addresses `pos + len` & `pos + len + 1`, which are out of bounds, and junk data. Thus it did not return `nullptr` as expected but returned pointer to `pos + len` (out of bounds).

Returning to `getDataAsString`, now we have `dataPos` that is equal to an out of bounds pointer.
This bypasses the `if(!dataPos)` check as an out of bounds pointer still evaluates to `true`.

Now we enter the branch (`removeEscapeCharacters == true`) where we try to read an out of bounds pointer.
In the old code that somehow didn't trigger the address sanitizer.

Example:
bug-ilustration

```
m_Data = `baseAddress` // base address is the start of the buffer
m_DataLen = 7 // 7 elements in the buffer
dataPos = m_Data + m_DataLen
```

In the original code in the loop condition `idx < m_DataLen - (dataPos - m_Data) + 1`:
- `m_DataLen - (dataPos - m_Data)` equals `0` with in the above situation.
The `+1` makes the loop iterate once so it reads the element `m_Data[m_DataLen]`, which is out of bounds. This is the off by one error that this PR initially fixed. Somehow that did not get flagged by the address sanitizer!

In the refactored code the loop uses iterators:
```cpp
auto const* beginIt = dataPos;
auto const* endIt = dataPos + m_DataLen - std::distance(m_Data, dataPos);
```
Which in this situation end up:
```
beginIt = ('baseAddress' + m_DataLen) /* dataPos */

endIt = ('baseAddress' + m_DataLen) /* dataPos */ + (m_DataLen) - (m_DataLen) /* std::distance(m_Data, dataPos) */
// Simplified from the above line.
endIt = (`baseAddress` + m_DataLen) /* dataPos */
```
Both `beginIt` and `endIt` end up equal to `dataPos` which is out of bounds in the copy call, which flagged the address sanitizer, causing Fuzz failure.

---
> I understand that asserts are removed to prevent overhead, but what I don't understand is what does "code has already been verified" mean? The only way to verify this code is by adding tests, if we don't add tests this code is never called in debug or in release...

The asserts act as mini tests for internal invariants of the library implementation during debug builds. Think of it like this:
- The unit tests test the public API of an object, end to end.
- The internal asserts test the internal state of the object between steps during the public method's execution.

For example:
We can't test `getNextDataField()` in unit tests standalone due to it being a private function. We can however put an assert after a call to it to ensure when the function is called internally, the result is validates so it is within the expected range.

This means that during development if `getNextDataField` is modified and returns an invalid result, it will be immediately caught and flagged at a place where it is called inside the library. This is in contrast to using only public API unit tests where an error in the internal implementation may still somehow work and return a valid result, thus missing the internal error in the code.

It is a tool to help isolate broken contracts in the internal implementation easier and more accurately.

It also has the benefit that the internal development asserts will also run on the debug builds of all user code, thus also implicitly testing the library internals on new situations that may not be in the unit tests. This flags the error closer to the point of failure to the users for their use case, allowing them to submit the bug report to us, so it can get fixed, instead of the library silently continuing to work with faulty data.

Packet++/src/TelnetLayer.cpp
Dimi1010: Sure, removed it. d28f96480ee77bb5c070477f8f709fb4e0f57880

Packet++/src/TelnetLayer.cpp
seladb: I guess it's fine, let's merge it 👍

seladb (approved):

PR #1932, opened 2025-08-23
Fix BSD CI

tigercosmos:
Verified in https://github.com/seladb/PcapPlusPlus/pull/1932/commits/9225c4344ba5e52f644e89f50d1c91c658e4f2f4

### DIFF
--- a/Tests/Packet++Test/Tests/BgpTests.cpp
+++ b/Tests/Packet++Test/Tests/BgpTests.cpp
@@ -119,7 +119,7 @@ PTF_TEST_CASE(BgpLayerParsingTest)
 	PTF_ASSERT_EQUAL(bgpRRLayer->getHeaderLen(), 23);
 	PTF_ASSERT_EQUAL(be16toh(bgpRRLayer->getRouteRefreshHeader()->afi), 1);
 	PTF_ASSERT_EQUAL(bgpRRLayer->getRouteRefreshHeader()->safi, 1);
-	PTF_ASSERT_EQUAL(bgpRRLayer->getRouteRefreshHeader()->reserved, 2); // TESTING!!!!!
+	PTF_ASSERT_EQUAL(bgpRRLayer->getRouteRefreshHeader()->reserved, 1);
 
 	// parse BGP UPDATE message with Withdrawn Routes
 


tigercosmos:
ummm seems there are other errors...

tigercosmos:
A lot of `[ERROR: PcapLiveDevice.cpp: sendPacketUnchecked:1087 ] Error sending packet: send: No buffer space available`

tigercosmos:
Open a issue for now: https://github.com/seladb/PcapPlusPlus/issues/1933

seladb:
> A lot of `[ERROR: PcapLiveDevice.cpp: sendPacketUnchecked:1087 ] Error sending packet: send: No buffer space available`

We might be sending packets too fast in this test. Here is what ChatGPT says:

> The error "No buffer space available" from pcap_sendpacket on FreeBSD typically indicates that the kernel's transmit buffer (or ring buffer) for the network interface is full and cannot accept more packets.
> 
> Common Causes:
> 
> Sending Packets Too Fast
> If you're sending packets in a tight loop without any pacing or rate limiting, the buffer fills up faster than the NIC (Network Interface Card) can transmit them.
> 
> Interface Not in Promiscuous or Up State
> If the interface is not properly configured (e.g. down or in an unexpected state), it might cause this kind of error.
> 
> Resource Limits (e.g., mbufs)
> FreeBSD relies on a pool of memory buffers (mbufs) for networking. If the system runs out of mbufs due to high network usage, you'll get this error.
> 
> pcap_sendpacket Queues Too Many Packets
> pcap_sendpacket() queues packets for transmission but does not block, so if the driver can't keep up, it returns an error.
> 
> Solutions:
> ✅ 1. Throttle Packet Sending
> 
> Add a small delay between packet sends to avoid overwhelming the interface.
> 
> ```
> usleep(100); // 100 microseconds
> ```
> 
> ✅ 2. Increase mbuf Pool Size
> 
> You can increase the number of available mbufs in FreeBSD:
> 
> Edit `/boot/loader.conf`:
>
> 
> ```
> kern.ipc.nmbclusters="32768"
> kern.ipc.nmbjumbop="2048"
> ```
> 
> Then reboot the system.
> 
> To check current usage:
> 
> ```
> vmstat -z | grep mbuf
> ```

seladb:
@tigercosmos are you planning to work on this PR or should we close it?

tigercosmos:
Let's close it.

PR #1935, opened 2025-08-24
Improve indication for skipped DPDK CI tests.

Dimi1010:


.github/workflows/build_and_test.yml
```diff
@@ -278,14 +286,20 @@ jobs:
         run: grep avx /proc/cpuinfo
```
clementperon: maybe we could remove this step no ?

seladb: I added it mostly for visibility, but I guess we can probably remove it

seladb (approved):

.github/workflows/build_and_test.yml
Dimi1010: I think if we want to keep it, it can be pulled into the new check and under a [debug message](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-setting-a-debug-message)? That way, it is only shown if the job is run with debugging enabled. It is somewhat superfluous most of the time.

seladb: Sure, we can do that too 👍

Dimi1010: Updated it. Didn't work out precisely as indented, but good enough for now?

PR #1942, opened 2025-09-02
Optimized getLayerOfType, getNextLayerOfType and getPrevLayerOfType.

Dimi1010:
Reworked the layer search algorithm to use a single `dynamic_cast` per layer instance.
Each `dynamic_cast` call incurs a runtime overhead, so it is best they are minimized.


Packet++/header/Packet.h
```diff
@@ -340,45 +343,45 @@ namespace pcpp
...
+		return searchLayerStackForType<TLayer>(curLayer, [](Layer* layer) { return layer->getNextLayer(); }, true);
 	}
 
 	template <class TLayer> TLayer* Packet::getPrevLayerOfType(Layer* curLayer) const
+	{
+		return searchLayerStackForType<TLayer>(curLayer, [](Layer* layer) { return layer->getPrevLayer(); }, true);
+	}
+
+	template <typename TLayer, typename NextLayerFn>
+	TLayer* Packet::searchLayerStackForType(Layer* curLayer, NextLayerFn getNextLayer, bool skipFirst)
```
seladb: The parameter name `getNextLayer` is different from the declaration name `nextLayerFn`

seladb (approved):
One comment, otherwise LGTM

Packet++/header/Packet.h
Dimi1010: Fixed 87707c37a06eb8d2f819b97301b0f5d3fc796bc0

PR #1945, opened 2025-09-05
[DRAFT] Add back Npcap when running in the main repo

seladb:


Dimi1010:
@seladb I assume we have an OEM key for this repo now?

Also, can we use [conditional job execution](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions) to control whether the NPcap tests run at all.

The a job separation might be needed because the `jobs.<job-id>.if` clause might be is evaluated before `strategy.matrix`

IMO, having a duplicate winpcap job masquerade as an npcap job will cause unnecessary confusion in downstream forks.

seladb:
> @seladb I assume we have an OEM key for this repo now?
> 
> Also, can we use [conditional job execution](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions) to control whether the NPcap tests run at all.
> 
> The a job separation might be needed because the `jobs.<job-id>.if` clause might be is evaluated before `strategy.matrix`
> 
> IMO, having a duplicate winpcap job masquerade as an npcap job will cause unnecessary confusion in downstream forks.

Yes, we got the OEM account back 🎉 

Ideally we want to run a matrix of MinGW/VS version and a pcap library (WinPcap/Npcap). 

An alternative could keep everything running with WinPcap and add additional jobs running with Npcap only in the main repo, but this will create unnecessary additional jobs

Dimi1010:
> Yes, we got the OEM account back 🎉
> 
> Ideally we want to run a matrix of MinGW/VS version and a pcap library (WinPcap/Npcap).
> 
> An alternative could keep everything running with WinPcap and add additional jobs running with Npcap only in the main repo, but this will create unnecessary additional jobs

Ideally yes, a matrix would be good, if it works.

But I think if the `<job-id>.if` condition does not work on matrix jobs, a cleaner solution would be to separate each large matrix into two smaller ones, rather than try to shoestring the job specifically meant to test npcap to work without npcap, making it meaningless.
- windows - mingw - winpcap
  - conditional: { always }
  - matrix: [x64, x32]
- windows - mingw - npcap
  - conditional: { has: NPcap OEM secret }
  - matrix: [x64, x32]
- windows - vs - winpcap
  - conditional: { always }
  - matrix: [ windows-version, vs-version ]
- windows - vs - npcap
  - conditional: { has: NPcap OEM secret}
  - matrix: [windows-version, vs-version]

seladb:
> > Yes, we got the OEM account back 🎉
> > Ideally we want to run a matrix of MinGW/VS version and a pcap library (WinPcap/Npcap).
> > An alternative could keep everything running with WinPcap and add additional jobs running with Npcap only in the main repo, but this will create unnecessary additional jobs
> 
> Ideally yes, a matrix would be good, if it works.
> 
> But I think if the `<job-id>.if` condition does not work on matrix jobs, a cleaner solution would be to separate each large matrix into two smaller ones, rather than try to shoestring the job specifically meant to test npcap to work without npcap, making it meaningless.
> 
>     * windows - mingw - winpcap
>       
>       * conditional: { always }
>       * matrix: [x64, x32]
> 
>     * windows - mingw - npcap
>       
>       * conditional: { has: NPcap OEM secret }
>       * matrix: [x64, x32]
> 
>     * windows - vs - winpcap
>       
>       * conditional: { always }
>       * matrix: [ windows-version, vs-version ]
> 
>     * windows - vs - npcap
>       
>       * conditional: { has: NPcap OEM secret}
>       * matrix: [windows-version, vs-version]

That's a good idea, let me try to do that!

seladb:
> That's a good idea, let me try to do that!

@Dimi1010 I'm not sure there's a good way to do that. I tried using AI to find a solution, but the only options it proposed are:
1. Duplicating the jobs and run some of them on forks, some on the main repo and some on both (instead of using `matrix`)
2. Duplicating a `if` condition on every step
3. Use a Composite Action which is like option 1 but without duplicating all the steps

All options aren't simple, clean or nice... 😕

Dimi1010:
@seladb Hmm, I guess we can do it the original way for now.

I think I found something that might work, but I have to experiment with it for a bit to see if it does:

https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#using-an-output-to-define-two-matrices

Dimi1010:
I think I got a working solution in #1950.

seladb:
Addressed in https://github.com/seladb/PcapPlusPlus/pull/1950

PR #1946, opened 2025-09-05
Reorder of PcapFileDevice.h to improve grouping.

Dimi1010:
Reordered the `PcapFileDevice.h` header to improve the grouping between the classes.
The new grouping groups the classes by the specific type of file they operate on, instead of on the type of device they are.
IMO, this improves the cohesion, as the concrete reader / writers of a type are much more relevant to each other as a pair than to the different writer classes of unrelated file type.

Old order:
- `IFileDevice`
- `IFileReaderDevice`
- File Readers...
- `IFileWriterDevice`
- File Writers...

New order:
- File Interfaces - `IFileDevice`, `IFileReaderDevice`, `IFileWriterDevice`
- `Pcap` reader / writer
- `PcapNG` reader / writer
- `Snoop` reader.

### DIFF
--- a/Pcap++/header/PcapFileDevice.h
+++ b/Pcap++/header/PcapFileDevice.h
@@ -92,6 +92,30 @@ namespace pcpp
 		static IFileReaderDevice* getReader(const std::string& fileName);
 	};
 
+	/// @class IFileWriterDevice
+	/// An abstract class (cannot be instantiated, has a private c'tor) which is the parent class for file writer
+	/// devices
+	class IFileWriterDevice : public IFileDevice
+	{
+	protected:
+		uint32_t m_NumOfPacketsWritten;
+		uint32_t m_NumOfPacketsNotWritten;
+
+		IFileWriterDevice(const std::string& fileName);
+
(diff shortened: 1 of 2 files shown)


seladb:
@Dimi1010 the diff is very hard to follow so it's almost impossible to review this PR. The number of lines is the same and all tests pass, but otherwise very difficult to review the changes

Dimi1010:
> @Dimi1010 the diff is very hard to follow so it's almost impossible to review this PR. The number of lines is the same and all tests pass, but otherwise very difficult to review the changes

I see. Would it help if it was done class by class? That way, diff would be pure removed and added lines, instead of mixed replace operations?

seladb:
> > @Dimi1010 the diff is very hard to follow so it's almost impossible to review this PR. The number of lines is the same and all tests pass, but otherwise very difficult to review the changes
> 
> I see. Would it help if it was done class by class? That way, diff would be pure removed and added lines, instead of mixed replace operations?

Yes, maybe if there are less changes it's easier to review...

PR #1949, opened 2025-09-07
Move PcapFileWriter to be next to PcapFileReader.

Dimi1010:


seladb:
The `.cpp` diff is really messy 😕 
I'm not sure if there's a way to make it cleaner...

Dimi1010:
Yeah... :/ I think it happens there because the functions blocks are next to each other originally, without something else to separate them.

PR #1950, opened 2025-09-07
Add dynamic matrix generation for winpcap / npcap backends.

Dimi1010:
This PR adds a new step to CI pipeline that dynamically generates the windows pcap backends matrix axis that are tested against.
- **Winpcap:** Always available as a tested backend.
- **Npcap:** Included as a backend if OEM credentials are provided to the workflow via repository secrets `NPCAP_USERNAME` and `NPCAP_PASSWORD`. These secrets are required for the silent installer to install the backend.

### DIFF
--- a/Packet++/header/IPReassembly.h
+++ b/Packet++/header/IPReassembly.h
@@ -284,7 +284,7 @@ namespace pcpp
 		/// @param[in] key A pointer to the identifier of the packet that is being dropped
 		/// @param[in] userCookie A pointer to the cookie provided by the user in IPReassemby c'tor (or nullptr if no
 		/// cookie provided)
-		typedef void (*OnFragmentsClean)(const PacketKey* key, void* userCookie);
+		using OnFragmentsClean = void (*)(const PacketKey* key, void* userCookie);
 
 		/// An enum representing the status returned from processing a fragment
 		enum ReassemblyStatus
--- a/Packet++/header/ProtocolType.h
+++ b/Packet++/header/ProtocolType.h
@@ -32,11 +32,11 @@ namespace pcpp
 
 	/// @typedef ProtocolType
 	/// Representing all protocols supported by PcapPlusPlus
-	typedef uint8_t ProtocolType;
+	using ProtocolType = uint8_t;
 
 	/// @typedef ProtocolTypeFamily
 	/// Representing a family of protocols
-	typedef uint32_t ProtocolTypeFamily;
+	using ProtocolTypeFamily = uint32_t;
 
 	/// Unknown protocol (or unsupported by PcapPlusPlus)
 	const ProtocolType UnknownProtocol = 0;
--- a/Packet++/header/SomeIpSdLayer.h
+++ b/Packet++/header/SomeIpSdLayer.h
@@ -508,10 +508,10 @@ namespace pcpp
 	public:
 		friend class SomeIpSdEntry;
 
-		typedef SomeIpSdEntry* EntryPtr;
-		typedef std::vector<EntryPtr> EntriesVec;
-		typedef SomeIpSdOption* OptionPtr;
-		typedef std::vector<OptionPtr> OptionsVec;
+		using EntryPtr = SomeIpSdEntry*;
+		using EntriesVec = std::vector<EntryPtr>;
+		using OptionPtr = SomeIpSdOption*;
+		using OptionsVec = std::vector<OptionPtr>;
 
 		/// A constructor that creates the layer from an existing packet raw data
 		/// @param[in] data A pointer to the raw data
--- a/Packet++/header/TcpReassembly.h
+++ b/Packet++/header/TcpReassembly.h
@@ -308,7 +308,7 @@ namespace pcpp
 		};
 
 		/// The type for storing the connection information
-		typedef std::unordered_map<uint32_t, ConnectionData> ConnectionInfoList;
+		using ConnectionInfoList = std::unordered_map<uint32_t, ConnectionData>;
 
 		/// @typedef OnTcpMessageReady
 		/// A callback invoked when new data arrives on a connection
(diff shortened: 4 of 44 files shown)


.github/workflows/build_and_test.yml
seladb, line 1: This is a great solution! However, it will add additional checks for PRs opened on the main repo vs. PRs opened in forks.
Ideally, I'd like the same number of checks, just use WinPcap instead of Npcap if not available

Dimi1010: Why the same number of checks?

If npcap isn't available, doing a CI run with it marked as winpcap is redundant as you are running the winpcap job twice in that case?

seladb: The way it works now is that we have a combination of x64 / x86 and debug / release configurations and we attach some of them to WinPcap and others to Npcap. The assumption is that for most PRs it doesn't matter, but it still good to run with both options. I don't think we should have more jobs running on the main repo, but rather replace Npap jobs with WinPcap jobs on forked repo

Dimi1010: We don't actually build release configs in the `build_and_test` CI as far as I know. Only debug.

It can be done so it uses Npcap backend if available and Winpcap if not. But that would have to be full replacement as it would be done in the matrix axis generation job, which would affect all Windows jobs.

I am worried as that would mean fully dropping Winpcap CI coverage on the main repo completely. My assumption is that most PRs from forked repos that want to contribute will target the main repo directly, so only main repo CI will run.

seladb: > We don't actually build release configs in the `build_and_test` CI as far as I know. Only debug.

You're right, we don't have release builds, we should probably have at least one...

> It can be done so it uses Npcap backend if available and Winpcap if not. But that would have to be full replacement as it would be done in the matrix axis generation job, which would affect all Windows jobs.

I don't think we should always use Npcap if available or WinPcap if not. Ideally we should have a mix of both, but in forked repos, fallback to WinPcap. For example:
```
  visual-studio:
    strategy:
      matrix:
        include:
          - os: windows-2025
            arch: Win32
            pcap_lib: "winpcap"
          - os: windows-2025
            arch: "x64"
            pcap_lib: "npcap"  # run on WinPcap in forked repos
          - os: windows-2022
            arch: "x64"
            pcap_lib: "npcap"  # run on WinPcap in forked repos 
```

In both main and forked repos we should have these 3 jobs, but in forked repos they should all use WinPcap

Dimi1010: > > It can be done so it uses Npcap backend if available and Winpcap if not. But that would have to be full replacement as it would be done in the matrix axis generation job, which would affect all Windows jobs.
> 
> I don't think we should always use Npcap if available or WinPcap if not. Ideally, we should have a mix of both, but in forked repos, fallback to WinPcap. For example:
> ```
>   visual-studio:
>     strategy:
>       matrix:
>         include:
>           - os: windows-2025
>             platform: "Visual Studio 17 2022"
>             arch: Win32
>             pcap_lib: "winpcap"
>           - os: windows-2025
>             platform: "Visual Studio 17 2022"
>             arch: "x64"
>             pcap_lib: "npcap"  # run on WinPcap in forked repos
>           - os: windows-2022
>             platform: "Visual Studio 17 2022"
>             arch: "x64"
>             pcap_lib: "npcap"  # run on WinPcap in forked repos 
> ```
> 
> In both main and forked repos we should have these 3 jobs, but in forked repos they should all use WinPcap

~~Well, that can't be done with the dynamic generation since that generates an axis. It can't generate individual jobs.~~ See edit.

IMO, having the extra Npcap jobs is preferable than having a CI job that says that it is an npcap job, but surprise, it actually isn't. That is from both coverage and maintenance perspectives. Debugging a failed CI job that silently replaces a dependency with something else would be much harder than just not having the run job in the first place because the dependency is missing.

The way the jobs get named, people will see "pcap_lib=npcap" in the name and, without looking further in the code, will assume it uses npcap.

The jobs already cancel when a new push is done on a PR, so even if the CI is somewhat longer, it will abort if it is unnecessary.

Edit: 
Upon further thought, I suppose you can set the `pcap_lib` to read from a variable for the flip flop jobs, but I kinda dislike that because of variable coverage. IMO, the main repo should not have less CI coverage than the forked repos. :/

Dimi1010: @seladb Managed to get it to work. Can you check the VS job?

seladb: Maybe we can have something like this?
```
  visual-studio:
    strategy:
      matrix:
        include:
          - os: windows-2025
            arch: Win32
            pcap_lib: "winpcap"
            run_on: all-repos
          - os: windows-2025
            arch: "x64"
            pcap_lib: "npcap"
            run_on: main-repo
          - os: windows-2025
            arch: "x64"
            pcap_lib: "winpcap"
            run_on: forked-repos
          - os: windows-2022
            arch: "x64"
            pcap_lib: "npcap"
            run_on: main-repo
          - os: windows-2022
            arch: "x64"
            pcap_lib: "winpcap"
            run_on: forked-repos
```

It's not ideal to have so many jobs, but maybe it's the most explicit option, and also not too hard to implement

Dimi1010: The most recent commit already does that.
The variable npcap-or-default resolves to 'npcap' if OEM credentials are available and 'winpcap' if not.

IMO, that is enough. Forked repos should be able to run the CI with Npcap with no changes, provided they supply their own OEM secret.

.github/workflows/build_and_test.yml
```diff
@@ -504,18 +504,50 @@ jobs:
...
-            sys: mingw64
-            pcap_lib: "winpcap"
-
+        pcap_lib: ${{ fromJSON(needs.define-windows-pcap-backend-matrix.outputs.pcap-backends) }}
+        sys: [mingw32, mingw64]
+        include: # Extends the matrix to add the env variable for each matching sys
+          - sys: mingw32
+            env: i686
+          - sys: mingw64
+            env: x86_64
```
seladb: Maybe we should define something similar to what we have in VS?
```
      matrix:
        include:
          - sys: mingw32
            env: i686
            pcap_lib: "winpcap"
          - sys: mingw64
            env: x86_64
            pcap_lib: ${{ needs.define-windows-pcap-backend-matrix.outputs.npcap-or-default }}
```

If we do that the `pcap-backends` output is not needed

Dimi1010: Updated 28a25f40212d7848c11087c4fd4f51d80ffd4c90 & 16aa358b06fe32fb59d4e704d116d879267bd98c

The name of the job is a bit inaccurate now, but it can probably do.

seladb:
Thank you for working on this @Dimi1010 ! 🙏

PR #1954, opened 2025-09-10
Add boundary checks to prevent memory safety issues in BgpLayer::getHeaderLen

### LINKED ISSUE #1864: [Bug]: Heap-Use-After-Free and Heap-Buffer-Overflow in `BgpLayer::getHeaderLen()`
### Summary

[pocs.tar.gz](https://github.com/user-attachments/files/21074165/pocs.tar.gz)
[asan_reports.tar.gz](https://github.com/user-attachments/files/21074166/asan_reports.tar.gz)

AddressSanitizer detected two critical memory safety issues in `BgpLayer.cpp` of the PcapPlusPlus project: **Heap-use-after-free** and **Heap-buffer-overflow** in `BgpLayer::getHeaderLen()`. These can both lead to crashes and potential security vulnerabilities.

#### Suspected Problematic Code
```C++
size_t BgpLayer::getHeaderLen() const
{
    if (m_DataLen < sizeof(bgp_common_header))
        return m_DataLen;

    uint16_t messageLen = be16toh(getBasicHeader()->length);
    if (m_DataLen < messageLen)
        return m_DataLen;

    return (size_t)messageLen;
}
```
#### Possible Explanation
- **For use-after-free:** The lifecycle of the packet/layer objects and their backing memory is not tightly controlled. After freeing the buffer, the code still tries to access fields via dangling pointers.
- **For buffer-overflow:** Insufficient bounds checking when parsing variable-length and potentially malformed BGP messages, especially if the internal `length` field is corrupted.

#### Suggested Fix
- **Use-after-free:**

  - Ensure that all layer/packet objects referencing a `RawPacket` are invalidated or detached before the underlying memory is freed or cleared.
  - Consider using smart pointers or explicit memory management to enforce ownership/lifetime guarantees.

- **Buffer-overflow:**

  - Always validate that accesses to `m_Data` and any structure members are within `m_DataLen`.
  - When reading the length field from the header, ensure the header is fully within the buffer, and that the field's value does not cause out-of-bounds accesses for further parsing.

### Platform
 - OS: `Ubuntu 22.04.5 LTS`
 - Clang version: `Ubuntu clang version 16.0.6 (++20231112100510+7cbf1a259152-1~exp1~20231112100554.106)`
 - Version commit `0588d88eca769a02e660161f8965ef6152ada90e`

### Steps to Reproduce

```shell
export SRC=/src OUT=/out
mkdir $SRC $OUT

apt-get update && apt-get install -y cmake autoconf flex bison zip
export CC=clang CXX=clang++ CFLAGS='-fsanitize=address -g -O1' CXXFLAGS='-fsanitize=address -g -O1' LDFLAGS="-fuse-ld=lld" LIB_FUZZING_ENGINE='-fsanitize=fuzzer'

git clone --depth=1 https://github.com/seladb/PcapPlusPlus PcapPlusPlus
git clone --depth=1 https://github.com/the-tcpdump-group/tcpdump.git tcpdump
git clone --depth=1 https://github.com/the-tcpdump-group/libpcap.git libpcap

cd $SRC/PcapPlusPlus
$SRC/PcapPlusPlus/Tests/Fuzzers/ossfuzz.sh

$OUT/FuzzTargetNg poc
```

### ASAN report
**Use-after-free:**
```
=================================================================
==530017==ERROR: AddressSanitizer: heap-use-after-free on address 0x613000000585 at pc 0x598652937a70 bp 0x7ffeb98a5560 sp 0x7ffeb98a5558
READ of size 2 at 0x613000000585 thread T0
    #0 0x598652937a6f in pcpp::BgpLayer::getHeaderLen() const /src/PcapPlusPlus/Packet++/src/BgpLayer.cpp:21:25
    #1 0x5986529f306c in pcpp::Packet::extendLayer(pcpp::Layer*, int, unsigned long) /src/PcapPlusPlus/Packet++/src/Packet.cpp:621:33
    #2 0x5986529d4823 in pcpp::Layer::extendLayer(int, unsigned long) /src/PcapPlusPlus/Packet++/src/Layer.cpp:85:20
    #3 0x59865293b3c4 in pcpp::BgpOpenMessageLayer::setOptionalParameters(std::vector<pcpp::BgpOpenMessageLayer::optional_parameter, std::allocator<pcpp::BgpOpenMessageLayer::optional_parameter>> const&) /src/PcapPlusPlus/Packet++/src/BgpLayer.cpp:267:15
    #4 0x5986528eb1ad in readParsedPacket(pcpp::Packet, pcpp::Layer*) /src/PcapPlusPlus/Tests/Fuzzers/ReadParsedPacket.h:440:22
    #5 0x5986528e5b50 in LLVMFuzzerTestOneInput /src/PcapPlusPlus/Tests/Fuzzers/FuzzTarget.cpp:66:5
    #6 0x5986527f4e90 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) crtstuff.c
    #7 0x5986527ded6f in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) crtstuff.c
    #8 0x5986527e4826 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) crtstuff.c
    #9 0x59865280dcf2 in main (/out/FuzzTargetNg+0x1a7cf2) (BuildId: 97bfbeafd983e755)
    #10 0x76efe31c8d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #11 0x76efe31c8e3f in __libc_start_main csu/../csu/libc-start.c:392:3
    #12 0x5986527d9c74 in _start (/out/FuzzTargetNg+0x173c74) (BuildId: 97bfbeafd983e755)

0x613000000585 is located 5 bytes inside of 337-byte region [0x613000000580,0x6130000006d1)
freed by thread T0 here:
    #0 0x5986528e2d5d in operator delete[](void*) (/out/FuzzTargetNg+0x27cd5d) (BuildId: 97bfbeafd983e755)
    #1 0x598652a23a08 in pcpp::RawPacket::clear() /src/PcapPlusPlus/Packet++/src/RawPacket.cpp:100:4
    #2 0x598652a23a08 in pcpp::RawPacket::~RawPacket() /src/PcapPlusPlus/Packet++/src/RawPacket.cpp:24:3
    #3 0x5986527f4e90 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) crtstuff.c
    #4 0x5986527ded6f in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) crtstuff.c
    #5 0x5986527e4826 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) crtstuff.c
    #6 0x59865280dcf2 in main (/out/FuzzTargetNg+0x1a7cf2) (BuildId: 97bfbeafd983e755)
    #7 0x76efe31c8d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16

previously allocated by thread T0 here:
    #0 0x5986528e250d in operator new[](unsigned long) (/out/FuzzTargetNg+0x27c50d) (BuildId: 97bfbeafd983e755)
    #1 0x598652a23bc7 in pcpp::RawPacket::copyDataFrom(pcpp::RawPacket const&, bool) /src/PcapPlusPlus/Packet++/src/RawPacket.cpp:60:16
    #2 0x598652a23bc7 in pcpp::RawPacket::RawPacket(pcpp::RawPacket const&) /src/PcapPlusPlus/Packet++/src/RawPacket.cpp:30:3

SUMMARY: AddressSanitizer: heap-use-after-free /src/PcapPlusPlus/Packet++/src/BgpLayer.cpp:21:25 in pcpp::BgpLayer::getHeaderLen() const
Shadow bytes around the buggy address:
  0x613000000300: 00 00 00 00 00 00 00 00 00 00 00 00 00 fa fa fa
  0x613000000380: fa fa fa fa fa fa fa fa 00 00 00 00 00 00 00 00
  0x613000000400: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x613000000480: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x613000000500: 00 00 01 fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0x613000000580:[fd]fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x613000000600: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x613000000680: fd fd fd fd fd fd fd fd fd fd fd fa fa fa fa fa
  0x613000000700: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
  0x613000000780: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x613000000800: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==530017==ABORTING
```

**Buffer-overflow:**
```
=================================================================
==530034==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6170000004d0 at pc 0x64a8cc0b5a70 bp 0x7fff24c584a0 sp 0x7fff24c58498
READ of size 2 at 0x6170000004d0 thread T0
    #0 0x64a8cc0b5a6f in pcpp::BgpLayer::getHeaderLen() const /src/PcapPlusPlus/Packet++/src/BgpLayer.cpp:21:25
    #1 0x64a8cc17256c in pcpp::Packet::shortenLayer(pcpp::Layer*, int, unsigned long) /src/PcapPlusPlus/Packet++/src/Packet.cpp:673:33
    #2 0x64a8cc1535e8 in pcpp::Layer::shortenLayer(int, unsigned long) /src/PcapPlusPlus/Packet++/src/Layer.cpp:114:20
    #3 0x64a8cc0bd645 in pcpp::BgpUpdateMessageLayer::setWithdrawnRoutes(std::vector<pcpp::BgpUpdateMessageLayer::prefix_and_ip, std::allocator<pcpp::BgpUpdateMessageLayer::prefix_and_ip>> const&) /src/PcapPlusPlus/Packet++/src/BgpLayer.cpp:581:15
    #4 0x64a8cc069b62 in readParsedPacket(pcpp::Packet, pcpp::Layer*) /src/PcapPlusPlus/Tests/Fuzzers/ReadParsedPacket.h:451:24
    #5 0x64a8cc063b50 in LLVMFuzzerTestOneInput /src/PcapPlusPlus/Tests/Fuzzers/FuzzTarget.cpp:66:5
    #6 0x64a8cbf72e90 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) crtstuff.c
    #7 0x64a8cbf5cd6f in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) crtstuff.c
    #8 0x64a8cbf62826 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) crtstuff.c
    #9 0x64a8cbf8bcf2 in main (/out/FuzzTargetNg+0x1a7cf2) (BuildId: 97bfbeafd983e755)
    #10 0x7429e2b78d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #11 0x7429e2b78e3f in __libc_start_main csu/../csu/libc-start.c:392:3
    #12 0x64a8cbf57c74 in _start (/out/FuzzTargetNg+0x173c74) (BuildId: 97bfbeafd983e755)

0x6170000004d0 is located 430 bytes after 674-byte region [0x617000000080,0x617000000322)
allocated by thread T0 here:
    #0 0x64a8cc06050d in operator new[](unsigned long) (/out/FuzzTargetNg+0x27c50d) (BuildId: 97bfbeafd983e755)
    #1 0x64a8cc1a2e70 in pcpp::RawPacket::reallocateData(unsigned long) /src/PcapPlusPlus/Packet++/src/RawPacket.cpp:144:24
    #2 0x64a8cc169a22 in pcpp::Packet::reallocateRawData(unsigned long) /src/PcapPlusPlus/Packet++/src/Packet.cpp:202:21
    #3 0x64a8cc170ed3 in pcpp::Packet::extendLayer(pcpp::Layer*, int, unsigned long) /src/PcapPlusPlus/Packet++/src/Packet.cpp
    #4 0x64a8cc152823 in pcpp::Layer::extendLayer(int, unsigned long) /src/PcapPlusPlus/Packet++/src/Layer.cpp:85:20
    #5 0x64a8cc0bd1f7 in pcpp::BgpUpdateMessageLayer::setWithdrawnRoutes(std::vector<pcpp::BgpUpdateMessageLayer::prefix_and_ip, std::allocator<pcpp::BgpUpdateMessageLayer::prefix_and_ip>> const&) /src/PcapPlusPlus/Packet++/src/BgpLayer.cpp:571:15
    #6 0x64a8cc069b62 in readParsedPacket(pcpp::Packet, pcpp::Layer*) /src/PcapPlusPlus/Tests/Fuzzers/ReadParsedPacket.h:451:24
    #7 0x64a8cc063b50 in LLVMFuzzerTestOneInput /src/PcapPlusPlus/Tests/Fuzzers/FuzzTarget.cpp:66:5
    #8 0x64a8cbf72e90 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) crtstuff.c
    #9 0x64a8cbf5cd6f in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) crtstuff.c
    #10 0x64a8cbf62826 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) crtstuff.c
    #11 0x64a8cbf8bcf2 in main (/out/FuzzTargetNg+0x1a7cf2) (BuildId: 97bfbeafd983e755)
    #12 0x7429e2b78d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16

SUMMARY: AddressSanitizer: heap-buffer-overflow /src/PcapPlusPlus/Packet++/src/BgpLayer.cpp:21:25 in pcpp::BgpLayer::getHeaderLen() const
Shadow bytes around the buggy address:
  0x617000000200: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x617000000280: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x617000000300: 00 00 00 00 02 fa fa fa fa fa fa fa fa fa fa fa
  0x617000000380: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x617000000400: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0x617000000480: fa fa fa fa fa fa fa fa fa fa[fa]fa fa fa fa fa
  0x617000000500: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x617000000580: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x617000000600: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x617000000680: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x617000000700: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==530034==ABORTING
```

PcapPlusPlus master branch

hash: `0588d88eca769a02e660161f8965ef6152ada90e`

Linux

_No response_

### Compiler version

clang version 16.0.6

_No response_

danasana:
Closes #1864 

This PR fixes heap-use-after-free and heap-buffer-overflow issues  in BgpLayer::getHeaderLen(), reported by AddressSanitizer when extending or shortening the BGP layer

Added boundary checks to ensure that requested offsets and lengths do not exceed:
  - the current layer length,
  - the total packet length,
  - the next layer boundary.

Added regression_samples


Packet++/src/Layer.cpp
```diff
@@ -78,6 +78,25 @@ namespace pcpp
 			return true;
 		}
 
+		if ((size_t)offsetInLayer > m_DataLen)
+		{
+			PCPP_LOG_ERROR("Requested offset is larger than data length");
+			return false;
+		}
```
Dimi1010: This check is identical to the one in `if (m_Packet == nullptr)` branch.
Maybe we can pull both checks to be done prior to the branch point?

Packet++/src/Layer.cpp
```diff
@@ -107,6 +126,32 @@ namespace pcpp
 			return true;
 		}
 
+		if ((size_t)offsetInLayer >= m_DataLen)
+		{
+			PCPP_LOG_ERROR("Requested offset is larger than data length");
+			return false;
+		}
```
Dimi1010: ditto, identical checks on both branches.

Packet++/src/Layer.cpp
Dimi1010, line 1: We should possibly check for if `Layer::extend/shorten` are called with negative values for `offsetInLayer`?

Packet++/src/Layer.cpp
```diff
@@ -78,6 +78,25 @@ namespace pcpp
...
 		}
 
+		if ((size_t)offsetInLayer > m_DataLen)
+		{
+			PCPP_LOG_ERROR("Requested offset is larger than data length");
+			return false;
+		}
+
+		if (m_Data - m_Packet->m_RawPacket->getRawData() + (ptrdiff_t)offsetInLayer 
+			> (ptrdiff_t)m_Packet->m_RawPacket->getRawDataLen())
```
Dimi1010: nit: Can we use distance here? IMO, it somewhat helps group the expression.

```suggestion
		if (std::distance(m_Packet->m_RawPacket->getRawData(), m_Data) + (ptrdiff_t)offsetInLayer 
			> (ptrdiff_t)m_Packet->m_RawPacket->getRawDataLen())
```

Packet++/src/Layer.cpp
```diff
@@ -107,6 +126,32 @@ namespace pcpp
...
+			return false;
+		}
+
+		if ((size_t)offsetInLayer + numOfBytesToShorten > m_DataLen)
+		{
+			PCPP_LOG_ERROR("Requested number of bytes to shorten is larger than data length");
+			return false;
+		}
+
+		if (m_Data - m_Packet->m_RawPacket->getRawData() + (ptrdiff_t)offsetInLayer + (ptrdiff_t)numOfBytesToShorten
```
Dimi1010: nit, ditto: std::distance.
```suggestion
		if (std::distance(m_Packet->m_RawPacket->getRawData(), m_Data) + (ptrdiff_t)offsetInLayer + (ptrdiff_t)numOfBytesToShorten
```

Packet++/src/Layer.cpp
```diff
@@ -107,6 +126,32 @@ namespace pcpp
 			return true;
 		}
 
+		if ((size_t)offsetInLayer >= m_DataLen)
```
Dimi1010: nit: We try to use C++ casts in new code and port the old C-casts to C++ casts.
Could you change the added casts to `static_cast`?

Dimi1010:
> I'm unsure why the CI has a failure on checkout on windows. :/

I think the error is that Windows can't contain `:` in the file names as that is reserved symbol.

Packet++/src/Layer.cpp
danasana: Since m_Packet->m_RawPacket->getRawData() is const and m_Data is not, we'll need to use a static_cast, which I think will make the expression less readable

Dimi1010: I see. Alright. We can leave it as is.

Dimi1010 (approved):

Packet++/src/Layer.cpp
```diff
@@ -78,6 +84,19 @@ namespace pcpp
...
+		{
+			PCPP_LOG_ERROR("Requested offset is larger than total packet length");
+			return false;
+		}
+
+		if (m_NextLayer != nullptr && static_cast<ptrdiff_t>(offsetInLayer) > m_NextLayer->m_Data - m_Data)
+		{
+			PCPP_LOG_ERROR("Requested offset exceeds current layer's boundary");
+			return false;
+		}
```
seladb: Why are these checks needed? All we need to check is that `offsetInLayer` is within the layer's boundary, which means it is smaller than or equal to `m_DataLen`. Why do we need to compare it to the whole packet or involve `m_NextLayer`?

Packet++/src/Layer.cpp
```diff
@@ -107,6 +132,28 @@ namespace pcpp
...
+
+		if (m_NextLayer != nullptr &&
+		    static_cast<ptrdiff_t>(offsetInLayer) + static_cast<ptrdiff_t>(numOfBytesToShorten) >
+		        m_NextLayer->m_Data - m_Data)
+		{
+			PCPP_LOG_ERROR("Requested number of bytes to shorten exceeds current layer's boundary");
+			return false;
+		}
```
seladb: ditto

Packet++/src/Packet.cpp
```diff
@@ -609,6 +609,9 @@ namespace pcpp
 			// assuming header length of the layer that requested to be extended hasn't been enlarged yet
 			size_t headerLen = curLayer->getHeaderLen() + (curLayer == layer ? numOfBytesToExtend : 0);
 			dataPtr += headerLen;
+
+			if (dataPtr > m_RawPacket->getRawData() + m_RawPacket->getRawDataLen())
+				break;
```
seladb: How can such a thing happen? `curLayer->getHeaderLen()` should never exceed the packet data, unless we have a bug in one of the layers

Packet++/src/Packet.cpp
```diff
@@ -660,6 +663,8 @@ namespace pcpp
 			// assuming header length of the layer that requested to be extended hasn't been enlarged yet
 			size_t headerLen = curLayer->getHeaderLen() - (curLayer == layer ? numOfBytesToShorten : 0);
 			dataPtr += headerLen;
+			if (dataPtr > m_RawPacket->getRawData() + m_RawPacket->getRawDataLen())
+				break;
```
seladb: ditto

Packet++/src/Layer.cpp
danasana: I analyzed several malformed packets that triggered sanitizer errors, and identified some of the root causes.

When iterating over layers during recalculation after shortening a layer, the pointer to the next layer is computed as `dataPtr` (current layer pointer) + `headerLen`. For BGP, `headerLen` corresponds either to the length field of the `bgp_common_header` structure or to the actual layer length `m_DataLen `— whichever is smaller. In malformed packets, the header length may not match the actual layer length, or may contain arbitrary values (possibly due to previous incorrect shortening or extension), which can lead to several issues

- `numOfBytesToShorten` is subtracted from the `m_DataLen` of all layers preceding the shortened one. If `numOfBytesToShorten` is greater than `m_DataLen`, this causes `m_DataLen` to wrap around, resulting in an extremely large value. In such cases, when calculating `headerLen`, the header field is selected, which may be invalid, leading to out-of-bounds memory access.

- In the shortened layer, `numOfBytesToShorten` is first subtracted from `m_DataLen`, then again from `headerLen`. `BgpLayer::getHeaderLen()` returns new `m_DataLen` as the smallest, which results in `numOfBytesToShorten` being subtracted twice, potentially producing a negative value.

I believe `headerLen` calculation could be moved before updating the current layer’s `m_DataLen`. However, this would not resolve the issue described in the next point.

- Malformed packets may also contain nested BGP layers, which is inherently invalid and can trigger the errors described earlier.

Packet++/src/Layer.cpp
danasana: Length-related errors introduced during layer shortening can result in the same sanitizer issues when the layer is later extended

Packet++/src/Layer.cpp
seladb: > * `numOfBytesToShorten` is subtracted from the `m_DataLen` of all layers preceding the shortened one. If `numOfBytesToShorten` is greater than `m_DataLen`

How can `numOfBytesToShorten` be greater than `m_DataLen`? 🤔 



> * In the shortened layer, `numOfBytesToShorten` is first subtracted from `m_DataLen`, then again from `headerLen`. `BgpLayer::getHeaderLen()` returns new `m_DataLen` as the smallest, which results in `numOfBytesToShorten` being subtracted twice, potentially producing a negative value.

I believe this is only relevant to the changes in `Packet.cpp`, right?

> * Malformed packets may also contain nested BGP layers, which is inherently invalid and can trigger the errors described earlier.

By nested BPG layers, do you mean a BGP layer that comes after another BGP layer? If yes, this is actually a valid scenario that we support: https://github.com/seladb/PcapPlusPlus/blob/32384c2915d2c08efc6578349b51596971d76bb0/Packet%2B%2B/src/BgpLayer.cpp#L80
What's the issue in this case?

Packet++/src/Layer.cpp
seladb: I'm not sure I understand... can you elaborate?

Packet++/src/Layer.cpp
danasana: > How can numOfBytesToShorten be greater than m_DataLen? 🤔

I'll try to explain with an example. The diagrams show the packet layout in memory. Each dot represents the start of a layer, layer addresses are represented by the last two bytes, and for some points I've calculated the actual offset from the beginning of the packet for clarity.
image


Let's consider a packet from the file `*d94f8b`  from regression_samples before shortening that caused sanitizer error.
The fragment that should be removed is highlighted in bold (`atIndex = 148`, `numOfBytesToShorten = 46`).

The first thing that stands out is that it gets to the next layer. Simply adding a check like `(size_t)offsetInLayer + numOfBytesToShorten >= m_DataLen` won't be sufficient in this case, because `m_DataLen` of BGP layers doesn't match the actual layer size in memory:
102 - 66 = 36, but DataLen = 95
183 - 125 = 58, but DataLen = 73

However, the error doesn't occur because of this. After removing the fragment, `m_DataLen` of previous layers is recalculated according to the condition:
```
if (!passedExtendedLayer)
    curLayer->m_DataLen -= numOfBytesToShorten;
```

Since the layer at address `afe6` has length of 23, subtracting 46 causes an overflow. The `headerLen` is taken from the `bgp_common_header` structure (which is 907 in this case), `dataPtr` points beyond the packet boundaries, and subsequent call to BgpLayer::getHeaderLen() results in a heap-buffer-overflow error.

> By nested BPG layers, do you mean a BGP layer that comes after another BGP layer?

I mean BGP layers where one layer is inside another, i.e. in the BGP payload there is another BGP layer.
For example, in file `*d6be97` the following situation occurs:

image

> I believe this is only relevant to the changes in Packet.cpp, right?

If you mean moving the calculation of `headerLen`, then yes, this applies to changes in Packet.cpp.

Packet++/src/Layer.cpp
danasana: I wanted to say that problems with `m_DataLen` not matching the actual layer length and the length from the `bgp_common_header`, as in the examples below, can lead to problems with exceeding the boundaries of the layer or packet during extension.

Packet++/src/Layer.cpp
seladb: Thanks for the thorough and detailed explanation, much appreciated! 🙏 

I have 2 follow-up questions:
1. Since the fixes are within `extendLayer()` and `shortenLayer()` I assume you found them when trying to edit packets, is that correct? As far as I know, our regression tests don't edit packets, so the files you added to `Tests/Fuzzers/RegressionTests/regression_samples` don't actually test the scenario being fixed
2. Is there any way to make these fixes inside `BgpLayer` or do they have to be `Layer` and `Packet`?

danasana: 1. This PR addresses bug fix #1864, and the files added to regression_samples are poc files that triggered sanitizer errors. I can write additional tests if needed

2. As for the BGP layer, I believe the following checks could be implemented:

- Disallow nested BGP layers.
- Do not shorten the length of the BGP layers that precede the layer that is actually being shortened.
- Ensure the actual length matches the `m_DataLen`.
- Update the length field in the header struct.
- Disallow BGP layers shorter than the mandatory header length (Header + Mandatory Attributes).

However, I'm not 100% sure these measures will be sufficient. There might be some combination of shorten/extend that could still lead to memory access errors. Also, shorten/extend operations with other layers could cause the same type of issues.

seladb: > 1. This PR addresses bug fix https://github.com/seladb/PcapPlusPlus/issues/1864, and the files added to regression_samples are poc files that triggered sanitizer errors. I can write additional tests if needed

I apologize, our fuzz tests do modify packets: https://github.com/seladb/PcapPlusPlus/blob/32384c2915d2c08efc6578349b51596971d76bb0/Tests/Fuzzers/FuzzTarget.cpp#L66
So adding these files should be ok.

BTW, did you have a chance to run these files before the fix and they failed?

> 2\. As for the BGP layer, I believe the following checks could be implemented

What I meant is fix `BgpOpenMessageLayer::setOptionalParameters` to disallow calling `shortenLayer` and `extendLayer` if the packet data is invalid. Would it fix the issue?

danasana: > BTW, did you have a chance to run these files before the fix and they failed?

Yes, each of these files triggered a sanitizer error (heap-buffer-overflow or heap-use-after-free) when calling `pcpp::BgpLayer::getHeaderLen()` on the problematic layer.

> What I meant is fix `BgpOpenMessageLayer::setOptionalParameters` to disallow calling `shortenLayer` and `extendLayer` if the packet data is invalid.

The checks could be moved into the BGP layer. To do that, they’d need to be extracted into a separate helper method, since `shortenLayer`/`extendLayer` can be called from:

- `BgpOpenMessageLayer::setOptionalParameters`
- `BgpUpdateMessageLayer::setWithdrawnRoutes`
- `BgpUpdateMessageLayer::setPathAttributes`
- `BgpUpdateMessageLayer::setNetworkLayerReachabilityInfo`
- `BgpNotificationMessageLayer::setNotificationData`

Also, `shortenLayer`/`extendLayer` are used by other protocols as well, so similar issues could potentially appear there too.

At the very least, it makes sense to keep in `shortenLayer` a check to ensure the number of bytes being removed doesn’t exceed the layer’s length:
```
(size_t)offsetInLayer + numOfBytesToShorten > m_DataLen
```

And it would be better to move the `headerLen` calculation before updating the current layer `m_DataLen`, to avoid double subtraction.

seladb: [This PR](https://github.com/seladb/PcapPlusPlus/pull/1812) fixes the issue in `setNetworkLayerReachabilityInfo` 🙂 

I think we can do the same for the rest of the setters in BPG layer. In general, we typically try to fix bugs where they occur, so lower-level methods such as `shortenLayer` and `extendLayer` don't receive "garbage data".

> At the very least, it makes sense to keep in `shortenLayer` a check to ensure the number of bytes being removed doesn’t exceed the layer’s length

I guess we can do that, sure 👍 
I'd only keep this check and remove the rest

danasana: So, we’re keeping the check for the number of bytes being removed in `Layer.cpp`, and moving the rest into the BGP layer?

seladb: Yes, I think we can keep this specific check, but fix the real issue in the setters of BGP layer

Packet++/src/Packet.cpp
```diff
@@ -656,20 +662,29 @@ namespace pcpp
...
+			size_t headerLen = curLayer->getHeaderLen();
+
 			// change the data length only for layers who come before the shortened layer. For layers who come after,
 			// data length isn't changed
 			if (!passedExtendedLayer)
 				curLayer->m_DataLen -= numOfBytesToShorten;
 
 			// assuming header length of the layer that requested to be extended hasn't been enlarged yet
-			size_t headerLen = curLayer->getHeaderLen() - (curLayer == layer ? numOfBytesToShorten : 0);
+			// size_t headerLen = curLayer->getHeaderLen() - (curLayer == layer ? numOfBytesToShorten : 0);
```
seladb: Can we remove this commented-out line?

Packet++/src/BgpLayer.cpp
```diff
@@ -117,6 +118,50 @@ namespace pcpp
 		}
 	}
 
+	bool BgpLayer::isValidExtendRange(int offsetInLayer, size_t numOfBytesToExtend) const
```
seladb: The `numOfBytesToExtend` param isn't used, do we need it?

Packet++/src/BgpLayer.cpp
```diff
@@ -117,6 +118,50 @@ namespace pcpp
...
+	{
+		int rawPacketLen = m_Packet->getRawPacket()->getRawDataLen();
+		const uint8_t* rawPacketPtr = m_Packet->getRawPacket()->getRawData();
```
seladb: `m_Packet` doesn't necessarily exist. It could be `nullptr` if the layer is initialized without a packet

Packet++/src/BgpLayer.cpp
```diff
@@ -117,6 +118,50 @@ namespace pcpp
...
+			return false;
+		}
+
+		return true;
+	}
+
+	bool BgpLayer::isValidShortenRange(int offsetInLayer, size_t numOfBytesToShorten) const
+	{
+		int rawPacketLen = m_Packet->getRawPacket()->getRawDataLen();
+		const uint8_t* rawPacketPtr = m_Packet->getRawPacket()->getRawData();
```
seladb: ditto

Packet++/header/BgpLayer.h
```diff
@@ -113,6 +113,10 @@ namespace pcpp
 		}
 
 		void setBgpFields(size_t messageLen = 0);
+
+		bool isValidExtendRange(int offsetInLayer, size_t numOfBytesToExtend) const;
+
+		bool isValidShortenRange(int offsetInLayer, size_t numOfBytesToShorten) const;
```
seladb: Maybe we can override `extendLayer` and `shortenLayer`, run the additional checks and then call `Layer::extendLayer()` / `Layer::shortenLayer()`?

seladb:
Thank you @danasana for working on this fix, much appreciated! 🙏

PR #1957, opened 2025-09-10
Added const qualified match methods to BfpFilterWrapper.

Dimi1010:
This PR updates BpfFilterWrapper to allow matching with a packet while the filter is marked as `const`.

### DIFF
--- a/Pcap++/header/PcapFilter.h
+++ b/Pcap++/header/PcapFilter.h
@@ -78,31 +78,35 @@ namespace pcpp
 	/// A wrapper class for BPF filtering. Enables setting a BPF filter and matching it against a packet
 	class BpfFilterWrapper
 	{
-	private:
-		std::string m_FilterStr;
-		LinkLayerType m_LinkType;
-		std::unique_ptr<bpf_program, internal::BpfProgramDeleter> m_Program;
-
-		void freeProgram();
-
 	public:
-		/// A c'tor for this class
-		BpfFilterWrapper();
+		/// @brief An enum with possible behaviours in case of link type missmatch
+		enum class LinkMissmatchBehaviour
+		{
+			/// @brief Attempt to recompile the filter with the new link type
+			RecompileFilter,
+			/// @brief Do not attempt to recompile the filter, just return false.
+			NoMatch
+		};
 
-		/// A copy constructor for this class.
-		/// @param[in] other The instance to copy from
+		/// @brief Creates a new instance with no filter.
+		BpfFilterWrapper() = default;
+
+		/// @brief Creates a new instance with the given filter string.
+		/// @param filter A filter in BPF syntax
+		/// @param linkType An optional parameter to set the filter's link type. The default is LINKTYPE_ETHERNET
+		/// @throws std::runtime_error if the filter could not be compiled
+		BpfFilterWrapper(std::string filter, LinkLayerType linkType = LINKTYPE_ETHERNET);
+
 		BpfFilterWrapper(const BpfFilterWrapper& other);
-
-		/// A copy assignment operator for this class.
-		/// @param[in] other An instance of IPNetwork to assign
-		/// @return A reference to the assignee
+		BpfFilterWrapper(BpfFilterWrapper&&) noexcept = default;
 		BpfFilterWrapper& operator=(const BpfFilterWrapper& other);
+		BpfFilterWrapper& operator=(BpfFilterWrapper&&) noexcept = default;
 
 		/// Set a filter. This method receives a filter in BPF syntax (https://biot.com/capstats/bpf.html) and an
 		/// optional link type, compiles them, and if compilation is successful it stores the filter.
 		/// @param[in] filter A filter in BPF syntax
 		/// @param[in] linkType An optional parameter to set the filter's link type. The default is LINKTYPE_ETHERNET
-		/// @return True if compilation is successful and filter is stored in side this object, false otherwise
+		/// @return True if compilation is successful and filter is stored inside this object, false otherwise
 		bool setFilter(const std::string& filter, LinkLayerType linkType = LINKTYPE_ETHERNET);
 
 		/// Match a packet with the filter stored in this object. If the filter is empty the method returns "true".
@@ -124,6 +128,44 @@ namespace pcpp
 		/// could not be compiled
 		bool matchPacketWithFilter(const uint8_t* packetData, uint32_t packetDataLength, timespec packetTimestamp,
 		                           uint16_t linkType);
+
+		/// @brief Match a packet with the filter stored in this object.
+		///
+		/// If the filter is empty the method returns "true".
+		/// If the link type of the raw packet is different than the one set in setFilter():
+		/// - If onLinkMissmatch is set to RecompileFilter, the filter will be re-compiled and stored in the object.
+		/// - If onLinkMissmatch is set to NoMatch, the method will return "false"
+		///
+		/// @param[in] rawPacket The raw packet to match the filter against
+		/// @param[in] onLinkMissmatch The behaviour in case of link type missmatch
+		/// @return True if the filter matches (or if it's empty). False otherwise
+		bool matches(const RawPacket& rawPacket,
+		             LinkMissmatchBehaviour onLinkMissmatch = LinkMissmatchBehaviour::RecompileFilter) const;
+
+		/// @brief Match a raw buffer of packet data against the filter stored in this object.
+		///
+		/// If the filter is empty the method returns "true".
+		/// If the link type provided is different than the one set in setFilter():
+		/// - If onLinkMissmatch is set to RecompileFilter, the filter will be re-compiled and stored in the object.
+		/// - If onLinkMissmatch is set to NoMatch, the method will return "false"
+		///
+		/// @param[in] packetData A pointer to the raw packet data
+		/// @param[in] packetDataLength The length of the raw packet data in bytes
+		/// @param[in] timestamp Timestamp to be associated with the packet
+		/// @param[in] linkType The link type of the packet
+		/// @param[in] onLinkMissmatch The behaviour in case of link type missmatch
+		/// @return True if the filter matches (or if it's empty). False otherwise
+		bool matches(const uint8_t* packetData, uint32_t packetDataLength, timespec timestamp, uint16_t linkType,
+		             LinkMissmatchBehaviour onLinkMissmatch = LinkMissmatchBehaviour::RecompileFilter) const;
+
+	private:
+		using BpfProgramUPtr = std::unique_ptr<bpf_program, internal::BpfProgramDeleter>;
+
+		std::string m_FilterStr;
+		mutable LinkLayerType m_CachedProgramLinkType = LinkLayerType::LINKTYPE_ETHERNET;
+		mutable BpfProgramUPtr m_CachedProgram;
+
+		static BpfProgramUPtr compileFilter(std::string const& filter, LinkLayerType linkType);
 	};
 
 	/// @class GeneralFilter
--- a/Pcap++/src/PcapFilter.cpp
+++ b/Pcap++/src/PcapFilter.cpp
@@ -38,17 +38,23 @@ namespace pcpp
 		}
 	}  // namespace internal
 
-	BpfFilterWrapper::BpfFilterWrapper() : m_LinkType(LinkLayerType::LINKTYPE_ETHERNET)
-	{}
-
-	BpfFilterWrapper::BpfFilterWrapper(const BpfFilterWrapper& other) : BpfFilterWrapper()
+	BpfFilterWrapper::BpfFilterWrapper(std::string filter, LinkLayerType linkType)
+	    : m_FilterStr(std::move(filter)), m_CachedProgramLinkType(linkType),
+	      m_CachedProgram(compileFilter(m_FilterStr, linkType))
 	{
-		setFilter(other.m_FilterStr, other.m_LinkType);
+		if (!m_FilterStr.empty() && m_CachedProgram == nullptr)
+		{
+			throw std::runtime_error("Couldn't compile BPF filter: '" + m_FilterStr + "'");
+		}
 	}
+
+	BpfFilterWrapper::BpfFilterWrapper(const BpfFilterWrapper& other)
+	    : BpfFilterWrapper(other.m_FilterStr, other.m_CachedProgramLinkType)
+	{}
 
 	BpfFilterWrapper& BpfFilterWrapper::operator=(const BpfFilterWrapper& other)
 	{
-		setFilter(other.m_FilterStr, other.m_LinkType);
+		setFilter(other.m_FilterStr, other.m_CachedProgramLinkType);
 		return *this;
 	}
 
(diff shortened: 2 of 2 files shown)


Pcap++/header/PcapFilter.h
```diff
@@ -124,6 +128,44 @@ namespace pcpp
...
+		/// - If onLinkmismatch is set to NoMatch, the method will return "false"
+		///
+		/// @param[in] packetData A pointer to the raw packet data
+		/// @param[in] packetDataLength The length of the raw packet data in bytes
+		/// @param[in] timestamp Timestamp to be associated with the packet
+		/// @param[in] linkType The link type of the packet
+		/// @param[in] onLinkmismatch The behaviour in case of link type mismatch
+		/// @return True if the filter matches (or if it's empty). False otherwise
+		bool matches(const uint8_t* packetData, uint32_t packetDataLength, timespec timestamp, uint16_t linkType,
+		             LinkMismatchBehaviour onLinkmismatch = LinkMismatchBehaviour::RecompileFilter) const;
```
Dimi1010: @seladb I would like a second opinion on if the `onLinkMismatch` is a good addition.

I can see it going two ways:
1. The filter is considered `filter string` + `compiled BPF program`. Then it would make sense to control if a recompile is desired. In that case, having a flag to control it is a potential improvement.

2. The filter is considered only `filter string` and the `compiled program` is considered part of the match procedure. If that is the case, then the program being cached in the object is an optimization. Recompiling it as needed is fine, as the program is expected to be compiled every time anyway. In this case, the user does not need an ability to affect the recompilation, as that is internal implementation.

Pcap++/header/PcapFilter.h
```diff
@@ -132,25 +158,47 @@ namespace pcpp
...
 		/// @param[in] rawPacket A pointer to the raw packet to match the BPF filter with
 		/// @return True if a raw packet matches the BPF filter or false otherwise
 		bool matchPacketWithFilter(RawPacket* rawPacket);
 
-		GeneralFilter()
-		{}
+		/// @brief Match a raw packet against the filter.
+		/// @param rawPacket The raw packet to match against.
+		/// @return True if the filter matches (or if it's empty). False otherwise.
+		bool matches(RawPacket const& rawPacket) const;
```
seladb, line 178: This is an overload of `matchPacketWithFilter`, no? Why not keep the same name?
Also - maybe add `const` qualifier to the existing methods?

Pcap++/src/PcapFilter.cpp
```diff
@@ -38,95 +27,176 @@ namespace pcpp
...
+	bool GeneralFilter::cacheFilterInternal() const
+	{
+		std::string filterStr;
+		parseToString(filterStr);
+		if (!m_BpfWrapper.setFilter(filterStr))
+			return false;
+
+		m_CachedFilter = true;
+		return true;
 	}
```
seladb: Why do we need this method? `BpfFilterWrapper::setFilter()` already prevents resetting the filter if it hasn't changed

Pcap++/header/PcapFilter.h
```diff
@@ -132,25 +158,47 @@ namespace pcpp
...
 
-		/// Virtual destructor, frees the bpf program
-		virtual ~GeneralFilter() = default;
+	protected:
+		/// @brief Parse the filter and cache the compiled BPF program.
+		/// @return True if the filter was successfully parsed and cached, false otherwise.
+		bool cacheFilter()
+		{
+			return cacheFilterInternal();
+		}
```
seladb: If we decide to keep this method, we can probably unify both of them and keep just one

Pcap++/header/PcapFilter.h
```diff
@@ -124,6 +119,37 @@ namespace pcpp
...
+		///
+		/// If the filter is empty the method returns "true".
+		/// If the link type provided is different than the one set in setFilter() the filter will be recompiled.
+		///
+		/// @param[in] packetData A pointer to the raw packet data
+		/// @param[in] packetDataLength The length of the raw packet data in bytes
+		/// @param[in] timestamp Timestamp to be associated with the packet
+		/// @param[in] linkType The link type of the packet
+		/// @return True if the filter matches (or if it's empty). False otherwise
+		bool matches(const uint8_t* packetData, uint32_t packetDataLength, timespec timestamp, uint16_t linkType) const;
```
seladb, line 146: ditto: if these are overloads of `matchPacketWithFilter` let's keep the same name?
We can also add `const` qualifier to the existing methods

Pcap++/header/PcapFilter.h
```diff
@@ -78,31 +78,26 @@ namespace pcpp
 	/// A wrapper class for BPF filtering. Enables setting a BPF filter and matching it against a packet
 	class BpfFilterWrapper
 	{
-	private:
-		std::string m_FilterStr;
-		LinkLayerType m_LinkType;
-		std::unique_ptr<bpf_program, internal::BpfProgramDeleter> m_Program;
-
-		void freeProgram();
-
```
seladb, line 87: Moving things around makes the changes harder to follow (here, but also in other classes). Maybe we can separate this PR into multiple PRs with fewer changes that are easier to review?

seladb (commented):
I left a few com

Pcap++/header/PcapFilter.h
Dimi1010: I suppose it can be done, but I made a new method for two reasons.

For one, `matchPacketWithFilter()` takes a pointer param, `matches()` takes a const reference parameter.

The second one is that IMO `matchPacketWithFilter` is a bit overly verbose for a method to a filter object? The object itself is already a filter, so there isn't much else to match with? The flow would then be `filter -> matches -> rawPacket`, which IMO would read cleaner. I figured that since the parameter signature is changing, might as well do both changes at once.

Pcap++/src/PcapFilter.cpp
Dimi1010: The new code prevents the need to recompute the filter string, which was previously done on every call, to check if the filter needs to be reset. Now it is just a bool check.

Pcap++/header/PcapFilter.h
seladb: Having methods with different names that do more or less the same thing is confusing. We can deprecate the old methods and generate `matches` methods instead, but I think it'd be easier to just stick with the same name

Pcap++/src/PcapFilter.cpp
seladb: I don't think recomputing the filter string is a big deal since it is not done in real time

Pcap++/header/PcapFilter.h
Dimi1010: They are separated into `cacheFilter` and `cacheFilterInternal` because the internal one is marked as `const` to be able to recompile the filter in `matches` / `matchPacketWithFilter`. 

My idea when I split them was to minimize the area of exposure of the cache modification when the object is const qualified. The way it is done currently, only the match method modifies the cache while the object is `const` qualified, because only that method needs to be able to do it.

Pcap++/src/PcapFilter.cpp
Dimi1010: Wdym not done in real time?

seladb: Setting a filter (usually) isn't time-critical, meaning that spending a few more cycles on parsing the string shouldn't be an issue. As opposed to parsing a packet which is done in real-time so should be as fast as possible

Pcap++/header/PcapFilter.h
seladb: Why can't `cacheFilterInternal` be marked as `const`?

Pcap++/header/PcapFilter.h
Dimi1010: I'm fine with deprecating the old ones. I think `matches` makes for a more user friendly API since it is more concise.

While we are on the topic. I think the `IPcapDevice::matchPacketWithFilter` should probably be deprecated too. It is a static method that is somewhat redundant. It requires the user to pass a filter object and a packet. If the user already has that, he can just call `filter.match***(packet)` instead, which IMO would be more readable than `IPcapDevice::matchPacketWithFilter(filter, packet)`.

Pcap++/header/PcapFilter.h
Dimi1010: `cacheFilterInternal` is `const`, `cacheFilter` isn't.

The difference is that `cacheFilter` is `protected` while `cacheFilterInternal` is `private`. Having `cacheFilter` `const` and `protected` means that every derived class can call it in their own `const` methods. I suppose that is fine, but I prefer to reduce the number of places a mutable const operation can take place.

Pcap++/src/PcapFilter.cpp
Dimi1010: The issue was that the previous code recomputed the filter BPF string on every call of `matchPacketWithFilter`.

The `GeneralFilter::matchPacketWithFilter` control flow was thus:
```
1. Update filter if needed
1.1. parseToString(bfpString) # Allocates and concatenates strings. (may be heavy)
1.2. setFilter(bfpString)
1.2.1 compare with compiled filter
1.2.1.1 recompile if different
2. Perform match (actual operation)
```

When filtering with the filter object itself and not a backend filter engine (libpcap network devices, etc).
Step 1.1 was always done, on every **match** call, not just the `setFilter`. The change skips the entire step 1 if the bool flag is set, so the filter string does not have to be recomputed to compare it with the compiled filter.

Pcap++/header/PcapFilter.h
Dimi1010: Reverted the reordered methods. Should be easier to compare now.

Pcap++/header/PcapFilter.h
seladb: > I'm fine with deprecating the old ones. I think `matches` makes for a more user friendly API since it is more concise.

I guess we can deprecate them? 🤔 



> While we are on the topic. I think the `IPcapDevice::matchPacketWithFilter` should probably be deprecated too. It is a static method that is somewhat redundant. It requires the user to pass a filter object and a packet. If the user already has that, he can just call `filter.match***(packet)` instead, which IMO would be more readable than `IPcapDevice::matchPacketWithFilter(filter, packet)`.

I agree, I think we can deprecate it. To be honest, I don't remember why it was added, there was probably a reason but I don't recall what it was

Pcap++/src/PcapFilter.cpp
seladb: Got it. I dug deeper into it, and the complexity is about invalidating the cache when the filter changes - we need to call `invalidateCache()` in every setter, and we might miss one (for example, you missed calling it in `setDirection()` 🙂 ).
I'm not sure how to solve it though...

Pcap++/header/PcapFilter.h
seladb: `cacheFilter` is only used once in `verifyFilter`. I think we can keep just `cacheFilterInternal` (or rename it to `cacheFilter`)

Dimi1010: Merged both functions fc45b30f9574f582e6365eac922dd7a81e37f4a6 & 4e801218a2116db1fa21d96d86aace4b1ac5cda8 (fix)
Also marked `invalidateCache` as `const` too, to keep the cache API consistent.

Pcap++/src/PcapFilter.cpp
Dimi1010: Fixed the invalidate miss. Thanks! b06fcef1ba14c29584c61ab063b2c13df2f293df

I am not sure there is a better way to do it. With this optimization the derived classes need to signal the base implementation to regenerate the cache somehow, since they add new public mutators that the base doesn't know about.

Pcap++/header/PcapFilter.h
Dimi1010: @seladb Do you want to deprecate them here or in a new PR so it is cleaner? IMO, a clean PR afterwards would be better for finding the deprecations later.

Pcap++/header/PcapFilter.h
Dimi1010: Added the `const` qualifiers to the old methods. f58ce68adf25fe1218d02d8dacab1c7dea73d89a

Pcap++/header/PcapFilter.h
seladb: It's ok to open a new PR for the deprecations 👍

Pcap++/src/PcapFilter.cpp
seladb: I asked ChatGPT, and it suggested using a `Property<T>` Template Wrapper:

```cpp
#include <iostream>
#include <functional>

template<typename T>
class Property {
    T value;
    std::function<void()> beforeSetter;
public:
    Property(std::function<void()> beforeSetterFunc = nullptr)
        : beforeSetter(beforeSetterFunc) {}

    T get() const {
        return value;
    }

    void set(const T& val) {
        if (beforeSetter) beforeSetter();
        value = val;
    }

    // Optional: Implicit conversion to T
    operator T() const {
        return value;
    }
};

class Base {
protected:
    void beforeSetter() {
        std::cout << "[Base] beforeSetter called\n";
    }

public:
    Property<int> x;

    Base() : x([this]() { beforeSetter(); }) {}

    // Optional convenience
    void setX(int val) { x.set(val); }
    int getX() const { return x.get(); }
};

class DerivedA : public Base {
protected:
    void beforeSetter() {
        std::cout << "[DerivedA] beforeSetter called\n";
    }

public:
    Property<std::string> name;

    DerivedA() : name([this]() { beforeSetter(); }) {}
    
    void setName(const std::string& n) { name.set(n); }
    std::string getName() const { return name.get(); }
};
```

Similarly, it suggested a `PropertyVector<T>` for vectors:
```cpp
#include <vector>
#include <algorithm>

template<typename T>
class PropertyVector {
    std::vector<T> value;
    std::function<void()> beforeSetter;
public:
    PropertyVector(std::function<void()> beforeSetterFunc = nullptr)
        : beforeSetter(beforeSetterFunc) {}

    const std::vector<T>& get() const {
        return value;
    }

    void set(const std::vector<T>& vec) {
        if (beforeSetter) beforeSetter();
        value = vec;
    }

    void add(const T& element) {
        if (beforeSetter) beforeSetter();
        value.push_back(element);
    }

    bool remove(const T& element) {
        if (beforeSetter) beforeSetter();
        auto it = std::find(value.begin(), value.end(), element);
        if (it != value.end()) {
            value.erase(it);
            return true;
        }
        return false;
    }

    std::size_t size() const {
        return value.size();
    }

    T& operator[](std::size_t idx) {
        return value[idx];
    }

    const T& operator[](std::size_t idx) const {
        return value[idx];
    }
};
```

It's a nice pattern but I'm not sure if it's better than just calling `invalidateCache()`, what do you think?

Dimi1010: Looks nice, but I think its not worth it in the current case. `invalidateCache()` will most likely be inlined as it is a single bool operation. Going through `std::function` removes that possibility as that is a call through function pointer, which the compiler doesn't know anything about at compile time.

Dimi1010: There is also the fact that this will also need to be setup in the derived classes, which makes it no different than simply calling `invalidateCache()`.

Pcap++/src/PcapFilter.cpp
```diff
@@ -38,95 +27,176 @@ namespace pcpp
...
+		{
+			auto newProgram = compileFilter(m_FilterStr, static_cast<LinkLayerType>(linkType));
+			if (newProgram == nullptr)
+			{
+				PCPP_LOG_ERROR("Couldn't compile BPF filter: '" << m_FilterStr << "' for link type: " << linkType);
+				return false;
+			}
+			m_CachedProgram = std::move(newProgram);
+			m_CachedProgramLinkType = static_cast<LinkLayerType>(linkType);
 		}
```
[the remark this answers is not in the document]
Dimi1010: The underlying BPF program is single thread anyway (I think?). The entire class is thread unsafe.

Pcap++/header/PcapFilter.h
```diff
@@ -79,30 +79,34 @@ namespace pcpp
...
+		BpfFilterWrapper() = default;
 
-		/// A copy constructor for this class.
-		/// @param[in] other The instance to copy from
-		BpfFilterWrapper(const BpfFilterWrapper& other);
+		/// @brief Creates a new instance with the given filter string.
+		/// @param filter A filter in BPF syntax
+		/// @param linkType An optional parameter to set the filter's link type. The default is LINKTYPE_ETHERNET
+		/// @throws std::runtime_error if the filter could not be compiled
+		explicit BpfFilterWrapper(std::string filter, LinkLayerType linkType = LINKTYPE_ETHERNET);
```
seladb: We never use this constructor, why do we need it?
It may throw an exception, which we almost don't do in this file, so it's better to avoid the c'tor altogether

Pcap++/src/PcapFilter.cpp
```diff
@@ -38,95 +27,176 @@ namespace pcpp
...
 		if (m_FilterStr.empty())
 			return true;
 
-		if (!setFilter(std::string(m_FilterStr), static_cast<LinkLayerType>(linkType)))
+		// This should never happen, but just in case
+		if (m_CachedProgram == nullptr)
 		{
-			return false;
+			throw std::runtime_error("No compiled BPF program available");
+		}
```
seladb: This could potentially happen if `BpfFilterWrapper` is created using the default c'tor and then the user calls `matches()`. I wouldn't throw an exception here to make it compatible with the rest of the file, instead, I'd write an error log and return `false`

Pcap++/src/PcapFilter.cpp
```diff
@@ -38,95 +27,176 @@ namespace pcpp
...
+
+	bool GeneralFilter::matches(RawPacket const& rawPacket) const
+	{
+		if (!m_CachedFilter)
+		{
+			if (!cacheFilter())
+			{
+				return false;
+			}
+		}
```
seladb: Maybe we can do:
```cpp
if (!m_CachedFilter && !cacheFilter())
{
    return false;
}
```

Pcap++/src/PcapFilter.cpp
```diff
@@ -38,95 +27,176 @@ namespace pcpp
...
+
+		return m_BpfWrapper.matches(rawPacket);
+	}
+
+	bool GeneralFilter::cacheFilter() const
+	{
+		std::string filterStr;
+		parseToString(filterStr);
+		if (!m_BpfWrapper.setFilter(filterStr))
+			return false;
```
seladb, line 172: nit: in newer code I prefer using curly brackets `{}` even if an `if` clause has just one line

Pcap++/src/PcapFilter.cpp
Dimi1010: Huh, I didn't notice that. Good catch.

If it is a valid case, then it can be merged into the recompile condition. If there is no compiled filter, it will attempt to recompile and proceed as normal from there.

Pcap++/header/PcapFilter.h
Dimi1010: I try to add RAII constructors when I can, so full initialization can be done in a single step.

Conceptually, the default constructor creates an "always true" noop filter. It makes sense to have a separate constructor that allows a specific filter to be set from the time of initialization instead of having to do a second call to `setFilter°.

seladb: I am all for adding constructors and overloads if we need them, but if we don't, and especially when `BpfFilterWrapper` is mainly used internally, I don't see the point in adding code that we don't need

Dimi1010: Hmm, sure, I will remove it then.

Dimi1010: Correction, it is actually used by the copy constructor, as that delegates to the new constructor.

Pcap++/src/PcapFilter.cpp
Dimi1010: Fixed fd6165a3d4d6419f20c056ae775eb9cf493cf5eb

Pcap++/src/PcapFilter.cpp
Dimi1010: Fixed 3c17eb77e10abf0b59fcfb42480cdce82340aeb6

Pcap++/src/PcapFilter.cpp
Dimi1010: Fixed 7efcc7be9f7d0368c2163b156f38ee6ce8d18c8d

Pcap++/header/PcapFilter.h
seladb: Can't the copy constructor call `setFilter()` as it did before, and then we can remove this new constructor?

Dimi1010: I suppose it can, but if the `setFilter()` fails inside the copy constructor, you have no indication that it did, like when you are calling it externally, unless you throw an exception again...

It shouldn't happen generally, but it can.

seladb: The `operator=` still calls `setFilter()`. I guess we can check the return value and throw an exception in this case?

Dimi1010: Updated 7705ff2c6943f56005bb822f726cb6a37c4e681d

PR #1963, opened 2025-09-12
More C++ Migration

tigercosmos:
by AI

### DIFF
--- a/Pcap++/header/XdpDevice.h
+++ b/Pcap++/header/XdpDevice.h
@@ -219,7 +219,7 @@ namespace pcpp
 		/// this timeout. The default value is 5000 ms
 		/// @return True if all packets were sent, or if waitForTxCompletion is true - all sent packets were confirmed.
 		/// Returns false if an error occurred or if poll timed out.
-		bool sendPackets(RawPacket* packets, size_t packetCount, bool waitForTxCompletion = false,
+		bool sendPackets(RawPacket packets[], size_t packetCount, bool waitForTxCompletion = false,
 		                 int waitForTxCompletionTimeoutMS = 5000);
 
 		/// @return A pointer to the current device configuration. If the device is not open this method returns nullptr
--- a/Pcap++/src/XdpDevice.cpp
+++ b/Pcap++/src/XdpDevice.cpp
@@ -317,7 +317,7 @@ namespace pcpp
 		                   [&]() { return packets.size(); }, waitForTxCompletion, waitForTxCompletionTimeoutMS);
 	}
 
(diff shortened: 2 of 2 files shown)


Pcap++/header/XdpDevice.h
```diff
@@ -219,7 +219,7 @@ namespace pcpp
 		/// this timeout. The default value is 5000 ms
 		/// @return True if all packets were sent, or if waitForTxCompletion is true - all sent packets were confirmed.
 		/// Returns false if an error occurred or if poll timed out.
-		bool sendPackets(RawPacket packets[], size_t packetCount, bool waitForTxCompletion = false,
+		bool sendPackets(RawPacket* packets, size_t packetCount, bool waitForTxCompletion = false,
```
seladb: keeping it as an array is clearer I think

Pcap++/src/XdpDevice.cpp
```diff
@@ -317,7 +317,7 @@ namespace pcpp
 		                   [&]() { return packets.size(); }, waitForTxCompletion, waitForTxCompletionTimeoutMS);
 	}
 
-	bool XdpDevice::sendPackets(RawPacket packets[], size_t packetCount, bool waitForTxCompletion,
+	bool XdpDevice::sendPackets(RawPacket* packets, size_t packetCount, bool waitForTxCompletion,
```
seladb: ditto

Pcap++/header/XdpDevice.h
Dimi1010: imo, the pointer pattern is more common than the `[]` in parameters, but I am fine with either.

Dimi1010 (approved):

Pcap++/header/XdpDevice.h
tigercosmos: The GPT answer:
```
Which is more idiomatic in C++?

int* arr
🔹 More common in C and C++ when you want to emphasize “this is just a pointer.”
🔹 Matches with pointer arithmetic and new[]/delete[].

int arr[]
🔹 Sometimes used in function parameters to emphasize "I intend to use this like an array."
🔹 But it’s misleading, because it doesn’t actually carry array size info.
```

seladb: I don't think it's misleading - an array in C++ is actually a pointer so both options are valid, but using `[]` is a way to say "I intend to use this like an array."

Dimi1010: Ideally it should be a `std::span<T>` or a similar class.

tigercosmos: std::array or std::vector should also be better here.

tigercosmos: anyway, @seladb  I've reverted it.

PR #1964, opened 2025-09-12
Fix inclusive `parseUntil` condition in the main packet parse loop.

Dimi1010:
The `parseUntil` is now fully inclusive and parses until it exhausts a sequence of target protocol types. This allows specifying target protocols that allow chaining of the same protocol types.

Simplified the rollback mechanism.

### DIFF
--- a/Packet++/src/Packet.cpp
+++ b/Packet++/src/Packet.cpp
@@ -63,39 +63,95 @@ namespace pcpp
 
 		m_FirstLayer = createFirstLayer(linkType);
 
-		m_LastLayer = m_FirstLayer;
-		Layer* curLayer = m_FirstLayer;
+		// As the stop conditions are inclusive, the parse must go one layer further and then roll back if needed
+		bool rollbackLastLayer = false;
+		bool foundTargetProtocol = false;
+		for (Layer* curLayer = m_FirstLayer; curLayer != nullptr; curLayer = curLayer->getNextLayer())
+		{
+			// Mark the current layer as allocated in the packet
+			curLayer->m_IsAllocatedInPacket = true;
+			m_LastLayer = curLayer;  // Update last layer to current layer
+
+			// If the current layer is of a higher OSI layer than the target, stop parsing
+			if (curLayer->getOsiModelLayer() > parseUntilLayer)
+			{
+				rollbackLastLayer = true;
+				break;
+			}
+
+			// If we are searching for a specific layer protocol, record when we find at least one target.
+			const bool matchesTarget = curLayer->isMemberOfProtocolFamily(parseUntil);
+			if (parseUntil != UnknownProtocol && matchesTarget)
+			{
+				foundTargetProtocol = true;
+			}
+
+			// If we have found the target protocol already, we are parsing until we find a different protocol
+			if (foundTargetProtocol && !matchesTarget)
+			{
+				rollbackLastLayer = true;
+				break;
+			}
+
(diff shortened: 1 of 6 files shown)


Packet++/src/Packet.cpp
```diff
@@ -63,39 +63,95 @@ namespace pcpp
...
+		// Roll back one layer, if parsing with search condition as the conditions are inclusive.
+		// Don't delete the first layer. If already past the target layer, treat the same as if the layer was found.
+		if (rollbackLastLayer && m_LastLayer != m_FirstLayer)
+		{
+			m_LastLayer = m_LastLayer->getPrevLayer();
+			delete m_LastLayer->m_NextLayer;
+			m_LastLayer->m_NextLayer = nullptr;
+		}
+
+		/*
```
Dimi1010: The old code is kept mostly so it does not interfere with the diff.
It would probably be better to be removed prior to merge.

Packet++/src/Packet.cpp
```diff
@@ -63,39 +63,95 @@ namespace pcpp
 
 		m_FirstLayer = createFirstLayer(linkType);
 
-		m_LastLayer = m_FirstLayer;
-		Layer* curLayer = m_FirstLayer;
+		// As the stop conditions are inclusive, the parse must go one layer further and then roll back if needed
+		bool rollbackLastLayer = false;
+		bool foundTargetProtocol = false;
+		for (Layer* curLayer = m_FirstLayer; curLayer != nullptr; curLayer = curLayer->getNextLayer())
```
seladb: nit: we can use `auto*` instead of `Layer*`

Packet++/src/Packet.cpp
seladb, line 1: Can we add tests to prove correct parsing when there is chaining of the same protocol type (both with `parseUntil` and `parseUntilLayer`)?

Dimi1010: Sure, BGP will probably be easiest to set up.

Dimi1010: Added 9bec9c7c73fdabe68e511cb1cc6bf15bc2a2d869

Packet++/src/Packet.cpp
Dimi1010: Fixed ffeed09e616dafe6e0975e4399a53ff2cc1ad92b

Tests/PcppTestUtilities/Resources.h
Dimi1010, line 1: The new `saveResource` method is mostly for easy test data creation. By default all resource providers should be constructed frozen, so they don't modify the resources.

I made and used it to dump the BGP multilayer packet from the creation test before I figured out it was already available.

seladb: We don't use this new method anywhere, so why add it? 🤔

Dimi1010: Same reason we have  `savePacketToPcap(...);` in TestUtils.h in the packet tests.

Utility for creating test resources. They aren't actually used inside the tests themselves but are for easier creation of resources when making the test cases.

seladb (approved):
Let's remove the commented-out code before merging: https://github.com/seladb/PcapPlusPlus/pull/1964#discussion_r2345328513

Otherwise, LGTM!

Packet++/src/Packet.cpp
Dimi1010: Removed code. e4515e0cb517fe8f2a1bfc6eef9927bc8578983e

PR #1838, opened 2025-05-31
Cleanup of `PcapLiveDevice` capture machinery.

Dimi1010:
This PR moves the implementation of the capture procedures to internally linked functions in `PcapLiveDevice.cpp` and aims to reduce the the need for extensive bookkeeping of capture thread only variables inside the device object's fields.

The static member function callback routines for `pcap_dispatch` have been replaced with internal linkage free functions to improve readability and encapsulation. The user token is now expected to be a pointer to a specific capture context object instead of the device instance.

Replacement list:
- `onPacketArrivesCallback` replaces `onPacketArrives`. Expects user token of type `CaptureContext`
- `onPacketArrivesAccumulator` replaces `onPacketArrivesNoCallback`. Expects user token of type `AccumulatorCaptureContext`
- `onPacketArrivesCallbackWithCancellation` replaces `onPacketArrivesBlockingMode`. Expects user token of type `CaptureContextST`
- Added new callback handler `onPacketArrivesNoop` to be used when no handling is required.

Replaced `captureThreadMain` static private function with internal linkage free functions. The free functions take their inputs as parameters instead of relying on data from the implicit `this` pointer, simplifying data ownership and reducing risk of data races.
- `captureThreadMain` - used for callback or noop capture.
- `captureThreadMainAccumulator` - used for accumulating the packets into a `RawPacketVector`.

Changed blocking mode capture to utilize `onPacketArrivesCallbackWithCancellation`. The new infrastructure provides a layer of separation between the device threading machinery and the cancellation requests by relying on `requestStop` flag to record a cancellation request that is then forwarded to `m_StopThread` after dispatch finishes instead of directly writing to `m_StopThread`.

Removed `PcapLiveDevice` member fields previously only used to store data for the capture thread as that data is now stored in context objects created on demand during capture startup.

PR #1966, opened 2025-09-15
Move PcapLiveDevice statistics machinery entirely inside the .cpp file.

Dimi1010:
Split of #1838, since that one has been sitting for a while.

This part moves only the statistics worker to be entirely internal to the PcapLiveDevice translation unit.

### DIFF
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -82,44 +82,6 @@ namespace pcpp
 			bool isLoopback;
 		};
 
-		/// @brief A worker thread that periodically calls the provided callback with updated statistics.
-		class StatisticsUpdateWorker
-		{
-		public:
-			/// @brief Constructs and starts a worker thread that periodically calls the provided callback with updated
-			/// statistics.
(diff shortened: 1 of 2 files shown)


seladb (commented):
This PR reverts many of the changes done here: https://github.com/seladb/PcapPlusPlus/pull/1785, I'm curious why we're moving back and forth? 🤔

Dimi1010:
> This PR reverts many of the changes done here: https://github.com/seladb/PcapPlusPlus/pull/1785, I'm curious why we're moving back and forth? 🤔 

The first change was an attempt to separate the variables only needed for the thread procedures away from the rest of the device.

As I was reworking the capture thread variables, I figured that having said variables grouped as contexts and passed to a pure function procedure would be better for internal representation. That way, the level of indirection done by the polymorphic `Worker` class is removed while also keeping the encapsulation of the thread variables used only inside the procedure.

Since the workers are one-offs and won't be reused, it is a simpler solution that would only keep `std::thread` and `std::atomic<bool>` inside the device class.
Additionally, it would allow reuse of the stop flag, as the same flag can be passed to multiple procedures. In the worker scenario, each worker created its own stop flag.

seladb (approved):

PR #1970, opened 2025-09-20
Cleanup of CI run_tests and run_tests_windows.

Dimi1010:
The PR cleans up the CI scripts that run the tests.

- Added a `TcpReplay` class to abstract tcpreplay executable and operations on it.
- Moved code that runs the tests on windows to separate functions for easier readability.


ci/run_tests/tcp_replay_utils.py
```diff
@@ -0,0 +1,87 @@
...
+        completed_process = subprocess.run(
+            [self.executable, "--listnics"],
+            capture_output=True,
+        )
+        if completed_process.returncode != 0:
+            raise RuntimeError('Error executing "tcpreplay --listnics"!')
+
+        raw_nics_output = completed_process.stdout.decode("utf-8")
+        nics = []
+        for row in raw_nics_output.split("\n")[2:]:
```
tigercosmos, line 75: It's not a bug for this PR, but I notice that `[2:]` could be a potential issue.

tigercosmos (approved):

ci/run_tests/tcp_replay_utils.py
Dimi1010: In what way?

tigercosmos: Can we always assume the output has more than two lines?

Dimi1010: Probably no, but here it uses slice indexing. If the list has less than 3 items, it will return an empty list. Thus, the for loop will not run.

ci/run_tests/run_tests_windows.py
```diff
@@ -28,38 +27,99 @@ def get_ip_by_guid(guid):
...
-                    shell=True,
-                )
-                if completed_process.returncode != 0:
-                    continue
-                return interface, ip_address
-            except Exception:
-                pass
-    return None, None
+
+def run_packet_coverage():
```
seladb, line 58: This is a method of just one action, why should it be its own method?

ci/run_tests/run_tests_windows.py
```diff
@@ -28,38 +27,99 @@ def get_ip_by_guid(guid):
...
+            "--",
+            os.path.join("Bin", "Packet++Test"),
+        ],
+        cwd=os.path.join("Tests", "Packet++Test"),
+        shell=True,
+        check=True,  # Raise exception if the worker returns in non-zero status code
+    )
+
+
+def run_pcap_tests(ip_address: str, skip_tests: list[str]):
```
seladb, line 82: ditto

ci/run_tests/tcp_replay_utils.py
seladb, line 1: I think that having this class instead of a simple `@contextmanager` and `subprocess.run()` to get the nics is an overkill. Usually less code is better 🙂

ci/run_tests/run_tests_windows.py
```diff
@@ -28,38 +27,99 @@ def get_ip_by_guid(guid):
...
+            "-x",
+            ";".join(skip_tests),
+        ],
+        cwd=os.path.join("Tests", "Pcap++Test"),
+        shell=True,
+        check=True,  # Raise exception if the worker returns in non-zero status code
+    )
+
+
+def run_pcap_coverage(ip_address: str, skip_tests: list[str]):
```
seladb, line 97: ditto: This is a method of just one action, why should it be its own method?

ci/run_tests/tcp_replay_utils.py
Dimi1010: It's not just the nics. Those were mostly an afterthought. It's so the tcp replay executable passed around as a structured object. IMO, that is better than having a path in the parameters that you need to know is the executable.

The differences between unix and windows calls are handled internally instead of having to do that all over the place or having duplicated procedures.

It also allows the places where it is used not to worry about how exactly the replay executable is found.

ci/run_tests/run_tests_windows.py
Dimi1010: IMO, It's easier to read at the call site. The main function is already quite complex, and it is easier to understand what `run_packet_***()` does at a glance than having the raw subprocess run command with all the arguments and having to understand what it does.

ci/run_tests/tcp_replay_utils.py
seladb: > It's not just the nics. Those were mostly an afterthought. It's so the tcp replay executable passed around as a structured object. IMO, that is better than having a path in the parameters that you need to know is the executable.

The only parameter that is passed around is the tcpreplay path, but I think it's ok because it's just a string.

> The differences between unix and windows calls are handled internally instead of having to do that all over the place or having duplicated procedures.

I actually think this is the main problem with the new class - methods like `get_nic_list()` only run on Windows and otherwise throw an exception, which is not great from OOP perspective. The constructor also has a few `if..else` per platform - again, not great OOP. I don't think this new class is really needed

ci/run_tests/run_tests_windows.py
seladb: I don't think the functions are that complex, moving back and forth between functions is also not very easy to read

ci/run_tests/tcp_replay_utils.py
Dimi1010: > The only parameter that is passed around is the tcpreplay path, but I think it's ok because it's just a string.

Eh, its workable, but IMO a lot less explicit. A string also won't be caught by the type system if it is passed in the wrong place.

> I actually think this is the main problem with the new class - methods like get_nic_list() only run on Windows and otherwise throw an exception, which is not great from OOP perspective. 

There is nothing preventing the method `get_nic_list()` to work on unix, as far as I know. It is just that the method was not used in the unix part so I didn't bother. I can add it if it is an issue.

> The constructor also has a few if..else per platform - again, not great OOP.

Why is per platform, `if else` not great OOP exactly?

The only thing that is "per platform" per se, is that on windows:
- the executable path is formatted correctly with an `.exe` suffix.
- the kill process command uses the `taskkill`.

Those are both basic cross platform patches that don't modify anything else in the functionality of the class, and would need to be done regardless if the code is structured as `tcp_replay_worker` function or as the `TcpReplay` object since the user only submits an optional directory, if the function is to be cross platform.

ci/run_tests/run_tests_windows.py
Dimi1010: I can revert it, but the way I see it the `main` function's focus is orchestration and the `run_***` functions are the actual implementations of how the actual actions are achieved. The orchestration shouldn't need to worry about how the actions are done, and the actions shouldn't need to be concerned with the orchestration.

It also mirrors the way it is written in `run_tests.py`, and having the two scripts somewhat similar would be nice.

PS: The tcp replay worker start / stop should probably be moved under their respective `run_pcap_***` operations too.

ci/run_tests/tcp_replay_utils.py
seladb: > Eh, its workable, but IMO a lot less explicit. A string also won't be caught by the type system if it is passed in the wrong place.

A path is a string... we can convert the string to Python's [`Path` class](https://docs.python.org/3/library/pathlib.html) but I don't think it's necessary here

> There is nothing preventing the method `get_nic_list()` to work on unix, as far as I know. It is just that the method was not used in the unix part so I didn't bother. I can add it if it is an issue.

To be honest, I didn't check if it can run on Linux, I assume it does and in that case the `if sys.platform != "win32"` is not needed

> Why is per platform, `if else` not great OOP exactly?

In "clean OOP" there should be minimum `if..else` clauses and you'd expect to have separate classes for different platforms that share a base class with the common functionality. However, as I mentioned before, I don't think all of this is needed here

Dimi1010: > To be honest, I didn't check if it can run on Linux, I assume it does and in that case the if sys.platform != "win32" is not needed

The call to `tcpreplay --listnics` works but the format is different.

It returns
```
tcpreplay --listnics
Warning in interface.c:get_interface_list() line 80:
May need to run as root to get access to all network interfaces.
Warning in interface.c:get_interface_list() line 80:
May need to run as root to get access to all network interfaces.
Available network interfaces:
eth0
any
bluetooth-monitor
nflog
nfqueue
dbus-system
dbus-session
```
instead of `\\Device\\NPF_***` strings, so the code after the call is win32 only. That is why the platform is restricted atm.

> In "clean OOP" there should be minimum if..else clauses and you'd expect to have separate classes for different platforms that share a base class with the common functionality. However, as I mentioned before, I don't think all of this is needed here

Fair enough. I also agree that full derived classes aren't needed here. 

Having said that, we can't escape the `if sys.platform` branches, if we want to keep the code duplication low and routines cross platform. It is either that or have full duplicated routines for a minor difference. We need to have the enforcement / compatibility checks somewhere. 

I used an object so the code is mostly at one place and to maintain the class invariants. It allows the checks / compatibility layer to be centralized. Otherwise it would require every routine that is cross-platform to do the checks in the [`__init__()`](https://github.com/seladb/PcapPlusPlus/pull/1970/files#diff-1bafab8bb0876248fb4831b51baa8c5fcf914278d6d1988848ab573fa62703d9R28-R34) method and the checks in the [`_kill_process()`](https://github.com/seladb/PcapPlusPlus/pull/1970/files#diff-1bafab8bb0876248fb4831b51baa8c5fcf914278d6d1988848ab573fa62703d9R83-R87) method. Currently that would only be `replay` / `tcp_replay_worker` but future maintainability suffers if more functions need to be added.

seladb: What I'm saying is that the current script isn't too complex to read or maintain, which is why I think this refactoring is not really needed. I'd think we can focus our energy in areas that can be more beneficial for the project and its users 🙂

Dimi1010: Eh... sure I guess...

PR #1971, opened 2025-09-21
Refactor worker thread capture procedures as internal functions.

Dimi1010:
Split of #1838 

Moves the machinery used when packets are collected on a separate worker thread to internal linkage functions.

### DIFF
--- a/Pcap++/src/PcapLiveDevice.cpp
+++ b/Pcap++/src/PcapLiveDevice.cpp
@@ -272,6 +272,142 @@ namespace pcpp
 			}
 			PCPP_LOG_DEBUG("Ended periodic statistics update procedure");
 		}
+
+		struct CaptureContext
+		{
+			PcapLiveDevice* device = nullptr;
+			OnPacketArrivesCallback callback;
+			void* userCookie = nullptr;
+		};
+
+		struct AccumulatorCaptureContext
+		{
+			PcapLiveDevice* device = nullptr;
+			RawPacketVector* capturedPackets = nullptr;
+		};
+
+		// A noop function to be used when no callback is set
+		void onPacketArrivesNoop(uint8_t* user, const pcap_pkthdr* pkthdr, const uint8_t* packet)
+		{}
+
+		// @brief Wraps the raw packet data into a RawPacket instance and calls the user callback
+		// @param user A pointer to a CaptureContext instance
+		// @param pkthdr A pointer to the pcap_pkthdr struct
+		// @param packet A pointer to the raw packet data
+		void onPacketArrivesCallback(uint8_t* user, const pcap_pkthdr* pkthdr, const uint8_t* packet)
+		{
+			auto* context = reinterpret_cast<CaptureContext*>(user);
+			if (context == nullptr || context->device == nullptr || context->callback == nullptr)
+			{
+				PCPP_LOG_ERROR("Unable to extract PcapLiveDevice instance or callback");
+				return;
+			}
+
+			try
+			{
+				RawPacket rawPacket(packet, pkthdr->caplen, pkthdr->ts, false, context->device->getLinkType());
+				context->callback(&rawPacket, context->device, context->userCookie);
+			}
+			catch (const std::exception& ex)
+			{
+				PCPP_LOG_ERROR("Exception occurred while invoking packet arrival callback: " << ex.what());
+			}
+			catch (...)
+			{
+				PCPP_LOG_ERROR("Unknown exception occurred while invoking packet arrival callback");
+			}
+		}
+
+		/// @brief Wraps the raw packet data into a RawPacket instance and adds it to the captured packets vector
+		/// @param user A pointer to an AccumulatorCaptureContext instance
+		/// @param pkthdr A pointer to the pcap_pkthdr struct
+		/// @param packet A pointer to the raw packet data
+		void onPacketArrivesAccumulator(uint8_t* user, const pcap_pkthdr* pkthdr, const uint8_t* packet)
+		{
+			auto* context = reinterpret_cast<AccumulatorCaptureContext*>(user);
+			if (context == nullptr || context->device == nullptr || context->capturedPackets == nullptr)
+			{
+				PCPP_LOG_ERROR("Unable to extract PcapLiveDevice instance or captured packets vector");
+				return;
+			}
+
+			try
+			{
+				uint8_t* packetData = new uint8_t[pkthdr->caplen];
+				std::memcpy(packetData, packet, pkthdr->caplen);
+				auto rawPacket = std::make_unique<RawPacket>(packetData, pkthdr->caplen, pkthdr->ts, true,
+				                                             context->device->getLinkType());
+				context->capturedPackets->pushBack(std::move(rawPacket));
+			}
+			catch (const std::exception& ex)
+			{
+				PCPP_LOG_ERROR("Exception occurred while invoking packet arrival callback: " << ex.what());
+			}
+			catch (...)
+			{
+				PCPP_LOG_ERROR("Unknown exception occurred while invoking packet arrival callback");
+			}
+		}
+
+		/// @brief A procedure that dispatches packets to a user-defined callback function whenever a packet arrives.
+		///
+		/// The procedure runs in a loop until the `stopFlag` is set to true.
+		/// The procedure will set the `hasStarted` flag to true at the start of the procedure.
+		///
+		/// @param stopFlag A reference to an atomic boolean that indicates when to stop capturing packets
+		/// @param hasStarted A reference to an atomic boolean that indicates when the capture thread has started
+		/// @param pcapDescriptor A reference to the pcap handle on which to call pcap_dispatch
+		/// @param context A CaptureContext instance that holds the device, callback and user cookie.
+		void captureThreadMain(std::atomic_bool& stopFlag, std::atomic_bool& hasStarted,
+		                       internal::PcapHandle const& pcapDescriptor, CaptureContext context)
+		{
+			PCPP_LOG_DEBUG("Started capture thread for device '" << context.device->getName() << "'");
+			hasStarted.store(true);
+
+			// If the callback is null, we use a no-op handler to avoid unnecessary overhead
+			// Statistics only capture still requires pcap_dispatch to be called, but we don't need to process
+			// packets.
+			pcap_handler callbackHandler = context.callback ? onPacketArrivesCallback : onPacketArrivesNoop;
+
+			while (!stopFlag.load())
+			{
+				if (pcap_dispatch(pcapDescriptor.get(), -1, callbackHandler, reinterpret_cast<uint8_t*>(&context)) ==
+				    -1)
+				{
+					PCPP_LOG_ERROR("pcap_dispatch returned an error: " << pcapDescriptor.getLastError());
+					stopFlag.store(true);
+				}
+			}
+
+			PCPP_LOG_DEBUG("Ended capture thread for device '" << context.device->getName() << "'");
+		}
+
+		/// @brief A procedure that accumulates captured packets into a vector.
+		///
+		/// The procedure runs in a loop until the `stopFlag` is set to true.
+		/// The procedure will set the `hasStarted` flag to true at the start of the procedure.
+		///
+		/// @param stopFlag A reference to an atomic boolean that indicates when to stop capturing packets
+		/// @param hasStarted A reference to an atomic boolean that indicates when the capture thread has started
+		/// @param pcapDescriptor A reference to the pcap handle on which to call pcap_dispatch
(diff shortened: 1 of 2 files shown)


Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -272,6 +272,142 @@ namespace pcpp
...
+				context->callback(&rawPacket, context->device, context->userCookie);
+			}
+			catch (const std::exception& ex)
+			{
+				PCPP_LOG_ERROR("Exception occurred while invoking packet arrival callback: " << ex.what());
+			}
+			catch (...)
+			{
+				PCPP_LOG_ERROR("Unknown exception occurred while invoking packet arrival callback");
+			}
```
seladb: I wonder if we should return when an exception is thrown?
Exceptions slow down high performance applications like this...

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -272,6 +272,142 @@ namespace pcpp
...
+				context->capturedPackets->pushBack(std::move(rawPacket));
+			}
+			catch (const std::exception& ex)
+			{
+				PCPP_LOG_ERROR("Exception occurred while invoking packet arrival callback: " << ex.what());
+			}
+			catch (...)
+			{
+				PCPP_LOG_ERROR("Unknown exception occurred while invoking packet arrival callback");
+			}
```
seladb: Is there any reason to catch an exception in this case?

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -272,6 +272,142 @@ namespace pcpp
...
+		void captureThreadMain(std::atomic_bool& stopFlag, std::atomic_bool& hasStarted,
+		                       internal::PcapHandle const& pcapDescriptor, CaptureContext context)
+		{
+			PCPP_LOG_DEBUG("Started capture thread for device '" << context.device->getName() << "'");
+			hasStarted.store(true);
+
+			// If the callback is null, we use a no-op handler to avoid unnecessary overhead
+			// Statistics only capture still requires pcap_dispatch to be called, but we don't need to process
+			// packets.
+			pcap_handler callbackHandler = context.callback ? onPacketArrivesCallback : onPacketArrivesNoop;
```
seladb, line 354: I'm not sure I understand why we support null callback and a noop 🤔

Dimi1010: From what I understand, if the user just wants statistics, they don't update without running a receive thread. The old code also did that.

The change from the old code is that instead of having a `if (pThis->m_cbOnPacketArrives != nullptr)` check on each receive handle operation (which happens after the `RawPacket` object has been constructed), we just provide an noop handler since we don't have a callback to forward to anyway.

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: The catch block logs an error and essentially terminates the hander's current context as there isn't any other code after it. Terminating the entire entire receive loop would be wrong as the error is only local to the currently received packet.

As far as I know, in most implementations try-catch blocks have little to no performance impact if no exception happens.

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: Propagating an exception from C++ compiled code to C compiled code is considered unsafe. We have to catch any and all exceptions in the handler before returning.

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: https://github.com/seladb/PcapPlusPlus/pull/1971#discussion_r2447598706 also applies.

seladb: What I mean is that we shouldn't catch the exception in `onPacketArrivesCallback` but rather in `captureThreadMain` after calling `pcap_dispatch`, and terminate the capture thread. The reason is that catching exceptions slows down the packet capture. We can let the users catch and handle exceptions in `context->callback` if they choose to

Pcap++/src/PcapLiveDevice.cpp
seladb: ditto - we can catch exceptions in `captureThreadMainAccumulator` and terminate the thread

Pcap++/src/PcapLiveDevice.cpp
seladb: You're right! In the current code we simply don't call the callback if it's null, maybe `captureThreadMain` can return (and stop the thread) if the callback is null?

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: > What I mean is that we shouldn't catch the exception in `onPacketArrivesCallback` but rather in `captureThreadMain` after calling `pcap_dispatch`, and terminate the capture thread. 

https://github.com/seladb/PcapPlusPlus/pull/1971#discussion_r2447598706

The referenced issue overrides that. We are crossing a C to C++ function boundary as the callback pointer is invoked from C code. 

We **have** to catch the exception before returning from the handler. The C compiled code might or might not know what to do with a thrown exception, but ultimately, we have no guarantee, so it's unsafe.

What we do with the exception after catching is irrelevant.

> The reason is that catching exceptions slows down the packet capture. We can let the users catch and handle exceptions in `context->callback` if they choose to

I would argue that stopping the capture thread on the first encountered exception in the loop slows down the packet capture even more (to 0). 

The nominal fast path case should be that the no exceptions happen / reach outside of the user callback to pcpp lib code.

IMO, if the user callback is throwing exceptions everywhere for some reason, slow capture is still better than no capture.

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: The issue is that `pcap_dispatch` still needs to be called for the pcap statistics to get updated by libpcap. If we don't run the capture thread, we effectively don't capture anything, and statistics aren't updated, so the user does not get meaningful updates.

Pcap++/src/PcapLiveDevice.cpp
seladb: > We **have** to catch the exception before returning from the handler. The C compiled code might or might not know what to do with a thrown exception, but ultimately, we have no guarantee, so it's unsafe.
> 
> What we do with the exception after catching is irrelevant.

This is a fair statement. Maybe catching an exception should turn on a flag that will signal the capture thread to stop capturing.

> I would argue that stopping the capture thread on the first encountered exception in the loop slows down the packet capture even more (to 0).

I agree that catching exceptions on a few packets has very minimal effect; however, the current implementation doesn't "count" how many callbacks threw exceptions, so it'll still work the same way if 50% of the callbacks throw exceptions

> The nominal fast path case should be that the no exceptions happen / reach outside of the user callback to pcpp lib code.
> 
> IMO, if the user callback is throwing exceptions everywhere for some reason, slow capture is still better than no capture.

I agree that the fast path shouldn't throw exceptions, which is why I think it should be the user's decision whether to avoid or catch exceptions in their callbacks. Our "opinion" should be clear: exceptions are not acceptable in fast path code and they will stop the capture (with an appropriate log message of course)

Pcap++/src/PcapLiveDevice.cpp
seladb: Hmm ok, that's a fair point. I guess we can leave it like this

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: > I agree that catching exceptions on a few packets has very minimal effect; however, the current implementation doesn't "count" how many callbacks threw exceptions, so it'll still work the same way if 50% of the callbacks throw exceptions

Hmm, it can be done, but I don't fully agree its the correct decision.

If the user is throwing exceptions 50% of the time, that is his problem that he needs to fix so it does not throw 50% of the time. And if he doesn't want to fix it, who are we to tell him otherwise?

Our code should still continue to work, even if he does not catch them himself. Might not get the best performance, but it is out of our hands. We did the best we could to give him a functional solution for the conditions he wants to run under. IMO, that is better than an non-functional solution as it allows greater scope of use.

> I agree that the fast path shouldn't throw exceptions, which is why I think it should be the user's decision whether to avoid or catch exceptions in their callbacks. Our "opinion" should be clear: exceptions are not acceptable in fast path code and they will stop the capture (with an appropriate log message of course)

If you throw an exception you are, by definition, no longer on the "fast path" and you should expect suboptimal performance. I disagree on stopping the capture altogether, tho.

I think of the current callback more in line with an event handler (think like C#'s events). An event handler (user callback) failing should not crash the publisher process (the capture loop) unless it performs some critical function that the loop can't do without. The current case does not fit in the exception, since the capture loop is perfectly capable of continuing along as the error is isolated to the user code.

IMO, we shouldn't limit the users too much, especially since this is the "main" capture device of the library. We should prioritize some ease of use and robustness, if it does not affect the performance in the general case (no exceptions occur). Halting the entire capture on an exception in the event handler might be unexpected for users as it is not a fatal error with respect to the capture loop, but only to the currently captured packet.

Below is a diagram. The critical section is the user callback, any errors that happen in the critical section should not escape it, and it should not interfere with the code running outside of it.

```mermaid
sequenceDiagram
  actor usr as User
  participant ct as Capture Thread Main
  participant pcap as Libpcap
  participant callback as On Packet Arrives Callback
  participant usrCallback as User Callback
  
  usr-)ct: Start capture
  activate ct
  ct-->ct: Setup procedure
  loop Until stop signal
    ct->>pcap: pcap_dispatch(descriptor, userCallback)
    activate pcap
    loop Until packet buffer is empty
        Note over callback: This is ultimately an event handler.<br/>Errors in it should not interfere with the invoking function.
        pcap->>callback: On packet arrives event
        activate callback
        callback->>callback: Unwrap event cookie
        callback->>callback: Construct RawPacket
        critical Invoke and handle user callback
          callback->>usrCallback: Invoke callback with packet
          activate usrCallback
          usrCallback->>usrCallback: Userland code
          usrCallback-->>callback:
          deactivate usrCallback
        option Exception Thrown
          note right of callback: We are no longer on the "fast path". <br/>Performance may be degraded for the cycle.
          callback->>callback: Log Error
        end
        callback-->>pcap: Event handler completed
        deactivate callback
    end
    pcap-->>ct: pcap_dispatch() result code
    deactivate pcap
  end
  ct-->ct: Cleanup procedure
  deactivate ct
```

Dimi1010: Hmm, on the other hand, I suppose we can have a policy of stopping the capture loop on a thrown exception if we treat it as a contract violation to force all users to handle their own exceptions? But it would need to be documented in the callback documentation.

Dimi1010: > Maybe catching an exception should turn on a flag that will signal the capture thread to stop capturing.

We can add a `stopRequested` flag in the context that we set to `true` on exception.
It will also require a call `pcap_breakloop()` to terminate the `pcap_dispatch` early, which would mean that the pcap descriptor also needs to be passed to the callback. 

Additional updates to the control synchronization will be needed as early termination of the thread will not update `captureActive()`.

seladb: Yes, we probably need to call `pcap_breakloop()` in this case, basically like we do in `stopCapture()`.

But maybe we can break it up into 2 PRs: one that doesn't handle exceptions (which is the current behavior), and a second PR to add the `stopRequested` flag and handle exceptions

Dimi1010: Sure, we can do that in a separate PR. The current one shouldn't have any issues working as is.

Dimi1010: > one that doesn't handle exceptions (which is the current behavior)

By the current behaviour, did you mean the base behaviour or the current PR state?
Coz by not having any issues, I meant the current PR state.

seladb: I meant the base behavior (before this PR). We don't handle exceptions there...

Dimi1010: That does leave the issue of UB as the exception is going through the C code, tho. :/

seladb: Yes, but we can handle that in the second PR, no?

Dimi1010: Updated 6d03945414624ac693cf40addf254d9c4566c24d
Added a TODO reminder tho.

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: Updated 6d03945414624ac693cf40addf254d9c4566c24d

seladb (approved):

seladb:
@Dimi1010 this PR is approved. Should we merge it?

PR #1977, opened 2025-09-26
Fixes `PcapFileWriter` inconsistently writing nanosecond precision.

### LINKED ISSUE #1975: `PcapFileWriter` does not actually write nanosecond precision files if used with Winpcap backend.
The latest Winpcap release does not support nanosecond precision pcap files, due to being based on libpcap 1.0.x. Nanosecond precision was added in 1.5.x+.

The currently used tests are flawed, due to not using a static nanosecond pcap file to compare to, but instead writing a new one and reading from it.

The `PcapFileWriter` implementation is flawed, due to providing little feedback if a nanosecond precision file is requested, but the runtime does not support it. Only an error message is provided, while the error makes the writer unusable for the requested purpose.

If used it instead silently falls back on using `pcap_open_dead` which will use the default precision.

### Fix recommendation:
- Add static nanoseconds pcap file to be tested and compared against.
- Convert the error message in `PcapFileWriterDevice`'s constructor to an exception to fail loudly at construction, as the constructed device instance is ill-formed with respect to its construction parameters.

_Originally posted by @Dimi1010 in https://github.com/seladb/PcapPlusPlus/issues/1962#issuecomment-3337317147_

Dimi1010:
Fixes #1975 

This PR applies the following changes:
- `PcapFileWriterDevice` constructor will now raise `std::runtime_error` when creating a nanosecond precision writer in an environment that does not support it.
Precision tests are split into two cases - microsecond precision and nanosecond precision.
- Added static pcap samples to compare against micro and nanosecond precision. 
- Changed the writer device path to write to `EXAMPLE_PCAP_*_WRITE_PATH`, `EXAMPLE_PCAP_*_PATH` is to be left read only for static file path.
- Nanosecond precision test now checks successful failure reporting if the test runtime does not support the feature.

### Instructions for updating code
The change makes `PcapFileWriterDevice` throw if nanosecond precision is unavailable instead of silently falling back to microsecond.
If the previous behavior is desired, the following snippet can be used to replicate it:
```cpp
PcapFileWriterDevice dev("filepath", LINKLAYER_ETHERNET, PcapFileWriterDevice::isNanoSecondPrecisionSupported());
```

### DIFF
--- a/Tests/Pcap++Test/Tests/FileTests.cpp
+++ b/Tests/Pcap++Test/Tests/FileTests.cpp
@@ -5,7 +5,6 @@
 #include "../Common/PcapFileNamesDef.h"
 #include <array>
 #include <fstream>
-#include <iostream>
 
 class FileReaderTeardown
 {


Tests/Pcap++Test/Tests/FileTests.cpp
```diff
@@ -99,21 +100,78 @@ PTF_TEST_CASE(TestPcapFileReadWrite)
...
+	for (auto const path : { EXAMPLE_PCAP_MICRO_PATH, EXAMPLE_PCAP_MICRO_WRITE_PATH })
+	{
+		// Read micro precision file
+		pcpp::PcapFileReaderDevice readerDevMicro(path);
+		PTF_ASSERT_EQUAL(readerDevMicro.getTimestampPrecision(), pcpp::FileTimestampPrecision::Unknown, enumclass);
+		PTF_ASSERT_TRUE(readerDevMicro.open());
+		PTF_ASSERT_EQUAL(readerDevMicro.getTimestampPrecision(),
+		                 pcpp::PcapFileReaderDevice::isNanoSecondPrecisionSupported()
+		                     ? pcpp::FileTimestampPrecision::Nanoseconds
+		                     : pcpp::FileTimestampPrecision::Microseconds,
```
seladb: Since both files (original and copy) were written with microseconds precision, shouldn't the precision always be `pcpp::FileTimestampPrecision::Microseconds`?

Tests/Pcap++Test/Tests/FileTests.cpp
```diff
@@ -99,21 +100,78 @@ PTF_TEST_CASE(TestPcapFileReadWrite)
...
+PTF_TEST_CASE(TestPcapFileNanoPrecision)
+{
 	// Writer precision support should equal to reader precision support
 	PTF_ASSERT_EQUAL(pcpp::PcapFileWriterDevice::isNanoSecondPrecisionSupported(),
 	                 pcpp::PcapFileReaderDevice::isNanoSecondPrecisionSupported());
 
+	if (!pcpp::PcapFileWriterDevice::isNanoSecondPrecisionSupported())
+	{
+		std::cout << "Pcap nano precision is not supported on the current platform! "
+		             "The test will check for proper failure messages.\n";
```
seladb: I don't think this message is needed, but if you insist, you can use `PTF_PRINT_VERBOSE`

Tests/Pcap++Test/Tests/FileTests.cpp
```diff
@@ -99,21 +100,78 @@ PTF_TEST_CASE(TestPcapFileReadWrite)
...
+
+	pcpp::RawPacket rawPacketMicro(testPayload.data(), testPayload.size(), timeval({ 1, 2 }), false);     // 1.000002000
+	pcpp::RawPacket rawPacketNano(testPayload.data(), testPayload.size(), timespec({ 1, 1234 }), false);  // 1.000001234
+
 	// Write nano precision file
-	pcpp::Logger::getInstance().suppressLogs();
-	pcpp::PcapFileWriterDevice writerDevNano(EXAMPLE_PCAP_NANO_PATH, pcpp::LINKTYPE_ETHERNET, true);
-	pcpp::Logger::getInstance().enableLogs();
+	pcpp::PcapFileWriterDevice writerDevNano(EXAMPLE_PCAP_NANO_WRITE_PATH, pcpp::LINKTYPE_ETHERNET, true);
 	PTF_ASSERT_EQUAL(writerDevNano.getTimestampPrecision(),
```
seladb: If we got here, it means that `pcpp::PcapFileWriterDevice::isNanoSecondPrecisionSupported()` is `true` so the precision should always be `pcpp::FileTimestampPrecision::Nanoseconds`. Same for line 181-185

Tests/Pcap++Test/Tests/FileTests.cpp
Dimi1010: Yes, that should be correct. I will change it. Thanks.

Dimi1010: @seladb Ok, so apparently that wasn't correct.

`PcapFileReaderDevice::open()` opens the pcap handle this way:
```cpp
#if defined(PCAP_TSTAMP_PRECISION_NANO)
		auto pcapDescriptor = internal::PcapHandle(
		    pcap_open_offline_with_tstamp_precision(m_FileName.c_str(), PCAP_TSTAMP_PRECISION_NANO, errbuf));
#else
		auto pcapDescriptor = internal::PcapHandle(pcap_open_offline(m_FileName.c_str(), errbuf));
#endif
```

So for microsecond pcaps, the read precision is always nanosecond if the system supports it.
Nanosecond pcaps are always nanosecond tho.

Tests/Pcap++Test/Tests/FileTests.cpp
Dimi1010: Fixed 95f74316c48560218d81cdb492512d5d48c46035

Tests/Pcap++Test/Tests/FileTests.cpp
Dimi1010: Fixed 3b02cfa739a91b62fe61c509d61e0b8de809dba8

Tests/Pcap++Test/Tests/FileTests.cpp
Dimi1010: Updated 95f74316c48560218d81cdb492512d5d48c46035. Made the expected precision into a variable to keep the assert body simpler.

Tests/Pcap++Test/Tests/FileTests.cpp
```diff
@@ -5,6 +5,7 @@
 #include "../Common/PcapFileNamesDef.h"
 #include <array>
 #include <fstream>
+#include <iostream>
```
seladb: Is this include still needed?

seladb (approved):

Tests/Pcap++Test/Tests/FileTests.cpp
Dimi1010:
```suggestion
```

Dimi1010: Shouldn't be, it was for the `std::cout`. Fixed f226a9e

PR #1981, opened 2025-09-30
Cleanup of `Packet::createFirstLayer`

Dimi1010:
- Changes the if-else chain into a switch case for better readability.
 - Removes the duplicated `new PayloadLayer(...)` calls. If all validations for a link layer fail, it exits the case statement and the payload layer is created by the tail call. IMO, that is easier to read than the multiline ternary operator + static casts calls.

### DIFF
--- a/Packet++/src/Packet.cpp
+++ b/Packet++/src/Packet.cpp
@@ -767,8 +767,7 @@ namespace pcpp
 		}
 		case LinkLayerType::LINKTYPE_NULL:
 		{
-			// Check if rawDataLen is too small fit Null/Loopback
-			if (rawDataLen >= sizeof(uint32_t))
+			if (NullLoopbackLayer::isDataValid(rawData, rawDataLen))
 			{
 				return new NullLoopbackLayer(const_cast<uint8_t*>(rawData), rawDataLen, this);
 			}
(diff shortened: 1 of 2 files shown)


Packet++/src/Packet.cpp
```diff
@@ -749,77 +751,79 @@ namespace pcpp
...
 			{
-				return IPv4Layer::isDataValid(rawData, rawDataLen)
-				           ? static_cast<Layer*>(
-				                 new IPv4Layer(const_cast<uint8_t*>(rawData), rawDataLen, nullptr, this))
-				           : static_cast<Layer*>(
-				                 new PayloadLayer(const_cast<uint8_t*>(rawData), rawDataLen, nullptr, this));
+				return new IPv4Layer(const_cast<uint8_t*>(rawData), rawDataLen, nullptr, this);
 			}
-			else if (ipVer == 0x60)
+			else if (ipVer == 0x60 && IPv6Layer::isDataValid(rawData, rawDataLen))
```
seladb: nit: the `else` keyword is redundant

Dimi1010: I suppose it is. I kept it mostly to more explicitly indicate that only one of the branches is taken even if the code is later modified. Can remove it if you want. There is a similar case in L750 too.

seladb: Yes, we can remove the `else` in that line too

Dimi1010: Fixed 4c147af3f896921e183427c1517ce94c780238ca

PR #1986, opened 2025-10-02
Deprecate match filter with packet

Dimi1010:
As per this comment chain: https://github.com/seladb/PcapPlusPlus/pull/1957#discussion_r2365562736

Deprecated all instances of `matchPacketWithFilter`.


Examples/PcapSplitter/SimpleSplitters.h
```diff
@@ -131,7 +131,7 @@ class BpfCriteriaSplitter : public Splitter
 	 */
 	int getFileNumber(pcpp::Packet& packet, std::vector<int>& filesToClose)
 	{
-		if (pcpp::IPcapDevice::matchPacketWithFilter(filter, packet.getRawPacket()))
+		if (packet.getRawPacket() != nullptr && filter.matches(*packet.getRawPacket()))
```
Dimi1010, line 134: Added the nullptr check here to follow the original design. `matchPacketWithFilter` returns `false` on nullptr packet. Dunno if its the best behaviour here tho. Maybe throw an exception instead?

seladb: I think we can keep it as is 👍

Dimi1010:
> @Dimi1010 can we modify the tests (`FilterTests.cpp`) to also use `match()` instead of `matchPacketWithFilter()`?

Sure, we will drop coverage for the old methods, but I guess it's fine?

Dimi1010:
> @Dimi1010 can we modify the tests (`FilterTests.cpp`) to also use `match()` instead of `matchPacketWithFilter()`?

Sure, we will drop coverage for the old methods, but I guess it's fine?

Pcap++/header/PcapDevice.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "Device.h"
+#include "DeprecationUtils.h"
```
seladb, line 4: nit: Let's update the comment in line 17?

Tests/Pcap++Test/Tests/FilterTests.cpp
```diff
@@ -825,16 +825,16 @@ PTF_TEST_CASE(TestPcapFilters_LinkLayer)
 	fileReaderDev1.close();
 
 	int validCounter = 0;
-	for (pcpp::RawPacketVector::VectorIterator iter = rawPacketVec.begin(); iter != rawPacketVec.end(); iter++)
+	for (auto* rawPacketPtr : rawPacketVec)
```
seladb, line 828: nit: let's fix the comment on line 818?

seladb (approved):
Few nit comments, otherwise LGTM

Pcap++/header/PcapDevice.h
Dimi1010: That comment is still correct, no? After the deprecated method is removed, the forward, declare isn't needed?

Technically, it isn't needed now either, as the filters are included in the definition of IFilterableDevice header, which is then included here?

seladb: Ah ok 👍

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: Fixed c0313adc78d50a2f98933061f0ad1ce829a92ea8

Issue #1992, opened 2025-10-09
Xdp Rx Queue


JasMetzger:
### Question

I am exploring XDP for high speed network processing and considering PCAPPlusPlus
which has wrapper for the user side. Perhaps I don't understand XDP enough but I think the AF_XDP socket  is associated with an rx queue but I don't see any reference to that in the open() function which configures it. How are rx queues designated? Thanks.  

Update: 
I think the uint32_t rxId = 0; in the receivePackets function indicates the rx queue. Any reason why this isn't an argument for the function so that the application can define multiple threads each one to handle a receive queue for performance or is that not going to help. Thanks.

### Operating systems

Linux

Dimi1010:
Hi @JasMetzger ,

Not really that familiar with XDP, but from what I understand the AF_XDP concept is based on 4 queue types:
- **FillRing:** Used for transfer of memory blocks from User space -> Kernel space
- **CompletionRing:** Used for transfer of memory blocks from Kernel space -> User space
- **TxRing:** Packets to be transmitted by the kernel are submitted here by the user space.
- **RxRing:** Received packets by the kernel are received here by the user space.

The queues can be further divided by owner:
- **FillRing** & **CompletionRing** are owned by the Umem region. They are created in `xsk_umem__create`.
- **TxRing** & **RxRing** are owned by the respective AF_XDP socket. Each socket can have a Tx queue, Rx queue or both. But from what I understand they only have up to 1 of each type. They are created during the socket setup in `xsk_socket_create`.

There are 2 independent flows in which the Umem frames are passed around in these queues:
- **Rx Flow:** User space -> FillRing -> Kernel -> RxRing -> User space
- **Tx Flow:** User space -> TxRing -> Kernel -> CompletionRing -> User space

All of the abovementioned queues are **single-producer / single-consumer**, which means that a single socket can only have 1 thread per flow, assuming the flows use separate frames and the receiver procedure does not require access to the transmission. However, I am not sure if this will work without data races through the Pcpp API. It probably should, but I can't confirm it without looking more thoroughly.

> I think the uint32_t rxId = 0; in the receivePackets function indicates the rx queue. 

I am pretty sure that is the index of the slot in the RxRing. It is initialized to 0 and then updated by `xsk_ring_cons__peek` with the index of the start of the slots available to dequeue from the ring.


[...]

PR #2000, opened 2025-10-20
Sockets dimensioned by Num queues

JasMetzger:
The device is construed as an interface and Number of Queues specified in the Configuration.  the open will create the sockets for each queue. The ids are assumed 0..., Nqueue - 1 which is like 99% of the cases. 

Aside: If just going to outfit the existing solution with a single queueid it might as well be 0 and it might as well be hardwired; no change would be necessary. Dimensioning seemed the right way to go.  

Everything should be compatible with existing API if the default is one queue and queueid = 0. Socket, stats, are all dimensioned by Number of Queues. Any queueid is checked to be less than the number of queues in configuration and will log an error if out of bounds.  If the queueid is illegal for getStatistics it returns a nulled out Device Stats since that function returns something. 

When I write the kernel side I can test but wanted to push this for consideration before I go further.

### DIFF
--- a/Pcap++/header/XdpDevice.h
+++ b/Pcap++/header/XdpDevice.h
@@ -102,6 +102,7 @@ namespace pcpp
 			/// @param[in] txSize The size of the TX ring used by the AF_XDP socket. The default value is 2048
 			/// @param[in] rxTxBatchSize The max number of packets to be received or sent in one batch. The default
 			/// value is 64
+			/// @param[in] numQueues The number of queues, and therefore sockets, to configure with device. The default is 1
 			explicit XdpDeviceConfiguration(AttachMode attachMode = AutoMode, uint16_t umemNumFrames = 0,
 			                                uint16_t umemFrameSize = 0, uint32_t fillRingSize = 0,
 			                                uint32_t completionRingSize = 0, uint32_t rxSize = 0, uint32_t txSize = 0,
@@ -248,7 +249,7 @@ namespace pcpp
 		XdpDeviceConfiguration* getConfig() const
 		{
 			// TODO: Return a copy or const ref to avoid user modifying config?
-			return m_Config.get();
+			return m_Config;
 		}
 
 		/// @return Current device statistics for queue id zero
@@ -311,8 +312,9 @@ namespace pcpp
 
 		/// A c'tor for this class. Please note that calling this c'tor doesn't initialize the AF_XDP socket. In order
 		/// to set up the socket call configureSocket().
-		/// @param[in] interfaceName The interface name to open the AF_XDP sockets on
-		explicit XdpSocket(XdpDevice *device, uint32_t qid);
+		/// @param[in] device The custodial XdpDevice
+		/// @param[in] queueid The queue id for the socket
+		explicit XdpSocket(XdpDevice *device, uint32_t queueid);
 
 		/// A d'tor for this class. It deletes the underlying socket resource and frees allocated UMEM
 		~XdpSocket();


Pcap++/header/XdpDevice.h
```diff
@@ -10,6 +10,10 @@
 /// @
 namespace pcpp
 {
+
+	// used to dimension sockets
+	#define MAXIMUM_NUMBER_QUEUES 8
```
Dimi1010: For hard constants, prefer using C++ `const` / `constexpr` variables.

If macros are needed in public headers, prefix them with `PCPP_` so they don't conflict with other macros the users might have.

Pcap++/header/XdpDevice.h
```diff
@@ -196,11 +205,11 @@ namespace pcpp
 		/// ms
 		/// @return True if stopped receiving packets because stopReceivePackets() was called or because timeoutMS
 		/// passed, or false if an error occurred.
-		bool receivePackets(OnPacketsArrive onPacketsArrive, void* onPacketsArriveUserCookie, int timeoutMS = 5000);
+		bool receivePackets(OnPacketsArrive onPacketsArrive, void* onPacketsArriveUserCookie, int timeoutMS = 5000, uint32_t queueid = 0);
 
 		/// Stop receiving packets. Call this method from within the callback passed to receivePackets() whenever you
 		/// want to stop receiving packets.
-		void stopReceivePackets();
+		void stopReceivePackets(uint32_t queueid = 0);
```
Dimi1010: How would the user call the correct queue id from inside the receive loop (`OnPacketsArrive` callback)? 🤔 
The users are expected to call `stopReceivePackets` inside the receive callback for their own queue.

Of course, they can save it externally or in the user cookie, but that seems like a code smell to me, as it should be provided to them in the callback.

@seladb IMO, it would be useful if the callback signature be changed to have a param `XdpReceiveContext&`, instead of having a pointer to `XdpDevice`. It would allow some methods to be contextualized for the receive op. Access to the device can also be through the context. For example in the current case, `XdpReceiveContext::stopReceive()`, that automatically cancels further receives. It would also allow cleaner addition and removal of parameters to the callback. Downside is, that it is a breaking change.

Alternative is to have the callback require a `bool` "continue receiving" return value, which is easier for simpler cases, but does not help with future maintenance if other changes are needed.

#### `XdpReceiveContext` version sample:
Unsure if the raw packets should be added to the context. There are pros and cons to both approaches:
- Packets inside context pros:
  - More consistent stable API signature. No need to change signature if packets structure changes.
  Each param change is breaking, while adding and deprecating accessors is non-breaking.
- Packets as separate params pros:
  - Better focus on the packet variables. They are generally the primary parameters the users want.
  - Independent lifetime can allow reuse of the context from the previous callback, as packets change more often than the rest of the context values.

```cpp
void handlePacket(RawPacket packets[], uint32_t packetCount, XdpReceiveContext& ctx)
{
  XdpDevice* device = ctx.getDevice(); // Get the device if needed.
  
  void* userCookie = ctx.getUserCookie();
  T* userCookie = ctx.getUserCookieAs<T>(); // Utility method to reinterpret_cast the cookie to T?
  
  // Assume foo is a future value that is added later. Notice how it does not modify the callback signature
  // and keeps backwards compatibility. Extra parameter would have required all callbacks to be modified.
  Foo* futureValue = ctx.getFoo();

   /* handle packet */
   
   // Stop operation sample.
   if(stopCondition)
   {
     ctx.requestStopReceiving();
   }
}
```

#### "Continue receiving" `bool`  version sample:
```cpp
bool handlePacket2(RawPacket packets[], uint32_t packetCount, XdpDevice* device, void* userCookie)
{
  /* handle packet */

  if (shouldStopCondition)
  {
    return false;
  }

  return true;
}
```

Pcap++/header/XdpDevice.h
```diff
@@ -294,21 +303,22 @@ namespace pcpp
...
-		XdpUmem* m_Umem;
-		void* m_SocketInfo;
-		XdpDeviceStats m_Stats;
-		XdpPrevDeviceStats m_PrevStats;
+		uint32_t m_NumQueues;
+		bool m_ReceivingPackets[MAXIMUM_NUMBER_QUEUES];
+		XdpUmem* m_Umem[MAXIMUM_NUMBER_QUEUES];
+		void* m_SocketInfo[MAXIMUM_NUMBER_QUEUES];
+		XdpDeviceStats m_Stats[MAXIMUM_NUMBER_QUEUES];
+		XdpPrevDeviceStats m_PrevStats[MAXIMUM_NUMBER_QUEUES];
```
Dimi1010: We prefer `std::array<T, N>` over raw arrays in new code. It gives better API for basically no extra cost.

IMO, the main benefit is the addition of `std::array<T, N>::size()` which makes it easy to get the size of the array without external variables.

Dimi1010: Additionally, I am debating on the viability of encapsulating a single open socket in `XdpSocket`.
It would allow simpler methods on a per socket basis. The `XdpDevice` can store `std::vector<XdpSocket> m_sockets`.

The current API can either forward to the selected socket (via queueId), or remain unchanged and be made to operate on all open sockets (except receive and send, those default to queue 0 for backwards compat). If option 2 is taken, `XdpDevice` can expose `XdpSocket* getSocket()` (WIP name), to allow per socket configurations.

The idea is because most elements are accessed on a "per thread queue" basis instead of "per array type" and would somewhat clean up the `[queueId]`s sprinkled everywhere. It would also allow easier addition (if needed) of sharing Umem regions between sockets as they aren't required to be on a 1-1 relationship with the open sockets.

Any thoughts?

Pcap++/header/XdpDevice.h
Dimi1010: PS: IMO, the context idea is relevant to all device type callback signatures, not just `XdpDevice`.

Dimi1010 (commented):
Thanks for taking the time to submit this PR. 🙂 
I added some comments on the implementation.

Pcap++/header/XdpDevice.h
JasMetzger: I agree that the OnPacketArrive should indicate the queue they are arriving on. This and any other information should passback to the callback.

Pcap++/header/XdpDevice.h
JasMetzger: ok---perhaps with the std::array I don't need this definition any more

Pcap++/header/XdpDevice.h
JasMetzger: Yes. I was just seeing how things propagate with the added dimension. I will make this change on my fork.

JasMetzger: I think so. Would the Xdp hold the socket info, queueid, and the umem parts? I am still groping around with XDP but it seems to be a natural progression.

Dimi1010: Essentially, yes. The `XdpSocket` holds:
- queue id
- socket info
- stats
- prev stats
- bool receivingPackets
- any other per socket variables
- (if needed) a pointer to the parent xdp device.

It can be a POD struct that just holds them for the device to use, or it can be a full class with its own methods (essentially some of the current `XdpDevice` methods that operate on a single socket) so it can be exposed to the user as a full XdpSocket object.

I somewhat like the full object solution, as it would provide simpler methods in the API.
You get the socket object from the device, store a reference to it somewhere, and then operate on it. There is no need to keep track of queue ids after fetching the open socket from the device for simple operations that run on a single queue.

JasMetzger: I would make it a class subordinate to the XdpDevice namespace.  I can start there and see how it pans out. For now might we define the Umem region part of the XdpSocket too since it's dimensioned that way currently,

JasMetzger: I pushed something that encapsulates in a Socket class.

Pcap++/header/XdpDevice.h
```diff
@@ -184,58 +187,12 @@ namespace pcpp
...
-		/// @param[in] waitForTxCompletion Wait for confirmation from the kernel that packets were sent. If set to true
-		/// this method will wait until the number of packets in the completion ring is equal or greater to the number
-		/// of packets sent. The default value is false
-		/// @param[in] waitForTxCompletionTimeoutMS If waitForTxCompletion is set to true, poll the completion ring with
-		/// this timeout. The default value is 5000 ms
-		/// @return True if all packets were sent, or if waitForTxCompletion is true - all sent packets were confirmed.
-		/// Returns false if an error occurred or if poll timed out.
-		bool sendPackets(RawPacket packets[], size_t packetCount, bool waitForTxCompletion = false,
-		                 int waitForTxCompletionTimeoutMS = 5000);
-
```
Dimi1010, line 230: Can we keep this API, but forward to `XdpSocket m_Sockets[0]` if called to keep the existing compatibility?

They can still fail if the device is not opened and there are no sockets.

Pcap++/header/XdpDevice.h
```diff
@@ -184,58 +187,12 @@ namespace pcpp
...
 		/// @return A pointer to the current device configuration. If the device is not open this method returns nullptr
 		XdpDeviceConfiguration* getConfig() const
 		{
 			return m_Config;
 		}
 
-		/// @return Current device statistics
-		XdpDeviceStats getStatistics();
```
Dimi1010, line 239: ditto, API compat.

Pcap++/header/XdpDevice.h
```diff
@@ -290,25 +247,132 @@ namespace pcpp
...
-		bool sendPackets(const std::function<RawPacket(uint32_t)>& getPacketAt,
-		                 const std::function<uint32_t()>& getPacketCount, bool waitForTxCompletion = false,
-		                 int waitForTxCompletionTimeoutMS = 5000);
-		bool populateFillRing(uint32_t count, uint32_t rxId = 0);
-		bool populateFillRing(const std::vector<uint64_t>& addresses, uint32_t rxId);
-		uint32_t checkCompletionRing();
-		bool configureSocket();
-		bool initUmem();
+		uint32_t m_NumQueues; // number of queues
+		std::array<XdpSocket*, PCPP_MAXIMUM_NUMBER_QUEUES> m_Socket;
```
Dimi1010: Btw, any specific reason `std::array<XdpSocket*, N>` was chosen instead of `std::vector<XdpSocket>`?

Vector would allow keeping only the active sockets, without having to deal with `nullptr` or manually deleting them. It would also remove a level of indirection when accessing `XdpSocket` through the container.

Also, since we are mapping thread queue to socket 1:1, using a `std::vector<XdpSocket>` we can use `m_Socket.size()` to fetch the number of active queues, removing the need for `m_NumQueues`.

Pcap++/header/XdpDevice.h
```diff
@@ -290,25 +247,132 @@ namespace pcpp
...
+
+			const XdpDevice *getDevice() { return m_Device; }
+
+			/// @typedef OnPacketsArrive
+			/// The callback that is called whenever packets are received on the socket
+			/// @param[in] packets An array of the raw packets received
+			/// @param[in] packetCount The number of packets received
+			/// @param[in] device The XdpDevice packets are received from (represents the AF_XDP socket)
+			/// @param[in] userCookie A pointer to an object set by the user when receivePackets() started
+			typedef void (*OnPacketsArrive)(RawPacket packets[], uint32_t packetCount, XdpSocket* socket, void* userCookie);
```
Dimi1010: nit: Not critical, but having socket nested to the class might make the signature a bit long? 🤔 

External code would have to define it as:
 `void handler(pcpp::RawPacket packets[], uint32_t packetCount, pcpp::XdpDevice::XdpSocket* socket, void* userCookie)`.
 
Anyway, its not critical and can be decided later.

Pcap++/header/XdpDevice.h
```diff
@@ -290,25 +247,132 @@ namespace pcpp
...
+			{
+				m_Device = nullptr;
+				m_Queueid = 0;
+
+				m_ReceivingPackets = false;
+				m_Umem = nullptr;
+				m_SocketInfo = nullptr;
+				memset(&m_Stats, 0, sizeof(XdpDeviceStats));
+				memset(&m_PrevStats, 0, sizeof(XdpPrevDeviceStats));
+			}
```
Dimi1010: nit: This appears to only be used once in the `XdpSocket` constructor?
Will it be used more, or can it be integrated directly?

Pcap++/header/XdpDevice.h
```diff
@@ -290,25 +247,132 @@ namespace pcpp
...
+			/// this timeout. The default value is 5000 ms
+			/// @return True if all packets were sent, or if waitForTxCompletion is true - all sent packets were confirmed.
+			/// Returns false if an error occurred or if poll timed out.
+			bool sendPackets(RawPacket packets[], size_t packetCount, bool waitForTxCompletion = false,
+							int waitForTxCompletionTimeoutMS = 5000);
+
+			/// @return Current device statistics
+			XdpDeviceStats getStatistics();
+
+			bool configure();
```
Dimi1010: This generally should not be accessible by public user code.
It is better to move it to `private` and add `XdpDevice` as a friend of `XdpSocket`.

Pcap++/header/XdpDevice.h
Dimi1010: PS: It should probably allow further changes to `XdpSocket` to be fully RAII compliant, eliminating the need for `configure` and having it happen in the constructor, but that can be done later.

Pcap++/header/XdpDevice.h
```diff
@@ -290,25 +247,132 @@ namespace pcpp
...
 
+
+
+
+	public:
+
+		class XdpSocket
+		{
+		public:
+			XdpSocket(XdpDevice *device, uint32_t qid);
```
Dimi1010: ditto: Private / Protected accessibility and `friend class XdpDevice`.
Code outside of `XdpDevice` should not be able to create an `XdpSocket`.

Pcap++/header/XdpDevice.h
Dimi1010: Also, breaking change in the signature with the existing callback API, but not much that can be done here.

If we are keeping the old `XdpDevice` API that works on Queue 0, we can probably provide a wrapper for code that essentially does:
```cpp
void handler(RawPacket packets[], uint32_t packetCount, XdpSocket* socket, void* userCookie)
{
  oldCallback(packets, packetCount, socket->getDevice(), userCookie);
}
```
to allow code using the old API to work seamlessly if it does not need extra queues.

It would probably require storing the callback in `std::function` instead of raw function pointer, but that has the secondary effect of allowing user code to use lambdas. For function pointers the difference between `std::function` and a raw function pointer is pretty negligible as most mainstream implementations have that as an optimization.

Pcap++/src/XdpDevice.cpp
```diff
@@ -523,15 +574,17 @@ namespace pcpp
...
+		if (initConfig())
 		{
-			if (m_Umem)
+			// configure for each socket
+			for(uint32_t i = 0; i < m_NumQueues; i++)
 			{
-				delete m_Umem;
-				m_Umem = nullptr;
+				m_Socket[i] = new XdpSocket(this, i);
+				m_Socket[i]->configure();
```
Dimi1010: Configure can fail. It should be handled by aborting the operation, cleaning up already opened resources and returning `false`.

Pcap++/header/XdpDevice.h
Dimi1010: It also removes the MAX_QUEUES limitation.

Pcap++/src/XdpDevice.cpp
```diff
@@ -117,21 +118,41 @@ namespace pcpp
...
+		initialize();
+
+		m_Device = device;
+		m_Queueid = qid;
+	}
+
+	XdpDevice::XdpSocket::~XdpSocket()
+	{
+		auto socketInfo = static_cast<xsk_socket_info*>(m_SocketInfo);
+		xsk_socket__delete(socketInfo->xsk);
```
Dimi1010: This should check for `nullptr` before delete.

Pcap++/src/XdpDevice.cpp
```diff
@@ -117,21 +118,41 @@ namespace pcpp
...
+
+		delete m_Umem;
+		m_Umem = nullptr;
```
Dimi1010: ditto: `nullptr` check before delete.

Tests/Pcap++Test/Tests/XdpTests.cpp
Dimi1010, line 1: Related to: https://github.com/seladb/PcapPlusPlus/pull/2000/files#r2446086115

If we return the old API, revert the changes to the tests for it. `XdpSocket` can have new tests.

Pcap++/header/XdpDevice.h
JasMetzger: Yeah---I was beginning to think that too. But it also suggests that the socket is subordinate to the device which it kind of is, and I thought that was a good idea. But the order of things in the XdpDevice class got messy.  I also thought it should be in its own file set which might make things tidier; I'll have to figure that out if you agree.

Pcap++/header/XdpDevice.h
JasMetzger: I thought about these things. Because the index into the array is also the queueid I thought the array was more natural but we could do the same thing with vector. I also thought of configure as part of constructor but the current code has these as separate phases. I will give this another go---this was the initial concept of XdpSocket.

Pcap++/header/XdpDevice.h
JasMetzger: This is a habit of mine in case of multiple constructors. I call initialize() in each one. But it's not necessary if there will be only one constructor

Pcap++/src/XdpDevice.cpp
JasMetzger: Yep--I have to go over this again. It was my initial description

Pcap++/src/XdpDevice.cpp
JasMetzger: Agreed.

Pcap++/src/XdpDevice.cpp
JasMetzger: I will look into this. Perhaps configure can go away if the constructor can do stuff. I will investigate

Pcap++/header/XdpDevice.h
Dimi1010: I see. Assuming there is no complex default code in construction, if all constructors require the same initial value defaults, would it not be better to define the values directly in the member declaration?

Then, you can override the member initializer list in the ctor if some specific constructors change some values.

I personally prefer to use that alongside delegating constructors if possible.

Pcap++/header/XdpDevice.h
Dimi1010: Hmm. I don't think we need a new file for now.

I think the following order would work:
```
root (namespace pcpp)
- XdpDevice - forward declare.
- XdpSocket - full declaration.
  - XdpUmem - nested, private visibility.
- XdpDevice - full declaration.
```

I think this should be fine since `XdpUmem` was currently only used inside `XdpSocket`? (Famous last words)

seladb:
@JasMetzger didn't we want to just add the RX queue to `XdpDeviceConfiguration` and use this value in `xsk_socket__create` instead of always passing `0` ([link](https://github.com/seladb/PcapPlusPlus/issues/1992#issuecomment-3417753006))?

This PR seems to do much more than just this...

Dimi1010:
> @JasMetzger didn't we want to just add the RX queue to `XdpDeviceConfiguration` and use this value in `xsk_socket__create` instead of always passing `0` ([link](https://github.com/seladb/PcapPlusPlus/issues/1992#issuecomment-3417753006))?
> 
> This PR seems to do much more than just this...

@seladb Fair enough, we did initially talk about that.

IMO, this is a slightly better approach as it separates the `XdpDevice`, which would be the NIC, from the open sockets on it — `XdpSocket`.

Having the user create multiple devices from the same interface for different thread queues seems confusing to me as all other `***Device` implementations seem to match the NIC it is used on 1:1 (one NIC makes 1 device instance). It would also be slightly brittler as we can't really prevent the user from opening a thread queue twice that way.

seladb:
> > @JasMetzger didn't we want to just add the RX queue to `XdpDeviceConfiguration` and use this value in `xsk_socket__create` instead of always passing `0` ([link](https://github.com/seladb/PcapPlusPlus/issues/1992#issuecomment-3417753006))?
> > This PR seems to do much more than just this...
> 
> @seladb Fair enough, we did initially talk about that.
> 
> IMO, this is a slightly better approach as it separates the `XdpDevice`, which would be the NIC, from the open sockets on it — `XdpSocket`.
> 
> Having the user create multiple devices from the same interface for different thread queues seems confusing to me as all other `***Device` implementations seem to match the NIC it is used on 1:1 (one NIC makes 1 device instance). It would also be slightly brittler as we can't really prevent the user from opening a thread queue twice that way.

This PR introduces a breaking change, which might not be great for users already using this device. While I agree we can think about the bigger refactoring, I'd start simple by allowing RX queue ID configuration, then see the demand for the bigger change. Please let me know what you think

Dimi1010:
> This PR introduces a breaking change, which might not be great for users already using this device. While I agree we can think about the bigger refactoring, I'd start simple by allowing RX queue ID configuration, then see the demand for the bigger change. Please let me know what you think

Do you mean the breaking changes in the callbacks and API of XdpDevice? If so, I have ideas for mitigation.

We keep the old API. It forwards its operations to socket[0] to keep old behaviour that operated on queue 0.

For the callbacks, the mitigation is:
- XdpDevice::OnPacketArrives remains unchanged and receives a device pointer.
- XdpSocket::OnPacketArrives has `XdpSocket` pointer instead of `XdpDevice` pointer.
- When the user uses the old `XdpDevice::receivePackets`, the user callback is wrapped in a lambda that does the following:
```cpp
[userCallback](RawPacket* packets, int packetCount, XdpSocket* socket, void userCookie)
{
   return userCallback(packets, packetCount, socket->getDevice(), userCookie);
}
```

Alternatively, it may capture the device from the `this` pointer as it will be created inside receivePackets, to avoid a `const cast` (XdpSocket returns a const device).

This should keep the changes backward compatible. Let me know if I missed something.

Dimi1010:
> This PR introduces a breaking change, which might not be great for users already using this device. While I agree we can think about the bigger refactoring, I'd start simple by allowing RX queue ID configuration, then see the demand for the bigger change. Please let me know what you think

@seladb Do you mean the breaking changes in the callbacks and API of XdpDevice? If so, I have ideas for mitigation.

We keep the old API. It forwards its operations to socket[0] to keep old behaviour that operated on queue 0.

For the callbacks, the mitigation is:
- XdpDevice::OnPacketArrives remains unchanged and receives a device pointer.
- XdpSocket::OnPacketArrives has `XdpSocket` pointer instead of `XdpDevice` pointer.
- When the user uses the old `XdpDevice::receivePackets`, the user callback is wrapped in a lambda that does the following:
```cpp
[userCallback](RawPacket* packets, int packetCount, XdpSocket* socket, void userCookie)
{
   return userCallback(packets, packetCount, socket->getDevice(), userCookie);
}
```

Alternatively, it may capture the device from the `this` pointer as it will be created inside receivePackets, to avoid a `const cast` (XdpSocket returns a const device).

This should keep the changes backward compatible. Let me know if I missed something.

seladb:
I want to make this PR small and simple. Later on, we can discuss bigger changes if needed...

Dimi1010:
> I want to make this PR small and simple. Later on, we can discuss bigger changes if needed...

@seladb, the small change would make this refactor a breaking change. This means we are locking ourselves out of this cleaner solution without a bigger breaking change, while adding this initially makes this possible without breaking changes.

seladb:
> > I want to make this PR small and simple. Later on, we can discuss bigger changes if needed...
> 
> @seladb, the small change would make this refactor a breaking change. This means we are locking ourselves out of this cleaner solution without a bigger breaking change, while adding this initially makes this possible without breaking changes.

Since this feature was never brought up until now, I wonder how much it's going to be used Even @JasMetzger is not sure he really needs it. I suggest we start small. Worst case if this feature becomes popular, we can figure it out later

JasMetzger:
If all you want to do is set a single rxqueue id in the config, I'm not sure that will do anything because the likelihood of that rxqueue id remaining zero is extremely high. Just leave alone.  To achieve wirespeeds of processing (e.g. 10G) likely require the multiple queues. 

The code is not ready. I'll integrate the suggested changes and leave it.

Dimi1010:
> Since this feature was never brought up until now, I wonder how much it's going to be used Even @JasMetzger is not sure he really needs it. I suggest we start small. Worst case if this feature becomes popular, we can figure it out later

IMO, figuring it out after it becomes popular is a bad way to go, as then you have to work around breaking user code that depends on the initial implementation of the change and all the headaches that brings. In the initial addition, you have a lot more freedom, as no code depends on it, so getting a good initial solution that is flexible and maintainable is important.

At its core, the PR is not that complex, only somewhat long. Wrap the logic that operates on a single socket in `XdpSocket` and make `XdpDevice` be able to hold more than one socket.

It can even be split in two stages:
1. Wrapping the logic inside `XdpSocket`
2. Making `XdpDevice` be able to hold more than 1 socket.

PS: Sorry for the mentions in the quoted text, Jas. 😅

JasMetzger:
I pushed changes that preserve the original API, tests, and integrate the suggested changes. Also did doxygen. Apologies for the thrash---this is the first time I've done this sort of thing.

seladb:
> If all you want to do is set a single rxqueue id in the config, I'm not sure that will do anything because the likelihood of that rxqueue id remaining zero is extremely high. Just leave alone. To achieve wirespeeds of processing (e.g. 10G) likely require the multiple queues.
> 
> The code is not ready. I'll integrate the suggested changes and leave it.

@JasMetzger I never tested high traffic applications, but I assume users can generate multiple devices with different RX queus which will open different sockets. Is there any reason why it wouldn't work?

seladb:
> IMO, figuring it out after it becomes popular is a bad way to go, as then you have to work around breaking user code that depends on the initial implementation of the change and all the headaches that brings. In the initial addition, you have a lot more freedom, as no code depends on it, so getting a good initial solution that is flexible and maintainable is important.

@Dimi1010 I don't necessarily agree, especially because I don't know eBPF / AF_XDP well enough to say if / how multiple RX queues should be used. Apparently there is not a lot of documentation about it as well. Since @JasMetzger is interested in configuring the RX queue, I think we can make this small change to allow configuring it. 

If someone has the time to explore how to use multiple RX queues, and have the hardware and time to test it, I'm open to the broader change

JasMetzger:
>@JasMetzger I never tested high traffic applications, but I assume users can generate multiple devices with different RX queus which will open different sockets. Is there any reason why it wouldn't work?

I think it might. I will explore that. Instead of an array of sockets in the XdpDevice I would need an array of XdpDevices in my application. Other than construing a device as an interface + queueid (which I don't think is the understanding of a XDP device but so be it) I think it will do what i want. 

Perhaps I will start over with this PR and do the queueid in the config and see where it leads. It should be easy.

Dimi1010:
> @Dimi1010 I don't necessarily agree, especially because I don't know eBPF / AF_XDP well enough to say if / how multiple RX queues should be used. Apparently there is not a lot of documentation about it as well. Since @JasMetzger is interested in configuring the RX queue, I think we can make this small change to allow configuring it.
> 
> If someone has the time to explore how to use multiple RX queues, and have the hardware and time to test it, I'm open to the broader change

Fine, we can do a `queueId` parameter in the `XdpDevice`'s constructor or `open()` as an initial test (and keep it as a device parameter), but can we please mark the change as experimental (and NOT part of the stable API) in the docs, so we can more easily break it if we do a reworking in the future to a `1 Device <-1:N-> N sockets <-1:1-> N queues` implementation.

```cpp
class XdpDevice
{
  XdpDevice(interface);
  XdpDevice(interface, threadQueue); // <- maybe this?
  
  bool open();
  bool open(threadQueue); // <- maybe this?
  bool open(XdpConfiguration);
  bool open(XdpConfiguration, threadQueue); // <- maybe this?
}

```

I really dislike the architectural discrepancy of a single `XdpDevice` being bound to an interface + a single hardware queue and requiring multiple instances bound to same device interface, considering that `DpdkDevice` supports multiple queues on the same device with a single instance. The entire purpose of the Device classes are that you only get 1 per NIC.

JasMetzger:
I still need to investigate but I think it boils down to nomenclature. If XdpDevice renamed everywhere XdpSocket in the current release and given a queue id, that might comport better with the XDP idea. For now keeping it XdpDevice and some documentation may be the best approach for now. 

I figured on renaming my current fork and re-forking. But that doesn't seem to work as it won't let me refork if the old one persists even though renamed. What happens if I delete the old fork and is there a better way to do this preserving what was done in the initial fork?

JasMetzger:
I will delete the fork and start over

seladb:
> but can we please mark the change as experimental (and NOT part of the stable API) in the docs, so we can more easily break it if we do a reworking in the future to a `1 Device <-1:N-> N sockets <-1:1-> N queues` implementation

@Dimi1010 I have no issues making it experimental, and we can break it later if we want to 👍 

> I still need to investigate but I think it boils down to nomenclature. If XdpDevice renamed everywhere XdpSocket in the current release and given a queue id, that might comport better with the XDP idea. For now keeping it XdpDevice and some documentation may be the best approach for now.

@JasMetzger I understand that the name `XdpDevice` is confusing... the reason we called it a "device" is to stay consistent with the other "backends" supported in PcapPlusPlus - for example: libpcap live device, pcap/ng file device, DPDK device, PF_RING device, and raw socket device. Maybe a better name could have been `XdpSocketDevice`...

seladb:
> I will delete the fork and start over

Sure, no problem! 👍

JasMetzger:
I am making the mods in the new fork--pretty simple. 

When vetting the config information, the system checks the legitimacy of configuration parameters. For the queue id, there are a number of possible checks:

* Check against the number of cores (std::thread::hardware_concurrency). This is easy to do but not necessarily a legitimate test, although typical in a production system for queues and thread =< cores. 
* Check against the number of hardware queues of the given interface. This is a bit trickier to obtain. xsk_ has no such function. A function could be written that examines the linux file system for the telltale definitions using a try catch mechanism.
* Check against a reasonable maximum. The DPDK device code defines Maximums for queues and this would follow suit.
* Don't check at all and let the socket creation return an error if the queueid exceeds what is allowed.

Dimi1010:
I think we can do 3 or 4 initially. It is simple to do and should work.

We can possibly refine the validation to utilize option 2 later down the line. 

Option 1 feels weird to me as the logical core count should have no relation to the possible number of available queues? It is true that for performance reasons, most common config is 1 to 1 maping, but that shouldn't impose a hard cap?

JasMetzger:
> I think we can do 3 or 4 initially. It is simple to do and should work.
> 
> We can possibly refine the validation to utilize option 2 later down the line.
> 
> Option 1 feels weird to me as the logical core count should have no relation to the possible number of available queues? It is true that for performance reasons, most common config is 1 to 1 maping, but that shouldn't impose a hard cap?

I agree. I'll define a reasonable maximum, perhaps 64 since the highest end may have that many.

seladb:
I think we can go with (2), I asked ChatGPT and it suggested the following code (which I didn't test...). Overall seems quite simple to me:
```cpp
#include <iostream>
#include <dirent.h>
#include <string>
#include <regex>

int main(int argc, char* argv[]) {
    if (argc != 2) {
        std::cerr << "Usage: " << argv[0] << " <interface>\n";
        return 1;
    }

    std::string path = "/sys/class/net/" + std::string(argv[1]) + "/queues/";
    DIR* dir = opendir(path.c_str());
    if (!dir) {
        perror("opendir");
        return 1;
    }

    std::regex rx_regex("^rx-\\d+$");
    int rx_count = 0;

    struct dirent* entry;
    while ((entry = readdir(dir)) != nullptr) {
        if (std::regex_match(entry->d_name, rx_regex)) {
            rx_count++;
        }
    }

    closedir(dir);
    std::cout << "RX queues: " << rx_count << "\n";
    return 0;
}
```

JasMetzger:
I'll give it  a try

seladb:
@JasMetzger can we close this PR? It should be replaced by https://github.com/seladb/PcapPlusPlus/pull/2009

JasMetzger:
Yes. Sorry for delay, I was away all week.

seladb:
Closing this PR as it will be replaced by #2009

PR #2002, opened 2025-10-23
handle FIN flag in order of the sequence numbers

### LINKED ISSUE #1979: TcpReassembly: data frames after FIN
### Bug description

I am trying to analyze FTP transfers with PcapPlusPlus. To extract the files from the passive FTP transfers, I used the TcpReassembly example as a reference. Basically, the extraction of the files works via the pcpp::TcpReassembly class. Unfortunately, in a longer test, too little data is extracted from the TCP stream in 14 of 355 transfers. TcpReassembly then reports missing data in the OnTcpMessageReady callback via the TcpStreamData object with the GetMissingByteCount() function. In a debug log, I see the following messages: 

`Got a packet after FIN or RST were already seen on this side (1). Ignoring this packet`

On the hardware side, a TAP is connected between the switch and one of the two participants. From the TAP, the transmission and reception directions are separated via 2x USB Ethernet adapters (LAN9005A chipset) to an ARM Linux computer. There, I combine the two interfaces via the bonding kernel module for the PcapPlusPlus library. The FTP server is Wind River's VwWorks RTOS.

In all 14 “faulty” transmissions, I see a frame with a FIN flag shortly before the end, followed by more data frames. The frame with the FIN flag belongs at the end of the sequence number, but it arrives a few frames too early. See screenshots of three transmissions: (I cannot publish the complete dumps).

Image
Image
Image

By recording via two different network interfaces, I can understand when the ACK packets from the other side are delayed.

Does such an “out-of-order” FIN frame still comply with the TCP specification? Can I expect the other side to understand this normally? (I do not have access to the other side.)

How complex is it to take this case into account in the pcpp::TcpReassembly class?

Other (specify in the next question)

v25.05

Linux

Yocto scarthgap armhf

### Compiler version

13.2.0

libpcap

matthiasklein:
This change resolves #1979

All 14 of my “faulty” transmissions that I had recorded are decoded correctly with this change.


Dimi1010:
Changing base branch to `dev` as per `Contributing.md`.

seladb:
@matthiasklein the `X509Toolkit` tests fail, I assume that this change is related. Can you please take a look?

matthiasklein:
@seladb I corrected the number of messages in the test. The 31 extracted certificates still have exactly the same content.

matthiasklein:
@seladb I have corrected the source code indentations that were complained about. Can you run the tests again?

seladb:
@matthiasklein can you please walk me through this fix?
Here is what I'm trying to understand: if a FIN packet has the expected sequence number, we still close the connection as we used to do, so additional data packets won't be captured on this flow.
If the FIN packet has a sequence number that doesn't match the expected one, we add it to the out-of-order fragment list and handle the connection close when going over the out-of-order list?

matthiasklein:
@seladb Yes, it is as you assumed: if the FIN packet has the correct sequence number, it is treated as before.

If the sequence number of the FIN packet is not as expected, it is added to the fragment list and only processed when it is in order based on the sequence number.

This ensures that we receive the complete data for all of our problem transmissions.

seladb:
@matthiasklein thanks, I think I understand this solution. Why not treat RST packets the same?
Also - do you have a pcap that you can add that tests this scenario?

matthiasklein:
@seladb I think the RST flag could be treated similarly. But in our particular case, it only ever affected the FIN flag.

I extracted a single FTP transfer from a longer recording that sends such a FIN too early, and anonymized almost everything. Even the content of the transferred file itself. Only the number of frames and the lengths are still the same so that the sequence numbers still match.

Would the file be OK?
[fin_too_early.zip](https://github.com/user-attachments/files/23930617/fin_too_early.zip)

seladb:
@matthiasklein this pcap file contains 2 flows - I believe the one that is sending a FIN packet out-of-order is the one with 117 packets, and the early FIN is packet 112?
image

Do you think you can find a short flow with less packets?
Also - it'd be good to only include 1 flow in the pcap file.

Another thing we can do if you don't have such a pcap file is to try to reproduce the issue with one of the existing pcap files by changing the order of the packets.

matthiasklein:
@seladb Yes, the flow with the early FIN in packet 112 is the problematic one.

This is the shortest real flow I have recorded with such an early FIN.
How do we proceed from here?

yvyw:
If you don't mind, I'd like to share a few thoughts on the subject.
- I think the `flowKey` should be passed as an additional argument to `checkOutOfOrderFragments`. There is no need to save it in memory one more time.
- To improve the `TcpFragment` layout, it would be better to move `isFin` and `sequence` to the end of the structure: https://godbolt.org/z/187xajcxf
- I agree with @seladb that FIN and RST should be handled similarly.
- In the `OnTcpConnectionEnd` callback, it could also be useful to know whether the connection finished with a FIN or RST. A connection that was closed by RST has been interrupted for some reason, so it may only contain part of the conversation.

PR #2004, opened 2025-10-24
Improve encapsulation of logic relating to Layer attachment to Packet.

Dimi1010:
## Summary
This PR intended to improve the encapsulation and clarity of Layer field data. As per the C++ core guidelines `protected` data fields are generally undesired as they represent difficulty in maintaining base class invariants across derived classes.

The main object of this PR is the introduction of a new helper struct `LayerAllocationInfo` in the internal namespace, to hold allocation information for Layer objects and provide behavior helper methods. The helper structure replaces is contained as a field in a `Layer` to represent its current allocation state.

### `LayerAllocationInfo` and chosen fields.
The information structure contains 2 members:
- `Packet* attachedPacket`, replacing `m_Packet` pointer inside `Layer`.
- `bool managedByPacket`, replacing `m_isAllocatedInPacket` flag inside `Layer`.

The replacements were chosen due to their relevancy to the allocation behaviour of `Layer` and little else. Those fields were previously intermixed with other fields in the `Layer` object causing confusion when determining their use and scope.

- `attachedPacket` (previously `m_Packet`) is exclusively used for indication if the layer's data is owned by a packet and to forward requests for structural data changes to the underlying packet buffer, if available. The new name was chosen to better represent that use.
- `managedByPacket` (previously `m_IsAllocatedInPacket`) is exclusively used as flag to determine if the `Layer` object's lifecycle is managed by the `Packet` instance it is attached to. The field intricately tied to the state of `attachedPacket` and as such is included to the helper structure.

The new fields include extensive documentation regarding their use, improving the maintainability of the code. ( Think about the you in 5 months reading it 🙂 )

### Behaviour methods instead of direct field mutation
The new helper structure adds two behavior methods:
- `attachPacket(Packet* packet, bool managed, bool force = false)`: This method encapsulates the behavior of attaching a layer to a Packet instance. It allows a `Packet` instance attaching a `Layer` to itself to update the information in one step instead of manually having to directly update `private` fields to the `Layer` instance. The method also provides integrated sanity checks if a `Layer` is being attached to a `Packet` instance twice.
- `detach()`: Encapsulates the behaviour of detaching a `Layer` from `Packet`. It allows a `Packet` instance detaching a `Layer` to inform it of its new state in one step.

### Protected data accessors and private fields
The new `LayerAllocationInfo` structure is held as a private field inside `Layer`. The change improves data integrity as the allocation information (packet or managed flag) is unavailable for accidental modification by derived classes.

To keep existing behaviour of a `Layer` instance forwarding its attachment to any other instance it constructs during parsing, a new protected accessor `getAttachedPacket()` has been added and instances of direct use of `m_Packet` are replaced by it.

### Bug Fixes
- Fixed bug due to lack of zeroing of `m_IsAllocatedInPacket` flag when detaching a layer. If said layer is attached to a new packet instance, the old implementation disregarded the ownership flag and considered the layer instance as owned by the packet, due to the layer flag being set from the previous packet.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -289,12 +289,6 @@ jobs:
           cd Tests/Packet++Test
           Bin/Packet++Test
 
-      - name: Test Pcap++
-        if: env.avx512 == 'true'
-        run: |
-          cd Tests/Pcap++Test
-          Bin/Pcap++Test
-
       - name: Tests skipped (no AVX-512 hardware support)
         if: env.avx512 == 'false'
         run: echo "Skipping tests since AVX-512 is not supported on the current runner"
@@ -501,7 +495,7 @@ jobs:
     name: Determine Windows Pcap Backend Matrix
     runs-on: ubuntu-latest
     outputs:
-        npcap-or-default: ${{ steps.generate-backends.outputs.npcap_or_default}}
+      npcap-or-default: ${{ steps.generate-backends.outputs.npcap_or_default}}
     steps:
       - name: Initialize NPcap backend variable
         run: echo "npcap_available=false" >> $GITHUB_ENV
@@ -705,42 +699,82 @@ jobs:
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
+  windivert:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Add MSBuild to PATH
+        uses: microsoft/setup-msbuild@6fb02220983dee41ce7ae257b6f4d8f9bf5ed4ce # v2.0.0
+
+      - name: Install WinPcap
+        run: |
+          ci\install_winpcap.bat
+          echo "PCAP_SDK_DIR=C:\\WpdPack" >> $env:GITHUB_ENV
+      - name: Set WinDivert root
+        run: echo "WINDIVERT_DIR=C:\\WinDivert" >> $env:GITHUB_ENV
+
+      - name: Install WinDivert
+        shell: pwsh
+        run: .\ci\install_windivert.ps1 -Target "$env:WINDIVERT_DIR"
+
+      - name: Configure PcapPlusPlus (WinDivert)
+        run: cmake -A x64 -G "Visual Studio 17 2022" -DPCAP_ROOT=${{ env.PCAP_SDK_DIR }} -DPCAPPP_USE_WINDIVERT=ON -DWinDivert_ROOT=${{ env.WINDIVERT_DIR }} -S . -B "$env:BUILD_DIR"
+
+      - name: Build PcapPlusPlus
+        run: cmake --build $env:BUILD_DIR -j
+
+      - name: Install tcpreplay
+        run: ci\install_tcpreplay.bat
+
+      - name: Test PcapPlusPlus
+        shell: pwsh
+        run: |
+          Copy-Item "${{ env.WINDIVERT_DIR }}\x64\WinDivert.dll" "Tests\Pcap++Test\Bin\WinDivert.dll" -Force
+          Copy-Item "${{ env.WINDIVERT_DIR }}\x64\WinDivert64.sys" "Tests\Pcap++Test\Bin\WinDivert64.sys" -Force
+
+          python -m pip install -r ci\run_tests\requirements.txt
+          python ci\run_tests\run_tests_windows.py --include-tests windivert
+
   freebsd:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         version: ["14.1", "13.4"]
     steps:
-    - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-    - name: Test in FreeBSD
-      id: test
-      uses: vmactions/freebsd-vm@8873d98fd1413b5977cb2f7348fe329775159892 # v1.1.9
-      with:
-        release: ${{ matrix.version }}
-        usesh: true
-        prepare: |
-          pkg install -y python bash cmake git gmake gsed libpcap tcpreplay libffi openssl py311-cryptography
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - name: Test in FreeBSD
+        id: test
+        uses: vmactions/freebsd-vm@8873d98fd1413b5977cb2f7348fe329775159892 # v1.1.9
+        with:
+          release: ${{ matrix.version }}
+          usesh: true
+          prepare: |
+            pkg install -y python bash cmake git gmake gsed libpcap tcpreplay libffi openssl py311-cryptography
+            sysctl kern.ipc.maxsockbuf=16777216
+          run: |
+            set -e
 
-        run: |
-          echo "Building PcapPlusPlus"
-          chmod a+rw /dev/bpf*
-          cmake -S . -B Dist
-          cmake --build Dist -j$(sysctl -n hw.ncpu)
+            echo "Building PcapPlusPlus"
+            chmod a+rw /dev/bpf*
+            cmake -S . -B Dist
+            cmake --build Dist -j$(sysctl -n hw.ncpu)
 
-          echo "Setting up the network interface for the tests"
-          # Get the first interface name that is not 'lo'
-          interface_name=$(ifconfig -l | tr ' ' '\n' | grep -v '^lo' | head -n 1)
-          ifconfig "$interface_name" promisc
+            echo "Setting up the network interface for the tests"
+            # Get the first interface name that is not 'lo'
+            interface_name=$(ifconfig -l | tr ' ' '\n' | grep -v '^lo' | head -n 1)
+            ifconfig "$interface_name" promisc
 
-          echo "Testing PcapPlusPlus"
-          python -m ensurepip
-          python -m pip install -r ci/run_tests/requirements.txt
-          python ci/run_tests/run_tests.py --interface "$interface_name"
+            echo "Testing PcapPlusPlus"
+            python -m ensurepip
+            python -m pip install -r ci/run_tests/requirements.txt
+            python ci/run_tests/run_tests.py --interface "$interface_name"
 
-          echo "Testing PcapPlusPlus examples"
-          cd Tests/ExamplesTest
-          python -m pip install -r requirements.txt
-          python -m pytest --interface "$interface_name" --root-path=../../Dist/examples_bin
+            echo "Testing PcapPlusPlus examples"
+            cd Tests/ExamplesTest
+            python -m pip install -r requirements.txt
+            python -m pytest --interface "$interface_name" --root-path=../../Dist/examples_bin
 
   android:
     strategy:
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -35,7 +35,7 @@ jobs:
         dry-run: false
         sanitizer: ${{ matrix.sanitizer }}
     - name: Upload Crash
-      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+      uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
       if: failure() && steps.build.outcome == 'success'
       with:
         name: artifacts
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -29,7 +29,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@3599b3baa15b485a2e49ef411a7a4bb2452e7f93 # v3.29.5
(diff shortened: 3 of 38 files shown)


Packet++/src/Layer.cpp
```diff
@@ -28,16 +26,18 @@ namespace pcpp
 		if (this == &other)
 			return *this;
 
+		// Should this really always delete m_Data? What if the layer is attached to a packet?
 		if (m_Data != nullptr)
 			delete[] m_Data;
```
Dimi1010, line 31: See comment? This would cause an error if called on a Layer that does not own its data span.

Packet++/src/Packet.cpp
```diff
@@ -497,15 +493,15 @@ namespace pcpp
...
+		if (tryToDelete && layer->m_AllocationInfo.managedByPacket)
 		{
 			delete layer;
 		}
 		// if layer was not allocated by this packet or the tryToDelete is not set, detach it from the packet so it can
 		// be reused
 		else
 		{
-			layer->m_Packet = nullptr;
+			layer->m_AllocationInfo.detach();
```
Dimi1010, line 504: This line in particular created a subtle bug in the old implementation.

If a layer is originally owned by the packet instance, and detached, the `m_IsAllocatedInPacket` flag was not zeroed. If the layer has then been attached to another Packet. The attached layer was considered as if managed by the new packet even if attached with `addLayer(layer, false)` which should not transfer ownership of the layer object.

Packet++/header/Layer.h
```diff
@@ -30,6 +30,59 @@ namespace pcpp
...
+			Packet* attachedPacket = nullptr;
+
+			/// @brief Controls if the layer object is considered managed by the attached Packet
+			///
+			/// If 'true', the Layer object is considered managed by the attached Packet and will be tracked and freed
+			/// by it
+			///
+			/// If 'false', the Layer object is considered unmanaged and the user is responsible for freeing it.
+			/// This is commonly the case for layers created on the stack and attached to a Packet.
+			bool managedByPacket = false;
```
Dimi1010: Maybe `ownedByPacket` is a more descriptive name? 🤔

Tests/Packet++Test/Tests/PacketTests.cpp
```diff
@@ -327,9 +327,9 @@ PTF_TEST_CASE(RemoveLayerTest)
 	PTF_ASSERT_NULL(vxlanPacketOrig.getLayerOfType(pcpp::ICMP));
 
 	pcpp::Packet packetWithoutTunnel;
-	PTF_ASSERT_TRUE(packetWithoutTunnel.addLayer(vxlanEthLayer));
-	PTF_ASSERT_TRUE(packetWithoutTunnel.addLayer(vxlanIP4Layer));
-	PTF_ASSERT_TRUE(packetWithoutTunnel.addLayer(vxlanIcmpLayer));
+	PTF_ASSERT_TRUE(packetWithoutTunnel.addLayer(vxlanEthLayer, true));
+	PTF_ASSERT_TRUE(packetWithoutTunnel.addLayer(vxlanIP4Layer, true));
+	PTF_ASSERT_TRUE(packetWithoutTunnel.addLayer(vxlanIcmpLayer, true));
```
Dimi1010, line 332: Or should we keep the test as is, but correctly delete the layers afterwards?

seladb:
@Dimi1010 I'm not sure why replacing 2 protected members (`m_Packet`, `m_IsAllocatedInPacket`) with 1 private struct (`m_AllocationInfo`) that contains both of them improves anything... if we want to avoid using protected methods (which I'm not sure why it is better), we can just introduce `getAttachedPacket()`, but TBH I'm not sure this change is needed also

Dimi1010:
> @Dimi1010 I'm not sure why replacing 2 protected members (`m_Packet`, `m_IsAllocatedInPacket`) with 1 private struct (`m_AllocationInfo`) that contains both of them improves anything... [...]

@seladb It mitigates the risk of this issue https://github.com/seladb/PcapPlusPlus/pull/2004#discussion_r2462149580 which was a bug that can be hard to track and easy to happen and completely broke the memory ownership model.

Having them in the same struct indicates that the two variables are much more tightly related compared to let's say, `m_Packet` and `m_Data`. It makes it a bit harder to overwrite only one of the variables, forgetting to update the other.

If the allocation info needs to be updated, it is more likely to be done through behaviour actions (`m_AllocInfo.attachPacket()` / `detach()`) that conceptualize the entire operation or completely overwriting the allocation info object.

IMO, this is easier to understand and less bug-prone at the point of use, than having to remember to flip N variables manually every time the operation needs to be performed.

> [...] if we want to avoid using protected methods (which I'm not sure why it is better), we can just introduce `getAttachedPacket()`, but TBH I'm not sure this change is needed also

Protected data make it a pain to ensure invariants in the base classes of an inheritance hierarchy. That is especially true for large hierarchies like `Layer` [1]. Let's take the allocation info for example. It is solely managed by the `Layer` base class (+ Packet). Derived classes shouldn't modify it, because they shouldn't deal with allocation logic. They should just add parsing logic of to the data buffer.

If it was `protected` data, I can at **ANY POINT** in **ANY** derived class, write `m_AllocationInfo = AllocationInfo{}` or `[...] = AllocationInfo{ someOtherPacketPtr, true }`. Congratulations, the invariant is broken. Everything that relied on `m_AllocationInfo` being synchronized with other logic (e.g the actual packet that holds the layer) can go and burn. It is the reason why it was added in the C++ core guidelines in the segment ["avoid protected data"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-protected).

Making it private with RO accessors (if RO access is needed), changes the situation from "It can happen, but shouldn't." to "It can't happen". Removing the need to scan every code change for if a bug of such nature was introduced.

PS:
[1] - IMO the Layer hierarchy is somewhat too large already, as separate payload messages should not be represented via inheritance as they are not truly a new network protocol type. I'm looking at `HTTPRequest` and `HTTPRespose`, etc. IMO, it should be `HTTPLayer` with `HTTPRequestPayload[View]` / `HTTPResposePayload[View]`, you can bind to the `HTTPLayer` to mutate the data buffer.

seladb:
> @seladb It mitigates the risk of this issue [#2004 (comment)](https://github.com/seladb/PcapPlusPlus/pull/2004#discussion_r2462149580) which was a bug that can be hard to track and easy to happen and completely broke the memory ownership model.

There are very few places that need to update these members, and most (if not all) of them reside in the `Packet` class. So if there's a bug it should not be very hard to find and fix it

> Making it private with RO accessors (if RO access is needed), changes the situation from "It can happen, but shouldn't." to "It can't happen". Removing the need to scan every code change for if a bug of such nature was introduced.

I agree that layers shouldn't deal with the allocation logic, and I'm pretty sure none of them do. In practice, they never need to unless there's a very special use case. I guess we can leave the RO method although I don't think it's a real pain for people writing protocol parsers - I don't recall a case where I asked someone not to change these members...

Dimi1010:
> There are very few places that need to update these members, and most (if not all) of them reside in the `Packet` class. So if there's a bug it should not be very hard to find and fix it

It wouldn't have been that simple to find. I found it only after I grouped the variables and a test suddenly started memory leaking due to being incorrect. It was attaching the layer without ownership and then not deleting the layer. Everything still appeared to run fine because the packet deleted it erroneously.

Also, IMO that is only treating the symptom. In memory ownership schemes, generally keeping things simple at the point of use is key to prevent errors. Having to manually remember which variables to flip is not simple at the point of use, even if its only a few places. Having the complex actions encapsulated in methods is better due to explicit requirement for all parameters to be provided before executing the action.

> I agree that layers shouldn't deal with the allocation logic, and I'm pretty sure none of them do. In practice, they never need to unless there's a very special use case. I guess we can leave the RO method although I don't think it's a real pain for people writing protocol parsers - I don't recall a case where I asked someone not to change these members...

In practice, if the fields aren't needed by the derived classes then they shouldn't be exposed to them. I don't think we will have any special use cases where a derived class will need to change its owned packet from its internals. The only need they have for the packet is to forward the pointer to the next layer which is RO operation.

seladb:
> It wouldn't have been that simple to find. I found it only after I grouped the variables and a test suddenly started memory leaking due to being incorrect. It was attaching the layer without ownership and then not deleting the layer. Everything still appeared to run fine because the packet deleted it erroneously.

You're saying the bug wasn't exposed until you refactored the code and wrote a specific test? It's a good sign that it's an edge case that would rarely happen. We have many protocols using this infrastructure, and they all work fine.

That's another indication that this refactoring is not needed - "if it ain't broken don't fix it" 😄 

> In practice, if the fields aren't needed by the derived classes then they shouldn't be exposed to them. I don't think we will have any special use cases where a derived class will need to change its owned packet from its internals. The only need they have for the packet is to forward the pointer to the next layer which is RO operation.

I don't disagree with you, but I don't think it's "painful" enough that we need to modify this code

Dimi1010:
> You're saying the bug wasn't exposed until you refactored the code and wrote a specific test? It's a good sign that it's an edge case that would rarely happen. We have many protocols using this infrastructure, and they all work fine.
> 
> That's another indication that this refactoring is not needed - "if it ain't broken don't fix it" 😄

@seladb I am saying that the old test was written wrong and did not catch the bug in memory semantics. The bug **still existed**. The thing that happened was that the refactoring fixed the bug in the Layer's attach/detach procedure and the test started failing because it had **another** bug of not cleaning up the layers it detached.

It was a classic case of 2 bugs covering each other in that particular situation.

This is the test case: https://github.com/seladb/PcapPlusPlus/pull/2004/files#diff-995c851e4dd7dc72ed1ba9d8bf87403a50d00e69f92d31cfcab69577b42763a1L310-R332

```cpp
// The layers are detached from the packet here.
pcpp::EthLayer* vxlanEthLayer = (pcpp::EthLayer*)vxlanPacketOrig.detachLayer(pcpp::Ethernet, 1);
pcpp::IcmpLayer* vxlanIcmpLayer = (pcpp::IcmpLayer*)vxlanPacketOrig.detachLayer(pcpp::ICMP);
pcpp::IPv4Layer* vxlanIP4Layer = (pcpp::IPv4Layer*)vxlanPacketOrig.detachLayer(pcpp::IPv4, 1);

/* Assertion code irrelevant to the issue */
/* [...] */

// This is the test old code. The commented false is because that is the default value for take ownership.
// See how we instruct the attach operation to NOT take ownership of the layer when attaching it again.
PTF_ASSERT_TRUE(packetWithoutTunnel.addLayer(vxlanEthLayer, /* false */));
PTF_ASSERT_TRUE(packetWithoutTunnel.addLayer(vxlanIP4Layer, /* false */));
PTF_ASSERT_TRUE(packetWithoutTunnel.addLayer(vxlanIcmpLayer, /* false */));
```
This should cause the the test to memory leak, because the layers aren't cleaned up explicitly in the test afterwards.
This is a error in the test code.

Instead what happened was that the test runs fine without memory leaks, because the layer's ownership gets **erroneously** transferred anyway, because the flag `m_IsAllocatedInPacket`:
- on detach the flag wasn't properly reset, keeping its old value of `true`.
- on attach the flag is only updated if the second param is `true`. On false it is untouched, so it keeps its old value.

This leads to the flag keeping its old value of `true` after being attached. The packet instance thus considers that its responsible for it and cleans it up, even though it shouldn't.

In a correctly written test case that properly handled the cleanup of the layers, it would have lead to double free bug.

This attach / detach bug is very easily reproduced by having:
1. Make a layer
2. Attach the layer to a packet and have the packet take ownership of the layer
3. Detach the layer from the packet
4. Attack the layer to a packet and have the packet NOT take ownership.
5. Delete the packet
5.1 See how the packet still deletes the layer.
6. Attempt to use the layer and explode because that is now a dangling pointer, because Packet deleted the layer object.

Here is the sample reproducing the memory ownership bug. Run this on the current `dev` or `master`.
It only needs a change the visibility of `m_IsAllocatedInPacket` to `public` to compile as it can't be accessed otherwise.
```cpp
int main(int argc, char* argv[])
{
	using namespace pcpp;

	Packet* pkt = new Packet();
	Layer* layer = new ArpLayer(ArpReply(MacAddress(), IPv4Address(), MacAddress(), IPv4Address()));
	
	// Attach the layer to the packet, ownership should transfer to the packet
	pkt->addLayer(layer, true);
	std::cout << "This should be true:" << layer->m_IsAllocatedInPacket << std::endl;

	// Detach the layer from the packet, ownership should transfer to the user
	pkt->detachLayer(layer);
	std::cout << "This should be false:" << layer->m_IsAllocatedInPacket << std::endl;

	// Attach the layer to the packet, ownership should NOT transfer to the packet
	pkt->addLayer(layer, false);
	std::cout << "This should be false:" << layer->m_IsAllocatedInPacket << std::endl;

	// Delete packet. In the erogenous case this will also delete the layer object.
	delete pkt;

	// In the correct case, this should work, as the layer should still be alive.
	// In the erogenous case. This will dereference a dangling pointer and crash.
	layer->m_IsAllocatedInPacket;

	// This should be done by the user since ownership was never transferred back to the packet
	// In the erogenous case, it will cause a double free and crash.
	delete layer;

	return 0;
}
```

---
This entire issue is why I prefer my ownership mechanics to be somewhat encapsulated into simple objects / operations then relying on flipping flags manually.

It might be a somewhat uncommon case of detaching and reattaching a layer, but IMO it is unacceptable behaviour as it breaks the memory ownership model which is of critical importance.

seladb (approved):
Sure, we can go with this change 👍

PR #2005, opened 2025-10-25
Remove possibility of `setFilter` and `clearFilter` shadowing in derived devices.

Dimi1010:
## Summary

The PR focuses on streamlining the logic for setting device filters, and removing possibility of errors. It overhauls the `IFilterableDevice` to utilize a [Template Method Pattern](https://en.wikipedia.org/wiki/Template_method_pattern), instead of direct pure virtual methods. The cause for the change is the problematic shadowing of methods when a derived class declares a override a virtual method. It fixes one such issue that was left unmitigated in `PcapNG` type devices hiding `setFilter(GeneralFilter)`.

### The problem
The shadowing issue becomes relevant when a base class declares two or more overloads of a `virtual` method. If a derived class overrides only one of the overloads, the C++ language spec hides all other overloads from the derived class's scope by default.

For example:
```cpp
struct Base {
   virtual void foo(int x);
   virtual void foo(std::string x);
}

struct Derived : public Base {
   void foo(int x) override;
}

int main() {
  Derived d;
  static_cast<Base&>(d).foo(std::string{}); // <-This will compile. foo(std::string) is visible for the base class.
  d.foo(std::string{}); // <- This will not compile. foo(std::string) is shadowed.
}
```
Such issues are hard to spot due to their subtlety. 
Such was the case in `PcapNGFileReaderDevice` and `PcapNGFileWriterDevice` which only declared `setFilter(std::string)`, making the interface `setFilter(GenericFilter const&)` unavailable through the types. 

### Solutions
There are three possible ways to fix this issue:
1. Ensure all overloads are overridden in a derived class.
2. Ensure all derived classes that override at least one method have a `using Base::foo;` declaration.
3. Keep all overloaded methods as non-virtual and extract uniquely named virtual methods to be overridden for derived classes.

### Proposed solution
Previously, most of the devices utilized option 2, but that is brittle as it requires developer knowledge of the issue and scrutinous review as observed due to existing issues with PcapNG-type reader devices. 

This PR refactors the code to utilize option 3 as an alternative. This was selected due to the overload `setFilter(GeneralFilter)` cleanly delegating the majority of the work to `setFilter(std::string)`.

The solution follows a [Template Method Pattern](https://en.wikipedia.org/wiki/Template_method_pattern) by introducing a protected hook method `doUpdateFilter(std::string)` and keeping the public API non-virtual.

The change allows all calls from `setFilter` and `clearFilter` to be delegated to the device hook, while keeping the public API unchanged. The following change has the following benefits:
- Removes the possibility of shadowing `override` overloads, as they are no longer `virtual`. It is still possible to shadow a method by adding another overload in a derived class, but that situation is a lot more rare.
- Minimized the number of methods that need to be `overriden` for a specific device.
- Provides a common place that is fully controlled by the base class for operations that are shared between device implementations. (e.g. possibly a device open check)

### Behaviour changes
- `clearFilter` now mirrors `setFilter` and requires the device to be in an open state due to reusing the same filter hook. This differs from the previous implementation where `clearFilter` could possibly succeed depending on the device's implementation of the method.

### DIFF
--- a/Pcap++/header/KniDevice.h
+++ b/Pcap++/header/KniDevice.h
@@ -232,10 +232,10 @@ namespace pcpp
 	private:
 		/// All instances of this class MUST be produced by KniDeviceList class
 		KniDevice(const KniDeviceConfiguration& conf, size_t mempoolSize, int unique);
-		/// This class is not copyable
-		KniDevice(const KniDevice&);
-		/// This class is not copyable
-		KniDevice& operator=(const KniDevice&);
+
+		KniDevice(const KniDevice&) = delete;
+		KniDevice& operator=(const KniDevice&) = delete;
+
 		/// All instances of this class MUST be destroyed by KniDeviceList class
 		~KniDevice();
 
--- a/Pcap++/header/LinuxNicInformationSocket.h
+++ b/Pcap++/header/LinuxNicInformationSocket.h
@@ -29,6 +29,10 @@ namespace pcpp
 		/// Tries to open handled socket on construction.
 		/// If fails prints the debug message
 		LinuxNicInformationSocket();
+
+		LinuxNicInformationSocket(const LinuxNicInformationSocket&) = delete;
+		LinuxNicInformationSocket operator=(const LinuxNicInformationSocket&) = delete;
+
 		/// Closes handled socket on destruction.
 		/// If no socket was opened prints the debug message
 		~LinuxNicInformationSocket();
@@ -50,10 +54,6 @@ namespace pcpp
 		bool makeRequest(const char* nicName, const IoctlType ioctlType, ifreq* request);
 
 	private:
-		/// Hidden copy constructor. This structure is not copyable
-		LinuxNicInformationSocket(const LinuxNicInformationSocket&);
-		/// Hidden copy assignment operator. This structure is not copyable
-		LinuxNicInformationSocket operator=(const LinuxNicInformationSocket&);
 		LinuxSocket m_Socket;
 	};
 }  // namespace pcpp


Pcap++/header/Device.h
```diff
@@ -53,22 +53,38 @@ namespace pcpp
...
-		virtual bool setFilter(std::string filterAsString) = 0;
+		bool setFilter(std::string const& filterAsString)
+		{
+			return doUpdateFilter(filterAsString);
+		}
 
 		/// Clear the filter currently set on the device
 		/// @return True if filter was removed successfully or if no filter was set, false otherwise
-		virtual bool clearFilter() = 0;
+		bool clearFilter()
```
seladb, line 75: The problem as presented doesn't apply to `clearFilter()` because it only has one overload so nothing is shadowed. Why not keep it abstract?

Using `""` as an indication to clear the filter seems a bit weird to me. I think `clearFilter()` is more explicit

Dimi1010: > The problem as presented doesn't apply to `clearFilter()` because it only has one overload so nothing is shadowed. Why not keep it abstract?

Simplification of the internal interface mostly. Having a single extension point for a certain functionality makes it harder to forget about part of it when overriding. This had happened in a couple of places already.

In the old code `PcapNgFileReaderDevice` and `PcapNgFileWriterDevice` only overrode `setFilter` to target `m_BpfWrapper`. Calling `clearFilter` would have had no effect on the filter, since that still used the `IPcapDevice::clearFilter` implementation that worked on `m_PcapDescriptor`.

My reasoning was that both `setFilter` and `clearFilter` can be conceptualized by the primitive action "update the filter", just with different parameters. This was also a factor when deciding to make the hook as `doUpdateFilter` and not for example `doSetFilter`.

The single extension point makes the above issue impossible, since to override the filter mechanism you have to update the entire hook. There is nothing to forget to override.

Afterwards `clearFilter` was made non-virtual to remove the possibility of overriding the method, thus potentially bypassing the base implementation, which is undesirable. Under the design derived classes should only modify the hook implementation.

> Using `""` as an indication to clear the filter seems a bit weird to me. I think `clearFilter()` is more explicit

For public API, I agree, `clearFilter` is more user friendly.

However, I primary made the the hook signature to fit the aforementioned design goals. It wasn't designed to be exposed to the public.

The choice of `""` for "no filter" was due to `PcapHandle` and libpcap already treating it as `match any packet`, when passed to `pcap_compile` so I thought it might fit.

seladb: I agree that a single abstract method is simpler than two when it comes to overriding and making sure one doesn't accidentally forget to override the other, but when looking at the implementation, in which `""` symbolizes clearing a filter, I think that `clearFilter` is better. It is an internal API, but it might be used by anyone who wants to extend our packet capture capabilities (including our future selves 🙂 ). Using `doUpdateFilter("")` seems a bit rigid and might limit future functionality (for example - if a `""` is a valid filter in one of the packet engines).

I suggest we keep `clearFilter` an abstract method - it doesn't make things much more complex and at the same time provides better flexibility.

Dimi1010: Fair enough. I see the reasoning about  potentially `""` being valid filter.

Tbh, this is one of those times where C++17's `std::optional` would be really nice. We could have `doUpdateFilter(std::optional<std::string> filter)`.

Maybe we can have a simple struct that mimics it until we upgrade to 17? I would prefer to keep the number of virtual methods to a minimum if we can. 

`string const*` would techinially work too since we don't need to store the value (or we would copy it anyway), `nullptr` being no filter. It would be better from the no copy perspective compared to `optional`.

If we can't, I guess we can keep the `clearFilter` abstract until c++17.

Dimi1010: Updated `doUpdateFilter` to take `std::string const* filterStr` as an filter parameter, with `nullptr` being the "no-filter" value. This should allow `""` to be used in an implementation defined way, while minimising the virtual interface surface area.

46f0374fa7eac6e1b64c178f8c95b00de231649a

PR #2009, opened 2025-10-27
Specify queue id as a configuration parameter

### LINKED ISSUE #1992: Xdp Rx Queue

JasMetzger:
The queue id can be specified as an additional configuration parameter. It is checked against the maximum queues available by the device using a utility provided as a static function. In the open() function, the configured queue id is used in place of the hardwired zero value that was used before. Added a convenience function to obtain the queue id from device. 

Community may consider renaming the XdpDevice to XdpSocketDevice as each instance of the XdpDevice class pertains to a single queue id / socket.

### DIFF
--- a/Pcap++/header/XdpDevice.h
+++ b/Pcap++/header/XdpDevice.h
@@ -11,8 +11,6 @@
 /// @
 namespace pcpp
 {
-#define XDP_MAX_RXTX_QUEUES 16
-
 	/// @class XdpDevice
 	/// A class wrapping the main functionality of using AF_XDP (XSK) sockets
 	/// which are optimized for high performance packet processing.
@@ -268,7 +266,7 @@ namespace pcpp
 		/// @param[in] interfaceName The interface name to use to detect hardware queues
 		/// @param[in] tx  If true, return TX queues, otherwise RX. Default is false
 		/// @return The number of hardware queues associated with the device.
-		static uint32_t numQueues(const std::string& interfaceName, bool tx = false);
+		static uint32_t numOfHardwareQueues(const std::string& interfaceName, bool tx = false);
 
 	private:
 		class XdpUmem
--- a/Pcap++/header/XdpDevice.h
+++ b/Pcap++/header/XdpDevice.h
@@ -102,7 +102,8 @@ namespace pcpp
 			explicit XdpDeviceConfiguration(AttachMode attachMode = AutoMode, uint16_t umemNumFrames = 0,
 			                                uint16_t umemFrameSize = 0, uint32_t fillRingSize = 0,
 			                                uint32_t completionRingSize = 0, uint32_t rxSize = 0, uint32_t txSize = 0,
-			                                uint16_t rxTxBatchSize = 0, uint16_t frameHeadroomSize = 0, uint32_t queueId = 0)
+			                                uint16_t rxTxBatchSize = 0, uint16_t frameHeadroomSize = 0,
+			                                uint32_t queueId = 0)
 			{
 				this->attachMode = attachMode;
 				this->umemNumFrames = umemNumFrames;
@@ -255,10 +256,9 @@ namespace pcpp
 		/// @return Return queue identifier for underlying socket
 		uint32_t getQueueId() const
 		{
-			if(m_Config)
+			if (m_Config)
 			{
 				return m_Config->queueId;
-
 			}
 
 			return 0;
--- a/Pcap++/header/XdpDevice.h
+++ b/Pcap++/header/XdpDevice.h
@@ -259,6 +259,7 @@ namespace pcpp
 		}
 
 		/// Get number of RX or TX hardware queues for device
+		/// @param[in] interfaceName The interface name to use to detect hardware queues
 		/// @param[in] tx  If true, return TX queues, otherwise RX. Default is false
 		/// @return The number of hardware queues associated with the device. 
 		static uint32_t getNumQueues(const std::string& iface, bool tx = false);
--- a/Pcap++/header/XdpDevice.h
+++ b/Pcap++/header/XdpDevice.h
@@ -262,7 +262,7 @@ namespace pcpp
 		/// @param[in] interfaceName The interface name to use to detect hardware queues
 		/// @param[in] tx  If true, return TX queues, otherwise RX. Default is false
 		/// @return The number of hardware queues associated with the device. 
-		static uint32_t getNumQueues(const std::string& iface, bool tx = false);
+		static uint32_t getNumQueues(const std::string& interfaceName, bool tx = false);
 
 	private:
 		class XdpUmem
--- a/Pcap++/header/XdpDevice.h
+++ b/Pcap++/header/XdpDevice.h
@@ -78,6 +78,10 @@ namespace pcpp
 			/// The max number of packets to be received or sent in one batch
 			uint16_t rxTxBatchSize;
 
+			/// This parameter specifies the amount of space designated for prepending data to the packet in the frame.
+			/// NOTE: the headroom size should be less than the frame size
+			uint16_t frameHeadroomSize;
+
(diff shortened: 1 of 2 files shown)


Pcap++/header/XdpDevice.h
```diff
@@ -238,6 +246,24 @@ namespace pcpp
 		/// @return Current device statistics
 		XdpDeviceStats getStatistics();
 
+		/// @return Return queue identifier for underlying socket
+		uint32_t getQueueId()
```
Dimi1010: nit: I think this method can be marked as `const`?

Pcap++/src/XdpDevice.cpp
```diff
@@ -636,4 +648,36 @@ namespace pcpp
 		return m_Stats;
 	}
 
+	uint32_t XdpDevice::getNumQueues(const std::string& iface, bool tx)
+	{
+		// returns number of hardware queues associated with the device
+		uint32_t rxtxqueues = 0;
```
Dimi1010: nit: can we rename this to `numQueues`?

Pcap++/src/XdpDevice.cpp
```diff
@@ -636,4 +648,36 @@ namespace pcpp
...
+		std::string prefix = tx ? "tx-" : "rx-";
+		std::string path = "/sys/class/net/" + iface + "/queues/";
+		DIR *dir = opendir(path.c_str());
+
+		if(dir)
+		{
+			std::regex rxtx_regex("^" + prefix + "[0-9]+$");
+
+			struct dirent* entry;
```
Dimi1010, line 674: nit: Curious why the `struct dirent*`? Can't we just do `dirent*`?

Dimi1010 (commented):
Looks good overall. Some minor notes.

XDP CI is failing tho. If you could look into that. 🤔

seladb:
@JasMetzger, please notice that clang-format fails in CI

JasMetzger:
Sorry for the delay--- I was out of town. Where is this test failing?

seladb:
> Sorry for the delay--- I was out of town. Where is this test failing?

@JasMetzger you can see the pre-commit build that failed: https://github.com/seladb/PcapPlusPlus/actions/runs/19214326473/job/54921310413?pr=2009

There are trailing white-spaces and formatting issues. To fix formatting, you should run clang-format as mentioned in https://github.com/seladb/PcapPlusPlus/blob/master/CONTRIBUTING.md

JasMetzger:
When I run the pre-commit, there are errors in files that I did not modify. Shall I address those?

seladb:
> When I run the pre-commit, there are errors in files that I did not modify. Shall I address those?

@JasMetzger there shouldn't be errors in files you did not modify, please avoid updating them in this PR

seladb:
@JasMetzger should we move this PR to DRAFT until it's ready for review?
The CI is currently failing...

JasMetzger:
When I do a clean make, there are warnings in files I haven't touched, so not sure what to do about that code. The first warning is freebsd running bash but I don't understand how it is failing BSD or what it is doing. Do I need to create a FreeBSD environment? Is it compiling it? I think I need some help with the CI. I got it to pass clang.

seladb:
There might be issues with the FreeBSD tests, I need to look into it... 😕

Pcap++/header/XdpDevice.h
```diff
@@ -11,6 +11,8 @@
 /// @
 namespace pcpp
 {
+#define XDP_MAX_RXTX_QUEUES 16
```
seladb: This const isn't used anywhere, should we remove it?

Pcap++/src/XdpDevice.cpp
```diff
@@ -403,8 +406,8 @@ namespace pcpp
 		auto umemInfo = static_cast<xsk_umem_info*>(m_Umem->getInfo());
 
 		struct xsk_socket_config xskConfig;
-		xskConfig.rx_size = m_Config->txSize;
-		xskConfig.tx_size = m_Config->rxSize;
+		xskConfig.rx_size = m_Config->rxSize;
+		xskConfig.tx_size = m_Config->txSize;
```
seladb, line 411: This was already fixed in https://github.com/seladb/PcapPlusPlus/pull/2030

Pcap++/header/XdpDevice.h
```diff
@@ -238,11 +253,28 @@ namespace pcpp
...
+			}
+
+			return 0;
+		}
+
+		/// Get number of RX or TX hardware queues for device
+		/// @param[in] interfaceName The interface name to use to detect hardware queues
+		/// @param[in] tx  If true, return TX queues, otherwise RX. Default is false
+		/// @return The number of hardware queues associated with the device.
+		static uint32_t numQueues(const std::string& interfaceName, bool tx = false);
```
seladb: Why is this a public method? Shouldn't it be private?

Pcap++/src/XdpDevice.cpp
```diff
@@ -499,13 +511,23 @@ namespace pcpp
 			return false;
 		}
 
+		unsigned int nhwqueues = numQueues(m_InterfaceName);
```
seladb: nit: let's rename to `numOfHardwareQueues`?

Pcap++/header/XdpDevice.h
JasMetzger: I will remove. I think at first I was simply setting a hard limit before I created a function that can detect it. If the function cannot detect it, it returns 0 which may not be desirable. Maybe it should return 1 if nothing else. That will be the identical behavior in the release version.

Pcap++/src/XdpDevice.cpp
JasMetzger: Will that be a problem if I leave it like this? A merge issue?

Pcap++/header/XdpDevice.h
JasMetzger: It may be useful for configuring the user space, such as limiting the number of threads and resources.

Pcap++/src/XdpDevice.cpp
JasMetzger: Will do.

Pcap++/header/XdpDevice.h
JasMetzger: I made the default number 1. So a qid of zero, the default, and as things worked before, will pass the test.

JasMetzger:
Redirecting a packet to a particular rxqueue and associated socket requires a "map" between kernel and userspace. Entries in the map are created on the user space side using the socket's file descriptor.  Unless there is a different way, I am considering a function in XdpDevice that passes in a pointer to int for the file descriptor to be written. The function returns a bool, true if the fd was set ok, false otherwise.

seladb:
> Redirecting a packet to a particular rxqueue and associated socket requires a "map" between kernel and userspace. Entries in the map are created on the user space side using the socket's file descriptor. Unless there is a different way, I am considering a function in XdpDevice that passes in a pointer to int for the file descriptor to be written. The function returns a bool, true if the fd was set ok, false otherwise.

@JasMetzger do you have a code example somewhere that shows how it should be done in AF_XDP?
How should this map be created? How do you pass it to the kernel?
I'm not familiar with this area, so providing more context would definitely help.

PR #2019, opened 2025-11-09
Add support for WinDivert packet capture engine

seladb:
This PR integrates WinDivert as another packet capture engine in PcapPlusPlus, enabling packet capture/injection on Windows via the WinDivert driver.

### What is WinDivert

WinDivert is an open-source Windows library (kernel + user-mode) that allows applications to intercept, modify, drop or inject network packets traversing the Windows network stack. It is designed for use cases such as packet sniffing, firewalling, NAT-/VPN-tunneling, loopback traffic inspection, etc. 

Key features include:
- Capturing both inbound and outbound packets (and loopback) on Windows 7/8/10/11. 
- Support for IPv4 and IPv6, and a simple filtering language. 
- User-mode API (via windivert.h / WinDivert.dll) that interacts with a kernel-mode driver. 

### Project Links

- Project website: https://reqrypt.org/windivert.html
- GitHub repository: https://github.com/basil00/WinDivert
- API / documentation: https://reqrypt.org/windivert-doc.html

### Testing

This PR includes basic tests for the `WinDivertDevice`. However, it also adds a lightweight abstraction over the WinDivert API using internal interfaces. It enables testing `WinDivertDevice` logic without the real driver by providing mock implementations. These mock tests aren't implemented in this PR, but can be added later.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
         args: ['--fix=lf']
       - id: trailing-whitespace
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.3
+    rev: v0.14.7
     hooks:
       - id: ruff # Run the linter.
         types_or: [ python ]
@@ -38,7 +38,7 @@ repos:
       - id: cppcheck
         args: ["--std=c++14", "--language=c++", "--suppressions-list=cppcheckSuppressions.txt", "--inline-suppr", "--force"]
   - repo: https://github.com/BlankSpruce/gersemi
-    rev: 0.22.3
+    rev: 0.23.2
     hooks:
     - id: gersemi
       args: ["-c"]
@@ -54,6 +54,6 @@ repos:
         args: ['--config=typos-config.toml']
         pass_filenames: false
   - repo: https://github.com/lovesegfault/beautysh
-    rev: v6.4.1
+    rev: v6.4.2
     hooks:
       - id: beautysh
--- a/3rdParty/OUIDataset/PCPP_OUIDataset.json
+++ b/3rdParty/OUIDataset/PCPP_OUIDataset.json
@@ -28920,7 +28920,7 @@
         "vendor": "Shenzhen Huapu Digital Co., Ltd"
     },
     "9674": {
-        "vendor": "Laird Connectivity"
+        "vendor": "Ezurio, LLC"
     },
     "9675": {
         "vendor": "Reiner SCT"
@@ -33482,6 +33482,9 @@
     "30925": {
         "vendor": "Ignition Design Labs"
     },
+    "31396": {
+        "vendor": "FRITZ! Technology GmbH"
+    },
     "31512": {
         "vendor": "SENTRY Co., LTD."
     },
@@ -34322,6 +34325,9 @@
     "35002": {
         "vendor": "Nc&C"
     },
+    "35273": {
+        "vendor": "Extreme Networks Headquarters"
+    },
     "35413": {
         "vendor": "Huawei Device Co., Ltd."
     },
@@ -35137,6 +35143,9 @@
     },
     "37370": {
         "vendor": "Synapse Product Development"
+    },
+    "37429": {
+        "vendor": "Apple, Inc."
     },
     "37501": {
         "vendor": "Ficosa Internationa(Taicang) C0.,Ltd."
@@ -39173,6 +39182,9 @@
     "267737": {
         "vendor": "Viwone"
     },
+    "268108": {
+        "vendor": "Nanjing SCIYON Wisdom Technology Group Co.,Ltd."
+    },
     "268214": {
         "vendor": "Smart Innovation LLC"
     },
@@ -39319,6 +39331,9 @@
     },
     "275651": {
         "vendor": "Qingdao Goertek Horizons Tecnology Co.,LTD"
+    },
+    "275663": {
+        "vendor": "Apple, Inc."
     },
     "275702": {
         "vendor": "Motorola (Wuhan) Mobility Technologies Communication Co., Ltd."
@@ -39884,6 +39899,9 @@
     "308478": {
         "vendor": "AVM Audiovisuelles Marketing und Computersysteme GmbH"
     },
+    "308658": {
+        "vendor": "Apple, Inc."
+    },
     "308673": {
         "vendor": "Itel Mobile Limited"
     },
@@ -40006,6 +40024,9 @@
     },
     "313817": {
         "vendor": "Dish Technologies Corp"
+    },
+    "313822": {
+        "vendor": "Qingdao HaierTechnology Co.,Ltd"
     },
     "313997": {
         "vendor": "Enfabrica"
@@ -41885,11 +41906,17 @@
     "669212": {
         "vendor": "Smardii"
     },
+    "678139": {
+        "vendor": "Purple WiFi Limited"
+    },
     "689974": {
         "vendor": "IEEE 1901 Working Group"
     },
     "691810": {
         "vendor": "Delta Solutions LLC"
+    },
+    "693030": {
+        "vendor": "Qorvo, Inc."
     },
     "696891": {
         "vendor": "Vitex LLC"
@@ -41950,6 +41977,9 @@
     },
     "790134": {
         "vendor": "D-Link International"
+    },
+    "790219": {
+        "vendor": "Huawei Device Co., Ltd."
     },
     "790789": {
         "vendor": "Akuvox (Xiamen) Networks Co., Ltd"
@@ -42473,6 +42503,9 @@
     "821106": {
         "vendor": "Fujian Star-Net Communication Co.,Ltd"
     },
+    "821298": {
+        "vendor": "Nokia Solutions and Networks India Private Limited"
+    },
     "821520": {
         "vendor": "Samsung Electronics Co.,Ltd"
     },
@@ -42924,7 +42957,7 @@
         "vendor": "Realme Chongqing Mobile Telecommunications Corp.,Ltd."
     },
     "845577": {
-        "vendor": "Fox Crypto B.V."
+        "vendor": "Sentyron B.V"
     },
     "845605": {
         "vendor": "Microsoft Corporation"
@@ -42988,6 +43021,9 @@
     },
     "848052": {
         "vendor": "Globalsat International Technology Ltd"
+    },
+    "848629": {
+        "vendor": "Sichuan AI-Link Technology Co., Ltd."
     },
     "848710": {
         "vendor": "Xiaomi Communications Co Ltd"
@@ -43164,7 +43200,7 @@
         "vendor": "Hangzhou Hikvision Digital Technology Co.,Ltd."
     },
     "1053489": {
-        "vendor": "Technicolor Delivery Technologies Belgium NV"
+        "vendor": "Vantiva Technologies Belgium"
     },
     "1053678": {
         "vendor": "Justec International Technology INC."
@@ -43718,6 +43754,9 @@
     "1083598": {
         "vendor": "Fiberhome Telecommunication Technologies Co.,LTD"
     },
+    "1083603": {
+        "vendor": "Huawei Technologies Co.,Ltd"
+    },
     "1083899": {
         "vendor": "Samsung Electronics Co.,Ltd"
     },
@@ -43966,6 +44005,9 @@
     },
     "1096691": {
         "vendor": "Hunan Fn-Link Technology Limited"
+    },
+    "1096758": {
+        "vendor": "Huawei Device Co., Ltd."
     },
     "1096855": {
         "vendor": "vivo Mobile Communication Co., Ltd."
@@ -44414,6 +44456,9 @@
     "1316650": {
         "vendor": "Fiberhome Telecommunication Technologies Co.,LTD"
     },
+    "1316902": {
+        "vendor": "Nokia"
+    },
     "1316932": {
         "vendor": "Xenon Smart Teknoloji Ltd."
     },
@@ -44669,6 +44714,9 @@
     "1330202": {
         "vendor": "Max Communication GmbH"
     },
+    "1330431": {
+        "vendor": "Grandstream Networks, Inc."
+    },
     "1330535": {
         "vendor": "Zioncom Electronics (Shenzhen) Ltd."
     },
@@ -44758,6 +44806,9 @@
     },
     "1334889": {
         "vendor": "Guangdong Oppo Mobile Telecommunications Corp.,Ltd"
+    },
+    "1334972": {
+        "vendor": "Huawei Technologies Co.,Ltd"
     },
     "1335188": {
         "vendor": "Huawei Technologies Co.,Ltd"
@@ -45266,6 +45317,9 @@
     "1365581": {
         "vendor": "D-Link International"
     },
+    "1365628": {
+        "vendor": "Uncord Technologies Private Limited"
+    },
     "1365797": {
         "vendor": "Barrot Technology Co.,Ltd."
     },
@@ -45553,6 +45607,9 @@
     },
     "1581694": {
         "vendor": "Samsung Electronics Co.,Ltd"
+    },
+    "1582137": {
+        "vendor": "Yippee Electronics Cp.,Limited"
     },
     "1582665": {
         "vendor": "Intel Corporate"
@@ -45856,6 +45913,9 @@
     },
     "1596339": {
         "vendor": "Samsung Electronics Co.,Ltd"
+    },
+    "1596577": {
+        "vendor": "Jiangxi Risound Electronics Co.,LTD"
     },
     "1596783": {
         "vendor": "N3com"
@@ -46274,6 +46334,9 @@
     "1619345": {
         "vendor": "I-Storm"
     },
+    "1619543": {
+        "vendor": "Huawei Technologies Co.,Ltd"
+    },
     "1619660": {
         "vendor": "We Corporation Inc."
     },
@@ -46344,7 +46407,7 @@
         "vendor": "SonicWall"
     },
     "1622675": {
-        "vendor": "Laird Connectivity"
+        "vendor": "Ezurio, LLC"
     },
     "1622719": {
         "vendor": "Buffalo.Inc"
@@ -46630,6 +46693,9 @@
     },
     "1835545": {
         "vendor": "Guangdong Oppo Mobile Telecommunications Corp.,Ltd"
+    },
+    "1836128": {
+        "vendor": "NXP Semiconductors Taiwan Ltd."
     },
     "1836471": {
         "vendor": "Chongqing Trantor Technology Co., Ltd."
@@ -47141,6 +47207,9 @@
     "1863714": {
         "vendor": "Murata Manufacturing Co., Ltd."
     },
+    "1863765": {
+        "vendor": "Huawei Technologies Co.,Ltd"
+    },
     "1863881": {
         "vendor": "Jiangsu Aisida Electronic Co., Ltd"
     },
@@ -47270,11 +47339,17 @@
     "1870831": {
         "vendor": "Beijing Xiaomi Electronics Co.,Ltd"
     },
+    "1871402": {
+        "vendor": "Apple, Inc."
+    },
     "1871452": {
         "vendor": "Huawei Technologies Co.,Ltd"
     },
     "1871502": {
         "vendor": "DB Communication & Systems Co., ltd."
+    },
+    "1871590": {
+        "vendor": "Vtech Telecommunications Limited"
     },
     "1871754": {
         "vendor": "Phase Motion Control SpA"
@@ -47539,6 +47614,9 @@
     },
     "1887825": {
         "vendor": "AzureWave Technology Inc."
+    },
+    "1888130": {
+        "vendor": "Palo Alto Networks"
     },
     "1888519": {
         "vendor": "Realme Chongqing Mobile Telecommunications Corp.,Ltd."
@@ -48528,7 +48606,7 @@
         "vendor": "SteelSeries ApS"
     },
     "2142209": {
-        "vendor": "Technicolor Delivery Technologies Belgium NV"
+        "vendor": "Vantiva Technologies Belgium"
     },
     "2142455": {
         "vendor": "Enclustra GmbH"
@@ -48550,6 +48628,9 @@
     },
     "2144299": {
         "vendor": "Sagemcom Broadband SAS"
+    },
+    "2144317": {
+        "vendor": "Unionman Technology Co.,Ltd"
     },
     "2144360": {
         "vendor": "Motorola Mobility LLC, a Lenovo Company"
@@ -48716,6 +48797,9 @@
     "2154652": {
         "vendor": "Nokia"
     },
+    "2154845": {
+        "vendor": "TP-Link Systems Inc."
+    },
     "2155176": {
         "vendor": "Apple, Inc."
     },
@@ -48730,6 +48814,9 @@
     },
     "2155876": {
         "vendor": "Commscope"
+    },
+    "2155931": {
+        "vendor": "Panasonic Automotive Systems"
     },
     "2156255": {
         "vendor": "eero inc."
@@ -48986,6 +49073,9 @@
     "2365810": {
         "vendor": "Quectel Wireless Solutions Co.,Ltd."
     },
+    "2365861": {
+        "vendor": "New H3C Technologies Co., Ltd"
+    },
     "2366092": {
         "vendor": "Squarehead Technology AS"
     },
@@ -49226,6 +49316,12 @@
     "2380735": {
         "vendor": "Enernet"
     },
+    "2380781": {
+        "vendor": "Dell Inc."
+    },
+    "2381210": {
+        "vendor": "Apple, Inc."
+    },
     "2381934": {
         "vendor": "zte corporation"
     },
@@ -49303,6 +49399,9 @@
     },
     "2384555": {
         "vendor": "Espressif Inc."
+    },
+    "2384582": {
+        "vendor": "Huawei Device Co., Ltd."
     },
     "2384590": {
         "vendor": "Hewlett Packard Enterprise"
@@ -49465,6 +49564,9 @@
     },
     "2392714": {
         "vendor": "Prowave Technologies Ltd."
+    },
+    "2392778": {
+        "vendor": "vivo Mobile Communication Co., Ltd."
     },
     "2393240": {
         "vendor": "Beijing Jiaoda Microunion Tech.Co.,Ltd."
@@ -49811,6 +49913,9 @@
     "2415286": {
         "vendor": "Sistemas de Gestión Energética S.A. de C.V"
     },
+    "2415508": {
+        "vendor": "Juniper Networks"
+    },
     "2415532": {
         "vendor": "Huawei Technologies Co.,Ltd"
     },
@@ -49959,7 +50064,7 @@
         "vendor": "Hewlett Packard Enterprise"
     },
     "2421469": {
-        "vendor": "Radiant Zemax LLC"
+        "vendor": "Radiant Vision Systems, LLC"
     },
     "2421510": {
         "vendor": "Itel Mobile Limited"
@@ -50158,6 +50263,9 @@
     },
     "2628897": {
         "vendor": "In One Smart Technology(H,K,)Limited"
+    },
+    "2629034": {
+        "vendor": "ASTI India Private Limited"
     },
     "2629115": {
         "vendor": "Huawei Technologies Co.,Ltd"
@@ -50822,6 +50930,9 @@
     "2668008": {
         "vendor": "Texas Instruments"
     },
+    "2668156": {
+        "vendor": "KEBODA Intelligent TECHNOLOGY CO., LTD."
+    },
     "2668585": {
         "vendor": "Juniper Networks"
     },
@@ -51113,6 +51224,9 @@
     "2681389": {
         "vendor": "Apple, Inc."
     },
+    "2681435": {
+        "vendor": "Samsara Networks Inc"
+    },
     "2681610": {
         "vendor": "Rolling Wireless S.a.r.l. Luxembourg"
     },
@@ -51235,6 +51349,9 @@
     },
     "2764675": {
         "vendor": "Uplink"
+    },
+    "2764770": {
+        "vendor": "onway ag"
     },
     "2770889": {
         "vendor": "Code Construct Pty Ltd"
@@ -51463,6 +51580,9 @@
     },
     "2895176": {
         "vendor": "Commend International GmbH"
+    },
+    "2895860": {
+        "vendor": "eero inc."
     },
     "2895898": {
         "vendor": "Technicolor CH USA Inc for Telus"
@@ -52496,6 +52616,9 @@
     "3149214": {
         "vendor": "Ruijie Networks Co.,LTD"
     },
+    "3149379": {
+        "vendor": "Apple, Inc."
+    },
     "3149496": {
         "vendor": "LG Electronics"
     },
@@ -53079,7 +53202,7 @@
         "vendor": "Skyworth Digital Technology(Shenzhen) Co.,Ltd"
     },
     "3182991": {
-        "vendor": "Technicolor Delivery Technologies Belgium NV"
+        "vendor": "Vantiva Technologies Belgium"
     },
     "3183350": {
         "vendor": "Shanghai Sunmon Communication Technogy Co.,Ltd"
@@ -53321,6 +53444,9 @@
     "3197750": {
         "vendor": "Belden Singapore Pte. Ltd."
     },
+    "3197833": {
+        "vendor": "OnLogic Inc"
+    },
     "3197895": {
         "vendor": "Cambium Networks Limited"
     },
@@ -53477,6 +53603,9 @@
     "3206268": {
         "vendor": "Shenzhen Along Electronics Co., Ltd"
     },
+    "3206307": {
+        "vendor": "Alfatron Electronics INC"
+    },
     "3206550": {
         "vendor": "LS Mecapion"
     },
@@ -53602,6 +53731,9 @@
     },
     "3408539": {
         "vendor": "Plexonics Technologies LImited"
+    },
+    "3408540": {
+        "vendor": "D-Link Corporation"
     },
     "3408862": {
         "vendor": "Texas Instruments"
@@ -53977,6 +54109,9 @@
     },
     "3429780": {
         "vendor": "Fujian Star-Net Communication Co.,Ltd"
+    },
+    "3429861": {
+        "vendor": "SJIT Co., Ltd."
     },
     "3430142": {
         "vendor": "Cisco Meraki"
@@ -55118,6 +55253,9 @@
     "3682732": {
         "vendor": "Weg"
     },
+    "3683269": {
+        "vendor": "Microsoft Corporation"
+    },
     "3683835": {
         "vendor": "Sagemcom Broadband SAS"
     },
@@ -55177,6 +55315,9 @@
     },
     "3687397": {
         "vendor": "Grotech Inc"
+    },
+    "3687614": {
+        "vendor": "Espressif Inc."
     },
     "3687739": {
         "vendor": "Ruckus Wireless"
@@ -55886,6 +56027,9 @@
     "3730247": {
         "vendor": "Huawei Technologies Co.,Ltd"
     },
+    "3730439": {
+        "vendor": "Motorola Mobility LLC, a Lenovo Company"
+    },
     "3730445": {
         "vendor": "Apple, Inc."
     },
@@ -55999,6 +56143,9 @@
     },
     "3734292": {
         "vendor": "Huawei Technologies Co.,Ltd"
+    },
+    "3734432": {
+        "vendor": "Shenzhen Baseus Technology CoLtd"
     },
     "3734580": {
         "vendor": "Huawei Device Co., Ltd."
@@ -56231,6 +56378,9 @@
     "3938974": {
         "vendor": "VitalThings AS"
     },
+    "3939020": {
+        "vendor": "Quectel Wireless Solutions Co.,Ltd."
+    },
     "3939320": {
         "vendor": "Hangzhou Hikvision Digital Technology Co.,Ltd."
     },
@@ -56284,6 +56434,9 @@
     },
     "3942566": {
         "vendor": "ALE International"
+    },
+    "3943091": {
+        "vendor": "Telesystem communications Pte Ltd"
     },
     "3943156": {
         "vendor": "Brother Industries, LTD."
@@ -56437,6 +56590,9 @@
     },
     "3950903": {
         "vendor": "ASSMANN Electronic GmbH"
+    },
+    "3951103": {
+        "vendor": "Unionman Technology Co.,Ltd"
     },
     "3951250": {
         "vendor": "Hewlett Packard"
@@ -56674,6 +56830,9 @@
     },
     "3965098": {
         "vendor": "Ransnet Singapore Pte Ltd"
+    },
+    "3965225": {
+        "vendor": "Inheco Industrial Heating and Cooling GmbH"
     },
     "3965400": {
         "vendor": "Sagemcom Broadband SAS"
@@ -57194,6 +57353,9 @@
     "3993484": {
         "vendor": "Zhejiang Dahua Technology Co., Ltd."
     },
+    "3993509": {
+        "vendor": "Cloud Network Technology Singapore Pte. Ltd."
+    },
     "3993617": {
         "vendor": "Intel Corporate"
     },
@@ -57257,6 +57419,9 @@
     "3996288": {
         "vendor": "Huawei Technologies Co.,Ltd"
     },
+    "3996418": {
+        "vendor": "Apple, Inc."
+    },
     "3996508": {
         "vendor": "Fiberhome Telecommunication Technologies Co.,LTD"
     },
@@ -57307,6 +57472,9 @@
     },
     "4196471": {
         "vendor": "Xiaomi Communications Co Ltd"
+    },
+    "4197095": {
+        "vendor": "BSH Hausgeräte GmbH"
     },
     "4197648": {
         "vendor": "Commscope"
@@ -57496,6 +57664,9 @@
     },
     "4208557": {
         "vendor": "Macro Image Technology, Inc."
+    },
+    "4208642": {
+        "vendor": "Silicon Laboratories"
     },
     "4209531": {
         "vendor": "Huawei Device Co., Ltd."
@@ -58538,6 +58709,9 @@
     "4463180": {
         "vendor": "xFusion Digital Technologies Co.,Ltd."
     },
+    "4463196": {
+        "vendor": "Cisco Systems, Inc"
+    },
     "4463236": {
         "vendor": "Quectel Wireless Solutions Co.,Ltd."
     },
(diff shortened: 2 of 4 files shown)


Pcap++/src/WinDivertDevice.cpp
```diff
@@ -0,0 +1,726 @@
...
+				}
+
+				DWORD packetLen = 0;
+				if (GetOverlappedResult(winDivertHandle->get(), &m_Overlapped, &packetLen, FALSE))
+				{
+					return { OverlappedResult::Status::Success, static_cast<uint32_t>(packetLen), 0 };
+				}
+
+				return { OverlappedResult::Status::Failed, 0, static_cast<uint32_t>(GetLastError()) };
+			}
```
Dimi1010: As far as I know, Win32 `GetOverlappedResult` does not modify anything in the result structure, so I think the whole operation can be `const`.

Also, wouldn't a reference param be better instead of pointer? We don't want `nullptr` handles being passed?

seladb: `GetOverlappedResult ` updates `m_Overlapped` so I guess it shouldn't be made const, no? 🤔 

This method gets a pointer because we want to be able to cast it to `WinDivertHandle`

.github/workflows/build_and_test.yml
```diff
@@ -289,12 +289,6 @@ jobs:
           cd Tests/Packet++Test
           Bin/Packet++Test
 
-      - name: Test Pcap++
-        if: env.avx512 == 'true'
-        run: |
-          cd Tests/Pcap++Test
-          Bin/Pcap++Test
```
seladb: This is not directly related to the PR, but CI failed a few times - probably because `env.avx512 == 'true'` and `Bin/Pcap++Test` cannot run without the `-i` argument. I figured that running it is not needed because it can't run the DPDK tests anyway

Tests/Pcap++Test/Tests/FilterTests.cpp
```diff
@@ -106,9 +106,9 @@ PTF_TEST_CASE(TestPcapFiltersLive)
 	andFilter.parseToString(filterAsString);
 	PTF_ASSERT_TRUE(liveDev->setFilter(andFilter));
 	PTF_ASSERT_TRUE(liveDev->startCapture(capturedPackets));
-	PTF_ASSERT_TRUE(sendURLRequest("www.walla.co.il"));
+	PTF_ASSERT_TRUE(sendURLRequest("www.google.com"));
 	// let the capture work for couple of seconds
-	totalSleepTime = incSleep(capturedPackets, 2, 7);
+	totalSleepTime = incSleep(capturedPackets, 2, 20);
```
seladb, line 111: This test also failed in CI so I made it more robust

ci/run_tests/run_tests_windows.py
```diff
@@ -72,6 +72,14 @@ def main():
...
     )
+    parser.add_argument(
+        "--include-tests",
+        "-t",
+        type=str,
+        nargs="+",
+        default=[],
+        help="Pcap++ tests to include",
+    )
     parser.add_argument(
```
seladb, line 83: This change was needed to support running only WinDivert tests in the `windivert` job in CI

Pcap++/src/WinDivertDevice.cpp
Dimi1010: > GetOverlappedResult  updates m_Overlapped so I guess it shouldn't be made const, no? 

Sure, we can leave it non-const then.

> This method gets a pointer because we want to be able to cast it to WinDivertHandle

We can also cast to `WinDivertHandle` if we take a reference, as those are technically a pointer under the hood.
```cpp
OverlappedResult getOverlappedResult(const IWinDivertHandle& handle) override
{
	auto winDivertHandle = dynamic_cast<const WinDivertHandle*>(&handle);			
	if (winDivertHandle == nullptr)
	{
		throw std::runtime_error("Failed to get WinDivertHandle");
	}

	/* ... */
}
```

Also if we are just throwing an exception on cast failure can't we just use:
```cpp
auto& winDivertHandle = dynamic_cast<const WinDivertHandle&>(handle);
```
This has a built-in `std::bad_cast` throw if the cast fails, which IMO is more descriptive than `std::runtime_error`?

.github/workflows/build_and_test.yml
Dimi1010: Can we update / remove `Test skipped` message in the next step then?

ci/install_windivert.bat
Dimi1010, line 1: Curious can't we write the entire script in powershell instead of having batch script that runs an embedded powershell script?
GH actions do support running a step directly in powershell.

Pcap++/header/WinDivertDevice.h
```diff
@@ -0,0 +1,408 @@
...
+	namespace internal
+	{
+		/// @brief Opaque handle wrapper for an opened WinDivert driver instance.
+		///
+		/// Implementations encapsulate the native WinDivert handle and its lifetime.
+		class IWinDivertHandle
+		{
+		public:
+			virtual ~IWinDivertHandle() = default;
+		};
```
Dimi1010: Do we plan on having multiple handle implementations?

If not, can't we just use an opaque pointer instead?

Pcap++/header/WinDivertDevice.h
```diff
@@ -0,0 +1,408 @@
...
+			                                               uint64_t flags) = 0;
+			virtual uint32_t close(const IWinDivertHandle* handle) = 0;
+			virtual uint32_t recvEx(const IWinDivertHandle* handle, uint8_t* buffer, uint32_t bufferLen,
+			                        size_t addressesSize, IOverlappedWrapper* overlapped) = 0;
+			virtual std::vector<WinDivertAddress> recvExComplete() = 0;
+			virtual uint32_t sendEx(const IWinDivertHandle* handle, uint8_t* buffer, uint32_t bufferLen,
+			                        size_t addressesSize) = 0;
+			virtual std::unique_ptr<IOverlappedWrapper> createOverlapped() = 0;
+			virtual bool getParam(const IWinDivertHandle* handle, WinDivertParam param, uint64_t& value) = 0;
+			virtual bool setParam(const IWinDivertHandle* handle, WinDivertParam param, uint64_t value) = 0;
```
Dimi1010: ditto: Windivert handle pointer -> reference. The API is meaningless for `nullptr` handle.

Dimi1010: Also, can't those functions be member functions of the `IWinDivertHandle` wrapper instead? That way we don't have to dynamic cast inside this class's implementors.

Pcap++/header/WinDivertDevice.h
Dimi1010: Withdrawn in favor of https://github.com/seladb/PcapPlusPlus/pull/2019#discussion_r2522359794

Pcap++/header/WinDivertDevice.h
Dimi1010: Expanding on this, I think restructuring `IWinDivertImplementation` and `IWinDivertHandle` would be beneficial for code performance and readability.

- `IWinDivertImplementation` should just be a stateless factory for `IWinDivertHandle` objects.
Maybe `getNetworkInterfaces` should be kept in there too, as that seems non-specific to the handles. The change would allow a single factory object to be shared between device instances.

- `IWinDivertHandle` changes:
  - Operations directly applied to the handle should be transferred to the object (e.g. `recvEx`, `sendEx`, `setParam`, `getParam`, etc...). This would simplify the interface and remove the need to downcast inside `IWinDivertImplementation`, which is slow.
  - The current member `IWinDivertImplementation::close` should be removed. The handle would be closed by deleting the `IWinDivertHandle` object in RAII pattern.

Pcap++/header/WinDivertDevice.h
```diff
@@ -0,0 +1,408 @@
...
+	/// getWinDivertTimestamp() respectively.
+	class WinDivertRawPacket : public RawPacket
+	{
+	public:
+		WinDivertRawPacket(const uint8_t* pRawData, int rawDataLen, timespec timestamp, bool deleteRawDataAtDestructor,
+		                   LinkLayerType layerType, uint32_t interfaceIndex, uint64_t winDivertTimestamp)
+		    : RawPacket(pRawData, rawDataLen, timestamp, deleteRawDataAtDestructor, layerType),
+		      m_InterfaceIndex(interfaceIndex), m_WinDivertTimestamp(winDivertTimestamp)
+		{}
+
```
Dimi1010, line 217: nit:
```suggestion

		~WinDivertRawPacket() override = default;

```

Pcap++/src/WinDivertDevice.cpp
seladb: > We can also cast to `WinDivertHandle` if we take a reference, as those are technically a pointer under the hood.

I guess we can but it's not so nice to do `dynamic_cast<const WinDivertHandle*>(&handle);` 🤔 

> This has a built-in `std::bad_cast` throw if the cast fails, which IMO is more descriptive than `std::runtime_error`?

I think throwing a custom exception with a custom message is nicer, no?

.github/workflows/build_and_test.yml
seladb: We still keep the Packet++Test, which doesn't require a network so should work if `env.avx512 == 'true'`. However, if it's `false` the whole executable fails...

ci/install_windivert.bat
seladb: Our other scripts are batch scripts and not `ps1` scripts... maybe it's better to convert the powershell part to batch, what do you think?

Pcap++/header/WinDivertDevice.h
seladb: The idea was to mimic the WinDivert interface as much as possible - `IWinDivertHandle` is an abstraction of WinDivert's `HANDLE`, and `IWinDivertImplementation` should be a very thin wrapper around WinDivert interface. I did it on purpose so mocking WinDivert will be straightforward and most of the business logic will be inside `WinDivertDevice`

Pcap++/header/WinDivertDevice.h
seladb: Addressed in this commit: https://github.com/seladb/PcapPlusPlus/pull/2019/commits/f42df08e0dc6f5f22a2845b5f80cc02f77b60c29

Pcap++/header/WinDivertDevice.h
Dimi1010: I see.

I think we can still keep mostly the same API structure even with the proposed redesign, no? 

Won't be complete 1 to 1, but the main change would be the handle first param would be omitted from the signature.

My main issue is that the current design uses dynamic cast on every operation, which is unnecessary overhead and makes the thin wrapper heavier than it needs to be.

CMakeLists.txt
```diff
@@ -284,6 +285,14 @@ if(PCAPPP_BUILD_PCAPPP)
     endif()
     add_definitions(-DUSE_XDP)
   endif()
+
+  if(PCAPPP_USE_WINDIVERT)
+    find_package(WinDivert REQUIRED)
+    if(NOT WinDivert_FOUND)
+      message(FATAL_ERROR "WinDivert not found!")
+    endif()
```
Dimi1010: The required flag on find_package will make the process fail if WinDivert is missing. There is no need for the if not found branch.

Pcap++/header/WinDivertDevice.h
seladb: I like this approach and have implemented it in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/72403879b4af3c3ea1ac64a88268375488c256f8 !

One thing I'm not sure of is how to handle `IWinDivertHandle::close()`. It returns a success or fail result which I want to keep and use in `WinDiverDevice::close()` so I don't want to completely remove it. What do you think?

Pcap++/src/WinDivertDevice.cpp
seladb: This was also fixed as part of the changes in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/72403879b4af3c3ea1ac64a88268375488c256f8

CMakeLists.txt
seladb: If you look through the file, we always output a message if the package wasn't found. I don't know a lot about CMake. What is the default message if the package isn't found?

Dimi1010: This is the fail message that is generated from `find_package(WinDivert REQUIRED)` if it can't find it. I think its explicit enough.

```
[...]
[CMake] -- Ccache not found!
[CMake] -- Building from commit:b31b8694 on branch:HEAD
[CMake] -- WinDivert root hint: 
[CMake] -- Looking in arch dir: x64
[CMake] CMake Error at C:/Program Files/Microsoft Visual Studio/2022/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
[CMake]   Could NOT find WinDivert (missing: WinDivert_INCLUDE_DIR WinDivert_LIBRARY)
[CMake] Call Stack (most recent call first):
[CMake]   C:/Program Files/Microsoft Visual Studio/2022/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
[CMake]   cmake/modules/FindWinDivert.cmake:97 (find_package_handle_standard_args)
[CMake]   CMakeLists.txt:292 (find_package)
[CMake] -- Configuring incomplete, errors occurred!
```

Pcap++/header/WinDivertDevice.h
Dimi1010: Hmm, in which conditions can it fail to close a handle?

If we want to keep it, we can. but we need to call it in the destructor if it hasn't been called and we need to check all other methods if it has been called and throw exception?

ci/install_windivert.bat
Dimi1010: Personally I prefer powershell scripts over batch since I find them easier to read.

Pcap++/header/WinDivertDevice.h
seladb: > Hmm, in which conditions can it fail to close a handle?

To be honest, I'm not sure, but the WinDivert API `WinDivertClose` returns a result: https://reqrypt.org/windivert-doc.html#divert_close

> If we want to keep it, we can. but we need to call it in the destructor if it hasn't been called and we need to check all other methods if it has been called and throw exception?

Sure, I'll do that 👍

CMakeLists.txt
seladb: Thanks! Removed in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/bbd692f7f0446d51374bcd3e961e423dc792ae35

Pcap++/header/WinDivertDevice.h
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/bbd692f7f0446d51374bcd3e961e423dc792ae35

ci/install_windivert.bat
seladb: Sure, the batch file was replaced with a powershell script in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/bbd692f7f0446d51374bcd3e961e423dc792ae35

Pcap++/header/WinDivertDevice.h
Dimi1010: > To be honest, I'm not sure, but the WinDivert API `WinDivertClose` returns a result: https://reqrypt.org/windivert-doc.html#divert_close

Apparently, its just a wrapper over Win32 `CloseHandle`. https://github.com/basil00/WinDivert/blob/master/dll/windivert.c#L718-L721
```cpp
BOOL WinDivertClose(HANDLE handle)
{
    return CloseHandle(handle);
}
```

`CloseHandle` mostly fails only if used with invalid handles or handles that are part of other windows subsystems and such, so it shouldn't be common. 🙂

Pcap++/src/WinDivertDevice.cpp
```diff
@@ -0,0 +1,694 @@
...
+		m_IsReceiving = true;
+		while (m_IsReceiving)
+		{
+			auto result = receivePacketsInternal(timeout, batchSize, buffer, overlapped.get());
+
+			if (result.status != ReceiveResult::Status::Completed)
+			{
+				m_IsReceiving = false;
+				return { result.status, result.error, result.errorCode };
+			}
```
Dimi1010, line 418: This branch can get executed if the IO operation is still pending (ERROR_IO_PENDING + event wait timeout expired).

Unless the IO operation is somehow cancelled, which I don't believe we do currently, it is unsafe to return.
Returning would invalidate the overlapped structure held in `overlapped` as the `unique_ptr` would go out of scope and delete the structure.

Pcap++/src/WinDivertDevice.cpp
```diff
@@ -0,0 +1,694 @@
...
+
+		uint32_t receivedPacketCount = 0;
+		while (maxPackets == 0 || receivedPacketCount < maxPackets)
+		{
+			auto result = receivePacketsInternal(timeout, batchSize, buffer, overlapped.get());
+
+			if (result.status != ReceiveResult::Status::Completed)
+			{
+				return { result.status, result.error, result.errorCode };
+			}
```
Dimi1010, line 345: ditto: Overlapped, heap use after free.

Pcap++/header/WinDivertDevice.h
seladb: Still, I think it's nicer to stick to the API, especially if we later want to mock it in tests

Pcap++/src/WinDivertDevice.cpp
```diff
@@ -0,0 +1,694 @@
...
+	}
+
+	WinDivertDevice::ReceiveResultInternal WinDivertDevice::receivePacketsInternal(
+	    uint32_t timeout, uint8_t batchSize, std::vector<uint8_t>& buffer, internal::IOverlappedWrapper* overlapped)
+	{
+		auto result = m_Handle->recvEx(buffer.data(), buffer.size(), batchSize, overlapped);
+		if (result != internal::IWinDivertHandle::ErrorIoPending)
+		{
+			return { ReceiveResult::Status::Failed, "Error receiving packets: " + getErrorString(result), result };
+		}
```
Dimi1010, line 609: I think `recvEx` can technically succeed synchronously, thus the success state should also be handled as a separate case.

Pcap++/src/WinDivertDevice.cpp
seladb: Just to make sure I understand: in the case where the event timeout expires, we should call `ResetEvent()`?

Pcap++/src/WinDivertDevice.cpp
seladb: From what I've explored and seen, since we're using OVERLAPPED we should always expect the callback from the OVERLAPPED object.

[`WinDivertRecvEx`](https://reqrypt.org/windivert-doc.html#divert_recv_ex) can indeed work synchronously if we don't pass the OVERLAPPED object, but then it will never timeout, meaning if no packets are received, it will stay stuck indefinitely

Pcap++/src/WinDivertDevice.cpp
seladb: See this response/question: https://github.com/seladb/PcapPlusPlus/pull/2019#discussion_r2532326076

Pcap++/src/WinDivertDevice.cpp
Dimi1010: Underneeth all the code, WinDivert uses Win32's `DeviceIOControl` API call.

This is an extract from MS's Overlapped IO reference:

"When a function is called to perform an overlapped operation, the operation might be completed before the function returns. When this happens, the results are handled as if the operation had been performed synchronously. If the operation was not completed, however, the function's return value is FALSE, and the GetLastError function returns ERROR_IO_PENDING."
 https://learn.microsoft.com/en-us/windows/win32/sync/synchronization-and-overlapped-input-and-output#:~:text=When%20a%20function%20is%20called,GetLastError%20function%20returns%20ERROR_IO_PENDING.

We can not assume that the overlapped operation will always return IO pending. The overlapped API is more of a "you can do it async" than a "you must do it async."

Dimi1010: The provided overlapped object is always correctly populated, tho. Even if the operation is completed synchronously.

Pcap++/src/WinDivertDevice.cpp
Dimi1010: No. That won't help.

The issue is that Win32 overlapped ops are enqueued and stay pending indefinitely, unless manually cancelled. They don't have a built-in timeout. The event wait timing out does not cancel the Async IO. It just terminates the waiting operation (`WaitForSingleObject`).

This is the current flow that we have in case there are no incoming packets:
```mermaid
sequenceDiagram
  actor usr as User
  participant device as WinDivert Device
  participant handle as WinDivert Handle
  participant windivert as WinDivert Library
  participant win32 as Windows Runtime
  participant win32io as Windows Overlapped IO subsystem

  usr->>device: receivePackets
  activate device
  device->>handle: Make overlapped
  activate handle
  handle-->>device: Overlapped
  deactivate handle
  
  device->>device: receivePacketsInternal
  activate device
  device->>handle: recvEx
  activate handle
  handle->>windivert: WinDivertRecvEx
  activate windivert

  windivert->>win32: DeviceIOControl
  activate win32
  win32-)win32io: Begin Async operation
  activate win32io
  note over win32io: The async operation is started.
  note over win32io: With no new packet it will stay pending indefinitely.
  note over win32: Async IO was started:<br/> Return pending IO result code immediately.
  win32-->>windivert: Result Code ERROR_IO_PENDING
  deactivate win32
  windivert-->>handle: Result code ERROR_IO_PENDING
  deactivate windivert
  handle-->>device: Result code ERROR_IO_PENDING
  deactivate handle
  device->>device: Wait on the event.
  
  activate device
  note right of device: No packets, timeout will expire.
  note right of device: Notice how the Async IO is not cancelled.
  device-->>device: Error code: TIMEOUT
  deactivate device
 
  device-->>device: ReceiveResult::Status::Timeout
  deactivate device
  
  device->>device: Check receive result
  
  device-->>usr: Return result Timeout.
  note over device: Overlapped goes out of scope and gets deleted.

  deactivate device
  
  win32io->>win32io: Async IO is still running and may attempt to use Overlapped
  note left of win32io: Heap-use-after-free: Overlapped
  deactivate win32io
```

The issue is at the final operations. As we exit `receivePackets` the overlapped structure goes out of scope, but the async IO is still running. We either need to call `CancelIOEx` to actually cancel the operation or keep the overlapped structure alive until it finishes.

NB: `CancelIOEx` is a request for cancellation, we still have to fetch the result of `Overlapped`. It might have succeeded in the timeframe the cancellation request took to process.

PS: Isn't async fun? 😄

Pcap++/src/WinDivertDevice.cpp
```diff
@@ -0,0 +1,694 @@
...
+	{
+		auto result = m_Handle->close();
+		if (result != internal::IWinDivertHandle::SuccessResult)
+		{
+			PCPP_LOG_ERROR("Couldn't receive packet, status: " << getErrorString(static_cast<uint16_t>(result)) << "("
+			                                                   << static_cast<int>(result) << ")");
+		}
+	}
+
+	WinDivertDevice::ReceiveResult WinDivertDevice::receivePackets(WinDivertRawPacketVector& packetVec,
```
Dimi1010, line 308: Also, I think the whole receive procedure can be improved a bit.

For example, since we are using Async IO, why are we not taking advantage of it?
Currently the user still receives the packets in a blocking manner.

I think something like this can allow for greater flexibility and is generally more readable?

It would also allow the user to write its own receive procedures if needed, if he wants to utilize the Async IO to do something else in the mean time.

If you want I can make an implementation and submit a PR to this branch?

```cpp
enum class AsyncIOStatus
{
  Unknown = 0,
  Pending,
  Completed, // on successful complete.
  Failed,         // on exception or something.
  Cancelled,
}

enum class WaitStatus
{
  Unknown = 0,
  ResultReady,
  Timeout,
};

// Hold the buffers required about a receive operation.
// Created from factory method on IWinDivertHandle
class IWinDivertRxBuffer
{
  // Modify the RX buffer to hold at least batchSize elements.
  // Does not shrink the buffer capacity.
  virtual void resizeForBatch(uint32_t batchSize) = 0;  // maybe prepare for batch?
  
  // Parses the receive buffer data into RawPackets.
  virtual WinDivertRawPacketVector getPackets() = 0;
  // Appends the packets into the vector. Does not clear the vector.
  virtual void appendPacketsTo(WinDivertRawPacketVector& packets) = 0;

  // Resets the buffer to clear state.
  void reset() = 0;
}

// Somewhat WinDivert specific.
class AsyncReceiveOperation
{
public:   
    // Overloads are to allow reuse of existing buffers.
    // If they are `nullptr`, create from the handle.
    AsyncReceiveOperation(IWinDivertHandle& handle);
    AsyncReceiveOperation(IWinDivertHandle& handle, std::unique_ptr<IOverlapped> overlapped);
    AsyncReceiveOperation(IWinDivertHandle& handle, std::unique_ptr<IWinDivertRxBuffer> buffer, std::unique_ptr<IOverlapped> overlapped = nullptr);

    // If the Op is not done call cancel() or follow std::thread and call std::terminate on improper destruction?
    // I think std::terminate is better. Cancel is a bit complex and users should finish their work explicitly.
    ~AsyncReceiveOperation();

    // move only class

    void requestCancelation(); // non-blocking request for cancellation, return immediately. Check status afterwards.
    AsyncIOStatus cancel(); // blocking request for cancellation, does not return until the op is cancelled or completed.

    AsyncIOStatus queryStatus(); // gets the current status of the operation. completed / failed / pending.
    WaitStatus waitForCompletion(int timeout = 0); // wait for the operation to signalled (result ready) or timeout.
    
    // Unsafe to use while operation is pending.
    IWinDivertRxBuffer& rxBuffer() { 
      if(rxBuffer == nullptr) throw std::logic_error("No buffer");
      return *rxBuffer;
    }

    // For swapping the RxBuffer with a fresh buffer.
    // Example use:
    //  Have 2 buffers.
    //  Async Receive IO Complete -> SwapRxBuffer -> Start Async Receive IO -> Handle the completed buffer -> Loop.
    //  Can technically reduce the delay in which the system is not receiving IO.
    // Note: do not accept nullptr. The operation should always have a buffer.
    std::unique_ptr<IWinDivertRxBuffer> exchangeRxBuffer(std::unique_ptr<IWinDivertRxBuffer> buffer);

    // Resets the async status of the operation, to prepare it for new use.
    void resetState() {
      rxBuffer->reset();
      overlapped->reset();
    }
private:
    std::unique_ptr<IWinDivertRxBuffer> rxBuffer;
    std::unique_ptr<IOverlapped> overlapped;
    IWinDivertHandle* handle; // For cancel operations. It requires handle.
}

class WinDivertDevice
{
  // Async API for the device.
  // User should not call this more than once while there is an active operation running.
  // User is responsible for the operation.
  // Batch size is the number of packets to be received by the operation.
  AsyncReceiveOperation beginReceive(uint32_t batchSize /* = 1 (default ?)*/);
  // In case we want to reuse a operation object. Op object should be clear.
  void beginReceive(AsyncReceiveOperation& op, uint32_t batchSize /* = 1 (default ?) */);
  
  // Current blocking receive operations, to be implemented as procedures utilizing beginReceive and AsyncReceiveOp under the hood.
  ReceiveResult recivePackets(WinDivertRawPacketVector& packetVec, uint32_t timeout, uint32_t maxPackets,
	                                                               uint8_t batchSize)
  {
     // Guards here
     packetVec.clear(); // Clear the output buffer.

     AsyncReceiveOp op(handle);
     
     // Simple RAII wrapper, sets the flag to True, when it enters, sets the flag to False when it exits.
     // Thus we don't need to set receiving to false prior to every return.
     ScopedTrue rxScope(m_Receiving);
     
     while(packetVec.size() < maxPackets && m_Receiving)
     {
        beginReceive(op, batchSize);
        auto waitResult = op.waitForCompletion(timeout);
        AsyncIOStatus ioStatus;
        if (waitResult == Timeout)
        {
           ioStatus = op.cancel(); // Wait for Async IO status update (Cancel or Complete).
           
           if (ioStatus == Cancelled)
           {
             return Timeout; /* Or cancelled? */
           }
           
           // If we are here, it means that the IO operation either Completed or Failed.
           // Thus we need to process the result.
        }
        else if (waitResult == ResultReady)
        {
           ioStatus = op.queryStatus();  // Fetch the new status
        }
        else
        {
            throw std::logic_error("Unsupported wait status");
        }

        switch(ioStatus)
        {
        case Completed: 
        {
           // Drain the RX buffer
           op.rxBuffer().appendPacketsTo(packetVec);
           op.reset();
        }
        case Failed: return Failed;
        default: throw std::logic_error("Unexpected status");
        }
     }

     // We are finished receiving, all received packets up to date are in packetVec.
     return Complete;
  }
}
```

Pcap++/src/WinDivertDevice.cpp
seladb: I asked Claude and it suggested adding this to the destructor. Would that be sufficient?
```cpp
~WinDivertOverlappedWrapper() override
{
    // Cancel any pending I/O
    CancelIoEx(m_Handle, &m_Overlapped);
    
    // Wait for cancellation to complete
    WaitForSingleObject(m_Event, INFINITE);
    
    // Now safe to cleanup
    CloseHandle(m_Event);
}
```

Pcap++/src/WinDivertDevice.cpp
seladb: I'm not sure I understand what the fix would be - if the result isn't `ERROR_IO_PENDING` we should treat it as a sync call?
That would complicate the implementation by quite a lot... how common it is to not get `ERROR_IO_PENDING` as a result? I never saw it in my tests

Pcap++/src/WinDivertDevice.cpp
seladb: The reason I didn't implement this approach is that I want to make the API simple, like we have in other devices. I wouldn't go into the rabbit hole of using async, but that's the only way I saw to ensure the operation never gets stuck

Pcap++/src/WinDivertDevice.cpp
Dimi1010: > if the result isn't ERROR_IO_PENDING we should treat it as a sync call?

As far as I understand it, yes, but with caveats. 
- All the "out" parameters on the call are still populated as if a sync call was made.
- The overlapped structure is still correctly populated with the state of the task. 
You can still use `GetOverlappedResult` to fetch the result if you want.
- I am unsure if the event gets signaled or not. The docs don't say anything specific about it and GPT said yes. So I would assume it gets signalled.
- There are some other caveats for IOCP usage and completion routines but those aren't relevant here.

> How common it is to not get ERROR_IO_PENDING as a result?

Not sure how common it can happen in our case. Probably not very common, but it can still happen by the Win32 Overlapped IO Spec. Generally it relies on the driver having the data ready at the time of call or available with minimal delay or something. The only way to be sure it can't happen is if we go digging in the driver code, which would probably be more tiresome.

Pcap++/src/WinDivertDevice.cpp
Dimi1010: I see. I suppose we can just expose the sync methods at first? `beginReceive` can be a private helper method.

Even if we keep the `beginReceive()` as internal helper methods, I do think it makes the Async code a bit cleaner and easier to read, and that is usually the most complicated part. I suppose that can be added later tho.

Pcap++/src/WinDivertDevice.cpp
Dimi1010: If you don't care about the result after you call the cancel, yes. That should be fine.

If you do care about the result. The operation can technically still succeed even after you called `CancelIoEx` depending on scheduling. You don't know until you call `GetOverlappedResult` on the overlapped operation after the event gets signaled.

seladb: Since this is called in the destructor, no one will call `GetOverlappedResult` after so I guess it should be fine. I'll add this to the destructor

Pcap++/src/WinDivertDevice.cpp
seladb: Assuming we can still use `GetOverlappedResult` to fetch the result, the solution would be to accept a `Success` result as well as `ErrorIoPending` and continue as usual. I'll implement it and test

seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/0ab7d080b1b2baaf1756237cd28a066c8781d24b

Pcap++/src/WinDivertDevice.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/0ab7d080b1b2baaf1756237cd28a066c8781d24b

Pcap++/header/WinDivertDevice.h
```diff
@@ -0,0 +1,466 @@
...
+		/// @warning The returned pointer may become invalid after subsequent calls that refresh interfaces.
+		const NetworkInterface* getNetworkInterface(uint32_t interfaceIndex) const;
+
+		/// @brief Enumerate Windows network interfaces.
+		/// @return A vector of NetworkInterface entries.
+		std::vector<NetworkInterface> getNetworkInterfaces() const;
+
+		/// @brief Replace the underlying implementation (intended for testing/mocking).
+		/// @param[in] implementation An implementation of the WinDivert backend APIs.
+		void setImplementation(std::unique_ptr<internal::IWinDivertImplementation> implementation);
```
Dimi1010: IMO, this should use constructor dependency injection and disallow switching of implementation after a device has been constructed. It simplifies the lifecycle. 

Curently, this setter directly swaps the implementation. If that happens while the device is already open, it might cause a desync between implementation and handle? It's not a big deal currently, but it's still an inconsistency.

Pcap++/header/WinDivertDevice.h
```diff
@@ -0,0 +1,466 @@
...
+
+		/// @class IWinDivertImplementation
+		/// @brief Factory and system-query abstraction used by WinDivertDevice.
+		///
+		/// The sole responsibilities of this interface are:
+		/// - Creating IWinDivertHandle instances (which expose the WinDivert API surface).
+		/// - Enumerating relevant Windows network interfaces.
+		/// Keeping these responsibilities here keeps WinDivertDevice decoupled from concrete
+		/// system/driver calls and enables unit testing and alternative implementations.
+		class IWinDivertImplementation
```
Dimi1010: nit: "Implementation" seems a bit generic descriptor? Maybe `IWinDivertDriver` or `IWinDivertAPI`? I prefer driver, since its not supposed to be the pubic API we provide.

Pcap++/header/WinDivertDevice.h
```diff
@@ -0,0 +1,466 @@
...
+			QueueSize     ///< Maximum memory allocated for the queue (bytes)
+		};
+
+		/// @typedef WinDivertRawPacketVector
+		/// @brief Convenience alias for a vector of WinDivertRawPacket pointers with ownership semantics.
+		using WinDivertRawPacketVector = PointerVector<WinDivertRawPacket>;
+		/// @typedef ReceivePacketCallback
+		/// @brief Callback invoked with a batch of received packets when using the callback receive API.
+		/// The callback is called from the receiving loop until stopReceive() is invoked or an error/timeout occurs.
+		using ReceivePacketCallback = std::function<void(const WinDivertRawPacketVector& packetVec)>;
```
Dimi1010: Curious, do we want to add a context second parameter as future extension point? It can be an empty struct or it can hold a pointer to the device? Adding it initially would allow adding more context data in the future if needed without breaking the callback signature.

The other devices have a "Device" callback, but I find that a bit rigid if you want to add data only relevant to the callback context. 

```cpp
struct WinDivertReceiveCallbackCtx
{
   WinDivertDevice* device = nullptr; // Maybe?
};

std::function<void(const WinDivertRawPacketVector& packetVec, WinDivertReceiveCallbackCtx& ctx)>`
``

File: Pcap++/src/WinDivertDevice.cpp [hunk 16]
```diff
@@ -0,0 +1,708 @@
...
+				if (!m_Event)
+				{
+					throw std::runtime_error("Failed to create event");
+				}
+				m_Overlapped.hEvent = m_Event;
+			}
+
+			// Non-copyable
+			WinDivertOverlappedWrapper(const WinDivertOverlappedWrapper&) = delete;
+			WinDivertOverlappedWrapper& operator=(const WinDivertOverlappedWrapper&) = delete;
```
Dimi1010, line 35: nit: If we are holding the overlapped struct as value inside the object, perhaps mark the move ctors explicitly deleted too? We don't want the object to be moved either.

They won't get implicitly declared anyway, so its fine either way code wise.

Pcap++/src/WinDivertDevice.cpp
```diff
@@ -0,0 +1,708 @@
...
+
+				auto packetLength = std::get<1>(packetInfo);
+				receivedPackets.pushBack(new WinDivertRawPacket(curPacketPtr, static_cast<int>(packetLength),
+				                                                std::get<2>(packetInfo), false, linkType,
+				                                                address.interfaceIndex, address.timestamp));
+				curPacketPtr += packetLength;
+				remainingBytes -= packetLength;
+			}
+
+			callback(receivedPackets);
```
Dimi1010: Should probably check if the callback is empty. Calling an empty std::function is UB.

The check can be either:
- After the result status check, if for some reason receiving packets and doing nothing with them is valid.
- At the start if we must have a callback.

I think the start would be better

Pcap++/src/WinDivertDevice.cpp
```diff
@@ -0,0 +1,708 @@
...
+		{
+			return { SendResult::Status::Failed, 0, "Device is not open" };
+		}
+
+		if (batchSize == 0)
+		{
+			return { SendResult::Status::Failed, 0, "Batch size has to be a positive number" };
+		}
+
+		uint8_t buffer[WINDIVERT_BUFFER_LEN];
```
Dimi1010: nit: `std::array`?

cmake/modules/FindWinDivert.cmake
```diff
@@ -0,0 +1,120 @@
...
+if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+  set(_WinDivert_ARCH_DIR "x64")
+  set(_WinDivert_SYS_NAME "WinDivert64.sys")
+else()
+  set(_WinDivert_ARCH_DIR "x86")
+  set(_WinDivert_SYS_NAME "WinDivert32.sys")
+endif()
+
+# Normalize user-provided root path
+if(DEFINED WINDIVERT_ROOT)
```
Dimi1010: Can we make this WinDivert_ROOT to match the rest of the variables in the module?
As far as I know, it is expected for the variables to match precisely with the package name.

```suggestion
if(DEFINED WinDivert_ROOT)
```

cmake/modules/FindWinDivert.cmake
```diff
@@ -0,0 +1,120 @@
...
+  file(TO_CMAKE_PATH "${WINDIVERT_ROOT}" _WinDivert_ROOT_HINT)
+elseif(DEFINED ENV{WINDIVERT_ROOT})
```
Dimi1010: ditto: var name. 

Unsure about this one tho. Env vars are case sensitive on unix, and case-insensitive on windows. But probably better to match exact anyway.


```suggestion
elseif(DEFINED ENV{WinDivert_ROOT})
```

seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/1fde6a2314b0288be3e714150c905398267bf7b9

Pcap++/src/WinDivertDevice.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/aaba07833c1d0233f42e1852160873a0d87180f3

Pcap++/src/WinDivertDevice.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/aaba07833c1d0233f42e1852160873a0d87180f3

Pcap++/src/WinDivertDevice.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/aaba07833c1d0233f42e1852160873a0d87180f3

Pcap++/src/WinDivertDevice.cpp
```diff
@@ -0,0 +1,714 @@
...
+					adapter = adapter->Next;
+				}
+
+				return networkInterfaces;
+			}
+		};
+
+	}  // namespace internal
+
+#define WINDIVERT_BUFFER_LEN 65536
```
Dimi1010: nit: Constexpr?

Pcap++/src/WinDivertDevice.cpp
```diff
@@ -0,0 +1,714 @@
...
+				                                                std::get<2>(packetInfo), false, linkType,
+				                                                address.interfaceIndex, address.timestamp));
+				curPacketPtr += packetLength;
+				remainingBytes -= packetLength;
+			}
+
+			callback(receivedPackets);
+
+			overlapped->reset();
+		}
```
Dimi1010, line 446: nit:
I think we can optimize this to reuse the `receivedPackets` buffer?
Currently the buffer is re-created on each iteration. Instead we can pull it out of the receive loop and call `clear()` after the callback is done?

We could possibly optimize to reuse the `WinDivertRawPacket` instances too but that would be a bit tricker as it would involve amending it's implementation and can be left for later.

Pcap++/src/WinDivertDevice.cpp
Dimi1010: `WinDivertOverlappedWrapper& operator=(WinDivertOverlappedWrapper&&) = delete;` too.

Pcap++/src/WinDivertDevice.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/a540ce4c88a4ce804250a4f31c187bb92ad19aee

Pcap++/src/WinDivertDevice.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/a540ce4c88a4ce804250a4f31c187bb92ad19aee

Pcap++/src/WinDivertDevice.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/a540ce4c88a4ce804250a4f31c187bb92ad19aee

Pcap++/header/WinDivertDevice.h
seladb: Renamed in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/5cdf66d1e3401ac76700e0372630917d3c3f0e75

Pcap++/header/WinDivertDevice.h
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2019/commits/720b23ace4da539736b1da2ed7cd7fd211529320

Pcap++/header/WinDivertDevice.h
seladb: I think that's a good idea! Added here: https://github.com/seladb/PcapPlusPlus/pull/2019/commits/7f18ba1b7d76697858642cf8e605de0969cd51be

Dimi1010 (approved):
LGTM

PR #2025, opened 2025-11-19
Change `IFilterableDevice` to inherit from `IDevice`.

Dimi1010:
This PR changes `IFilterableDevice` to inherit from `IDevice`.

With both `IDevice` and `IFilterableDevice` being pure interfaces / abstract classes, it makes sense to have such an inheritance. All filterable devices are by definition also simple devices.

This change would simplify the inheritance chain of `PcapDevice` and `PfRingDevice` which would only inherit from `IFilterableDevice` instead of from both interfaces. This has the minor effect of removing 1 vtable pointer per instance for classes that inherit from `IDevice` and `IFilterableDevice`. Previously each interface would have generated a vtable pointer.

It would also allow up / down casting the chain of `IDevice` -> `IFilterableDevice` -> `ConcreteDevice`.
This would allow template methods in `IFilterableDevice` to check the general `IDevice` status.
Example: In #2005 's template template methods would be able to check `isOpened()` before proceeding to `doUpdateFilter`. This would reduce the boilerplate of needing to check if the device is open in `doUpdateFilter` implementations.

If the need arises for a non-device object to implement the filterable interface, it would be better for a interface that is unconnected with the device connotation to exist (e.g. `IBpfFilterable`).

### DIFF
--- a/Pcap++/header/Device.h
+++ b/Pcap++/header/Device.h
@@ -53,22 +53,39 @@ namespace pcpp
 		/// received
 		/// @param[in] filter The filter to be set in PcapPlusPlus' GeneralFilter format
 		/// @return True if filter set successfully, false otherwise
-		virtual bool setFilter(GeneralFilter& filter)
+		bool setFilter(GeneralFilter& filter)
 		{
 			std::string filterAsString;
 			filter.parseToString(filterAsString);
-			return setFilter(filterAsString);
+			return doUpdateFilter(&filterAsString);
 		}
 
 		/// Set a filter for the device. When implemented by the device, only packets that match the filter will be
-		/// received
+		/// processed.
 		/// @param[in] filterAsString The filter to be set in Berkeley Packet Filter (BPF) syntax
 		/// (http://biot.com/capstats/bpf.html)
 		/// @return True if filter set successfully, false otherwise
-		virtual bool setFilter(std::string filterAsString) = 0;
+		bool setFilter(std::string const& filterAsString)
+		{
+			return doUpdateFilter(&filterAsString);
+		}
 
 		/// Clear the filter currently set on the device
 		/// @return True if filter was removed successfully or if no filter was set, false otherwise
-		virtual bool clearFilter() = 0;
+		bool clearFilter()
+		{
+			return doUpdateFilter(nullptr);
+		}
+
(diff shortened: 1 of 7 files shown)


seladb:
@Dimi1010 I'm not totally against this change, but the idea was to have an interface that acts like a [Rust trait](https://doc.rust-lang.org/book/ch10-02-traits.html) - which defines a "filterable" behavior. Would renaming it to `IFilterable` make more sense?

You may claim, and I don't disagree, that this is not a common behavior and is only used in devices - which is why I'm not against this change 🙂 

What do you think?

Dimi1010:
> @Dimi1010 I'm not totally against this change, but the idea was to have an interface that acts like a [Rust trait](https://doc.rust-lang.org/book/ch10-02-traits.html) - which defines a "filterable" behavior. Would renaming it to `IFilterable` make more sense?

It would make a bit more sense since that would imply a standalone interface, unlinked from devices. I did mention `IBpfFilterable` as a potential interface name (since the filter is a bpf syntax).

> You may claim, and I don't disagree, that this is not a common behavior and is only used in devices - which is why I'm not against this change 🙂 
> 
> What do you think?

Currently, I don't think a standalone trait is all that useful for the current use cases. We have other filterable-like objects that don't use it at all since they don't need polymorphic behavior (`BpfWrapper`, `PcapHandle`).

Renaming the class has its own issues with the need to keep an alias of the new and old name for the deprecation period.

If we want to keep the standalone trait, I think adding `IFilterableDevice` to `IDevice` beirarchy and adding a new `IBpfFilterable` (that `IFilterableDevice` inherits) would create less breaking issues due to the change.

Still, I don't think standalone trait is currently necessary due to lack of demand. We don't really have any other public facing filterables outside of devices.

---

PS: What's up with DPDK device having `setFilter` methods that always fail with a not implemented error? Would it not be better to deprecate / remove them if it does not support filtering?

seladb (approved):

seladb:
> PS: What's up with DPDK device having `setFilter` methods that always fail with a not implemented error? Would it not be better to deprecate / remove them if it does not support filtering?

To be honest, I don't remember why I implemented it like this. I guess we can decide to remove it altogether

PR #2026, opened 2025-11-19
Cleanup `PcapLiveDevice` synchronous capture.

Dimi1010:
Part of #1838

This PR updates the synchronous capture method.


Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -813,14 +841,15 @@ namespace pcpp
...
-		m_cbOnPacketArrivesBlockingMode = std::move(onPacketArrives);
-		m_cbOnPacketArrivesBlockingModeUserCookie = userCookie;
-
 		m_CaptureThreadStarted = true;
 		m_StopThread = false;
 
-		const int64_t timeoutMs = timeout * 1000;  // timeout unit is seconds, let's change it to milliseconds
+		// A valid timeout is only generated when timeout is positive.
+		// This means that the timeout timepoint should be after the start time.
+		const bool hasTimeout = timeout > 0;
```
seladb, line 849: nit: I'd avoid this variable and use `timeout > 0` to indicate timeout is used.

I'd also do:
```cpp
auto timeoutTime = timeout > 0 ? startTime + std::chrono::milliseconds(static_cast<int64_t>(timeout * 1000)) : 0;
```

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -886,40 +930,46 @@ namespace pcpp
...
-		m_cbOnPacketArrivesBlockingModeUserCookie = nullptr;
 
 		if (shouldReturnError)
 		{
 			return 0;
 		}
 
-		if (std::chrono::duration_cast<std::chrono::milliseconds>(currentTime - startTime).count() >= timeoutMs)
+		// Check the time only if a valid timeout was specified. Otherwise it would always be true.
+		if (hasTimeout && currentTime >= timeoutTime)
```
seladb, line 959: If we set `timeoutTime = 0` if `timeout <= 0` we probably don't need `hasTimeout` here

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -310,6 +318,46 @@ namespace pcpp
...
+			{
+				PCPP_LOG_ERROR("Exception occurred while invoking packet arrival callback: " << ex.what());
+				context->requestStop = true;  // Stop capture on exception
+			}
+			catch (...)
+			{
+				PCPP_LOG_ERROR("Unknown exception occurred while invoking packet arrival callback");
+				context->requestStop = true;  // Stop capture on unknown exception
+			}
+		}
```
seladb, line 359: I think we decided not to catch exceptions here, no?

Dimi1010: We agreed to not catch exceptions for the other cases because they continued packet capture after catching it. We wanted it to stop which would have been harder to do for the other cases due to running on a separate thread.

In this case, the callback does terminate the packet capture after an exception is caught so its fine. It also solves the unresolved UB that was passing the exceptions through a C compiled code context.

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: If we have `timeoutTime = 0` then `currentTime >= timeoutTime` will always be true, thus always reporting timeout expired even when no timeout was set. We will also have to check `timeoutTime > 0` which in the current case is precomputed by the `hasTimeout` bool?

Pcap++/src/PcapLiveDevice.cpp
Dimi1010: > I'd avoid this variable and use timeout > 0 to indicate timeout is used.

We can do that, curious as to why? They are both the same thing but the variable is essentially a cached check that you can read better in the code?

> `auto timeoutTime = timeout > 0 ? startTime + std::chrono::milliseconds(static_cast<int64_t>(timeout * 1000)) : 0;`

Timepoint has no constructor from `int` value, so it would have to be something like:
```cpp
auto timeoutTime = timeout > 0 ? startTime + std::chrono::milliseconds(static_cast<int64_t>(timeout * 1000)) : std::chrono::steady_clock::time_point();
```
or have it default initialized and then set to an actual value as such:
```cpp
std::chrono::steady_clock::time_point timeoutTime;
if(timeout > 0) { /* ... */}
```

Timepoint is also a bit verbose to compare to 0 afterwards, as you either have to do `timeoutTime > std::chrono::steady_clock::time_point()` or `timeoutTime.time_since_epoch().count() > 0` to do so.

Issue #1933, opened 2025-08-23
FIx FreeBSD CI issue


tigercosmos:
There are a few error cases for FreeBSD CI that we didn't catch before. Check the error cases in [the WIP PR](https://github.com/seladb/PcapPlusPlus/pull/1932). You can open a new PR for this issue.

seladb:
Fixed in #2028

PR #2028, opened 2025-11-24
Fix FreeBSD CI

seladb:
Fixes https://github.com/seladb/PcapPlusPlus/issues/1933

- Add `set -e` to ensure the FreeBSD job will fail if the tests fail
- Refactor `TestSendPacket` and `TestSendPackets` to send fewer packets - the FreeBSD job can't send thousands of packets in a very short amount of time


tigercosmos (approved):
LGTM

PR #2031, opened 2025-11-30
Add comprehensive SCTP protocol support

jordanvrtanoski:
Implement full SCTP (Stream Control Transmission Protocol) layer with support for:

Core Protocol (RFC 9260):
- All standard chunk types (DATA, INIT, INIT-ACK, SACK, HEARTBEAT, ABORT, SHUTDOWN, ERROR, COOKIE-ECHO, COOKIE-ACK, ECNE, CWR, SHUTDOWN-COMPLETE)
- CRC32c checksum calculation and validation
- Chunk and parameter action bits helpers
- Bundling validation per RFC 9260
- Host Name Address deprecation detection

Extensions:
- AUTH chunk with HMAC-SHA1/SHA256 computation and verification (RFC 4895)
- ASCONF/ASCONF-ACK for dynamic address reconfiguration (RFC 5061)
- RE-CONFIG for stream reconfiguration (RFC 6525)
- FORWARD-TSN for partial reliability (RFC 3758)
- I-DATA and I-FORWARD-TSN for message interleaving (RFC 8260)
- PAD chunk (RFC 4820)
- NR-SACK experimental support (IANA registered, draft-based)
- Zero Checksum Acceptable parameter (RFC 9653)

Additional features:
- Parameter iterators for INIT, RE-CONFIG, and ASCONF chunks
- Error cause iterator for ABORT/ERROR chunks
- Extended PPID enum with 70+ protocol identifiers from IANA registry
- Chunk creation APIs for building SCTP packets

Includes 72 test cases covering parsing, creation, and validation.


Packet++/src/SctpLayer.cpp
```diff
@@ -0,0 +1,3690 @@
...
+{
+	// ==================== CRC32c Implementation ====================
+	// CRC32c (Castagnoli) polynomial: 0x1EDC6F41
+	// This is the polynomial used by SCTP as defined in RFC 9260
+
+	namespace
+	{
+		// CRC32c lookup table (Castagnoli polynomial)
+		// Pre-computed for polynomial 0x1EDC6F41 (reflected)
+		constexpr uint32_t crc32cTable[256] = {
```
Dimi1010, line 20: nit: Can we use `std::array` here?

Packet++/header/SctpLayer.h
```diff
@@ -0,0 +1,2412 @@
...
+
+	// Forward declarations
+	class SctpChunk;
+	class SctpLayer;
+	class SctpInitParameter;
+
+	/// @class SctpChunk
+	/// A wrapper class for SCTP chunks. This class does not create or modify chunk records,
+	/// but rather serves as a wrapper and provides useful methods for retrieving data from them
+	class SctpChunk
```
Dimi1010, line 1067: This is essentially a view class right?

Somewhat thinking out loud but can we split this class into several view subclasses for each chunk type?
Currently the class seems a bit unwieldy with all the different chunk type methods?

Can we instead have:
- SctpChunk - Common Methods (`getChunkType()`, etc)
  - SctpDataChunk - A data chunk.
  - SctpInitChunk   - Init/InitAck chunk.
  - ScptSackChunk - Sack chunk.
  - ...

Since the chunk objects only hold a pointer, they can be treated as value objects, similar to `std::string_view`.
```cpp
class SctpChunk
{
  /* Generic chunk methods */
  SctpChunkType getChunkType() const;
}

class SctpDataChunk : public SctpChunk {
  SctpDataChunk(SctpChunk chunk) { /* validate correct chunk type */ }

  /* Data specific chunk accessors */
}

// Other chunk types.
```
The inheritance in the above example is mostly for code reuse of the common methods, not for polymorphism, therefor the lack of any virtual methods.

Do you think that would work or is there something I am missing?

Dimi1010 (commented):
Thank you for the submission. :)

Packet++/src/SctpLayer.cpp
jordanvrtanoski: constexpr uint32_t  is used for compile-time optimization of the `calculateCrc32cSoftware`. It can be changed to std::array but that will prevent optimization.

Dimi1010: From what I understand, `std::array` also supports being defined as `constexpr std::array<uint8_t, N>`, or is the optimization something specific to the raw array?

Packet++/header/SctpLayer.h
jordanvrtanoski: That's good observation, let me see what it takes to refactor it to be more aligned to OO principles.

Packet++/src/SctpLayer.cpp
jordanvrtanoski: For std::array to be constexpr and considerd for compile-time optimization a minimum of C++17 is required. I am aware that even for C++17 there were discussions if the optimization is full equivalent of the "C" style array.
In addition, the `crc32cTable` is part of an anonymous namespace and will not be exposed to the user of the library.

Packet++/header/SctpLayer.h
jordanvrtanoski: The reason the SctpChunk is designed as a monolith class in the first pace is to avoid the heap overhead that inheritance would create, as well as it's impact on the cache as the objects in the hierarchy have bad locality. 

Having the chunks as views classes on the SctpChunk will address my original performance concerns, so I will refactor the code based on the suggestion.

jordanvrtanoski:
Converted the PR to draft to address the proposed refactoring.

Tests/Packet++Test/PacketExamples/SctpAuthPacket.dat
seladb, line 1: Can you please add the original pcap file used for these `.dat` files?

seladb:
@jordanvrtanoski this PR is opened for a long time, are you planning to continue working on it? Otherwise we can close it for now

jordanvrtanoski:
I will rebase it and prepare for merge, as I see all concerns ware addressed.

seladb:
> I will rebase it and prepare for merge, as I see all concerns ware addressed.

Thanks @jordanvrtanoski ! Can you please make sure CI passed as well?

jordanvrtanoski:
Sure

jordanvrtanoski:
@seladb, I have problems in my github runner with actions/setup-java@v5 test. I am getting "Error: Java setup process failed due to: error code: 520". The project is build fine but the test is failing as the JDK can not be set up properly. Have you faced this issue? This is blocking Android tests, all other tests are green. Any tips how to proceed?

jordanvrtanoski:
After multiple retries, one of the Android test passed, but the other 3 are still failing with the same issue, I assume the repository had reached some type of limitation.

jordanvrtanoski:
After a few more tries, all the checks on my repo passed!

jordanvrtanoski:
@seladb ready to merge

Packet++/src/SctpLayer.cpp
```diff
@@ -0,0 +1,3759 @@
...
+
+			chunk = getNextChunk(chunk);
+		}
+
+		// Cannot add INIT, INIT-ACK, or SHUTDOWN-COMPLETE to a packet with existing chunks
+		if (chunkType == SctpChunkType::INIT || chunkType == SctpChunkType::INIT_ACK ||
+		    chunkType == SctpChunkType::SHUTDOWN_COMPLETE)
+		{
+			return false;
+		}
```
Dimi1010, line 2363: nit: Perhaps move this condition above the O(N) while loop?

Packet++/header/SctpLayer.h
```diff
@@ -0,0 +1,3587 @@
...
+		uint8_t getFlags() const;
+
+		/// @return Chunk length (as specified in header, not including padding)
+		uint16_t getLength() const;
+
+		/// @return Total size including padding (aligned to 4 bytes)
+		size_t getTotalSize() const;
+
+		/// @return Pointer to chunk value/payload (after the 4-byte header)
+		uint8_t* getValue() const;
```
Dimi1010, line 1106: Would it be better to return `uint8_t const*`? 
Currently the returned buffer can be mutated, but I am not sure if that is intended behaviour.

PS: The question also applies for all other view types that return pointer to internals.

Dimi1010 (commented):
Well. There are some long files here. I did a first pass review. I will probably do another pass later, tho. I am really liking how the view classes turned out. I have one comment about the return type of pointers from them tho. The other things I looked at LGTM.

---
PS: @jordanvrtanoski if you make further edits to this, can you please not force push. It makes it really hard to find the edits since it deletes the old commits. It will all get squashed on merge, so it doesn't matter for the main branch, but it makes it easier to track changes during the review process.

PR #2034, opened 2025-12-30
Read and write pcap files without libpcap

seladb:
This PR implements parsing and writing pcap files without using libpcap. It's a rewrite of `PcapFileReaderDevice` and `PcapFileWriterDevice` classes.

This change preserves the interface of these classes and deprecates `PcapFileReaderDevice::isNanoSecondPrecisionSupported()` and `PcapFileWriterDevice::isNanoSecondPrecisionSupported()` which always return `true`.

The tests, specifically in `FileTests.cpp`, were heavily refactored, mostly added more tests, and changed tests that are less relevant.

After this change, the only part that will still use libpcap is setting and matching BPF filters. I'm thinking how to remove libpcap dependency here as well, but will be done in a separate PR.

### DIFF
--- a/Pcap++/src/PcapFileDevice.cpp
+++ b/Pcap++/src/PcapFileDevice.cpp
@@ -655,12 +655,13 @@ namespace pcpp
 		}
 
 		bool result = true;
-		std::for_each(packets.begin(), packets.end(), [&](const auto& packet) {
+		for (auto const& packet : packets)
+		{
 			if (!writePacket(*packet))
 			{
 				result = false;
 			}
-		});
+		}
 
 		return result;
 	}


Tests/Pcap++Test/Tests/FilterTests.cpp
seladb, line 1: The reason for this change is that the link type of `RAW_IP_PCAP_PATH` is `LINKTYPE_DLT_RAW1` (value of 12), which in libpcap is translated to `LINKTYPE_RAW` (value of 101): https://github.com/the-tcpdump-group/libpcap/blob/master/pcap-common.c#L1399-L1400

`LINKTYPE_RAW` doesn't have good support in BPF filters, so I replaced it with `LINKTYPE_NULL`

Dimi1010:
Can we deprecate the Nano precision supported methods instead of outright removing them (we can make them always return `true`)? I think they were part of the previous releases?

seladb:
> Can we deprecate the Nano precision supported methods instead of outright removing them (we can make them always return `true`)? I think they were part of the previous releases?

You're right, it's better to deprecate than to remove them altogether.

Added back in https://github.com/seladb/PcapPlusPlus/pull/2034/commits/d16a3614c15618a3858fb72b60ac21d5c512e83a

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -170,91 +288,217 @@ namespace pcpp
...
+			packetHeader.caplen = swap32(packetHeader.caplen);
+			packetHeader.len = swap32(packetHeader.len);
+		}
+
+		if (packetHeader.caplen > m_SnapshotLength)
+		{
+			PCPP_LOG_ERROR("Packet captured length " << packetHeader.caplen << " exceeds file snapshot length "
+			                                         << m_SnapshotLength);
 			return false;
 		}
```
Dimi1010: Technically this is a regression in the parser functionality.

Libpcap supports `caplen` values larger than `snapshot` if the `caplen` value is smaller than a constant MAX_SNAPLEN value (usually 262144 bytes). If so, the library reads the first `snaplen` bytes and discards the rest.

I think this is legacy compatibility behavior tho, as it specifically mentiones libpcap 0.4 and Solaris 2.3 and the `caplen` is set to `snapshot` if that branch is taken.

[Libpcap code](https://github.com/the-tcpdump-group/libpcap/blob/03713d712400fc0200b601bec49e8620f5f61c8a/sf-pcap.c#L510)

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -170,91 +288,217 @@ namespace pcpp
...
+		if (filterAsString == nullptr)
+		{
+			return m_BpfWrapper.setFilter("");
+		}
+
+		return m_BpfWrapper.setFilter(*filterAsString);
+	}
+
+	bool PcapFileReaderDevice::readNextPacket(timespec& packetTimestamp, uint8_t* packetData, uint32_t& capturedLength,
+	                                          uint32_t& frameLength)
```
Dimi1010: nit: Can we explicitly pass the `packetDataLen` as parameter instead of relying on `m_SnapshotLength` internally? IMO, it makes it less opaque what the size of the input `packetData` buffer is.

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -170,91 +288,217 @@ namespace pcpp
...
+
+		capturedLength = packetHeader.caplen;
+		if (capturedLength > 0)
+		{
+			if (!m_PcapFile.read(reinterpret_cast<char*>(packetData), packetHeader.caplen))
+			{
+				PCPP_LOG_ERROR("Failed to read packet data");
+				return false;
+			}
+		}
```
Dimi1010: nit: Perhaps use `packetHeader.caplen` and only write to `capturedLength` at the end after everything passed?

Dimi1010: nit: If conditions can be chained to reduce indents.

```suggestion
		capturedLength = packetHeader.caplen;
		if (capturedLength > 0 && !m_PcapFile.read(reinterpret_cast<char*>(packetData), packetHeader.caplen))
		{
			PCPP_LOG_ERROR("Failed to read packet data");
			return false;
		}
```

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -170,91 +288,217 @@ namespace pcpp
...
+		return (value >> 24) | ((value >> 8) & 0x0000FF00) | ((value << 8) & 0x00FF0000) | (value << 24);
+	}
+
 	bool PcapFileReaderDevice::open()
 	{
 		resetStatisticCounters();
 
-		if (m_PcapDescriptor != nullptr)
+		m_PcapFile.open(m_FileName.c_str(), std::ifstream::binary);
+		if (!m_PcapFile.is_open() || !m_PcapFile.good())
```
Dimi1010: Why not use `if(m_PcapFile.fail()) { ... }`? If `open()` call fails it automatically sets the failbit.

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -170,91 +288,217 @@ namespace pcpp
...
+		{
+			if (m_BpfWrapper.matches(packetData.get(), capturedLength, packetTimestamp, m_PcapLinkLayerType))
+			{
+				rawPacket.setRawData(capturedLength > 0 ? packetData.release() : nullptr, capturedLength,
+				                     packetTimestamp, m_PcapLinkLayerType, frameLength);
+				reportPacketProcessed();
+				return true;
+			}
+			PCPP_LOG_DEBUG("Packet doesn't match filter");
+		}
```
Dimi1010: Do we want to call `reportPacketDropped()` on packets that don't match the filter?
I'm fine either way, but I don't remember what the libpcap default behavior was.

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -170,91 +288,217 @@ namespace pcpp
...
 
+	uint16_t swap16(uint16_t value)
+	{
+		return static_cast<uint16_t>((value >> 8) | (value << 8));
+	}
+
+	uint32_t swap32(uint32_t value)
+	{
+		return (value >> 24) | ((value >> 8) & 0x0000FF00) | ((value << 8) & 0x00FF0000) | (value << 24);
+	}
```
Dimi1010: Change to internal linkage (`namespace { ... }` or `static`), to prevent issues with ODR during linking.

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -170,91 +288,217 @@ namespace pcpp
...
+		timespec packetTimestamp;
+		auto packetData = std::make_unique<uint8_t[]>(m_SnapshotLength);
+		uint32_t capturedLength = 0, frameLength = 0;
+
+		while (readNextPacket(packetTimestamp, packetData.get(), capturedLength, frameLength))
+		{
+			if (m_BpfWrapper.matches(packetData.get(), capturedLength, packetTimestamp, m_PcapLinkLayerType))
+			{
+				rawPacket.setRawData(capturedLength > 0 ? packetData.release() : nullptr, capturedLength,
+				                     packetTimestamp, m_PcapLinkLayerType, frameLength);
```
Dimi1010: Ambiguous ownership semantics. See #2035

Dimi1010 (commented):
Looked through the reader device. Will look through the writer later.

Pcap++/src/PcapFileDevice.cpp
seladb: My understanding is that although libpcap checks for `MAX_SNAPLEN` it still doesn't read more than the `snapshot` size:

Here it grows the buffer to `snapshot`:
https://github.com/the-tcpdump-group/libpcap/blob/master/sf-pcap.c#L668-L674

Then reads `snapshot` bytes from the file:
https://github.com/the-tcpdump-group/libpcap/blob/89e982c37c36ad0bf9f10b7ded421cb42422effa/sf-pcap.c#L679

And then discards the rest of the bytes:
https://github.com/the-tcpdump-group/libpcap/blob/master/sf-pcap.c#L701-L724

It also adjusts the `caplen` to be the `snapshot`:
https://github.com/the-tcpdump-group/libpcap/blob/89e982c37c36ad0bf9f10b7ded421cb42422effa/sf-pcap.c#L730

I'll fix the code to match this behavior.

seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2034/commits/3725a944a0f4d391c227ff177862efc969e2bc73

Pcap++/src/PcapFileDevice.cpp
seladb: I agree, addressed in https://github.com/seladb/PcapPlusPlus/pull/2034/commits/3725a944a0f4d391c227ff177862efc969e2bc73

Pcap++/src/PcapFileDevice.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2034/commits/3725a944a0f4d391c227ff177862efc969e2bc73

seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2034/commits/3725a944a0f4d391c227ff177862efc969e2bc73

Pcap++/src/PcapFileDevice.cpp
seladb: Previously, we didn't call `reportPacketDropped()` on packets that didn't match the filter. This is because it was internal to libpcap, and we couldn't get stats for unmatched packets. I think it's ok to preserve this behavior

Pcap++/src/PcapFileDevice.cpp
seladb: You're right, addressed in https://github.com/seladb/PcapPlusPlus/pull/2034/commits/fb95044a1e976a3e9bef90d00e0f0f5caf551f31

Pcap++/src/PcapFileDevice.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2034/commits/fb95044a1e976a3e9bef90d00e0f0f5caf551f31

Pcap++/src/PcapFileDevice.cpp
seladb: You're right, I wrote this method a long time ago, and it has issues 😕

Dimi1010: #2035 is closed now. You can use the overload with `takeOwnership = true`.

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -170,91 +288,222 @@ namespace pcpp
...
-		return checkNanoSupport();
+		timespec packetTimestamp;
+		auto packetData = std::make_unique<uint8_t[]>(m_SnapshotLength);
+		uint32_t capturedLength = 0, frameLength = 0;
+
+		while (readNextPacket(packetTimestamp, packetData.get(), m_SnapshotLength, capturedLength, frameLength))
+		{
+			if (m_BpfWrapper.matches(packetData.get(), capturedLength, packetTimestamp, m_PcapLinkLayerType))
+			{
+				rawPacket.setRawData(capturedLength > 0 ? packetData.release() : nullptr, capturedLength,
```
Dimi1010: If `captured length` is 0, do we even have to run the filter and create the packet?
We don't really have a "packet body" to display?

seladb: I guess we don't have to run the filter, but we should probably create the packet - the user would expect to get all packets in the file, even those that have no data...

I think it's a rare edge case and the BPF filter will probably return very quickly so maybe we don't need to change anything?

Pcap++/src/PcapFileDevice.cpp
seladb: Thanks! Added here: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/eeff1e4fe40352c6f121844121303db04c4ec9d0

Pcap++/header/PcapFileDevice.h
```diff
@@ -294,9 +306,37 @@ namespace pcpp
...
 		void flush();
 
+	protected:
+		bool doUpdateFilter(std::string const* filterAsString) override;
+
 	private:
+		LinkLayerType m_PcapLinkLayerType = LINKTYPE_ETHERNET;
+		bool m_NeedsSwap = false;
+		FileTimestampPrecision m_Precision = FileTimestampPrecision::Unknown;
+		std::ofstream m_PcapFile;
```
Dimi1010: Since in some append operations a `std::ifstream` is created to check stuff. I think its better to change `m_PcapFile` to `std::fstream` and just use the read pointer of that instead of reopening new streams.

seladb: @Dimi1010 do you mean that we first try to open `m_PcapFile` for reading, then close it, and open it again for writing/appending?

Dimi1010: Something like this. All `fstream / ifstream / ofstream` are wrappers over `filebuf`. We can theoretically use the same `fstream` object to `read somewhere -> move pointer -> write somewhere else`, if its open in read / write mode.

I think it would be nice but it's not critical and it might be easier to implement/review in a separate PR.

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -264,272 +513,326 @@ namespace pcpp
...
-			closeFile();
-			return false;
+	PcapFileWriterDevice::CheckHeaderResult PcapFileWriterDevice::checkHeader(const std::string& fileName,
+	                                                                          FileTimestampPrecision requestedPrecision,
+	                                                                          LinkLayerType requestedLinkType)
+	{
+		std::ifstream pcapFileTemp(fileName, std::ios::binary);
+		if (!pcapFileTemp.is_open())
+		{
+			return { CheckHeaderResult::Result::HeaderError, "Failed to open file: " + fileName };
```
Dimi1010: nit: A named constructor could improve readability. Something like:
```cpp
CheckHeaderResult CheckHeaderResult::fromError(std::string errorMessage)
{
  return { CheckHeaderResult::Result::HeaderError, std::move(errorMessage) };
}

CheckHeaderResult CheckHeaderResult::fromOk(bool needsSwap)
{
  CheckHeaderResult res;
  res.result = CheckHeaderResult::Result::HeaderOk;
  res.needsSwap = needsSwap;
  return res;
}
```

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -264,272 +513,326 @@ namespace pcpp
...
+			return { CheckHeaderResult::Result::HeaderNeeded };
 		}
 
-		if (fseek(m_File, 0, SEEK_END) == -1)
+		if (size < static_cast<std::streamsize>(sizeof(pcap_file_header)))
 		{
-			PCPP_LOG_ERROR("Cannot read pcap file '" << m_FileName << "' to it's end, error was: " << errno);
-			closeFile();
-			return false;
+			return { CheckHeaderResult::Result::HeaderError, "Malformed file header or not a pcap file" };
```
Dimi1010: nit, ditto: named ctor.

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -264,272 +513,326 @@ namespace pcpp
...
+		pcapFileTemp.seekg(0, std::ios::beg);
 
-		m_DeviceOpened = true;
-		PCPP_LOG_DEBUG("File writer device for file '" << m_FileName << "' opened successfully in append mode");
-		return true;
-	}
+		pcap_file_header pcapFileHeader{};
+		if (!pcapFileTemp.read(reinterpret_cast<char*>(&pcapFileHeader), sizeof(pcapFileHeader)))
+		{
+			return { CheckHeaderResult::Result::HeaderError, "Cannot read file header" };
```
Dimi1010: nit, ditto: named ctor.

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -264,272 +513,326 @@ namespace pcpp
...
+		}
+		case NSEC_TCPDUMP_MAGIC_SWAPPED:
+		{
+			needsSwap = true;
+			precisionFromHeader = FileTimestampPrecision::Nanoseconds;
+			break;
+		}
+		default:
+		{
+			return { CheckHeaderResult::Result::HeaderError, "Unsupported pcap file format" };
```
Dimi1010: nit, ditto: named ctor

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -264,272 +513,326 @@ namespace pcpp
...
+				case FileTimestampPrecision::Nanoseconds:
+					return "Nanoseconds";
+				default:
+					return "Unknown";
+				}
+			};
+			return { CheckHeaderResult::Result::HeaderError, "Existing file precision (" +
+				                                                 getPrecisionStr(precisionFromHeader) +
+				                                                 ") does not match the requested device precision (" +
+				                                                 getPrecisionStr(requestedPrecision) + ")" };
```
Dimi1010: nit, ditto: named ctor

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -264,272 +513,326 @@ namespace pcpp
...
+		}
 
-		if (!m_AppendMode && m_PcapDumpHandler != nullptr)
+		if (pcapFileHeader.version_major != PCAP_MAJOR_VERSION || pcapFileHeader.version_minor != PCAP_MINOR_VERSION)
 		{
-			pcap_dump_close(m_PcapDumpHandler);
+			return { CheckHeaderResult::Result::HeaderError, "Unsupported pcap file version" };
```
Dimi1010: nit, ditto: named ctor

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -264,272 +513,326 @@ namespace pcpp
...
 		}
-		else if (m_AppendMode && m_File != nullptr)
+
+		if (pcapFileHeader.linktype != static_cast<uint32_t>(requestedLinkType))
 		{
-			// in append mode it's impossible to use pcap_dump_close, see comment above pcap_dump
-			fclose(m_File);
+			return { CheckHeaderResult::Result::HeaderError,
+				     "Existing file link type does not match the requested device link type" };
```
Dimi1010: nit, ditto: named ctor

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -264,272 +513,326 @@ namespace pcpp
...
 		}
 
-		m_PcapDumpHandler = nullptr;
-		m_File = nullptr;
-		PCPP_LOG_DEBUG("File writer closed for file '" << m_FileName << "'");
+		return { CheckHeaderResult::Result::HeaderOk, "", needsSwap };
```
Dimi1010: nit, ditto: named ctor

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -264,272 +513,326 @@ namespace pcpp
...
+		std::streamsize size = pcapFileTemp.tellg();
+
+		if (size == 0)
 		{
-			PCPP_LOG_ERROR(
-			    "Pcap file has a different link layer type than the one chosen in PcapFileWriterDevice c'tor, "
-			    << linkLayerType << ", " << m_PcapLinkLayerType);
-			closeFile();
-			return false;
+			return { CheckHeaderResult::Result::HeaderNeeded };
```
Dimi1010: nit, ditto: named ctor. Less priority as its only used once and keeps fields default.

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -264,272 +513,326 @@ namespace pcpp
...
+		if (wroteHeader)
 		{
-			return openWrite();
+			m_DeviceOpened = true;
 		}
+		return wroteHeader;
 	}
 
-	bool PcapFileWriterDevice::openWrite()
+	bool PcapFileWriterDevice::openAppend()
```
Dimi1010: IIRC, `openAppend` was added because previously it required a vastly different flow than `openWrite`.

From what I see, `append` now works as `write` when the file is empty.
Since they both now use similar flow, can't we merge them?

Pcap++/header/PcapFileDevice.h
seladb: `checkHeader()` is a static method and if we use `m_PcapFile` we need to pass both the file and the file name so it can open it in read mode (I don't think the method should assume the file is already open, no?).

I'd like to keep it static because in the future we might use it to detect if a file is a pcap file...

Dimi1010: If we inject the input stream as a reference to the check header, we don't need to pass the file name.

> (I don't think the method should assume the file is already open, no?)

The injected stream should absolutely be required to be open before injecting to the function. That being `std::ifstream& stream` that satisfies `is_open() == true`. That is great for separation of concerns.

It makes the `checkHeader` a pure data procedure ("I give you data stream, you give me the parsed result."), that doesn't care where the input is stored and decouples it from any responsibility to "know" anything more about it other than "It is a data stream that should hopefully be pcap data.".

```cpp
auto checkHeader(std::ifstream& stream, /* other params */)
{
    if(!stream.is_open())
    {
      return Error;
    }
    
    // Do pcap header checks here.
    return Success;
}
```

Pcap++/src/PcapFileDevice.cpp
seladb: They are somewhat similar, but not entirely:
1. `openAppend()` should check the header first - it'd write the header only if it doesn't exist, otherwise it needs to adjust fields like `m_NeedsSwap`
2. `openWrite()` opens the file in override mode, but `openAppend()` opens it in append mode (`std::ios::app`)

Initially I planned to merge them, but the amount of `if..else` would make it more complex and less readable. Please let me know what you think

Pcap++/src/PcapFileDevice.cpp
seladb: I like this idea 💯 ! Updated here: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/970f4f699f86cf6bc4c546807499eec812ae1f68

Pcap++/src/PcapFileDevice.cpp
seladb: Updated here: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/970f4f699f86cf6bc4c546807499eec812ae1f68

Pcap++/src/PcapFileDevice.cpp
seladb: Updated here: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/970f4f699f86cf6bc4c546807499eec812ae1f68

Pcap++/src/PcapFileDevice.cpp
seladb: Updated here: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/970f4f699f86cf6bc4c546807499eec812ae1f68

Pcap++/src/PcapFileDevice.cpp
seladb: Updated here: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/970f4f699f86cf6bc4c546807499eec812ae1f68

Pcap++/src/PcapFileDevice.cpp
seladb: Updated here: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/970f4f699f86cf6bc4c546807499eec812ae1f68

Pcap++/src/PcapFileDevice.cpp
seladb: Updated here: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/970f4f699f86cf6bc4c546807499eec812ae1f68

Pcap++/src/PcapFileDevice.cpp
seladb: Updated here: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/970f4f699f86cf6bc4c546807499eec812ae1f68

Pcap++/src/PcapFileDevice.cpp
seladb: Updated here: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/970f4f699f86cf6bc4c546807499eec812ae1f68

Pcap++/src/PcapFileDevice.cpp
Dimi1010: Hmm, the way I see the procedure flow is as such:
- Calculate `openMask` based on `appendMode`
  - append mode sets `std::ios::in | std::ios::out | std::ios::app` 
  Reading is allowed in append mode, but a write operation will reset the pointer to the end of the stream.
  - write mode sets `std::ios::out | std::ios::trunc`.
  Trunc is added to explicitly clear the file when opening it.

- Open `fstream` with `openMask`.
  - On `appendMode == true`, if the open fails, retry with `std::ios::out` and set `appendMode = false`
- On `appendMode == true`, check headers.
  - On `HeaderOk`, read from header, populate fields and return.
- On `appendMode == false` or `HeaderNeeded`, write headers, set defaults and return.

I think most of that procedure is straight forward. What do you think?

---
Btw, I have seen some software error if a file with the specified output name exists to not accidentally overwrite user data, due to user error. Should we do the same?

Pcap++/header/PcapFileDevice.h
seladb: Ok, I'm convinced 🙂 
Updated here: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/211ddd4859b0d9589d6e85c2ccb0a1087ed37495

Pcap++/src/PcapFileDevice.cpp
seladb: I tried to update the code, it'd look something like this:
```cpp
bool PcapFileWriterDevice::open(bool appendMode)
{
	if (m_DeviceOpened || m_PcapFile.is_open())
	{
		PCPP_LOG_ERROR("File already opened");
		return false;
	}

	auto flags = std::ios::binary | std::ios::out;
	if (appendMode)
	{
		flags |= std::ios::in | std::ios::app;
	}
	m_PcapFile.open(m_FileName, flags);

	if (!m_PcapFile.is_open())
	{
		PCPP_LOG_ERROR("Failed to open file: " << m_FileName);
		return false;
	}

	bool shouldWriteHeader = false;
	if (!appendMode)
	{
		shouldWriteHeader = true;
	}
	else
	{
		auto checkHeaderResult = checkHeader(m_PcapFile, m_Precision, m_PcapLinkLayerType);

		if (checkHeaderResult.result == CheckHeaderResult::Result::HeaderError)
		{
			PCPP_LOG_ERROR(checkHeaderResult.error);
			return false;
		}

		if (checkHeaderResult.needsSwap)
		{
			m_NeedsSwap = true;
		}

		if (checkHeaderResult.result == CheckHeaderResult::Result::HeaderNeeded)
		{
			shouldWriteHeader = true;
		}

		m_PcapFile.seekg(0, std::ios::end);
	}

	if (shouldWriteHeader && !writeHeader(m_PcapFile, m_Precision, PCPP_MAX_PACKET_SIZE, m_PcapLinkLayerType))
	{
		return false;
	}

	m_DeviceOpened = true;
	return true;
}
```

If you insist I will update the code, but I think separating them is nicer and easier to read. Please let me know what you think.

Dimi1010: We keep them separate, if you wish, but merging them does allow some code reuse.
I cleaned it up the code sample a bit. What do you think?

- There were some unnecessary `if` statements. E.g. `if (bool p) { bool x = true; }` can simplified to `bool x = p;` if `x` was always `false` before.
- `shouldWriteHeader` is now defaulted to `true` as the "appendMode" branch will always update it. This removes the "writeMode" branch in this section.

```cpp
bool PcapFileWriterDevice::open(bool appendMode)
{
	if (m_DeviceOpened || m_PcapFile.is_open())
	{
		PCPP_LOG_ERROR("File already opened");
		return false;
	}

	auto flags = std::ios::binary | std::ios::out;
	// Can also be the one liner. Tbh, I kinda prefer it, than the full `if`.
	// flags |= appendMode ? std::ios::in | std::ios::app : 0;
	if(appendMode)
	{
		flags |= std::ios::in | std::ios::app;
	}
	m_PcapFile.open(m_FileName, flags);
	
	if (!m_PcapFile.is_open())
	{
		PCPP_LOG_ERROR("Failed to open file: " << m_FileName);
		return false;
	}

	m_NeedsSwap = false;  // Default to no swap, write mode should always write in native order.
	bool shouldWriteHeader = true;  // Defaults to 'true' as write mode will always need it.
	if (appendMode)
	{
		auto checkHeaderResult = checkHeader(m_PcapFile, m_Precision, m_PcapLinkLayerType);

		if (checkHeaderResult.result == CheckHeaderResult::Result::HeaderError)
		{
			PCPP_LOG_ERROR(checkHeaderResult.error);
			return false;
		}

		// Update open params from header result.
		m_NeedsSwap = checkHeaderResult.needsSwap;
		shouldWriteHeader = checkHeaderResult.result == CheckHeaderResult::Result::HeaderNeeded;

		m_PcapFile.seekg(0, std::ios::end);
	}

	if (shouldWriteHeader && !writeHeader(m_PcapFile, m_Precision, PCPP_MAX_PACKET_SIZE, m_PcapLinkLayerType))
	{
		return false;
	}

	m_DeviceOpened = true;
	return true;
}
```

seladb: Thanks for cleaning it up! Updated in this commit: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/b79e248acd8f8489102069d514df32f0eed5906d

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -55,16 +63,126 @@ namespace pcpp
 		uint32_t len;
 	};
 
-	static bool checkNanoSupport()
+	LinkLayerType toLinkLayerType(uint32_t value)
```
tigercosmos, line 66: I think you can check if it is invalid at the beginning and simply call `static_cast<LinkLayerType>(value)` to avoid the long `switch`.

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -374,162 +533,280 @@ namespace pcpp
...
 	{
-		if (isOpened())
+		if (m_DeviceOpened || m_PcapFile.is_open())
 		{
-			// TODO: Ambiguity in API
-			//   If appendMode is required but the file is already opened in write mode.
-			PCPP_LOG_DEBUG("Pcap descriptor already opened. Nothing to do");
-			return true;
+			PCPP_LOG_ERROR("File already opened");
+			return false;
```
tigercosmos: Maybe seperate the checks to make the error message correct.

Pcap++/src/PcapFileDevice.cpp
Dimi1010: LinkLayerType is not a packed enum, tho. There are gaps, e.g. there are no enum names for values between 14(iirc) and 50.

Pcap++/src/PcapFileDevice.cpp
Dimi1010: Do we even need to store `m_DeviceOpened`? We could make it to instead infer if the device is opened purely, though `m_PcapFile.is_open()`?

Pcap++/src/PcapFileDevice.cpp
seladb: Yes, there are more gaps, for example: between 51 - 100, 101 - 104, 114 - 117, and more. There's is no easy way to do this conversion

Pcap++/src/PcapFileDevice.cpp
seladb: If we rely just on `m_PcapFile.is_open()` we need to always close the file if something goes wrong in `open()` after we open it. We can use a temp file for this and transfer the ownership to `m_PcapFile` in the end of `open()`, like this:
```cpp
std::fstream pcapFile;
pcapFile.open(m_FileName, flags);

// The rest of the method

m_PcapFile = std::move(pcapFile);
return true;
```

We can do the same in `PcapFileReaderDevice`. Please let me know what you think.

**EDIT**: if we do that we also need to override `isOpened()` and ignore `m_DeviceOpened` that is defined in `IFileDevice`...

seladb: Implemented in this commit: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/8174ba1dc75b41aebb0e808da5bba161f8976d76

If you think we shouldn't I can revert this commit.

Dimi1010: Hmm, I like the construct as local -> move, as we don't maintain any state if open fails midway, for example if header is partial, broken or something.

> EDIT: if we do that we also need to override isOpened() and ignore m_DeviceOpened that is defined in IFileDevice...

Honestly, forgot about that detail. I think it would be better to keep the `m_DeviceOpened` as it was for now to keep the PR simple.

When this PR is merged I think we can simplify some of the upper interfaces. Some examples:
- `IFileDevice` inherits `IPcapDevice` - this is only because `m_PcapDescriptor` was used in pcap parsing.
  This leads to all file devices having a pcap handle member. I think we can drop that and inherit directly from `IFilterableDevice`?
- Drop the `close` override on `IFileDevice` (and the destructor body) as there wouldn't be a pcap descriptor to close.

seladb: > When this PR is merged I think we can simplify some of the upper interfaces. Some examples:
> 
>     * `IFileDevice` inherits `IPcapDevice` - this is only because `m_PcapDescriptor` was used in pcap parsing.
>       This leads to all file devices having a pcap handle member. I think we can drop that and inherit directly from `IFilterableDevice`?
> 
>     * Drop the `close` override on `IFileDevice` (and the destructor body) as there wouldn't be a pcap descriptor to close.

Yes, I was planning to do it in follow up PRs. The ultimate goal is to not require libpcap at all, like the other capture engines, but it will take time because we need to figure out the BPF filtering...



> Honestly, forgot about that detail. I think it would be better to keep the `m_DeviceOpened` as it was for now to keep the PR simple.

So do you suggest I revert the override of `isOpened()` and continue using `m_DeviceOpened` for now?

Dimi1010: > The ultimate goal is to not require libpcap at all, like the other capture engines, but it will take time because we need to figure out the BPF filtering...

Hmm. Curious as to why tho. I see the benefit for standalone read / write, but I think keeping the `libpcap` dependency for filtering support seems ok? The only issue would be modularising the code so that it can be compiled without filter support.

> So do you suggest I revert the override of `isOpened()` and continue using `m_DeviceOpened` for now?

Yeah, we can leave it to use the logic from `IFileDevice` for now.

seladb: > Yeah, we can leave it to use the logic from `IFileDevice` for now.

Sure, done in this commit: https://github.com/seladb/PcapPlusPlus/pull/2034/commits/f6b92eaf54f8532f469067070cdf4bab253e5705

seladb: > Hmm. Curious as to why tho. I see the benefit for standalone read / write, but I think keeping the `libpcap` dependency for filtering support seems ok? The only issue would be modularising the code so that it can be compiled without filter support.

Today libpcap / Npcap / WinPcap are different from the other capture engines we support because they are always linked, even if an app doesn't need them. For example: if someone wants to write an application that uses PcapPlusPlus with DPDK, they still have to link libpcap.

The long term goal is to remove this dependency:
- People can build PcapPlusPlus with any capture engine, and libpcap is just one of the options
- libpcap is just used as a capture engine, same as DPDK, WinDivert, AF_XDP, etc. That's why handling files shouldn't depend on libpcap

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -374,162 +541,283 @@ namespace pcpp
...
 		}
 
-		m_PcapDumpHandler = reinterpret_cast<pcap_dumper_t*>(m_File);
+		bool result = true;
+		std::for_each(packets.begin(), packets.end(), [&](const auto& packet) {
+			if (!writePacket(*packet))
+			{
+				result = false;
+			}
+		});
```
Dimi1010: IMO, in post-C++11 `range-for` is a cleaner syntax than `std::for_each`, unless you want to use post-C++17 execution policies.

```suggestion
		for (auto const& packet : packets)
		{
			if (!writePacket(*packet))
			{
				result = false;
			}
		}
```

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -374,162 +541,283 @@ namespace pcpp
...
+		if (!m_PcapFile.write(reinterpret_cast<const char*>(&packetHeader), sizeof(packetHeader)))
+		{
+			PCPP_LOG_ERROR("Cannot write the packet header to file");
+			reportPacketDropped();
 			return false;
 		}
 
-		LinkLayerType linkLayerType = static_cast<LinkLayerType>(pcapFileHeader.linktype);
-		if (linkLayerType != m_PcapLinkLayerType)
+		if (!m_PcapFile.write(reinterpret_cast<const char*>(packet.getRawData()), packet.getRawDataLen()))
```
Dimi1010, line 638: Do we want to check if `packet.getRawData()` is `nullptr`?
What should happen on the following?
```cpp
RawPacket pack; // Default ctor makes data ptr = nullptr.
writer.writePacket(pack);
```

Pcap++/src/PcapFileDevice.cpp
seladb: Good catch! Fixed in https://github.com/seladb/PcapPlusPlus/pull/2034/commits/59d557f6fb4c6709afe8f38f0ee5ae06f702b6ff

Pcap++/src/PcapFileDevice.cpp
seladb: I'm not sure we need to check that. What would happen is that we only write the packet header but no packet data. I generated such a file and opened it in Wireshark. It opens and shows a "malformed packet":
image

PR #2042, opened 2026-01-08
Improve parsing options support.

Dimi1010:
This PR improves the parsing options by adding incremental parsing.

The parsing procedure has been extracted from the `setRawData` procedure and placed into `parsePacket`.
The new procedure can be called repeatedly on the same packet with different options and will incrementally parse the packet.

The packet parsing options have been grouped into the struct `ParseOptions` for better method readability.

### DIFF
--- a/.github/workflows/auto_update.yml
+++ b/.github/workflows/auto_update.yml
@@ -13,11 +13,11 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           ref: dev
       - name: Setup Python
-        uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6.1.0
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: "3.8.x"
       - name: Run update
@@ -25,7 +25,7 @@ jobs:
             pip install pre-commit
             pre-commit autoupdate
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@98357b18bf14b5342f975ff684046ec3b2a07725 # v8.0.0
+        uses: peter-evans/create-pull-request@c0f553fe549906ede9cf27b5156039d195d2ece0 # v8.1.0
         with:
           token: ${{ secrets.PAT }}
           author: GitHub <noreply@github.com>
@@ -46,11 +46,11 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           ref: dev
       - name: Setup Python
-        uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6.1.0
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: "3.9.x"
       - name: Run update
@@ -58,7 +58,7 @@ jobs:
             python3 3rdParty/OUIDataset/create_oui_data.py
             mv -f PCPP_OUIDataset.json 3rdParty/OUIDataset/PCPP_OUIDataset.json
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@98357b18bf14b5342f975ff684046ec3b2a07725 # v8.0.0
+        uses: peter-evans/create-pull-request@c0f553fe549906ede9cf27b5156039d195d2ece0 # v8.1.0
         with:
           token: ${{ secrets.PAT }}
           author: GitHub <noreply@github.com>
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -26,7 +26,7 @@ jobs:
     container: seladb/alpine317
     steps:
       - name: Checkout code
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       # Checkout is performed out of the container and doesn't match our user
       - name: Fix checkout ownership
@@ -92,7 +92,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       # Checkout is performed out of the container and doesn't match our user
       - name: Fix checkout ownership
@@ -106,7 +106,7 @@ jobs:
 
       - name: Restore Ccache
         id: ccache-restore
-        uses: actions/cache/restore@9255dc7a253b0ccc959486e2bca901246202afeb # v5.0.1
+        uses: actions/cache/restore@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: |
             ${{ env.CCACHE_DIR }}
@@ -171,7 +171,7 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
       - name: Save Ccache
-        uses: actions/cache/save@9255dc7a253b0ccc959486e2bca901246202afeb # v5.0.1
+        uses: actions/cache/save@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: ${{ env.CCACHE_DIR }}
           key: ${{ steps.ccache-restore.outputs.cache-primary-key }}
@@ -183,7 +183,7 @@ jobs:
       image: ubuntu2404
     steps:
       - name: Checkout code
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       # Checkout is performed out of the container and doesn't match our user
       - name: Fix checkout ownership
@@ -191,7 +191,7 @@ jobs:
 
       - name: Restore Ccache
         id: ccache-restore
-        uses: actions/cache/restore@9255dc7a253b0ccc959486e2bca901246202afeb # v5.0.1
+        uses: actions/cache/restore@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: |
             ${{ env.CCACHE_DIR }}
@@ -241,7 +241,7 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
       - name: Save Ccache
-        uses: actions/cache/save@9255dc7a253b0ccc959486e2bca901246202afeb # v5.0.1
+        uses: actions/cache/save@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: ${{ env.CCACHE_DIR }}
           key: ${{ steps.ccache-restore.outputs.cache-primary-key }}
@@ -267,11 +267,11 @@ jobs:
           fi
 
       - name: Checkout code
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Restore Ccache
         id: ccache-restore
-        uses: actions/cache/restore@9255dc7a253b0ccc959486e2bca901246202afeb # v5.0.1
+        uses: actions/cache/restore@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: ${{ env.CCACHE_DIR }}
           key: ${{ matrix.image }}-ccache-${{ github.run_id }}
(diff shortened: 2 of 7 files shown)


Packet++/header/Packet.h
```diff
@@ -150,6 +202,43 @@ namespace pcpp
...
+		/// Unless @p fullReparse is set to true, the procedure attempts to reuse existing layers where possible,
+		/// enabling incremental parsing.
+		///
+		/// If @p fullReparse is true, any existing layer pointers owned by this Packet instance are destroyed.
+		///
+		/// @param[in] options Parsing options to configure the parsing behavior.
+		/// @param[in] fullReparse If 'true', forces a complete re-parse of the packet, disregarding any existing layer
+		/// structure.
+		/// @throws std::runtime_error if there is no RawPacket associated with this Packet instance.
+		void parsePacket(ParseOptions options, bool fullReparse = false);
```
Dimi1010: Would the second parameter be more explicit as `fullReparse = false (default)` or `incrementalParse = true (default)`? 🤔

Packet++/header/Packet.h
```diff
@@ -103,6 +137,24 @@ namespace pcpp
...
+		/// when it is no longer needed. If 'false', the caller retains ownership and is responsible for its disposal.
+		/// @param options Parsing options to configure the parsing behavior.
+		explicit Packet(RawPacket* rawPacket, bool takeOwnership, ParseOptions options);
+
+		/// @brief A constructor for creating a packet out of already allocated RawPacket without parsing it.
+		///
+		/// @param[in] rawPacket The raw packet to associate with this Packet instance.
+		/// @param[in] takeOwnership If 'true' the Packet will take ownership of the rawPacket pointer and dispose of it
+		/// when it is no longer needed. If 'false', the caller retains ownership and is responsible for its disposal.
+		explicit Packet(NoParseTag, RawPacket* rawPacket, bool takeOwnership);
```
seladb: Why do we want to allow creating a `Packet` without parsing it? 🤔 
`RawPacket` is a non-parsed packet...

Packet++/header/Packet.h
```diff
@@ -103,6 +137,24 @@ namespace pcpp
...
+		/// @brief A constructor for creating a packet out of already allocated RawPacket.
+		///
+		/// The packet saves a reference to the RawPacket (data isn't copied) and the RawPacket is parsed according to
+		/// the provided parsing options.
+		///
+		/// @param[in] rawPacket The raw packet to parse.
+		/// @param[in] takeOwnership If 'true' the Packet will take ownership of the rawPacket pointer and dispose of it
+		/// when it is no longer needed. If 'false', the caller retains ownership and is responsible for its disposal.
+		/// @param options Parsing options to configure the parsing behavior.
+		explicit Packet(RawPacket* rawPacket, bool takeOwnership, ParseOptions options);
```
seladb: As long as we don't have more parsing options, this c'tor is a duplicate of the other c'tors which I think might be confusing to users.

In the future if we extend it to more options we may thing about adding this `ParseOptions` struct

Packet++/header/Packet.h
Dimi1010: Hmm, when it's put that way, it doesn't really make much sense at the moment. I will remove those overloads.

I think I was going with the idea of allowing the user to do a construct -> manually call `parsePacket`, but we can do without it.

Packet++/header/Packet.h
Dimi1010: The `ParseOptions` struct is mostly, so there isn't a need to deal with 3 overloads on `parsePacket` with different default parameters.

Otherwise, it would have:
- Protocol = unk, OSI = unk, incFlag = true
- Protocol, incFlag = true
- OSI, incFlag = true

I think having just one signature with an options param is cleaner?

The overloads in the constructor and `setRawPacket` are mostly for API symmetry. They can be removed, but that would mean that users can't pass the options in from that API.

seladb: Do we want to deprecate the other constructors? Having all of them are a bit confusing because because it's not easy for the user to understand which to use 🤔

Dimi1010: I'm not sure... I am unsure if `ParseOptions` has an easy way to set the options just yet, and I don't want to deprecate the old ones prior to that.

On the one hand we could do with constructors (`ParseOptions()` / `ParseOptions(Proto)` / `ParseOptions(OSI)` / `ParseOptions(Proto, OSI)`).

On the other hand I really like keeping the options a POD with aggregate initialization (no user constructors), because it gives flexibility for users that use >= C++20. It introduces [designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization.html#Designated_initializers:~:text=without%20any%20element-,Designated%20initializers,-The%20syntax%20forms) which allow stuff like this:

```cpp
pcpp::ParseOptions opts = {
	.parseUntil = pcpp::IPv6,
	// .parseUntilLayer is set to default
};

opts = {
	// .parseUntil is set to default
	.parseUntilLayer = pcpp::OsiModelNetworkLayer,
};

opts = {
	.parseUntil = pcpp::IPv6,
	.parseUntilLayer = pcpp::OsiModelNetworkLayer,
}
```
Which I think is very clean for configuration structs. Downside is that pre-C++20 has to either do full aggregate or use a temporary variable and set it manually.

There is a middle ground where we have no user constructors but have static factory methods, which would allow both, but is more complex?

What do you think?

Packet++/header/Packet.h
Dimi1010: Removed the overloads that allow deferred parsing. d44047c6ee1c54279c23493e8afe5878f614bbb2

Packet++/header/Packet.h
seladb: > I'm not sure... I am unsure if `ParseOptions` has an easy way to set the options just yet, and I don't want to deprecate the old ones prior to that.

Keeping both is confusing to users because it's unclear what's the difference between them. That's my main concern about this PR. This class already has 6 constructors (!!) so adding more won't help the users.



> What do you think?

If we eventually decide to do it, I'd keep is as a simple struct with no c'tors and default values, and people can use it however they want.

Dimi1010: > Keeping both is confusing to users because it's unclear what's the difference between them. That's my main concern about this PR. This class already has 6 constructors (!!) so adding more won't help the users.

OK, we can deprecate the rest and leave the main signature to be:
 `Packet(RawPacket* rawPacket, bool takeOwnership, ParseOptions options = {})`

I think this should simplify the signature and cut down the number of ctors by 3.
It would involve updating a bunch of usages for Packet, so I will remove the new ctors from this PR and readd them in a separate one. The main content of this PR is the separation of parsing into a dedicated function.

> If we eventually decide to do it, I'd keep is as a simple struct with no c'tors and default values, and people can use it however they want.

I agree. The working version I ended up with is this:
```cpp
struct ParseOptions
{
	ProtocolTypeFamily parseUntilProtocol = UnknownProtocol;
	OsiModelLayer parseUntilLayer = OsiModelLayerUnknown;

	static ParseOptions parseUntil(ProtocolTypeFamily protocol)
	{
		ParseOptions options;
		options.parseUntilProtocol = protocol;
		return options;
	}

	static ParseOptions parseUntil(OsiModelLayer osiLayer)
	{
		ParseOptions options;
		options.parseUntilLayer = osiLayer;
		return options;
	}
};
```

It keeps the aggregate, but allows simple updates of old code:
```cpp
// Old
Packet(rawPtr, false, pcpp::IPv4);
// New
Packet(rawPtr, false, ParseOptions::parseUntil(pcpp::IPv4));

// Old
Packet(rawPtr, false, OsiModelDataLinkLayer);
// New
Packet(rawPtr, false, ParseOptions::parseUntil(OsiModelDataLinkLayer));
```

The helper functions will be added in the separate PR too tho.

Dimi1010: Something I also noticed while rolling back the tests.

In the old constructors there is a silent bug where the main signature (`explicit Packet(RawPacket* rawPacket, bool freeRawPacket, ProtocolType parseUntil, OsiModelLayer parseUntilLayer);`)
takes `ProtocolType` and not `ProtocolTypeFamily` meaning that if a `ProtocolTypeFamily` is provided it silently gets casted to just the first element of the family. (e.g. pcpp::IP gets cut down to pcpp::IPv6).

Packet++/header/Packet.h
```diff
@@ -10,6 +10,32 @@
 /// @brief The main namespace for the PcapPlusPlus lib
 namespace pcpp
 {
+	/// @brief Options struct for configuring packet parsing behavior.
+	///
+	/// The parsing options are combined. If multiple stop conditions are specified, parsing stops
+	/// as soon as any condition is met.
+	struct ParseOptions
```
seladb: Won't it make sense to move this struct inside of the `Packet` class? 🤔

Packet++/src/Packet.cpp
seladb, line 1: General comment: a few methods were moved within the file making the actual changes harder to track. I'd advise keeping methods where they were

Tests/Packet++Test/Tests/PacketTests.cpp
```diff
@@ -1095,3 +1095,91 @@ PTF_TEST_CASE(PacketParseMultiLayerTest)
 
 	PTF_ASSERT_EQUAL(actualNumOfBgpMessages, expectedNumOfBgpMessages);
 }
+
+PTF_TEST_CASE(PacketIncrementalParseTest)
+{
+	using namespace pcpp;
```
seladb: We never do that, can you remove it?

Tests/Packet++Test/Tests/PacketTests.cpp
```diff
@@ -1095,3 +1095,91 @@ PTF_TEST_CASE(PacketParseMultiLayerTest)
...
+
+	PTF_ASSERT_EQUAL(igmpPacket.getLayerOfType<pcpp::IPLayer>(), ipLayer);
+
+	// Incremental parsing shouldn't remove layers previously parsed beyond the specified layer
+	PTF_ASSERT_EQUAL(igmpPacket.getLastLayer(), igmpPacket.getLayerOfType<pcpp::PacketTrailerLayer>());
+}
+
+PTF_TEST_CASE(PacketFullReparseTest)
+{
+	using namespace pcpp;
```
seladb: ditto

Tests/Packet++Test/Tests/PacketTests.cpp
```diff
@@ -1095,3 +1095,91 @@ PTF_TEST_CASE(PacketParseMultiLayerTest)
...
+	PTF_ASSERT_NOT_NULL(igmpPacket.getLayerOfType<pcpp::PacketTrailerLayer>());
+
+	// Disable incremental parsing and do a full reparse up to IP layer.
+	// The IGMP and Trailer layers shouldn't be present after the reparse.
+	ParseOptions pOpts;
+	pOpts.parseUntilProtocol = pcpp::IP;
+	igmpPacket.parsePacket(pOpts, false);
+
+	PTF_ASSERT_TRUE(igmpPacket.isPacketOfType(pcpp::IPv4));
+	PTF_ASSERT_TRUE(igmpPacket.isPacketOfType(pcpp::Ethernet));
```
seladb, line 1175: Maybe we should check that the pointers of these layers don't equal the pointers in the previous parsing?

Packet++/header/Packet.h
Dimi1010: Possibly, but it makes the code more verbose for not much benefit? 

From https://github.com/seladb/PcapPlusPlus/pull/2042#discussion_r2745853644 `Packet(rawPacket, true, ParseOptions { /* ... */})` would have to be `Packet(rawPacket, true, Packet::ParseOptions { /* ... */})` and so on.

I think keeping it separate is good enough and offers a bit more concise API?

Packet++/src/Packet.cpp
Dimi1010: Sure, I can revert that or extract it to a separate PR. The moved methods were reordering the constructors so they aren't scattered across the cpp.

Tests/Packet++Test/Tests/PacketTests.cpp
Dimi1010: Eh, we can, but it would not be entirely accurate assert.

There is a *very* small chance that the allocator will allocate the newly created layers in the same memory locations as the previous layers. That would still be valid behaviour as the previous layers would already be deleted at the start of the reparse procedure. If we have such an assert, in that situation the test would fail, even though it is valid?

Packet++/src/Packet.cpp
Dimi1010: Fixed d11460570bda72ee86e9a9027f3d10ae30c7eb63

Tests/Packet++Test/Tests/PacketTests.cpp
Dimi1010: Fixed 91c96b649128744d133b502660478c7c9f735bcd

Tests/Packet++Test/Tests/PacketTests.cpp
Dimi1010: Fixed 91c96b649128744d133b502660478c7c9f735bcd

Packet++/header/Packet.h
seladb: So maybe at least call it `PacketParseOptions`?
When looking at the API `pcpp::ParseOptions` seems a bit weird and out of context

Tests/Packet++Test/Tests/PacketTests.cpp
seladb: We do a similar check here:
https://github.com/Dimi1010/PcapPlusPlus/blob/91c96b649128744d133b502660478c7c9f735bcd/Tests/Packet%2B%2BTest/Tests/PacketTests.cpp#L1129-L1130

We can also say that if we have a bug in `parsePacket()` and it destroys all layers, there is a chance the pointers will still be equal. 

I think we should add the pointer check here and see if it fails, if it is we can remove it

Dimi1010: Fixed 26b21a7260fcd1bd1e441723e0740636d7861b2c

Packet++/header/Packet.h
Dimi1010: Fixed 1cfaf6aa99c3f01bff05c0f24c801ff8c043feb5

Tests/Packet++Test/Tests/PacketTests.cpp
Dimi1010: Welp, its failing more often than not. When running locally, I get like 3 fails, 1 success ratio. :/
I assume because most allocators try to reuse the memory they already have, and that deleted slot is "just" the right size to be thrown back in.

Dimi1010: Reverted 9bf29cd007be3a7baa24413d4a7b13e926dde8c1

seladb: Sure thanks! 👍

PR #2044, opened 2026-01-14
Cleanup file devices

seladb:
After [decoupling pcap files from libpcap](https://github.com/seladb/PcapPlusPlus/pull/2034) we can clean up some code in the file devices:
1. Change `IFileDevice` to inherit from `IFilterableDevice` and `IPcapStatisticsProvider` instead of `IPcapDevice`
2. Remove `m_DeviceOpened` from `IFileDevice` and implement  `isOpened()` in each file device using the specific handle
3. Add `BpfFilterWrapper` to `IFileDevice` and implement `doUpdateFilter()` once in this class
4. Update to C++11/14 - use `override` and `delete`
5. Remove unused includes
6. Add filtering to SNOOP reader device
7. Add tests for SNOOP reader device

**Note:** there is more cleanup and tests we can do for SNOOP devices, but I think we can leave it to a separate PR

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -751,7 +751,7 @@ jobs:
           release: ${{ matrix.version }}
           usesh: true
           prepare: |
-            pkg install -y pcre2 python bash cmake git gmake gsed libpcap tcpreplay libffi openssl py311-cryptography
+            pkg install -y pcre2 python bash cmake git gmake gsed libpcap tcpreplay libffi openssl rust
             sysctl kern.ipc.maxsockbuf=16777216
           run: |
             set -e


Pcap++/src/PcapFileDevice.cpp
```diff
@@ -189,32 +185,16 @@ namespace pcpp
 	// IFileDevice members
 	// ~~~~~~~~~~~~~~~~~~~
 
-	IFileDevice::IFileDevice(const std::string& fileName) : IPcapDevice()
+	IFileDevice::IFileDevice(const std::string& fileName) : IFilterableDevice(), IPcapStatisticsProvider()
```
Dimi1010: nit: I don't think we actually need the `: IFilterableDevice(), IPcapStatisticsProvider()` part. The base classes will be default constructed by default, if they aren't specified.

seladb: You're right! Addressed in https://github.com/seladb/PcapPlusPlus/pull/2044/commits/a94547f50516518d6a45fb511fe3444926993795

PR #2045, opened 2026-01-16
Fix run_tests_windows to search through more than 1 interface.

Dimi1010:
This PR refactors the `find_interface` function fixing and fixes a bug where the script only searched the first device interface repeatedly.

- Fixed getting a device IP from GUID, on machines with more than 1 interface. Previous behaviour did not respect GUID and always returned the ipv4 of the first device if any.
- Added `InterfaceList` to cache the list returned from `scapy.arch.windows.get_windows_if_list()`.
- Moved `get_ip_by_guid` to be a member function of `InterfaceList`.
- Removed `validate_ipv4_address` as it was inlined directly in `get_ip_by_guid`.

### DIFF
--- a/Packet++/header/Layer.h
+++ b/Packet++/header/Layer.h
@@ -264,6 +264,22 @@ namespace pcpp
 			return m_NextLayer != nullptr;
 		}
 
+		/// @brief Construct the next layer in the protocol stack. No validation is performed on the data.
+		///
+		/// This overload infers the Packet from the current layer.
+		///
+		/// @tparam T The type of the layer to construct
+		/// @tparam Args The types of the arguments to pass to the layer constructor
+		/// @param data The data to construct the layer from
+		/// @param dataLen The length of the data
+		/// @param extraArgs Extra arguments to be forwarded to the layer constructor
+		/// @return The constructed layer
+		template <typename T, typename... Args>
+		Layer* constructNextLayer(uint8_t* data, size_t dataLen, Args&&... extraArgs)
+		{
+			return constructNextLayer<T>(data, dataLen, getAttachedPacket(), std::forward<Args>(extraArgs)...);
+		}
+
 		/// Construct the next layer in the protocol stack. No validation is performed on the data.
 		/// @tparam T The type of the layer to construct
 		/// @tparam Args The types of the arguments to pass to the layer constructor
@@ -283,6 +299,27 @@ namespace pcpp
 			Layer* newLayer = new T(data, dataLen, this, packet, std::forward<Args>(extraArgs)...);
 			setNextLayer(newLayer);
 			return newLayer;
+		}
+
+		/// @brief Try to construct the next layer in the protocol stack with a fallback option.
+		///
+		/// This overload infers the Packet from the current layer.
+		///
+		/// The method checks if the data is valid for the layer type T before constructing it by calling
+		/// T::isDataValid(data, dataLen). If the data is invalid, it constructs the layer of type TFallback.
+		///
+		/// @tparam T The type of the layer to construct
+		/// @tparam TFallback The fallback layer type to construct if T fails
+		/// @tparam Args The types of the extra arguments to pass to the layer constructor of T
+		/// @param[in] data The data to construct the layer from
+		/// @param[in] dataLen The length of the data
+		///	@param[in] extraArgs Extra arguments to be forwarded to the layer constructor of T
+		/// @return The constructed layer of type T or TFallback
+		template <typename T, typename TFallback, typename... Args>
+		Layer* tryConstructNextLayerWithFallback(uint8_t* data, size_t dataLen, Args&&... extraArgs)
+		{
+			return tryConstructNextLayerWithFallback<T, TFallback>(data, dataLen, getAttachedPacket(),
+			                                                       std::forward<Args>(extraArgs)...);
 		}
 
 		/// Try to construct the next layer in the protocol stack with a fallback option.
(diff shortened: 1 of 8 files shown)


ci/run_tests/run_tests_windows.py
```diff
@@ -10,55 +10,61 @@
...
+
+    all_interfaces = scapy.arch.windows.get_windows_if_list()
+    for interface in all_interfaces:
+        if_guid: str | None = interface.get("guid", None)
+        if if_guid is None:
+            continue
+
+        # Not a NPF guid? Can we even get those on a machine with NPcap / WinPcap?
+        if if_guid not in npf_guids:
+            continue
```
Dimi1010: Any idea if scapy returns non-`NPF_` interfaces? If not, we could possibly fetch the devices directly from scapy, skipping the tcpreplay call.

Dimi1010:
@seladb Fix is because I had issues with testing the run script on my windows machine that had more than 1 interface.

seladb:
> @seladb Fix is because I had issues with testing the run script on my windows machine that had more than 1 interface.

@Dimi1010 which issues did you have? I'm not sure what this PR is trying to fix. Can you reduce the changes to only fix the issue you found so it's easier to review?

Dimi1010:
> > @seladb Fix is because I had issues with testing the run script on my windows machine that had more than 1 interface.
> 
> @Dimi1010 which issues did you have? I'm not sure what this PR is trying to fix. Can you reduce the changes to only fix the issue you found so it's easier to review?

The main issue is this segment:
```python
def get_ip_by_guid(guid): 
     interfaces = scapy.arch.windows.get_windows_if_list() 
     for iface in interfaces: 
         ips = iface.get("ips", []) 
         # Find the first valid IPv4 address inside iface["ips"]. If no address is found, return None 
         return next(filter(validate_ipv4_address, ips), None) 
     # Return None if no matching interface is found 
     return None
```

It always goes into the first device and either hits its ip or a `None`. It never goes and checks any other device. The GUID param is also ignored making the function name misleading. 

If you have 2 interfaces and the second one is connected, but the first one isn't (2 eth ports / eth + wifi) the functions will never find the connected interface.

The rest is mostly simplifications and inlining of the logic.

seladb:
> > > @seladb Fix is because I had issues with testing the run script on my windows machine that had more than 1 interface.
> > 
> > 
> > @Dimi1010 which issues did you have? I'm not sure what this PR is trying to fix. Can you reduce the changes to only fix the issue you found so it's easier to review?
> 
> The main issue is this segment:
> 
> ```python
> def get_ip_by_guid(guid): 
>      interfaces = scapy.arch.windows.get_windows_if_list() 
>      for iface in interfaces: 
>          ips = iface.get("ips", []) 
>          # Find the first valid IPv4 address inside iface["ips"]. If no address is found, return None 
>          return next(filter(validate_ipv4_address, ips), None) 
>      # Return None if no matching interface is found 
>      return None
> ```
> 
> It always goes into the first device and either hits its ip or a `None`. It never goes and checks any other device. The GUID param is also ignored making the function name misleading.
> 
> If you have 2 interfaces and the second one is connected, but the first one isn't (2 eth ports / eth + wifi) the functions will never find the connected interface.
> 
> The rest is mostly simplifications and inlining of the logic.

So maybe fix this issue only in this PR?

Dimi1010:
> So maybe fix this issue only in this PR?

Should be cleaner now.

Dimi1010:
No clue why the Fuzz CI fails, but it should be unrelated as it doesn't use windows at all.

ci/run_tests/run_tests_windows.py
```diff
@@ -10,25 +12,30 @@
...
+        if interface is None:
+            return None
 
+        ips = interface.get("ips", [])
+        for ip in ips:
+            try:
+                ipv4 = IPv4Address(ip)
+                return ipv4
+            except ValueError:
+                pass
```
seladb: nit: you can do:
```python
for ip in interface.get("ips", []):
    try:
        return IPv4Address(ip)
    except ValueError:
        pass
```

ci/run_tests/run_tests_windows.py
```diff
@@ -10,25 +12,30 @@
...
+    def get_ipv4_by_guid(self, guid: str) -> IPv4Address | None:
+        interface = self.interfaces.get(guid, None)
+        if interface is None:
+            return None
 
+        ips = interface.get("ips", [])
+        for ip in ips:
+            try:
+                ipv4 = IPv4Address(ip)
+                return ipv4
```
Dimi1010:
```suggestion
                return IPv4Address(ip)
```

ci/run_tests/run_tests_windows.py
Dimi1010: Fixed b4ac8199ee9c4d7de05cad54d1e316b3d8e8a047 (and f415d9a16ee7cad7fd5254bfffe4af6df3e7c691 since I forgot the for loop change)

ci/run_tests/run_tests_windows.py
```diff
@@ -10,25 +12,29 @@
...
+            iface["guid"]: iface for iface in interfaces if "guid" in iface
+        }
+
+    def get_ipv4_by_guid(self, guid: str) -> IPv4Address | None:
+        interface = self.interfaces.get(guid, None)
+        if interface is None:
+            return None
 
+        ips = interface.get("ips", [])
+        for ip in ips:
```
Dimi1010:
```suggestion
        for ip in interface.get("ips", []):
```

PR #2056, opened 2026-01-29
Remove `IPcapDevice` and move relevant logic to `PcapLiveDevice`

seladb:
- Remove `IPcapDevice` and move its responsibilities into `PcapLiveDevice`
- Move `internal::PcapHandle` into `PcapLiveDevice` and own it there.
- Standardize to `pcpp::PcapStats` across API; update callbacks and users.

Breaking changes:
- `IPcapDevice` removed
- `PcapStats` extracted from `IPcapDevice::PcapStats`

### DIFF
--- a/ci/clang-tidy-all.sh
+++ b/ci/clang-tidy-all.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -e
 
-IGNORE_LIST=".*dirent.* .*DpdkDevice* .*KniDevice* .*MBufRawPacket* .*PfRingDevice* .*RemoteDevice* .*XdpDevice* .*WinPcap* .*WinDivert* .*FuzzTarget*"
+IGNORE_LIST=".*dirent.* .*DpdkDevice* .*KniDevice* .*MBufRawPacket* .*PfRingDevice* .*RemoteDevice* .*XdpDevice* .*WinPcap* .*WinDivert* .*Fuzz*"
 
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "${SCRIPT}")


Pcap++/header/PcapLiveDevice.h
```diff
@@ -343,6 +435,11 @@ namespace pcpp
 		/// this list from PcapLiveDevice as well
 		const std::vector<IPv4Address>& getDnsServers() const;
 
+		/// A static method for retrieving pcap lib (libpcap/WinPcap/etc.) version information. This method is actually
+		/// a wrapper for [pcap_lib_version()](https://www.tcpdump.org/manpages/pcap_lib_version.3pcap.html)
+		/// @return A string containing the pcap lib version information
+		static std::string getPcapLibVersionInfo();
```
Dimi1010, line 437: Somewhat unrelated to this PR directly, but maybe we should change this to a free function?
The version of libpcap is not really related to the devices directly?

seladb: When I finish the refactoring (probably in the next PR) libpcap / WinPcap / Npcap will be optional and the main interface to libpcap will be `PcapLiveDevice` and its inherit classes. In that sense, keeping this method as a static method on the class makes sense. Another option is to put it under `PcapLiveDeviceList`, but the way it is is good enough...

Pcap++/header/PcapDevice.h
seladb, line 1: This file is only left with `PcapStats` and `IPcapStatisticsProvider` which isn't necessarily tied to libpcap, but in order to maintain backward compatibility I think we should keep these names for now. The question is should we move it to `Device.h` or keep it in `PcapDevice.h`. Please let me know what you think

Pcap++/header/PcapLiveDevice.h
Dimi1010: > When I finish the refactoring (probably in the next PR) libpcap / WinPcap / Npcap will be optional and the main interface to libpcap will be `PcapLiveDevice` and its inherit classes.

Wouldn't that be confusing due to `PcapLiveDevice` mixing both per-device methods and general libpcap methods?

Dimi1010 (approved):

Pcap++/header/PcapLiveDevice.h
seladb: I guess there are pros and cons but it's quite old API and people don't complain so I guess it's good enough 🙂

PR #2057, opened 2026-01-31
Make libpcap / WinPcap / Npcap optional

seladb:
Allow building PcapPlusPlus without libpcap / WinPcap / Npcap.

This PR includes:
- Adding a new `PCAPPP_USE_PCAP` CMake option with a default of `ON` to keep it backward compatible. Turning it off will build everything that doesn't require libpcap / WinPcap / Npcap
- When libpcap / WinPcap / Npcap is turned off - skip tests that depend on it
- When libpcap / WinPcap / Npcap is turned off - don't build examples that depend on it. Also skip relevant example tests
- Add CI job that tests `PCAPPP_USE_PCAP=OFF`

### DIFF
--- a/Packet++/header/Packet.h
+++ b/Packet++/header/Packet.h
@@ -10,6 +10,32 @@
 /// @brief The main namespace for the PcapPlusPlus lib
 namespace pcpp
 {
+	/// @brief Options struct for configuring packet parsing behavior.
+	///
+	/// The parsing options are combined. If multiple stop conditions are specified, parsing stops
+	/// as soon as any condition is met.
+	struct PacketParseOptions
+	{
+		/// @brief Defines the protocol type up to and including which the packet should be parsed (inclusive).
+		///
+		/// For large packets where only the lower layers are of interest, this option can be used to limit the parsing
+		/// procedure to the desired protocol. The packet is parsed until the specified protocol is reached and
+		/// consumed. Layers beyond this protocol are not parsed, which can improve performance and reduce
+		/// memory usage.
+		///
+		/// @remarks If multiple consecutive layers of the same protocol type exist, parsing stops after exhausting the
+		/// first contiguous sequence of that protocol type. This matters for protocols that may appear multiple times
+		/// in succession within a packet (e.g. BgpLayer).
+		ProtocolTypeFamily parseUntilProtocol = UnknownProtocol;
+
+		/// @brief Defines an OSI model layer up to and including which the packet should be parsed (inclusive).
+		///
+		/// For large packets where only the lower layers are of interest, this option can be used to limit the parsing
+		/// procedure to the desired OSI layer. The packet is parsed until all layers belonging to the specified OSI
+		/// layer are consumed. Layers beyond that are ignored, which can improve performance and reduce memory usage.
+		OsiModelLayer parseUntilLayer = OsiModelLayerUnknown;
+	};
+
 	/// @class Packet
 	/// This class represents a parsed packet. It contains the raw data (RawPacket instance), and a linked list of
 	/// layers, each layer is a parsed protocol that this packet contains. The layers linked list is ordered where the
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -163,6 +163,7 @@ cmake_dependent_option(
   "PCAPPP_USE_DPDK"
   OFF
 )
+option(PCAPPP_USE_PCAP "Setup PcapPlusPlus with libpcap / WinPcap / Npcap" ON)
```
Dimi1010, line 166: If libpcap is being transferred to optional dependency, I don't think we will need `PCAPPP_BUILD_PCAPPP` anymore since all targets would be able to be built without any packet capture engine dependencies.

seladb: I think it'd be nice to keep both - you can decide to only build Packet++, or build both Packet++ and Pcap++ but without libpcap

Dimi1010: I guess we can keep it but I think we can do that without the option.

The main reason the option existed was that cmake wouldn't generate the project successfully without libpcap before because Pcap++ had a mandatory dependency on it. That necessitated that the target being excluded from generation altogether. If the cmake project can be successfully generated without libpcap, then that makes the option unnecessary - you could build only Packet++ just as well by selecting that as the sole build target instead of building all targets.

I am a bit unsure about how install behaves, so I guess we can keep it as is for this PR and update it after this is merged.

Pcap++/src/PcapFilter.cpp
Dimi1010, line 1: Generally, I somewhat dislike the sprinkled #ifdef USE_PCAP across the filter code. IMO, it makes it hard to read.

It works for now, but we should probably find a better alternative.

Pcap++/src/PcapFilter.cpp
```diff
@@ -22,8 +24,12 @@ namespace pcpp
 	{
 		void BpfProgramDeleter::operator()(bpf_program* ptr) const noexcept
 		{
+#ifdef USE_PCAP
 			pcap_freecode(ptr);
 			delete ptr;
+#else
+			(void)ptr;
+#endif
```
Dimi1010, line 32: Shouldn't this still call `delete` on the pointer even if no pcap is available? Otherwise it can leak memory.

Pcap++/CMakeLists.txt
```diff
@@ -76,10 +76,12 @@ target_compile_features(Pcap++ PUBLIC cxx_std_14)
 if(APPLE)
   target_link_libraries(Pcap++ PRIVATE "-framework CoreFoundation" "-framework SystemConfiguration")
 elseif(WIN32)
-  target_compile_definitions(Pcap++ PUBLIC -DHAVE_REMOTE)
-  target_compile_definitions(Pcap++ PUBLIC -DWPCAP)
   target_link_libraries(Pcap++ PRIVATE ws2_32 iphlpapi)
-  target_link_libraries(Pcap++ PUBLIC Packet::Packet)
+  if(PCAPPP_USE_PCAP)
+    target_compile_definitions(Pcap++ PUBLIC -DHAVE_REMOTE)
+    target_compile_definitions(Pcap++ PUBLIC -DWPCAP)
```
Dimi1010: nit: Those two lines can be combined. Also by spec `-D` prefix is optional in this context.
```cmake
target_compile_definitions(Pcap++ PUBLIC HAVE_REMOTE WPCAP)
```

CMakeLists.txt
```diff
@@ -210,26 +211,30 @@ endif()
...
-    if(NOT Packet_FOUND)
-      message(FATAL_ERROR "Please specify Packet library -DPacket_ROOT=<NPCAP_SDK_PATH>")
+    # Look for Packet
+    if(WIN32)
+      find_package(Packet)
+      if(NOT Packet_FOUND)
+        message(FATAL_ERROR "Please specify Packet library -DPacket_ROOT=<NPCAP_SDK_PATH>")
+      endif()
     endif()
+    add_definitions(-DUSE_PCAP)
```
Dimi1010: The command `add_definitions` has generally been superseded by a selection of more specific commands added in cmake 3.12. `add_compile_definitions` is the recommended command for adding pre-processor definitions. [[1]](https://cmake.org/cmake/help/latest/command/add_definitions.html)

```suggestion
    add_compile_definitions(USE_PCAP)
```

Pcap++/src/PcapFilter.cpp
seladb: It's only for the parts that use libpcap. We could theoretically create a "null filter" but I don't think it's worth the efforts

Pcap++/src/PcapFilter.cpp
seladb: This doesn't work because `'ptr' has incomplete type`...

https://github.com/seladb/PcapPlusPlus/actions/runs/21740633235/job/62715174139?pr=2057

Pcap++/CMakeLists.txt
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2057/commits/6c45c9fdc5b7fa381f079f75757337e524cfff3c

CMakeLists.txt
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2057/commits/6c45c9fdc5b7fa381f079f75757337e524cfff3c

Dimi1010 (approved):
LGTM.

PR #2062, opened 2026-02-01
Add enum for RawPacket concrete implementation.

Dimi1010:
This PR replaces the raw packet type identifier integers `uint8_t` with strongly typed enum `internal::RawPacketImplType`.

The associated method `RawPacket::getObjectType()` has been deprecated in favor of `RawPacket::getImplType()`.

The new method `getImplType` has been documented as for internal use, and should not be considered part of the stable API.


Pcap++/src/DpdkDevice.cpp
```diff
@@ -1212,8 +1213,8 @@ namespace pcpp
 		for (RawPacketVector::ConstVectorIterator iter = rawPacketsVec.begin(); iter != rawPacketsVec.end(); iter++)
 		{
 			MBufRawPacket* rawPacket = nullptr;
-			uint8_t rawPacketType = (*iter)->getObjectType();
-			if (rawPacketType != MBUFRAWPACKET_OBJECT_TYPE)
+			auto rawPacketType = (*iter)->getImplType();
+			if (rawPacketType != internal::RawPacketImplType::DpdkMBuf)
```
seladb: @Dimi1010 I wonder if we shouldn't just deprecate `getObjectType()` without replacing it with another method. Here we can do something like this:
```cpp
if (dynamic_cast<MBufRawPacket*>(*iter) != nullptr)
{
    ...
}
```

Pcap++/src/DpdkDevice.cpp
```diff
@@ -1259,8 +1260,8 @@ namespace pcpp
 
 	bool DpdkDevice::sendPacket(RawPacket& rawPacket, uint16_t txQueueId, bool useTxBuffer)
 	{
-		uint8_t rawPacketType = rawPacket.getObjectType();
-		if (rawPacketType == MBUFRAWPACKET_OBJECT_TYPE)
+		auto rawPacketType = rawPacket.getImplType();
+		if (rawPacketType == internal::RawPacketImplType::DpdkMBuf)
```
seladb: ditto, we can do:
```cpp
if (dynamic_cast<MBufRawPacket*>(&rawPacket) != nullptr)
{
    ...
}
```

Pcap++/src/DpdkDevice.cpp
Dimi1010: I suppose we can, but we should profile it before that. I know that dynamic cast is tried to be avoided due to the overhead of the cast involving a trip to the RTTI infrastructure.

I am pretty sure that is the reason getObjectType was initially added.

seladb: I'm not sure that `dynamic_cast` performance is significantly poor comparing to a polymorphic `getObjectType()`, but we can profile it and check...

Dimi1010: Here is a simple [benchmark](https://quick-bench.com/q/_22yTSHoK0F-LM2e51Aq9aVJ_Dg). It is quite slower (x3.5), and that is why it is discouraged in hot loops (imo, `sendPacket` qualifies?)

The difference between dynamic cast and the virtual call is the mechanisms behind it.
- A dynamic cast essentially does a search through the inheritance tree of the object to find the target type. The precise method is implementation defined.
- The virtual pointer method just does a virtual call which is roughly equivalent to loading a pointer from the vtable and calling a function through a function pointer. Afterwards the code just assumes that object is of said type depending on the function result.

Generally the dynamic cast is safer as it "just works" TM, but the `getType` method is faster due to less overhead, but it is more error prone due to relying on manually enforced assumptions skipping overhead. Ideally, the code would be structured to require neither, but there are valid cases of use. 😅

seladb: Wow, it is more significant that I imagined... I guess we need to keep the enum

Packet++/header/RawPacket.h
```diff
@@ -253,6 +253,24 @@ namespace pcpp
...
 
+	namespace internal
+	{
+		/// @brief Type of RawPacket implementation
+		///
+		/// This is mainly used internally to distinguish between different implementations without using RTTI.
+		enum class RawPacketImplType : uint8_t
+		{
+			/// @brief Unknown type
+			Unknown = 0,
```
seladb, line 269: Why do we need `Unknown`?

Packet++/header/RawPacket.h
```diff
@@ -324,7 +342,16 @@ namespace pcpp
 		/// @return A pointer to the new RawPacket object which is a clone of this object
 		virtual RawPacket* clone() const;
 
+		/// @brief Get the type of RawPacket implementation.
+		/// @remarks This method is for internal use only and may be subject to change without prior notice.
+		virtual internal::RawPacketImplType getImplType() const
+		{
+			return internal::RawPacketImplType::Standard;
+		}
```
seladb: If this method is public it shouldn't return an internal enum. We should move `RawPacketImplType` to be not internal

seladb:
@Dimi1010 do you know why the "Generate Cobertura Report" phase randomly fails? It uses `gcovr` to generate the coverage report from the `.gcda` files generated during the build with `--coverage`. I tried checking with AI but not exactly sure why it could fail...

Packet++/header/RawPacket.h
Dimi1010: It is in `internal` by design. The method has been documented as an internal use method.

IMO, this isn't "public" API, and the users shouldn't rely on it or use it.

The accesibility should really be something like C#'s `internal` access level, but since we don't have such things, it needs to be public to allow access.

The return type being inside `internal` is precisely to discourage use.

Packet++/header/RawPacket.h
Dimi1010: It is so zero initialized values have a separate enum value that won't just be assumed to be a certain implementation. 

It is mostly a defensive programming technique.

Packet++/header/RawPacket.h
Dimi1010: Although, I suppose there is a way to do the thing semi-encapsulated and not be public.

It would be something like this:
```cpp

namespace internal
{
  enum class RawPacketImplType { /* Impl Here */ }

  RawPacketImplType getRawPacketImpl(RawPacket const& rawPacket)
  {
    return rawPacket.getImplType();
  }
}

class RawPacket
{
  friend RawPacketImplType getRawPacketImpl(RawPacket const& rawPacket);

protected:
  virtual internal::RawPacketImplType() getImplType() const;
}

class DerivedRawPacket : public RawPacket
{ /* ... */}
```

Dimi1010:
> @Dimi1010 do you know why the "Generate Cobertura Report" phase randomly fails? It uses `gcovr` to generate the coverage report from the `.gcda` files generated during the build with `--coverage`. I tried checking with AI but not exactly sure why it could fail...

Tbh, I'm not really sure either. I can try to debug it, but I'm not promising anything.

Packet++/header/RawPacket.h
Dimi1010: Updated 17d4f6750f65a3776b3aea8c864f99d335f4df08

Dimi1010:
> > @Dimi1010 do you know why the "Generate Cobertura Report" phase randomly fails? It uses `gcovr` to generate the coverage report from the `.gcda` files generated during the build with `--coverage`. I tried checking with AI but not exactly sure why it could fail...
> 
> Tbh, I'm not really sure either. I can try to debug it, but I'm not promising anything.

@seladb created a separate issue #2067 for that so it isn't scattered in PRs.

Packet++/header/RawPacket.h
seladb: Since the base class `RawPacket` is the default, I think it should be the default for the enum as well

seladb (approved):
Please see the comment below, otherwise LGTM

Packet++/header/RawPacket.h
Dimi1010: The `RawPacket` being nominal default implementation doesn't guarantee it is to be assumed everywhere tho?

I would prefer not to risk it in this case, as it can lead to memory corruption bugs due to improper down casting and those are hard to track down. It is done this way precisely to reduce general assumptions, which can be dangerous when working with memory.

seladb: Sure, we can keep the `Unknown` value if you prefer 👍

PR #2070, opened 2026-02-20
Add PostgreSQL Wire Protocol (PGWire) layer

seladb:
This PR adds a new PostgreSQL Wire Protocol Layer (PostgresLayer) to PcapPlusPlus, enabling parsing of PostgreSQL network traffic. This is the protocol used by PostgreSQL clients and servers to communicate over TCP.

This PR only parses the Postgres messages. Creating messages will be done in subsequent PRs

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -9,7 +9,7 @@ on:
 
 env:
   BUILD_DIR: Dist
-  GCOVR_FLAGS: --gcov-ignore-parse-errors --exclude-throw-branches --filter Common --filter Pcap --filter Packet --xml
+  GCOVR_FLAGS: --gcov-ignore-errors all --gcov-ignore-parse-errors all --merge-mode-functions=separate --exclude-throw-branches --filter Common --filter Pcap --filter Packet --xml
   CCACHE_DIR: ${{ github.workspace }}/.ccache
   CPPCHECK_VERSION: 2.9
 
@@ -57,6 +57,7 @@ jobs:
     runs-on: ${{ matrix.runner }}
     container: seladb/${{ matrix.image }}
     strategy:
+      fail-fast: false
       matrix:
         include: # Check the images at https://github.com/seladb/PcapPlusPlus-DockerImages
           - runner: ubuntu-latest
@@ -73,12 +74,15 @@ jobs:
           - runner: ubuntu-latest
             image: ubuntu2004
             config-zstd: OFF
+            gcovr-version: 8.2
           - runner: ubuntu-latest
             image: rhel94
             config-zstd: OFF
+            gcovr-version: 8.5
           - runner: ubuntu-latest
             image: ubuntu2004-zstd
             config-zstd: ON
+            gcovr-version: 8.2
           - runner: ubuntu-latest
             image: fedora42
             config-zstd: OFF
@@ -157,7 +161,7 @@ jobs:
       - name: Create Cobertura Report
         run: |
           . .venv/bin/activate
-          python3 -m pip install gcovr==8.2
+          python3 -m pip install gcovr==${{ matrix.gcovr-version || '8.6' }}
           gcovr -v -r . ${{ matrix.additional-gcov-flags }} $GCOVR_FLAGS -o coverage.xml
 
       - name: Upload Coverage Results
@@ -227,7 +231,7 @@ jobs:
       - name: Create Cobertura Report
         run: |
           . .venv/bin/activate
-          python3 -m pip install gcovr==8.2
+          python3 -m pip install gcovr==8.6
           gcovr -v -r . $GCOVR_FLAGS -o coverage.xml
 
       - name: Upload Coverage Results
@@ -530,7 +534,7 @@ jobs:
       - name: Create Cobertura Report
         if: ${{ matrix.host-arch == matrix.arch }}
         run: |
-          python3 -m pip install gcovr==8.2
+          python3 -m pip install gcovr==8.6
           gcovr -v -r . $GCOVR_FLAGS -o coverage.xml
 
       - name: Upload Coverage Results
@@ -651,7 +655,7 @@ jobs:
           python -m pytest --root-path=../../Dist/examples_bin
 
       - name: Install Coverage Requirements
-        run: python3 -m pip install gcovr==8.2
+        run: python3 -m pip install gcovr==8.6
 
       - name: Process Coverage Files
         shell: msys2 {0}
@@ -812,7 +816,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Test in FreeBSD
         id: test
-        uses: vmactions/freebsd-vm@ba6bedee4a4884da2b782a41a64329a1c8e42ffb # v1.3.8
+        uses: vmactions/freebsd-vm@c9f815bc7aa0d34c9fdd0619b034a32d6ca7b57e # v1.4.2
         with:
           release: ${{ matrix.version }}
           usesh: true
@@ -972,7 +976,7 @@ jobs:
 
       - name: Create Cobertura Report
         run: |
-          python -m pip install gcovr==8.2
+          python -m pip install gcovr==8.6
           gcovr -v -r . $GCOVR_FLAGS -o coverage.xml
 
       - name: Upload Coverage Results
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -35,7 +35,7 @@ jobs:
         dry-run: false
         sanitizer: ${{ matrix.sanitizer }}
     - name: Upload Crash
-      uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+      uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
       if: failure() && steps.build.outcome == 'success'
       with:
         name: artifacts
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -29,7 +29,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v3.29.5
+      uses: github/codeql-action/init@89a39a4e59826350b863aa6b6252a07ad50cf83e # v3.29.5
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
@@ -44,4 +44,4 @@ jobs:
         cmake --build build -j
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v3.29.5
+      uses: github/codeql-action/analyze@89a39a4e59826350b863aa6b6252a07ad50cf83e # v3.29.5
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -82,7 +82,7 @@ jobs:
 
       - name: Generate artifact attestation
         if: github.ref_type == 'tag'
-        uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
+        uses: actions/attest-build-provenance@a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32 # v4.1.0
         with:
           subject-path: "${{ env.BUILD_DIR }}/*.tar.gz,${{ env.BUILD_DIR }}/*.deb,${{ env.BUILD_DIR }}/*.rpm"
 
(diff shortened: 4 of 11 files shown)


Packet++/header/PostgresLayer.h
```diff
@@ -0,0 +1,412 @@
...
+			Backend_PortalSuspended,
+			/// Ready for query (idle state)
+			Backend_ReadyForQuery,
+			/// Row description (column definitions)
+			Backend_RowDescription,
+			/// Unknown backend message type
+			Backend_Unknown,
+		};
+
+		PostgresMessageType() : m_Value(Frontend_Unknown)
```
Dimi1010: nit: can also be marked constexpr.

Packet++/header/PostgresLayer.h
```diff
@@ -0,0 +1,412 @@
...
+		{
+			return OsiModelApplicationLayer;
+		}
+
+		/// @return Returns the protocol info as readable string
+		std::string toString() const override;
+
+	private:
+		PostgresMessageOrigin m_MessageOrigin;
+		mutable std::vector<std::unique_ptr<PostgresMessage>> m_Messages;
```
Dimi1010: Curious, why not use `PointerVector<PostgresMessage>`?

Packet++/header/PostgresLayer.h
seladb: Good catch! Addressed in https://github.com/seladb/PcapPlusPlus/pull/2070/commits/f21e018e21a7e08deb341e27b48268efdca65730

Packet++/header/PostgresLayer.h
seladb: `PostgresMessage` is a base class. In this PR we only have 2 derived classes: `PostgresStartupMessage` and `PostgresParameterStatus` but in the next PR I'll add more. Because it's a base class we have to keep a vector of pointer, no? 🤔

Dimi1010: Sure, I was just saying that `pcpp::PointerVector<T>` also does that. It internally has `std::vector<T*>`, but it has a bit more specialized API for work with pointers (`getAndDetach`, etc). Either is fine, I was just curious.

seladb: Oh, I didn't notice you wrote `PointerVector`. Yes, that's a good idea, I'll fix that

seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2070/commits/483f3abb8d22029c7bf9cac08c1d1e45e3ff81be

Packet++/header/PostgresLayer.h
```diff
@@ -0,0 +1,411 @@
...
+			return m_DataLen;
+		}
+
+		/// @brief Returns the raw payload bytes of the message
+		/// @return The raw payload bytes of the message
+		std::vector<uint8_t> getRawPayload() const;
+
+	protected:
+		PostgresMessage(const uint8_t* data, size_t dataLen, const PostgresMessageType& messageType)
+		    : m_Data(data), m_DataLen(dataLen), m_MessageType(messageType)
```
Dimi1010, line 244: Nit: Types under 8 bytes are generally cheaper to pass by value than const ref.

PS: I am unsure if the compiler does this automatically under optimizations for small types so it might be fine.

Packet++/header/PostgresLayer.h
```diff
@@ -0,0 +1,411 @@
...
+		/// @brief Get a PostgreSQL message by its type
+		/// @param[in] messageType The type of message to retrieve
+		/// @return A pointer to the message, or nullptr if not found
+		const PostgresMessage* getPostgresMessage(const PostgresMessageType& messageType) const;
+
+		/// @brief Get a PostgreSQL message by its type (template version)
+		/// @tparam TMessage The message type to retrieve (must derive from PostgresMessage)
+		/// @return A pointer to the message of the specified type, or nullptr if not found
+		template <class TMessage,
+		          typename std::enable_if<std::is_base_of<PostgresMessage, TMessage>::value>::type* = nullptr>
```
Dimi1010: Nit: Cpp14 allows the shortened `std::enable_if_t`

Also, I find the `std::enable_if_t<Cond, bool> = true` more common in this syntax. Either works, tho.

Packet++/src/PostgresLayer.cpp
```diff
@@ -0,0 +1,849 @@
...
+	constexpr char PostgresFrontendMessage_S = 'S';
+	constexpr char PostgresFrontendMessage_d = 'd';
+	constexpr char PostgresFrontendMessage_c = 'c';
+	constexpr char PostgresFrontendMessage_f = 'f';
+	constexpr char PostgresFrontendMessage_X = 'X';
+
+	constexpr uint32_t PostgresFrontendTag_SSLRequest = 80877103;
+	constexpr uint32_t PostgresFrontendTag_GSSENCRequest = 80877104;
+	constexpr uint32_t PostgresFrontendTag_CancelRequest = 80877102;
+	constexpr uint32_t PostgresFrontendTag_StartupMessage = 196608;
```
Dimi1010, line 50: Can we make all these have internal linkage?

Packet++/src/PostgresLayer.cpp
```diff
@@ -0,0 +1,849 @@
...
+			return nullptr;
+		}
+
+		if (dataLen < 5)
+		{
+			return new PostgresMessage(data, dataLen, PostgresMessageType::Backend_Unknown);
+		}
+
+		auto messageLength = be32toh(*reinterpret_cast<const uint32_t*>(data + 1));
+		if (dataLen < messageLength + 1)
```
Dimi1010, line 300: Does message length include its own 4 bytes in the length?

Packet++/src/PostgresLayer.cpp
```diff
@@ -0,0 +1,849 @@
...
+
+	uint32_t PostgresMessage::getMessageLength() const
+	{
+		if (m_Data == nullptr || m_DataLen < 4)
+		{
+			return 0;
+		}
+
+		const auto offset = (m_Data[0] == 0) ? 0 : 1;
+		return be32toh(*reinterpret_cast<const uint32_t*>(m_Data + offset));
```
Dimi1010: This can have out of bounds access on `m_DataLen=4` and `m_Data[0] != 0`.

Packet++/src/PostgresLayer.cpp
```diff
@@ -0,0 +1,849 @@
...
+		if (m_DataLen < 6)
+		{
+			return {};
+		}
+
+		const auto* start = reinterpret_cast<const char*>(m_Data + 5);
+		const auto maxLen = m_DataLen - 5;
+		const auto* end = static_cast<const char*>(memchr(start, '\0', maxLen));
+
+		return std::string(start, end != nullptr ? end : start + maxLen);
```
Dimi1010: This can be simplified to use `std::find` and use purely iterators.

```suggestion
		const auto* start = reinterpret_cast<const char*>(m_Data + 5);
		auto* end = start + m_DataLen - 5;
		end = std::find(start, end, '\0');
		return std::string(start, end);
```

Packet++/src/PostgresLayer.cpp
```diff
@@ -0,0 +1,849 @@
...
+
+	std::string PostgresParameterStatus::getParameterName() const
+	{
+		if (m_DataLen < 6)
+		{
+			return {};
+		}
+
+		const auto* start = reinterpret_cast<const char*>(m_Data + 5);
+		const auto maxLen = m_DataLen - 5;
```
Dimi1010: nit, magic value: Possibly export `5` into `constexpr size_t headerLen` as in `getParameterValue`?

Packet++/src/PostgresLayer.cpp
```diff
@@ -0,0 +1,849 @@
...
+		{
+			return "";
+		}
+
+		const size_t nameLen = static_cast<size_t>(nameEnd - base);
+		const char* valueStart = nameEnd + 1;
+		const size_t valueMaxLen = remaining - nameLen - 1;
+
+		const char* valueEnd = static_cast<const char*>(memchr(valueStart, '\0', valueMaxLen));
+		return std::string(valueStart, valueEnd != nullptr ? valueEnd : valueStart + valueMaxLen);
```
Dimi1010: Ditto, can be simplified to use pure iterator algorithms.

```suggestion
		const char* base = reinterpret_cast<const char*>(m_Data) + headerLen;
		const char* baseEnd = base + m_DataLen - headerLen;

		// Value start = nameEnd + 1;
		const char* nameEnd = std::find(base, baseEnd, '\0');
		if (nameEnd >= baseEnd || nameEnd + 1 >= baseEnd)
		{
			// Bounds check.
			// If find returns baseEnd - 1 or baseEnd, valueStart is out of bounds.
			return "";
		}
		const char* valueStart = nameEnd + 1;

		const char* valueEnd = std::find(valueStart, baseEnd, '\0');
		return std::string(valueStart, valueEnd);
```

Packet++/src/PostgresLayer.cpp
```diff
@@ -0,0 +1,849 @@
...
+			auto parseFunc = (m_MessageOrigin == PostgresMessageOrigin::Backend)
+			                     ? &PostgresMessage::parsePostgresBackendMessage
+			                     : &PostgresMessage::parsePostgresFrontendMessage;
+
+			auto data = m_Data;
+			auto dataLen = m_DataLen;
+
+			while (dataLen > 0)
+			{
+				auto curMessage = std::unique_ptr<PostgresMessage>(parseFunc(data, dataLen));
```
Dimi1010: `PostgresMessage::parsePostgresBackendMessage` and `PostgresMessage::parsePostgresFrontendMessage` are only used here. Why not make them return `std::unique_ptr<PostgresMessage>`?

Packet++/header/PostgresLayer.h
seladb: Are you referring to `PostgresMessageType`? It's a class, not an enum / enum class (which contains an inner enum). I think it's better to use a reference, no?

Packet++/src/PostgresLayer.cpp
seladb: What do you mean by internal linkage? 🤔

Packet++/src/PostgresLayer.cpp
seladb: Yes, please see here: https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS

For example:
```
Int32(8)

    Length of message contents in bytes, including self.
```

Packet++/header/PostgresLayer.h
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2070/commits/65f96fb005eec5a8512f5c2dcc1e27b96cb2faf5

Packet++/src/PostgresLayer.cpp
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2070/commits/65f96fb005eec5a8512f5c2dcc1e27b96cb2faf5

Packet++/src/PostgresLayer.cpp
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2070/commits/65f96fb005eec5a8512f5c2dcc1e27b96cb2faf5

Packet++/src/PostgresLayer.cpp
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2070/commits/65f96fb005eec5a8512f5c2dcc1e27b96cb2faf5

Packet++/src/PostgresLayer.cpp
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2070/commits/65f96fb005eec5a8512f5c2dcc1e27b96cb2faf5

Packet++/src/PostgresLayer.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2070/commits/282a8ab0b4e6af5b240a74920342fa480e258339

Packet++/header/PostgresLayer.h
Dimi1010: If it might grow in size in the future, ref would be better. If it won't either is fine. Generally I prefer to pass cheap to copy types by value if their total size is less than a pointer, but it is mostly a very minor performance difference.

We can keep it as reference, if you want.

Packet++/src/PostgresLayer.cpp
Dimi1010: Generally internal linkage is when the symbol is not exposed outside of its own translation unit (aka can't be accessed via `extern int foo;` from some other TU). This is generally what happens when non-member functions or global variables are marked as `static` or in anonymous namespace.

Upon further review, it might not actually be necessary here as apparently `const` qualified global variables [use internal linkage by default](https://en.cppreference.com/w/cpp/language/storage_duration.html#:~:text=non%2Dtemplate%20(since%20C%2B%2B14)variables%20of%20non%2Dvolatile%20const%2Dqualified%20type%2C%20unless).

PR #2072, opened 2026-02-24
Fixes performance regression during Pcap file reading.

### LINKED ISSUE #2071: Apparent performance regression
Commit 44e9085710177be4ed8369ee0ba4c772dbd450ff appears to have introduced a performance regression when compared with v25.05.
The performance regression is present in the current master branch.

Attached is a simple benchmark I used to measure the performance difference between `libpcap` and the current pcapplusplus.

## Platform details

```
((v25.05))> lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.5 LTS
Release:        22.04
Codename:       jammy
```

* GCC version: gcc version 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04.2)
* libpcap version: 1.10.1

## Build and comparison

Performance between the `master` branch and the `v25.05` tag were compared by building 

```
cmake -DCMAKE_INSTALL_PREFIX=(realpath ~/local_install/pcapplusplus/) -DPCAPPP_BUILD_EXAMPLES=ON -B build .
cd build
make
make install

g++ -O2 benchmark.cpp -o benchmark -I ~/local_install/pcapplusplus/include/pcapplusplus -L ~/local_install/pcapplusplus/lib/ -lPcap++ -lPacket++ -lCommon++ -lpcap -lpthread
```

### Regressed performance beginning in commit 44e9085710177be4ed8369ee0ba4c772dbd450ff
```
Running libpcap benchmark:
Benchmarking PCAP file with libpcap: pcap/file.pcap
Reading all packets...

=== libpcap Benchmark Results ===
Packets read: 485760
Total bytes: 232226544 bytes (221.47 MB)
Time taken: 0.077176 seconds
Packets/second: 6294184.72
Throughput: 2869.65 MB/s

--------------------------------------------------

Running PcapPlusPlus benchmark:

=== PcapPlusPlus Benchmark Results ===
Packets read: 485760
Total bytes: 232226544 bytes (221.47 MB)
Time taken: 1.857491 seconds
Packets/second: **261514.05**
Throughput: **119.23 MB/s**
```

### High performance in v25.05

```
Running libpcap benchmark:
Benchmarking PCAP file with libpcap: pcap/file.pcap
Reading all packets...

=== libpcap Benchmark Results ===
Packets read: 485760
Total bytes: 232226544 bytes (221.47 MB)
Time taken: 0.065889 seconds
Packets/second: 7372399.03
Throughput: 3361.24 MB/s

--------------------------------------------------

Running PcapPlusPlus benchmark:
=== PcapPlusPlus Benchmark Results === 
Packets read: 485760
Total bytes: 232226544 bytes (221.47 MB)
Time taken: 0.084271 seconds
Packets/second: 5764260.54
Throughput: 2628.05 MB/s
```

The attached benchmark.cpp is 80% Claude Sonnet written.

Dimi1010:
Fixes #2071
This PR fixes a regression introduced in 44e9085 due to extensive allocator pressure.

The regression was due to the always allocating `m_SnapshotLength` (nominal 262144 bytes) in the free store, causing excessive allocator pressure. As most packets are significantly smaller than that, the allocation is mostly waste and pure overhead.

The PR introduces an persistent per-device intermediate read buffer and restores the old implementation of then allocating a new buffer sized precisely to the captured packet.


Dimi1010:
To note, the read buffer could be further optimized for memory footprint, having it dynamically grow to the largest so-far encountered packet, but this is a much more minor issue than the large speed degradation that is addressed in this PR.

Packet++/src/BgpLayer.cpp
seladb, line 1: Are these change related to this fix?

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -1,5 +1,8 @@
 #define LOG_MODULE PcapLogModuleFileDevice
 
+#include <algorithm>
+#include <iterator>
```
seladb, line 4: Is `<iterator>` needed? 🤔

seladb:
@Dimi1010 I ran the Fuzz tests multiple times and they keep failing, I'm not sure why 🤔

Packet++/src/BgpLayer.cpp
Dimi1010: Related to trying to fix the fuzz.

Pcap++/src/PcapFileDevice.cpp
Dimi1010: `std::next` is in `<iterator>`

Dimi1010:
> @Dimi1010 I ran the Fuzz tests multiple times and they keep failing, I'm not sure why 🤔

@seladb The fuzz is failing because of multiple minor off by one errors that are finally being caught. I am unsure why they weren't being caught before...

The hack solution to make them pass for this PR is to add a 5~10 bytes padding after the buffer, but we probably want them fixed anyways. ¯\\\_(ツ)\_/¯

emeneker38a:
The changes seem to fix the performance regression.
With the changes from fix/pcap-read-regression, I get the following performance from the same benchmark code

```
Running libpcap benchmark:
Reading all packets...

=== libpcap Benchmark Results ===
Packets read: 485760
Total bytes: 232226544 bytes (221.47 MB)
Time taken: 0.065686 seconds
Packets/second: 7395183.14
Throughput: 3371.62 MB/s

--------------------------------------------------

Running PcapPlusPlus benchmark:

=== PcapPlusPlus Benchmark Results ===
Packets read: 485760
Total bytes: 232226544 bytes (221.47 MB)
Time taken: 0.103885 seconds
Packets/second: 4675939.74
Throughput: 2131.86 MB/s
```

Thank you @seladb and @Dimi1010 for the excellent library and the fast response.

Packet++/src/BgpLayer.cpp
seladb: I'm not sure why this suddenly fails because of the changes in the file device 🤔

seladb:
> > @Dimi1010 I ran the Fuzz tests multiple times and they keep failing, I'm not sure why 🤔
> 
> @seladb The fuzz is failing because of multiple minor off by one errors that are finally being caught. I am unsure why they weren't being caught before...
> 
> The hack solution to make them pass for this PR is to add a 5~10 bytes padding after the buffer, but we probably want them fixed anyways. ¯\_(ツ)_/¯

Do you know what in this PR cause them to fail?

Dimi1010:
> Do you know what in this PR cause them to fail?

@seladb Unsure of the precise cause exactly.

Also, we might have to roll back to the libpcap read pcap implementation. When checking the their [pcap parsing source](https://github.com/the-tcpdump-group/libpcap/blob/8c7496f7fd81c3bf2201194d71f57ae1f03e9997/sf-pcap.c#L510), they do a lot of other validation and cover edge cases, especially for link types other than Ethernet.

We can still keep it optional dependency, but we might have to drop pcap read / write support with that flag disabled. I think this would be a less maintenance heavy solution, since we don't have to actively cover all those cases again...

Packet++/src/BgpLayer.cpp
Dimi1010: Possibly because the buffer is now sized precisely to the captured region? If the buffer was previously larger, the address sanitizer wouldn't have flagged a buffer overrun as it still technically isn't a overrun from its point of view.

Dimi1010:
@seladb Ok... I figured out why the Fuzzers are suddenly failing at 50 billion places. It has to do with how the CI reports Fuzz failures.

Normally, when the Fuzz CI detects an error with the PR head, it retries the same corpus on the Base head. If the base also produces the same failure condition, it is ignored as the failure is not considered to be introduced in the affected PR.

See https://github.com/seladb/PcapPlusPlus/actions/runs/20675626091/job/59362879527?pr=2041 which is the action log directly before https://github.com/seladb/PcapPlusPlus/commit/44e9085710177be4ed8369ee0ba4c772dbd450ff which changed the pcap read packet pipeline. The thing that happened in 44e9085 tho, is that the newly allocated buffer wasn't precisely `caplen` but it was the enormous `m_SnapshotLength` (≈262144 bytes). This made all the buffer overruns that `AddressSanitizer` was detecting disappear, since it wasn't a buffer overrun from its perspective (it tracks allocated size).

The current PR, again restores the buffer length to the original `caplen`, but so all the buffer overruns are back. The issue is that when the Fuzz tries to reproduce them in the Base head, it can't since that is using the huge buffer block that `AddressSanitizer` is happy about.  So it flags every single fuzz failure it ignored because "it wasn't introduced in this PR" as "it was introduced in this PR", even though the current PR doesn't change anything directly related to the error.

seladb:
@Dimi1010 wow that's a great find! 💯 
I'm wondering how Fuzz passed before - maybe we force-merged it once and then subsequent PRs stop reporting these issues?

Do you see a simple way to fix it or should we force-merge this PR as well and fix the Fuzz issues in a separate PR?

Dimi1010:
> @Dimi1010 wow that's a great find! 💯 I'm wondering how Fuzz passed before - maybe we force-merged it once and then subsequent PRs stop reporting these issues?
> 
> Do you see a simple way to fix it or should we force-merge this PR as well and fix the Fuzz issues in a separate PR?

The "simple way" is to fix all the failures or add an oversized buffer again.
I suppose we can force push too, but I would prefer to avoid it if we can...

seladb:
> The "simple way" is to fix all the failures or add an oversized buffer again.
> I suppose we can force push too, but I would prefer to avoid it if we can...

Do you think fixing everything will create a reasonable size PR? 🤔

Dimi1010:
> > The "simple way" is to fix all the failures or add an oversized buffer again.
> > I suppose we can force push too, but I would prefer to avoid it if we can...
> 
> Do you think fixing everything will create a reasonable size PR? 🤔 

Truthfully, no. Atm, I think imma be done for this PR when I fix the SSL extension buffer overrun. The rest can be fixed later.

.github/workflows/cifuzz.yml
Dimi1010, line 1: Fail fast is disabled as every fuzz test gives different information.

Packet++/src/BgpLayer.cpp
```diff
@@ -263,8 +263,7 @@ namespace pcpp
 			return;
 		}
 
-		size_t optionalParamsLen = (size_t)be16toh(msgHdr->optionalParameterLength);
-
+		size_t optionalParamsLen = msgHdr->optionalParameterLength;
```
Dimi1010, line 266: `be16toh` is removed due to `optionalParamLength` being `uint8_t`

Packet++/src/RawPacket.cpp
```diff
@@ -139,6 +139,11 @@ namespace pcpp
 
 	void RawPacket::insertData(int atIndex, const uint8_t* dataToInsert, size_t dataToInsertLen)
 	{
+		if (atIndex > m_RawDataLen)
+		{
+			throw std::out_of_range("Insert index is out of raw packet bound. Inserts can only happen in range [0, " + std::to_string(m_RawDataLen) + ']');
+		}
```
Dimi1010, line 146: Throws exception because, ideally `atIndex` should be validated in the caller.
Also, no other way to return error without breaking API...

Dimi1010:
@seladb Ok, I think I am done with trying to do fixes on this PR. Might as well merge the rest even with the fuzz failure...

So far this PR fixes the Fuzz "heap out of bounds access" failures in the following layers:
- DHCP Layer: Added validation `isDataValid` prior to construction. UdpLayer updated to utilize it.
- Vxlan Layer: Added validation `isDataValid` prior to construction. UdpLayer updated to utilize it.
- BGP Layer: Fixed unsigned underflow when calculating the option parameters length.
- SSL Layer: Hardened the SSL extension parsing algorithm against partial extension data buffers.
- TLV Data: Fixed unsigned integer underflow in length check.

Partial fixes:
- DnsLayer: Attempted to fix. Did not work out but kept the conditions.
- TelnetLayer: Attempted to fix `distanceToNextIAC` as that was what appears to be causing the failure. Did not work. :/

Active Fuzz failures:
- VrrpLayer
- DnsLayer
- TelnetLayer

CI Notes:
- Disabled `fast-fail` on CIFuzz and Build fuzzers as each fuzzer utilizes a different sanitizer. Having all of them finish on failure can produce more informative result for fixing the issue.

Packet++/header/VxlanLayer.h
```diff
@@ -90,6 +90,11 @@ namespace pcpp
 
 		~VxlanLayer() override = default;
 
+		static bool isDataValid(uint8_t const* data, size_t dataLen)
```
seladb, line 97: Can we add doxgen documentation?

seladb:
@Dimi1010 if we can't fix all issues, maybe we can separate it to multiple PRs that are easier to review?
I'm ok with force-merging this PR. 

BTW - if we force-merge this PR, will the next PR (that is not related to fixing it) pass Fuzzing?

Dimi1010:
>  if we can't fix all issues, maybe we can separate it to multiple PRs that are easier to review?
I'm ok with force-merging this PR.

I suppose we can, but I don't think the changes are that hard to review? Most of the individual changes are localized to a single file or 2 in cases where `UdpLayer` is updated. We can do it if you want, but I would prefer to avoid the having to split this into 6 PRs if I can avoid it.

> BTW - if we force-merge this PR, will the next PR (that is not related to fixing it) pass Fuzzing?

It should pass, since it will go into the same cycle that we had before... where the failures are reproduced in the head too and ignored.

seladb:
Ok sure, I'll review all of the changes in this PR

Packet++/header/SSLHandshake.h
```diff
@@ -118,6 +137,29 @@ namespace pcpp
...
+		/// @param data Pointer to the raw data of the extension.
+		/// @param dataLen Max length of the span to be parsed.
+		/// @return The extension instance if the data buffer is valid, nullptr otherwise.
+		template <typename T, typename std::enable_if_t<std::is_base_of<SSLExtension, T>::value, bool> = true>
+		static std::unique_ptr<T> tryCreateExtension(uint8_t* data, size_t dataLen)
+		{
+			if (data == nullptr || dataLen < sizeof(SSLExtensionStruct))
+				return nullptr;
+
+			SSLExtensionStruct* extStruct = reinterpret_cast<SSLExtensionStruct*>(data);
```
seladb: nit: we can use `auto` here

Packet++/header/SSLHandshake.h
```diff
@@ -118,6 +137,29 @@ namespace pcpp
...
+			if (dataLen < sizeof(SSLExtensionStruct) + extStruct->getDataLength())
+				return nullptr;
```
seladb, line 161: nit: we can use curly brackets `{}` here

Packet++/src/SSLHandshake.cpp
```diff
@@ -1091,12 +1103,12 @@ namespace pcpp
 
 	uint16_t SSLExtension::getTypeAsInt() const
 	{
-		return be16toh(getExtensionStruct()->extensionType);
+		return getExtensionStruct()->getExtType();
```
seladb: `getExtType()` is only used here, is it really needed?

Packet++/src/SSLHandshake.cpp
```diff
@@ -1091,12 +1103,12 @@ namespace pcpp
...
 	}
 
 	uint16_t SSLExtension::getLength() const
 	{
-		return be16toh(getExtensionStruct()->extensionDataLength);
+		return getExtensionStruct()->getDataLength();
```
seladb, line 1111: ditto

Packet++/src/SSLHandshake.cpp
```diff
@@ -1337,6 +1363,72 @@ namespace pcpp
 	// SSLClientHelloMessage methods
 	// -----------------------------
 
+	namespace
```
seladb: Is this change needed? It make the diff much harder to review. Can we make the minimal changes possible?

Packet++/src/SSLHandshake.cpp
Dimi1010: It was added for symmetry as `getDataLength` was moved to the internal struct to allow fetching with conversion from the templated factory in the header.

This method itself isn't critical.

Packet++/src/SSLHandshake.cpp
Dimi1010: Needed as the struct data length is used in the templated factory                 `tryCreateExtension`. The header doens't have access to the be16toh macros, so that utility needed to be moved to the internal struct.

Packet++/src/SSLHandshake.cpp
Dimi1010: Probably. There were buffer overruns in in the SSL extension parsing.

The majority of the per-message validation is handed in `SSLExtension::tryCreateExtension<T>`, so we could probably do without the loop refactor, but it significantly simplifies the bounds calculation. I am also unsure if there were subtle bugs in the bounds checking in the original, since rewriting the loop to mimic standard a algorithm was easier for me, than trying to debug all the `pointer + range` arithmetic in the original loop condition.

Instead of having a `curPos` and recalculating its distance to the bounds on each cycle w.r.t. the extension data length and the total message length in this:
```cpp
(curPos - extensionPos) < (int)extensionLength && 
(curPos - m_Data) < (int)messageLen && 
(int)messageLen - (curPos - m_Data) >= (int)minSSLExtensionLen
```

The new parse procedure essentially operates on a memory span [first, last) like standard algorithms.
The memory span is pre-calculated in the following steps:
1. `first = extensionPos`: start of the data buffer.
2. `last = extensionPos + extensionLength`: end of the data buffer
3. Clamp last to be the smaller of `last` and `m_Data + messageLen`: This prevents a buffer overrun if the reported `extensionLength` in the message is larger than the overall message.

Afterwards the memory region is parsed to the parse procedure as a validated range [first, last). This allows to essentially always have 1 conceptual "valid" memory span, rather than having to track the 2 "ends" of the region, reducing cognitive load on understanding the loop bounds (IMO).

The export to a function helper is because the same code was duplicated in `SSLClientHelloMessage` and `SSLServerHelloMessage`. That can possibly be reverted?

seladb: If we can simplify the changes it'd be great and easier to review

Dimi1010: @seladb Reverted the export to a function helper in c741a1b0321343f1df19a65da3d78dab62a52e64
The changes should be closer to the old code now in the diff.

Packet++/src/SSLHandshake.cpp
seladb: I think we can revert it?

Packet++/header/SSLHandshake.h
Dimi1010: Fixed dcb0879695cfafef27901e852a9f0ee7e6ca8fc2

Packet++/header/SSLHandshake.h
Dimi1010: Fixed dcb0879695cfafef27901e852a9f0ee7e6ca8fc2

Packet++/src/SSLHandshake.cpp
Dimi1010: Reverted 3c96821cf58e1119e917b7a002042c93af8d2ee8

Packet++/header/VxlanLayer.h
Dimi1010: Fixed b516b47e5009894558d5127ed5b136eb5f780625

seladb (approved):
LGTM

PR #2073, opened 2026-02-27
Parse additional PostgreSQL Wire Protocol (PGWire) messages

seladb:
This PR parses the following messages:
- [`Query` (Frontend)](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-QUERY)
- [`RowDescription` (Backend)](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-ROWDESCRIPTION)
- [`DataRow` (Backend)](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-DATAROW)
- [`ErrorRespnse` (Backend)](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-ERRORRESPONSE)

### DIFF
--- a/Packet++/header/PostgresLayer.h
+++ b/Packet++/header/PostgresLayer.h
@@ -392,6 +392,10 @@ namespace pcpp
 			/// @return The raw column data as a vector of bytes
 			std::vector<uint8_t> getData() const
 			{
+				if (!m_Data)
+				{
+					return {};
+				}
 				return { m_Data, m_Data + m_DataLen };
 			}
 
--- a/Packet++/src/PostgresLayer.cpp
+++ b/Packet++/src/PostgresLayer.cpp
@@ -734,10 +734,9 @@ namespace pcpp
 		}
 
 		const char* queryStart = reinterpret_cast<const char*>(m_Data) + headerLen;
-		const size_t maxQueryLen = m_DataLen - headerLen;
-
-		const char* nullPos = static_cast<const char*>(memchr(queryStart, '\0', maxQueryLen));
-		return std::string(queryStart, nullPos != nullptr ? nullPos : queryStart + maxQueryLen);
+		const char* queryEnd = queryStart + m_DataLen - headerLen;
+		queryEnd = std::find(queryStart, queryEnd, '\0');
+		return { queryStart, queryEnd };
 	}
 
 	std::vector<PostgresRowDescriptionMessage::PostgresColumnInfo> PostgresRowDescriptionMessage::getColumnInfos() const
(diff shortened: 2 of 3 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -1,7 +1,7 @@
 name: Build and test
 on:
   push:
-    branches: ["master", "dev"]
```
seladb, line 4: I'll revert that when the [previous PR](https://github.com/seladb/PcapPlusPlus/pull/2070) is merged

Tests/ExamplesTest/tests/test_utils.py
```diff
@@ -77,7 +77,7 @@ def _root_path(self, request):
         self.root_path = request.config.getoption("--root-path")
 
     def run_example(
-        self, args, timeout=10, expected_return_code=0, requires_root=False
+        self, args, timeout=20, expected_return_code=0, requires_root=False
```
seladb, line 80: I'm not sure why we get a timeout sometimes, but for now increasing it to 20 secs seems to work 😕

Packet++/src/PostgresLayer.cpp
```diff
@@ -706,6 +724,204 @@ namespace pcpp
...
+			column.typeOID = be32toh(fixedData->typeOID);
+			column.typeSize = be16toh(fixedData->typeSize);
+			column.typeModifier = be32toh(fixedData->typeModifier);
+			auto formatCode = be16toh(fixedData->formatCode);
+			column.format =
+			    formatCode < 2 ? static_cast<PostgresColumnFormat>(formatCode) : PostgresColumnFormat::Unknown;
+
+			offset += sizeof(::internal::PostgresColumnFixedData);
+			columns.push_back(column);
+		}
```
Dimi1010, line 795: note: This can possibly be reworked to utilize a stream-like iterator parsing, instead of having to do `offset` and `remaining`, but it is fine for now.

Packet++/src/PostgresLayer.cpp
```diff
@@ -706,6 +724,204 @@ namespace pcpp
...
+		if (m_DataLen < headerLen + 1)
+		{
+			return "";
+		}
+
+		const char* queryStart = reinterpret_cast<const char*>(m_Data) + headerLen;
+		const size_t maxQueryLen = m_DataLen - headerLen;
+
+		const char* nullPos = static_cast<const char*>(memchr(queryStart, '\0', maxQueryLen));
+		return std::string(queryStart, nullPos != nullptr ? nullPos : queryStart + maxQueryLen);
```
Dimi1010: Simplification using iterator based approach.
```suggestion
		const char* queryStart = reinterpret_cast<const char*>(m_Data) + headerLen;
		const char* queryEnd = queryStart + m_DataLen - headerLen;
		queryEnd = std::find(queryStart, queryEnd, '\0');
		return std::string(queryStart, queryEnd);
```

Packet++/header/PostgresLayer.h
```diff
@@ -308,6 +308,189 @@ namespace pcpp
...
+			/// @param[in] data A pointer to the raw column data
+			/// @param[in] dataLen Size of the data in bytes
+			ColumnData(const uint8_t* data, size_t dataLen) : m_Data(data), m_DataLen(dataLen)
+			{}
+
+			/// @return The raw column data as a vector of bytes
+			std::vector<uint8_t> getData() const
+			{
+				return { m_Data, m_Data + m_DataLen };
+			}
```
Dimi1010, line 400: This will explode if called on `isNull() == true` column.
I think that is fine, but perhaps it would be nice to add that to the documentation of the method.

Alternatively, it could return empty vector?

Dimi1010, line 400: Possible enhancement: Depending on usage, we could possibly allow direct read-only access to the underlying data to save on copy operations? It is fine for now tho.

Dimi1010 (approved):
Minor note and suggestion on possible refactor, but overall LGTM.

Packet++/src/PostgresLayer.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2073/commits/26fa79073fcdcd3c619229ad1361974ac6d18d1c (I hope I understand the comment)

Packet++/src/PostgresLayer.cpp
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2073/commits/26fa79073fcdcd3c619229ad1361974ac6d18d1c

Packet++/header/PostgresLayer.h
seladb: Addressed in https://github.com/seladb/PcapPlusPlus/pull/2073/commits/26fa79073fcdcd3c619229ad1361974ac6d18d1c

seladb: I agree we can probably do it later if needed

Packet++/src/PostgresLayer.cpp
Dimi1010: Mostly yes. Opened #2080 with some further possible simplifications.

PR #2082, opened 2026-03-16
Deduplication of pcap header read procedures.

Dimi1010:
This PR deduplicates the pcap header read procedures shared between `PcapReaderDevice` and `PcapWriterDevice` in append mode.

The affected procedures have been moved to `readPcapHeader` and a mirror `writePcapHeader` functions inside `PcapFileDevice.cpp` that have internal linkage.
The PR also removes `PcapWriterDevice::checkHeader` as it can be integrated directly into `open()` by using `readPcapHeader`.

### DIFF
--- a/Pcap++/src/PcapFileDevice.cpp
+++ b/Pcap++/src/PcapFileDevice.cpp
@@ -281,15 +281,123 @@ namespace pcpp
 	// PcapFileReaderDevice members
 	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-	static uint16_t swap16(uint16_t value)
+	namespace
 	{
-		return static_cast<uint16_t>((value >> 8) | (value << 8));
-	}
+		uint16_t swap16(uint16_t value)
+		{
+			return static_cast<uint16_t>((value >> 8) | (value << 8));
+		}
 
-	static uint32_t swap32(uint32_t value)
-	{
-		return (value >> 24) | ((value >> 8) & 0x0000FF00) | ((value << 8) & 0x00FF0000) | (value << 24);
-	}
+		uint32_t swap32(uint32_t value)
+		{
+			return (value >> 24) | ((value >> 8) & 0x0000FF00) | ((value << 8) & 0x00FF0000) | (value << 24);
+		}
+
+		enum class PcapReadHeaderStatus
+		{
+			Ok,
+			NoData,
+			MalformedData,
+			UnsupportedFormat
+		};
+
+		/// @brief Reads a pcap file header from the given input stream and fills the provided header structure,
+		/// timestamp precision, and byte swap flag.
+		///
+		/// The function will populate the file header structure and convert it to host byte order.
+		///
+		/// @param inStream The input stream to read from.
+		/// @param header A pcap_file_header structure that will be filled with the header data read from the stream.
+		/// @param precision The precision of the timestamps in the pcap file.
+		/// @param needsSwap If the file contents need to be byte-swapped to match the host's endianness.
+		/// @return The status of the reading operation.
+		PcapReadHeaderStatus readPcapHeader(std::istream& inStream, pcap_file_header& header,
+		                                    FileTimestampPrecision& precision, bool& needsSwap)
+		{
+			inStream.read(reinterpret_cast<char*>(&header), sizeof(header));
+
+			auto readBytes = inStream.gcount();
+
+			// Clear any error flags that may have been set by read() to allow further operations on the stream
+			inStream.clear();
+			if (readBytes == 0)
+			{
+				return PcapReadHeaderStatus::NoData;
+			}
+			else if (readBytes < static_cast<std::streamsize>(sizeof(header)))
+			{
+				return PcapReadHeaderStatus::MalformedData;
+			}
+
+			switch (header.magic)
+			{
+			case TCPDUMP_MAGIC:
+			{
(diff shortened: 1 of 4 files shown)


Pcap++/src/PcapFileDevice.cpp
```diff
@@ -184,6 +189,19 @@ namespace pcpp
...
+		switch (precision)
+		{
+		case FileTimestampPrecision::Microseconds:
+			return "Microseconds";
+		case FileTimestampPrecision::Nanoseconds:
+			return "Nanoseconds";
+		default:
+			return "Unknown";
+		}
+	}
```
seladb, line 203: Why did you extract it from a lambda function?

seladb (commented):
@Dimi1010 I'm not sure we need this change. It does remove some code duplication, but with a price of extracting pcap format checks outside of the `PcapFile*Device` which I don't think it great

Dimi1010:
> @Dimi1010 I'm not sure we need this change. It does remove some code duplication, but with a price of extracting pcap format checks outside of the `PcapFile*Device` which I don't think it great

Generally, I think that the deduplication is the lesser price, since we only have to maintain it at 1 point, therefor less chance for bugs and easier maintenance.

As for extracting it outside of the device. I agree, its not ideal, but I think its acceptable to have helper functions outside of them if needed provided they are modular enough (like the pcap header code).

Personally, I view the ***Device classes more as a common API, rather than "must contain all implementation details inside". Having all implementation there is nice, but it can hurt maintainability, especially where complex duplicated logic is involved across devices.

Pcap++/src/PcapFileDevice.cpp
Dimi1010: Mostly cleanup to improve readability.

I moved it because its a `toString` that did not have any specific functionality to the location of the lambda. The returned strings are generic.

IMO, having it as a free function would be easier to use (and find that it exists), if it is needed again somewhere in the file, instead of having to either do another lambda.

seladb:
@Dimi1010 maybe we can contain an interface / class that contains the common logic? C++ has multiple inheritance which is not always recommended, but in this case maybe it makes sense to use it

Dimi1010:
> @Dimi1010 maybe we can contain an interface / class that contains the common logic? C++ has multiple inheritance which is not always recommended, but in this case maybe it makes sense to use it

@seladb I would prefer to avoid using inheritance purely for code reuse in favor of composition, since inheritance changes the externally observed object hierarchy and it's much harder to modify down the line. The new potential interfaces would also need to be exposed in the public headers for the devices to inherit from them.

If we want to contain the logic in a class, I suppose we can have a helper class local to the `PcapFileDevice.cpp` file, tho at that point it has little practical difference from the current implementation aside from grouping. 🤔 

The composition hierarchy I envisioned is something like this:
```
PcapHeaderLogic: A logical group of procedures for reading / writing a pcap header.
 ⤷ used in PcapWriterFileDevice: To read / write the header data
 ⤷ used in PcapReaderFileDevice: To read the header data
```

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -281,15 +299,120 @@ namespace pcpp
...
+		                                    FileTimestampPrecision& precision, bool& needsSwap)
+		{
+			inStream.read(reinterpret_cast<char*>(&header), sizeof(header));
+
+			auto readBytes = inStream.gcount();
+			if (readBytes == 0)
+			{
+				return PcapReadHeaderStatus::NoData;
+			}
+			else if (readBytes < static_cast<std::streamsize>(sizeof(header)))
```
seladb: nit: the `else` is redundant

Pcap++/src/PcapFileDevice.cpp
seladb: We only used it in one function so I don't see the point extracting it, but if you insist... 🤷‍♂️

Pcap++/src/PcapFileDevice.cpp
```diff
@@ -281,15 +299,120 @@ namespace pcpp
...
+				header.thiszone = swap32(header.thiszone);
+				header.sigfigs = swap32(header.sigfigs);
+				header.snaplen = swap32(header.snaplen);
+				header.linktype = swap32(header.linktype);
+			}
+
+			return PcapReadHeaderStatus::Ok;
+		}
+
+		bool writePcapHeader(std::ostream& outStream, FileTimestampPrecision precision, uint32_t snaplen,
```
seladb: This method is only used in `PcapFileWriter`, so maybe it can continue to be a member of the class

Pcap++/src/PcapFileDevice.cpp
Dimi1010: Eh, I am not strongly insisting on the change. I can revert it, but it seemed cleaner if it's not defined in-between the logic of the `open()`, since its not strictly needed to be defined there and `open` is somewhat long.

Pcap++/src/PcapFileDevice.cpp
Dimi1010: Reverted 6555932c5880fa837dff79b8099207affde1d1eb

Pcap++/src/PcapFileDevice.cpp
Dimi1010: Fixed e21b0133e6059439012853515d4c71f0d06992a7

seladb (approved):
LGTM

PR #2085, opened 2026-03-19
Improve benchmarks to support benching with PcapNG and Snoop files.

Dimi1010:
The PR adds support to specify `PcapNG` or `Snoop` files to the benchmark-google project to allow benchmarking of the file read/write speeds on those devices.

### DIFF
--- a/Examples/PcapPlusPlus-benchmark/README.md
+++ b/Examples/PcapPlusPlus-benchmark/README.md
@@ -11,10 +11,15 @@ A benchmark application used for measuring PcapPlusPlus performance can be found
 
 Another application integrates with the Google Benchmark library and can be found in `benchmark-google.cpp`. This application currently consists of five different benchmarks, and each benchmark can be influenced by various factors. These benchmarks aim to utilize different influence factors to provide accurate results for different scenarios. You can check the table below for more information. For performance-critical applications using PcapPlusPlus, it is recommended to run benchmarks in your specific environment for more accurate results. Using larger pcap files and those with diverse protocols and sessions can provide better insights into PcapPlusPlus performance in your setup.
 
-|     Benchmark        |   Operation   |  Influencing factors |
-|:-----------------:   |:-------------:|:--------------------:|
-| BM_PcapFileRead      |     Read      |  CPU + Disk (Read)   |
-| BM_PcapFileWrite     |     Write     |  CPU + Disk (Write)  |
-| BM_PacketParsing     | Read + Parse  |  CPU + Disk (Read)   |
-| BM_PacketPureParsing |     Parse     |        CPU           |
-| BM_PacketCrafting    |     Craft     |        CPU           |
+### Benchmarks and their influencing factors
+
+Supported files denote the file formats that can be used as a data source for the benchmark.
+
+
+|     Benchmark        |   Supported files   |   Operation   |  Influencing factors |
+|:--------------------:|:-------------------:|:-------------:|:--------------------:|
+| BM_FileRead          | pcap, pcapng, snoop |     Read      |  CPU + Disk (Read)   |
+| BM_FileWrite         |       N/A           |     Write     |  CPU + Disk (Write)  |
+| BM_PacketParsing     | pcap, pcapng, snoop | Read + Parse  |  CPU + Disk (Read)   |
+| BM_PacketPureParsing | pcap, pcapng, snoop |     Parse     |        CPU           |
+| BM_PacketCrafting    |       N/A           |     Craft     |        CPU           |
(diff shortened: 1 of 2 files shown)


Examples/PcapPlusPlus-benchmark/benchmark-google.cpp
```diff
@@ -53,18 +66,61 @@ static void BM_PcapFileRead(benchmark::State& state)
...
+	else if (endsWith(filepath, ".pcapng"))
 	{
-		state.SkipWithError("Cannot open pcap file for writing");
+		writer = std::make_unique<pcpp::PcapNgFileWriterDevice>(filepath);
+	}
+	else if (endsWith(filepath, ".snoop"))
+	{
+		state.SkipWithError("Snoop file writing is not supported");
+		return;
+	}
```
seladb: I don't think this is needed as we don't support snoop writing, we can treat it as any other file

Examples/PcapPlusPlus-benchmark/benchmark-google.cpp
```diff
@@ -91,24 +147,39 @@ static void BM_PcapFileWrite(benchmark::State& state)
 	state.SetBytesProcessed(totalBytes);
 	state.SetItemsProcessed(totalPackets);
 }
-BENCHMARK(BM_PcapFileWrite);
 
-static void BM_PacketParsing(benchmark::State& state)
+BENCHMARK_CAPTURE(BM_FileWrite, Pcap, "benchmark-output.pcap");
+BENCHMARK_CAPTURE(BM_FileWrite, PcapNg, "benchmark-output.pcapng");
+// BENCHMARK_CAPTURE(BM_FileWrite, Snoop, "benchmark-output.snoop"); // Snoop file writing is not supported
```
seladb: This is probably redundant

Examples/PcapPlusPlus-benchmark/README.md
```diff
@@ -11,10 +11,15 @@ A benchmark application used for measuring PcapPlusPlus performance can be found
...
-| BM_PacketCrafting    |     Craft     |        CPU           |
+### Benchmarks and their influencing factors
+
+Supported files denote the file formats that can be used as a data source for the benchmark.
+
+
+|     Benchmark        |   Supported files   |   Operation   |  Influencing factors |
+|:--------------------:|:-------------------:|:-------------:|:--------------------:|
+| BM_FileRead          | pcap, pcapng, snoop |     Read      |  CPU + Disk (Read)   |
+| BM_FileWrite         |       N/A           |     Write     |  CPU + Disk (Write)  |
```
seladb: Shouldn't `Supported files` be `pcap, pcapng`?

seladb (approved):
Please see a few comments, otherwise LGTM

Examples/PcapPlusPlus-benchmark/README.md
Dimi1010: Maybe? 

Generally the supported files column denominates the files that can be used as a "source" file to use for the reading benchmark through the CLI. The write operations write static payload to static files so IMO the column not applicable.

Examples/PcapPlusPlus-benchmark/benchmark-google.cpp
Dimi1010: Fixed b3040561b1340e1f0e30f26bd8483782b7232817

Examples/PcapPlusPlus-benchmark/benchmark-google.cpp
Dimi1010: Fixed b3040561b1340e1f0e30f26bd8483782b7232817

Examples/PcapPlusPlus-benchmark/README.md
seladb: Technically you're right, but since we run `BM_FileWrite` twice with pcap and pcapng files I think adding them to `"Supported files"` is more correct

Dimi1010: Fixed bed88b58d9dff5cf921a2811b6f963d27bc984a6

Dimi1010:
@seladb re-request because I changed the CLI scheme.

Instead of `--pcap-file [pcap filepath] --pcapng-file [pcapng filepath] --snoop-file [snoop filepath]`,
the CLI now only has `--pcap-file [any supported file]` that can be repeated an arbitrary amount of times.

The benchmarks now also include the specific file used as a source instead of just Pcap , Pcapng or Snoop.
(e.g `BM_FileRead/<filename>`)

seladb (approved):
LGTM

PR #2087, opened 2026-03-21
Extract logging tests to a separate executable.

### LINKED ISSUE #2083: Logging tests do not restore original logging executable configuration.
Currently the logging tests restore the "factory" logging configuration, not the actual one when the test was started.

Steps to reproduce: Enable '-b' (Debug flag) on the Pcap++ test executable. Observe how no debug logs are emitted after the logging tests finish.

Also, the logging test fails if the executable is started with '-b' flag.

Expected behaviour: The tests should restore the logging config as is was at the start of the test.

Additional note:
It might be a good idea to isolate the logging tests into a separate executable?

Dimi1010:
Fixes #2083

## Bug

After the logger tests are ran, the logger configuration is restored to its default state instead of to the state it was prior to the test being ran.

## Proposed fix

A proposed fix is to extract all logger tests to a separate executable. This is because the logger tests modify the logger configuration and callback, The callback specifically cannot be extracted from the logger, so a better solution would be to separate those tests in a new program.


Dimi1010:
Well, this is inconvenient. I reworked the multithreaded test to actually stress test the logger for synchronization issues, and apparently `MemPlumber` infrastructure starts to seg fault. Possibly because it can't handle 2 threads deallocating at once?

Tests/Logger++Test/LoggerTests.cpp
```diff
@@ -148,38 +169,51 @@ class LoggerCleaner
...
-	for (int i = 0; i < 1000; i++)
+
+	for (int i = 0; i < numMessages; i++)
 	{
 		pcpp::invokeErrorLog(threadIdAsString);
+
 		int sleepTime = dist(simpleRand);
 		std::this_thread::sleep_for(std::chrono::milliseconds(sleepTime));
 	}
 }
```
Dimi1010, line 192: This test is faulty as it does not test multithreaded calls reliably.

The current implementation does test calling the logger from different threads, but does not reliably perform them simultaneously due to the large sleep intervals. The chance of Thread1 and Thread2 calling the logger at the same time to test for actual sync issues is exceedingly rare.

It is also a large time sink in the CI pipeline, due to the test "faking" work by sleeping a lot.

---

Originally the PR included fixes for this, but I decided to separate those to keep the PR focused.

Tests/Logger++Test/LoggerTests.cpp
```diff
@@ -34,66 +35,86 @@ namespace pcpp
 class LogPrinter
```
Dimi1010, line 35: Converted `LogPrinter` to an instance to simplify the code as the logger can now take a lambda callback.

PR #2088, opened 2026-03-24
Wrap macro test driver code inside ptf::TestRunner

Dimi1010:
The PR wraps the test driver code that was previously embedded in the macros `PTF_START_RUNNING_TESTS`, `PTF_RUN_TEST` and `PTF_END_RUNNING_TESTS` inside the `ptf::TestRunner` class.

### Rationale
The change improves debugging UX by allowing step through debugging inside the test driver code.
Previously the embedded macro code was not "seen" by the IDE environment as actual code. This meant that debugging functionality like breakpoints and step through debuggers were unavailable inside those fragments. The entire macro code was treated as one line, instead of a code block.

### Changes

The PR moves the code from the embedded macros into a wrapper class `TestRunner`. This allows the IDE to treat the code as regular code instead of macro text, thus allowing improved debugger support.

The functionality has been moved according to the following table:
| Functionality | Original location  | New location |
|-- | ------------- | ------------- |
| Setup test environment | PTF_START_RUNNING_TESTS  | `TestRunner` constructor  |
| Run test case | PTF_RUN_TEST  | `TestRunner::runTest` |
| Display aggregated test results | PTF_END_RUNNING_TESTS  | `TestRunner::finalizeResults`  |

### Compatibility 
The old macros are kept as thin wrappers over the `TestRunner` instance to preserve compatibility with existing code.


seladb:
@Dimi1010 this is (yet another) PR with many changes that is hard to review... TBH I'm not sure what's the benefit of it...

Dimi1010:
> @Dimi1010 this is (yet another) PR with many changes that is hard to review... TBH I'm not sure what's the benefit of it...

The main benefit is improved IDE step through debugging support. When the code is directly embedded into the macro, it can't be stepped through via debugger (which I needed at the time I made this) since the IDE doesn't "see" the source code in the files. You also can't place breakpoints on it.

For example `PTF_RUN_TEST`, in the old version is treated as 1 line of code for the step through debugger. In the new version it can be stepped through line by line.

The changes are mainly just moving the code without functional changes.
`PTR_RUN_TEST` -> `TestRunner::runTest`
`PTF_START_RUNNING_TESTS` -> `TestRunner` ctor
`PTF_END_RUNNING_TESTS` ->`TestRunner::finalizeResults`

I can separate the rename of  `__ptfSplitString` to `ptfSplitString`, because that is independent. But the rest are tied together or it won't compile.

The reason behind the rename of `__ptfSplitString` and `__ptfCheckTags` is because the current naming breaks this rule in the C++ standard for reserved identifiers:

> The following identifiers are reserved and may not be declared in a program (doing so invokes undefined behavior):
> - [All identifiers that begin with an underscore followed by a capital letter or by another underscore (these reserved identifiers allow the library to use numerous behind-the-scenes non-external macros and functions).](https://en.cppreference.com/w/c/language/identifier.html#:~:text=All%20identifiers%20that%20begin%20with%20an%20underscore%20followed%20by%20a%20capital%20letter%20or%20by%20another%20underscore%20(these%20reserved%20identifiers%20allow%20the%20library%20to%20use%20numerous%20behind%2Dthe%2Dscenes%20non%2Dexternal%20macros%20and%20functions).)

Edit: The link is to the C reference, but it also applies to the C++ language spec [here](https://en.cppreference.com/w/cpp/language/name.html#:~:text=in%20the%20global%20namespace%2C%20identifiers%20that%20begin%20with%20an%20underscore).

> in the global namespace, identifiers that begin with an underscore

PS: I suppose I should have elaborated more in the PR description in the first place. Sorry.

Dimi1010:
@seladb Reverted the changes that aren't critical to the PR and improved the PR description. Hope it helps. :)

Dimi1010:
@seladb do you think we can merge this? It has been sitting ready for a while.

It is moving the test driver logic to sit in pure cpp code instead of inside macro defines.

PR #2099, opened 2026-04-11
Fix silent cast of `ProtocolTypeFamily` to `ProtocolType` in Packet ctor.

Dimi1010:
This PR changes the signature of the full Packet parsing constructor to use `ProtocolTypeFamily` instead of `ProtocolType`.

This is done to allow protocol "family" constants to be used in the main signature. Previously used family constants (`uint32_t`) were silently narrowed to the first protocol of the family (`uint8_t`). (e.g "IP" family would be treated as only "IPv4") if used in the full constructor signature.

### DIFF
--- a/Packet++/header/Packet.h
+++ b/Packet++/header/Packet.h
@@ -83,15 +83,16 @@ namespace pcpp
 		/// @param[in] rawPacket A pointer to the raw packet
 		/// @param[in] freeRawPacket Optional parameter. A flag indicating if the destructor should also call the raw
 		/// packet destructor or not. Default value is false
-		/// @param[in] parseUntil Optional parameter. Parse the packet until you reach a certain protocol (inclusive).
-		/// Can be useful for cases when you need to parse only up to a certain layer and want to avoid the performance
-		/// impact and memory consumption of parsing the whole packet. Default value is ::UnknownProtocol which means
-		/// don't take this parameter into account
+		/// @param[in] parseUntil Optional parameter. Parse the packet until you reach a certain protocol family
+		/// (inclusive). Can be useful for cases when you need to parse only up to a certain layer and want to avoid the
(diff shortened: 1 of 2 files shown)


Packet++/header/Packet.h
```diff
@@ -83,15 +83,16 @@ namespace pcpp
...
+		/// (inclusive). Can be useful for cases when you need to parse only up to a certain layer and want to avoid the
+		/// performance impact and memory consumption of parsing the whole packet. Default value is ::UnknownProtocol
+		/// which means don't take this parameter into account
 		/// @param[in] parseUntilLayer Optional parameter. Parse the packet until you reach a certain layer in the OSI
 		/// model (inclusive). Can be useful for cases when you need to parse only up to a certain OSI layer (for
 		/// example transport layer) and want to avoid the performance impact and memory consumption of parsing the
 		/// whole packet. Default value is ::OsiModelLayerUnknown which means don't take this parameter into account
-		explicit Packet(RawPacket* rawPacket, bool freeRawPacket = false, ProtocolType parseUntil = UnknownProtocol,
+		explicit Packet(RawPacket* rawPacket, bool freeRawPacket = false,
+		                ProtocolTypeFamily parseUntil = UnknownProtocol,
```
seladb: This changes the method's signature from `uint8_t` to `uint32_t`. I can't think of a specific use case, but I'm concerned it may break backward compatibility for users 🤔

Dimi1010: Hmm, we could alternatively add that as a separate overload? That would also fix the issue.

seladb: I guess that it might be better, I'm not sure though...

Dimi1010: Fixed 5120e8f641cc7465ec1cd24683f0b6a4b7775acb

seladb (approved):
LGTM

PR #2100, opened 2026-04-11
Fix include: Add <ostream> include to "Layer.h".

Dimi1010:
`std::ostream` is used in the overload for `std::ostream& operator<<(std::ostream&, Layer&)`. That will fail compilation if ostream is not defined, and the header currently relies on external definition from somewhere else.


seladb (approved):
LGTM

PR #2104, opened 2026-04-13
Simplify TcpReassembly connection purge to use multimap.

Dimi1010:
The change removes the need to create a list for situations where only 1 connection will be purged at time T.
Instead the additional flow keys at time T will be directly inserted into the multimap as additional nodes.

The change also simplifies the purge code down to a single loop over the range of nodes `[beginIt, upperBound(T))`.
Since `multimap<K, V>` is an ordered container, the connections are purged in the order: oldest expire time -> oldest insertion.

### DIFF
--- a/Packet++/header/TcpReassembly.h
+++ b/Packet++/header/TcpReassembly.h
@@ -483,7 +483,7 @@ namespace pcpp
 
 		void closeConnectionInternal(uint32_t flowKey, ConnectionEndReason reason);
 
-		void insertIntoCleanupList(uint32_t flowKey);
+		void scheduleCleanup(uint32_t flowKey);
 	};
 
 }  // namespace pcpp
--- a/Packet++/header/TcpReassembly.h
+++ b/Packet++/header/TcpReassembly.h
@@ -460,7 +460,7 @@ namespace pcpp
 		};
 
 		using ConnectionList = std::unordered_map<uint32_t, TcpReassemblyData>;
-		using CleanupList = std::map<time_t, std::list<uint32_t>>;
+		using CleanupList = std::multimap<time_t, uint32_t>;
 
 		OnTcpMessageReady m_OnMessageReadyCallback;
 		OnTcpConnectionStart m_OnConnStart;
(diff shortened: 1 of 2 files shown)


Packet++/header/TcpReassembly.h
```diff
@@ -460,7 +460,7 @@ namespace pcpp
 		};
 
 		using ConnectionList = std::unordered_map<uint32_t, TcpReassemblyData>;
-		using CleanupList = std::map<time_t, std::list<uint32_t>>;
+		using CleanupList = std::multimap<time_t, uint32_t>;
```
tigercosmos: Maybe rename it to `CleanupMultiMap`. The original naming is misleading.

Packet++/src/TcpReassembly.cpp
```diff
@@ -798,16 +798,14 @@ namespace pcpp
...
-		CleanupList::mapped_type& keysList = pair.first->second;
-		keysList.push_front(flowKey);
+		// m_CleanupList is a multimap with key of type time_t (expiration time) and a value type uint32_t (flow key).
+		// Due to being a multimap, multiple flow keys can have the same expiration time.
+		//
+		// During purge, up to 'maxNumToClean' entries with expiration time smaller than the current time will be
+		// removed from storage.
+
+		auto expireTime = time(nullptr) + m_ClosedConnectionDelay;
+		m_CleanupList.insert(std::make_pair(expireTime, flowKey));
```
tigercosmos: I wonder if `emplace` is better.

Dimi1010: Fixed b16dddbeb9da70678a560ff224ff1849c5e6447a

Packet++/header/TcpReassembly.h
Dimi1010: Fixed cffae2c7f462db924c4a38a0602a5b7fded863fe.

Also renamed `insertIntoCleanupList` to `scheduleCleanup` in b3c7020444297bfa9132b2bf0ceed71f2de1d19c due to being missleading after the refactor.

seladb (approved):

PR #2106, opened 2026-04-15
Add DPDK tests

seladb:
- Replace the existing DPDK CI job with one that runs the tests we have for DPDK
- Build DPDK locally on the GitHub Actions runner VM and cache it instead of using the docker image
- Use DPDK [AF_PACKET PMD](https://doc.dpdk.org/guides/nics/af_packet.html) in order to run the tests in CI with network traffic
- Setup 2 virtual Ethernet interfaces (`veth`) that are connected to each other: packets are injected to `veth1` which mirrors the traffic to `veth0`. DPDK listens to `veth0`
- Drop support for DPDK 22.11. I'll add the latest LTS version DPDK 25.11 in a subsequent PR
- Fix areas in the code that didn't work


cmake/modules/FindDPDK.cmake
seladb, line 1: `FindDPDK` didn't really work before if DPDK wasn't install in `/usr/local`. The changes here fix it

Tests/Pcap++Test/Common/TestUtils.cpp
seladb, line 1: This was added so we can pass the DPDK args that are used in the test:
- `--no-pci`
- `--vdev=net_af_packet0,iface=veth0`

Tests/Pcap++Test/Tests/DpdkTests.cpp
```diff
@@ -219,12 +219,6 @@ class DpdkTestWorkerThread : public pcpp::DpdkWorkerThread
 			uint16_t packetReceived = m_DpdkDevice->receivePackets(mBufArr, 32, m_QueueId);
 			lock.unlock();
 			m_PacketCount += packetReceived;
-			lock.lock();
-			uint16_t packetsSent = m_DpdkDevice->sendPackets(mBufArr, packetReceived, m_QueueId);
-			if (packetsSent != packetReceived)
-			{
-				return false;
-			}
```
seladb, line 227: Sending a lot of packets when using AF_PACKET seem to block the TX queue and the test gets stuck. I didn't manage to find a way to fix it, but I don't think TX really matters here because we have separate tests for sending packets

Pcap++/src/MBufRawPacket.cpp
```diff
@@ -328,17 +328,19 @@ namespace pcpp
...
 			return;
 		}
 
-		RawPacket::setRawData(rte_pktmbuf_mtod(mBuf, const uint8_t*), rte_pktmbuf_pkt_len(mBuf), timestamp,
-		                      LINKTYPE_ETHERNET);
 		m_MBuf = mBuf;
+		m_MbufDataSize = mBufDataSize;
+		auto mBufDataLen = rte_pktmbuf_pkt_len(mBuf);
+		m_RawDataLen = mBufDataLen;
+		RawPacket::setRawData(rte_pktmbuf_mtod(mBuf, const uint8_t*), mBufDataLen, timestamp, LINKTYPE_ETHERNET);
```
seladb, line 343: This fixes a bug: when creating `MBufRawPacket` and calling `setMBuf()` right away with an existing mbuf if wouldn't work because `m_MbufDataSize` is initialized to 0.

In general I think we should refactor some of the code here to make the internal API more clear, but we should do it in a separate PR

Pcap++/src/MBufRawPacket.cpp
Dimi1010: Doesn't this technically mutate the mempool of the `DpdkDevice` by allocating from it?
I think keeping the device non-const seems more correct in that case?

Pcap++/src/MBufRawPacket.cpp
Dimi1010: > In general I think we should refactor some of the code here to make the internal API more clear, but we should do it in a separate PR

I agree that the internal API needs some work. Perhaps we can reconsider #1899 to separate `RawPacket` and `MBufRawPacket` and have them both inherit from a new base class? It would fix a lot of the current debt due to needing the two classes to work around each other.

Pcap++/header/MBufRawPacket.h
```diff
@@ -87,7 +87,7 @@ namespace pcpp
 		/// @param[in] device The DpdkDevice which has the pool to allocate the mbuf from
 		/// @return True if initialization succeeded and false if this method was already called for this instance (and
 		/// an mbuf is already attached) or if allocating an mbuf from the pool failed for some reason
-		bool init(DpdkDevice* device);
+		bool init(const DpdkDevice* device);
```
Dimi1010: Future note:
I think it might be beneficial to look into changing the `init` method to either a constructor, have a factory method on the device like take over the initialization of a packet on the device buffer.

Some example signatures can be:
- `MBufRawPacket* DpdkDevice::createMBufRawPacket()` - `MBufRawPacket` proper allocated on the heap.
- `bool /* or status enum */ DpdkDevice::attachRawPacket(MBufRawPacket&)`
API primary for stack allocated packets. Attaches the raw packet to the current device's mempool. If it has an already allocated mempool attached, the procedure releases the old buffer back to its pool.

Dimi1010 (commented):
1 concern on `MBufRawPacket::init`'s const change and some notes on possible future improvement.

Otherwise LGTM.

Pcap++/src/MBufRawPacket.cpp
seladb: I don't think allocating from the mempool mutates it. I mean, it obviously changes it, but I think `const` is nicer. If you insist I can revert it

Pcap++/header/MBufRawPacket.h
seladb: Yes, I thought about it, but an empty constructor is beneficial for use-cases like initializing containers of `MBufRawPacket`: `std::vector<MBufRawPacket> vec`.

Having said that, we can think about it more

seladb:
> 1 concern on `MBufRawPacket::init`'s const change and some notes on possible future improvement.
> 
> Otherwise LGTM.

@Dimi1010 you didn't approve the PR so I can't merge it... if you think it's LGTM, can you approve it?

Pcap++/src/MBufRawPacket.cpp
Dimi1010: > I don't think allocating from the mempool mutates it. I mean, it obviously changes it, [...]

Isn't that contradictory? A mutation is a change in the internal state of an object?

The pool does presumably keep an internal state of which parts of it are used or not. So allocating from it would change that state.

Therefor the promise of state preservation done by `const` would be broken by the allocation.

I think it would be better to keep it non-const.

Pcap++/header/MBufRawPacket.h
Dimi1010: We can keep the empty ctor. That is what the `attach` method was meant to solve.

```cpp
MBufPacket pak; // default ctor
DpkdDevice::attachRawPacket(pak);
// ready for use
```

Dimi1010:
> @Dimi1010 you didn't approve the PR so I can't merge it... if you think it's LGTM, can you approve it?

I didn't set it because of the const concern. Once that is resolved, I will set it.

Pcap++/header/MBufRawPacket.h
seladb: i guess we can think about it. We `DpdkDevice` is the class that usually creates `MBufPacket`, that's why it uses `setMbuf()`...

Pcap++/src/MBufRawPacket.cpp
seladb: Update in https://github.com/seladb/PcapPlusPlus/pull/2106/commits/5730a5a343e7640900165028d0c712ccdb636950

Pcap++/src/DpdkDevice.cpp
```diff
@@ -974,9 +974,12 @@ namespace pcpp
 		{
 			struct rte_mbuf* mBuf = mBufArray[index];
 			if (rawPacketsArr[index] == nullptr)
+			{
 				rawPacketsArr[index] = new MBufRawPacket();
+				rawPacketsArr[index]->init(const_cast<DpdkDevice*>(this));
```
Dimi1010, line 979: It's only somewhat related, but why is `receivePackets` const, but `sendPackets` is not?

I would think that both would be non-const since they conceptually either push or pull data to/from the wire.

We should probably address that in the future.

Dimi1010 (approved):

Pcap++/src/DpdkDevice.cpp
seladb: These are old APIs so I don't remember 🤷‍♂️

Dimi1010:
Oops... accidentally clicked review on copilot on mobile...

PS: Apparently you can cancel the run in actions.

.github/workflows/build_and_test.yml
```diff
@@ -312,71 +312,98 @@ jobs:
...
+          python3 -m pip install gcovr==${{ matrix.gcovr-version || '8.6' }}
+          gcovr -v -r . ${{ matrix.additional-gcov-flags }} $GCOVR_FLAGS -o coverage.xml
+
+      - name: Upload Coverage Results
+        uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24 # v5.4.3
         with:
-          path: ${{ env.CCACHE_DIR }}
-          key: ${{ steps.ccache-restore.outputs.cache-primary-key }}
+          files: ./coverage.xml
+          flags: ${{ matrix.image }},unittest
```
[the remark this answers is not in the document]
seladb: Fixed in https://github.com/seladb/PcapPlusPlus/pull/2106/commits/3056b7f2c8eb4fafce8bf0e046d7ca77176d2eb6

PR #2108, opened 2026-04-21
Replace IPNetwork underlying network storage with a tagged union.

Dimi1010:
This PR replaces the way the underlying `IPv4Network` and `IPv6Network` objects are stored inside `IPNetwork`.

The old implementation used to store the two network types into discrete `std::unique_ptr<NetworkT>` types. While that storage type is compact for the `IPNetwork` struct, it introduces an indirection cost and mandatory heap allocation for no specific benefit due to `IPNetwork` utilizing value semantics for public access.

The new implementation features a tagged union that stores the underlying network type directly inside the `IPNetwork` object
removing the indirection cost and heap allocation, while preserving exiting functionality.

Additional note: The custom variant structure can be replaced by `std::variant` when C++17 becomes the minimum required standard.


Common++/header/IpAddress.h
```diff
@@ -762,125 +744,242 @@ namespace pcpp
...
+			}
+
+			return m_NetworkVariant.tryGetIPv6()->toString();
 		}
 
 	private:
-		std::unique_ptr<IPv4Network> m_IPv4Network;
-		std::unique_ptr<IPv6Network> m_IPv6Network;
+		// TODO: C++17 Replace with std::variant.
+		class NetworkVariant
```
seladb, line 901: I wonder if we need this internal class... why can't we change the private members of `IPNetwork` to be a union?

Dimi1010: We can probably do without it, but I made it so it would be simpler to replace with `std::variant` member later.

It also gets easier to shoot yourself in the foot with lifetime issues if a lot of logic is kept in the union class, so I thought it better to keep the union in an internal class which only manages the lifetime.

seladb: Ok that makes sense 👍

Common++/header/IpAddress.h
```diff
@@ -762,125 +744,242 @@ namespace pcpp
...
 		uint8_t getPrefixLen() const
 		{
-			return (m_IPv4Network != nullptr ? m_IPv4Network->getPrefixLen() : m_IPv6Network->getPrefixLen());
+			auto ip4 = m_NetworkVariant.tryGetIPv4();
+			if (ip4 != nullptr)
+			{
+				return ip4->getPrefixLen();
+			}
+
+			return m_NetworkVariant.tryGetIPv6()->getPrefixLen();
```
seladb: My IDE warns that `m_NetworkVariant.tryGetIPv6()` could be `nullptr`. While it can't be true in reality , I think we should think handle it somehow. Maybe introduce `getIPv6()` that throws an exception if the underlying type is wrong? 🤔

Dimi1010: Possibly fixed cecf3ebee1af2c632af2a68338f39a2233eac00e?

seladb (approved):
LGTM

PR #2111, opened 2026-04-23
Fix #2109: Recognize LLC payload in SLL2.

### LINKED ISSUE #2109: PcapPlusPlus does not create an LLC layer for SLL/SLL2 pcap files
### Bug description

**Describe the bug**
When reading SLL/SLL2 pcap files, PcapPlusPlus does not create the LLC layer and the ones following it.

**Steps to reproduce**
The problem can be easily reproduced with PcapPrinter using the attached file:

```
$ ./PcapPrinter sll2-stp.pcap
File summary:
~~~~~~~~~~~~~
   File name: sll2-stp.pcap
   File size: 270 bytes
   Link layer type: Linux cooked capture v2

Packet length: 66 [Bytes], Arrival time: 2026-03-30 16:58:46.046179000
Linux cooked header v2
Payload Layer, Data length: 46 [Bytes]

Packet length: 66 [Bytes], Arrival time: 2026-03-30 16:58:47.846305000
Linux cooked header v2
Payload Layer, Data length: 46 [Bytes]

Packet length: 66 [Bytes], Arrival time: 2026-03-30 16:58:49.647154000
Linux cooked header v2
Payload Layer, Data length: 46 [Bytes]

Finished. Printed 3 packets
```

[sll2-stp.pcap.tar.gz](https://github.com/user-attachments/files/26923490/sll2-stp.pcap.tar.gz)

PcapPlusPlus master branch

_No response_

Linux

_No response_

### Compiler version

g++ 13.3.0

_No response_

Dimi1010:
The PR addresses #2109 and adds checks for LLC payload inside an SLL2 frame.

The protocol type values for the LLC payload have been sourced from https://www.tcpdump.org/linktypes/LINKTYPE_LINUX_SLL2.html


Tests/Packet++Test/Tests/Sll2Tests.cpp
Dimi1010, line 1: Not a very comprehensive test, but it just needs to verify that the LLC layer is discovered.

Packet++/src/Sll2Layer.cpp
```diff
@@ -9,10 +9,39 @@
...
+		/// This function mimics the behavior of std::to_underlying from C++23.
+		///
+		/// @tparam E The enum class type.
+		/// @param e The value of the enum class to convert.
+		/// @return The underlying type value corresponding to the enum class value.
+		template <typename E> constexpr auto to_underlying(E e) noexcept
+		{
+			return static_cast<std::underlying_type_t<E>>(e);
+		}
+	}  // namespace
```
seladb, line 21: Why do we need all this complexity for a simple const? 🤔

Dimi1010: Since Sll2 has 7 exclusive protocol values (that are not part of Ethernet), I thought it might be nice to eventually have them grouped into an enum?

We don't really need it atm per se, so we can have a single const value for now, if you think its better.

seladb: const is simpler, and simpler is usually better 🙂 I think we can start with a const and if we need an enum we can do that in the future. It's not a public API so updating to an enum won't break anything

Dimi1010: Fixed b3f859c7cea0bcd74606640c32911388ab8aa456

seladb (approved):
LGTM

PR #2112, opened 2026-04-23
Add basic tool to dump pcap to test dat resource format.

Dimi1010:
This PR adds a test tool for dumping a packet to the test hex .dat format using the resource utility.

Since are mostly using the hex format to store packet samples for tests, I think it would be useful to have a standard tool to generate them.

PS: I think we don't have such a tool yet? Or at least I haven't found it?


seladb:
@Dimi1010 for creating the hex string I usually use Wireshark. I don't think we need this tool in PcapPlusPlus.

image

Dimi1010:
Huh... didn't know about that option. Sure, that works.

PR #1962, opened 2025-09-12
Use file content heuristics to decide file reader.

Dimi1010:
The PR adds heuristics based on the file content that is more robust than deciding based on the file extension.

The new decision model scans the start of the file for its magic number signature. It then compares it to the signatures of supported file types [[1]](https://en.wikipedia.org/wiki/List_of_file_signatures) and constructs a reader instance based on the result.

A new function `createReader` and `tryCreateReader` has been added due to changes in the public API of the factory.
The functions differ in the error handling scheme, as `createReader` throws and `tryCreateReader` returns `nullptr` on error.

Method behaviour changes during erroneous scenarios:
| Scenario | `getReader` | `createReader` | `tryCreateReader` |
| ------------- | ------------- | ------------- | ------------- |
| File not found | N/A | Throws exception | Return `nullptr` |
| Unsupported format | Return `PcapFileDeviceReader` | Throws exception | Return `nullptr` |

PR #2113, opened 2026-04-25
Fix #2081: Apply zstd compression level reliably across the documente…

### LINKED ISSUE #2081: Zstd compression level mapping
### Bug description

**Describe the bug**
[Documentation](https://pcapplusplus.github.io/docs/features#readingwriting-pcapng-files-with-compression) describes a 0-10 range which is not properly mapped to zstd's range of 1-22. The scaled value is saved to the PcapPlusPlus context but not the zstd compression parameter.

https://github.com/seladb/PcapPlusPlus/blob/d1263a24b271c828d981227e3abd437c7abc2b0b/3rdParty/LightPcapNg/LightPcapNg/src/light_zstd_compression.c#L60-L61

**Expected behavior**
The scaling that gets applied to the context. For reference [Technica-Engineering/LightPcapNG ](https://github.com/Technica-Engineering/LightPcapNg/blob/55c7f975cd97a43cca835da2f4dd5178364eaab2/src/light_io_zstd.c#L267-L270) uses a slightly different calculation

```
//Input is scale 0-9 but zstd goes 1-22!
compression_level = (compression_level * 2) + 1;
```

PcapPlusPlus master branch

_No response_

Linux

_No response_

### Compiler version

GCC 13.3.0

libpcap

MerinoSheep:
## Description                
  Fixes #2081.                  

  The zstd compression level passed to `PcapNgFileWriterDevice` was not being                                                      
  applied as documented. Two related defects in `light_zstd_compression.c`:

  ### Root Cause                
  1. `ZSTD_CCtx_setParameter` was wrapped in `assert()`, so the call (along                                                        
     with its side effect of actually setting the level) is stripped under                                                         
     `NDEBUG`. In Release builds the requested level was never set; zstd fell                                                      
     back to its default (3).       
  2. The 0–10 input range documented for `PcapNgFileWriterDevice` was scaled                                                       
     into the context (`* 2`) but the original unscaled value was the one                                                          
     passed to zstd, and the scaling did not span zstd's 1–22 range.                                                               

  ### Proposed Fix              
  - Extract `ZSTD_CCtx_setParameter` into a named local so its side effect                                                         
    survives `NDEBUG`, and check the return value separately.    
  - Map input `1..10` → zstd `3..21`, preserve `0` as the documented                                                               
    "default compression" sentinel, and pass the same scaled value to zstd                                                         
    that is stored on the context.                               

  ### Test                      
  Adds `TestPcapNgZstdCompressionLevels`: writes the same input pcapng at                                                   
  levels 1 and 10 and asserts the level-10 file is strictly smaller. With                                                          
  either bug present the two files are the same size and the assertion                                                             
  fires.

### DIFF
--- a/Tests/Pcap++Test/Tests/FileTests.cpp
+++ b/Tests/Pcap++Test/Tests/FileTests.cpp
@@ -1314,6 +1314,31 @@ PTF_TEST_CASE(TestPcapNgFileReadWrite)
 
 }  // TestPcapNgFileReadWrite
 
+PTF_TEST_CASE(TestPcapNgZstdCompressionLevels)
+{
+	// If the compression level is silently ignored, low and high levels
+	// produce identical output sizes.
+	int sizes[2] = { 0, 0 };
+	const int levels[2] = { 1, 10 };
+	for (int i = 0; i < 2; i++)
+	{
+		pcpp::PcapNgFileReaderDevice readerDev(EXAMPLE_PCAPNG_PATH);
+		PTF_ASSERT_TRUE(readerDev.open());
+
+		pcpp::PcapNgFileWriterDevice writerDev(EXAMPLE_PCAPNG_ZSTD_LEVELS_WRITE_PATH, levels[i]);
+		PTF_ASSERT_TRUE(writerDev.open());
+
+		pcpp::RawPacket rawPacket;
+		while (readerDev.getNextPacket(rawPacket))
+			PTF_ASSERT_TRUE(writerDev.writePacket(rawPacket));
+		readerDev.close();
+		writerDev.close();
+
+		sizes[i] = getFileLength(EXAMPLE_PCAPNG_ZSTD_LEVELS_WRITE_PATH);
+	}
+	PTF_ASSERT_GREATER_THAN(sizes[0], sizes[1]);
+}  // TestPcapNgZstdCompressionLevels
+
 PTF_TEST_CASE(TestPcapNgFileReadWriteAdv)
 {
 	pcpp::PcapNgFileReaderDevice readerDev(EXAMPLE2_PCAPNG_PATH);
--- a/Tests/Pcap++Test/Tests/FileTests.cpp
+++ b/Tests/Pcap++Test/Tests/FileTests.cpp
@@ -1317,12 +1317,15 @@ PTF_TEST_CASE(TestPcapNgFileReadWrite)
 PTF_TEST_CASE(TestPcapNgZstdCompressionLevels)
 {
 	// If the compression level is silently ignored, low and high levels
-	// produce identical output sizes.
-	int sizes[2] = { 0, 0 };
-	const int levels[2] = { 1, 10 };
-	for (int i = 0; i < 2; i++)
+	// produce identical output sizes. PcapPlusPlus exposes levels 0-10,
+	// which the zstd wrapper maps to zstd's native 1-22 range. Test three
+	// points to verify the level scales monotonically, not just that the
+	// endpoints differ.
+	int sizes[3] = { 0, 0, 0 };
+	const int levels[3] = { 1, 5, 10 };
+	for (int i = 0; i < 3; i++)
 	{
-		pcpp::PcapNgFileReaderDevice readerDev(EXAMPLE_PCAPNG_PATH);
+		pcpp::PcapNgFileReaderDevice readerDev(EXAMPLE2_PCAPNG_PATH);
 		PTF_ASSERT_TRUE(readerDev.open());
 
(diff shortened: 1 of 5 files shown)


Tests/Pcap++Test/Tests/FileTests.cpp
```diff
@@ -1314,6 +1314,31 @@ PTF_TEST_CASE(TestPcapNgFileReadWrite)
...
+	{
+		pcpp::PcapNgFileReaderDevice readerDev(EXAMPLE_PCAPNG_PATH);
+		PTF_ASSERT_TRUE(readerDev.open());
+
+		pcpp::PcapNgFileWriterDevice writerDev(EXAMPLE_PCAPNG_ZSTD_LEVELS_WRITE_PATH, levels[i]);
+		PTF_ASSERT_TRUE(writerDev.open());
+
+		pcpp::RawPacket rawPacket;
+		while (readerDev.getNextPacket(rawPacket))
+			PTF_ASSERT_TRUE(writerDev.writePacket(rawPacket));
```
Dimi1010, line 1338: nit: Can we add `{}` on this?

Dimi1010 (commented):
The added unit test seems to fail?

Tests/Pcap++Test/Tests/FileTests.cpp
MerinoSheep: fixed

MerinoSheep:
I made some tweak to the test and forgot to run it again ill fix it up.

Tests/Pcap++Test/Tests/FileTests.cpp
```diff
@@ -1314,6 +1314,37 @@ PTF_TEST_CASE(TestPcapNgFileReadWrite)
...
 
+PTF_TEST_CASE(TestPcapNgZstdCompressionLevels)
+{
+	// If the compression level is silently ignored, low and high levels
+	// produce identical output sizes. PcapPlusPlus exposes levels 0-10,
+	// which the zstd wrapper maps to zstd's native 1-22 range. Test three
+	// points to verify the level scales monotonically, not just that the
+	// endpoints differ.
+	int sizes[3] = { 0, 0, 0 };
+	const int levels[3] = { 1, 5, 10 };
```
seladb: nit: we can use `std::array` here

Tests/Pcap++Test/Tests/FileTests.cpp
```diff
@@ -1314,6 +1314,37 @@ PTF_TEST_CASE(TestPcapNgFileReadWrite)
...
+		{
+			PTF_ASSERT_TRUE(writerDev.writePacket(rawPacket));
+		}
+		readerDev.close();
+		writerDev.close();
+
+		sizes[i] = getFileLength(EXAMPLE_PCAPNG_ZSTD_LEVELS_WRITE_PATH);
+	}
+	PTF_ASSERT_GREATER_THAN(sizes[0], sizes[1]);
+	PTF_ASSERT_GREATER_THAN(sizes[1], sizes[2]);
```
seladb, line 1346: CI is failing for almost all platforms, maybe this is not always true? 🤔 
I asked Claude and it suggests checking the window size from the header. Compression levels 1, 5, and 10 _should_ result with different windows sizes

MerinoSheep: I took a look at `.github/workflows/build_and_test.yml` and it seems the only one that passes is the ubuntu container that has zstd configured on. From what I've seen if zstd is not available at run time the writer silently writes uncompressed files with the zstd extension.

Dimi1010: That appears to be because the compression context in LightPcapNG is `NULL` if its compiled without zst. It effectively performs a silent fallback to uncompressed, since `NULL` is also used if no compression is requested.

I suppose we could catch that in the constructor or open call of the PcapNg writer. 

PR #1962 does add some methods to detect if zst is detected at compile time, tho that one is still open.

seladb: Maybe the test can use the `USE_Z_STD` macro that is used in LighPcapNg, otherwise it can skip the test:
```cpp
PTF_TEST_CASE(TestPcapNgZstdCompressionLevels)
{
#ifdef USE_Z_STD
	// The test...
#else
	PTF_SKIP_TEST("Zstd is not configured");
#endif
}
```

MerinoSheep: Passing now 😄

Dimi1010 (approved):
LGTM. Thank you for the effort!

seladb (approved):
LGTM

seladb:
@MerinoSheep thank you so much for working on it, much appreciated! 🙏

PR #2114, opened 2026-04-26
Add DPDK 25.11 support

### LINKED ISSUE #2066: DPDK 25.11 removal of `rte_eth_stats` `q_***` members.
The previously deprecated in [DPDK 23.11](https://doc.dpdk.org/guides-23.07/rel_notes/deprecation.html#:~:text=ethdev%3A%20Queue%20specific,will%20be%20removed.) `rte_eth__stats` `q_***` members used to fetch queue specific statistics have been removed in 25.11. This API is currently used in `DpdkDevice::getStatistics()` to fetch per queue stats.

The instructions for updating in the deprecation notice recommend fetching queue statistics via xstats API.

This issue needs to be resolved to support building PCPP with with DPDK 25.11.

seladb:
Fixes https://github.com/seladb/PcapPlusPlus/issues/2066.

Per-queue stats were removed in DPDK 25.11 (as mentioned in section 1.2 [here](https://doc.dpdk.org/guides-25.11/rel_notes/release_25_11.html)), and it's now required to use DPDK xstats API instead.

The xtats API is more complex than the previous API. Each stat in each queue has a name and an ID. In order to collect stats we need to find their IDs which are determined in runtime. Hence, the following steps are needed:
1. When the device is initiated, call `rte_eth_xstats_get_id_by_name()` for each stat in each RX and TX queue we want to collect stats for. Save these IDs on the `DpdkDevice`
2. When collecting the stats gather all stat IDs in one array and call `rte_eth_xstats_get_by_id()` which retrieves the value for each stat ID
3. Translate this structure into the array of `RxTxStats` we have for RX and TX queue

This logic was wrapped in a new private class `QueueStatsCollector`. It's initialized when the device is opened and gathered the stat IDs. Then when collecting the stats it uses the IDs to get the values.


.github/workflows/build_and_test.yml
```diff
@@ -341,7 +342,7 @@ jobs:
             && tar -xvf dpdk-${{ matrix.dpdk_version }}.tar.xz \
             && rm -rf dpdk-${{ matrix.dpdk_version }}.tar.xz \
             && mv dpdk-stable-${{ matrix.dpdk_version }} dpdk && cd dpdk \
-            && meson setup build --prefix=${{ github.workspace }}/dpdk/install \
+            && meson setup build -Dmachine=default --prefix=${{ github.workspace }}/dpdk/install \
```
seladb, line 345: This change is needed because we cache the DPDK build and apparently GitHub Actions runners don't guarantee a specific CPU architecture so the cached build might not work. This happened in this build for example: https://github.com/seladb/PcapPlusPlus/actions/runs/24953406570/job/73067473423

Pcap++/src/DpdkDevice.cpp
```diff
@@ -1579,6 +1566,112 @@ namespace pcpp
...
+		}
+	}
+
+	void DpdkDevice::QueueStatsCollector::resolveStatId(const std::string& statName, uint64_t& idOut) const
+	{
+		if (rte_eth_xstats_get_id_by_name(m_PortId, statName.c_str(), &idOut) != 0)
+		{
+			throw std::runtime_error("Failed to resolve xstat ID for: " + statName);
+		}
+	}
```
Dimi1010, line 1649: Can't we use return value instead of out param here? It would be simpler API.

Pcap++/src/DpdkDevice.cpp
```diff
@@ -1579,6 +1566,112 @@ namespace pcpp
...
+		for (uint16_t qId = 0; qId < m_TxQueueCount; ++qId)
+		{
+			resolveStatId("tx_q" + std::to_string(qId) + "_packets", txPacketIds[qId]);
+			resolveStatId("tx_q" + std::to_string(qId) + "_bytes", txByteIds[qId]);
+		}
+
+		m_RxPacketIds = rxPacketIds;
+		m_RxByteIds = rxByteIds;
+		m_TxPacketIds = txPacketIds;
+		m_TxByteIds = txByteIds;
```
Dimi1010: The `vector` buffers are created and then copied.

It is probably better to use the member variables directly and call `resize` or use move assignment instead of copy assignment. Resize is more efficient on reusing existing buffers if any, and move assign preserves the old values if the resolution fails midway.

Pcap++/src/DpdkDevice.cpp
```diff
@@ -1579,6 +1566,112 @@ namespace pcpp
...
+	void DpdkDevice::QueueStatsCollector::collect(DpdkDeviceStats& stats, const DpdkDeviceStats& prevStats,
+	                                              double secsElapsed) const
+	{
+		std::memset(stats.rxStats, 0, sizeof(stats.rxStats));
+		std::memset(stats.txStats, 0, sizeof(stats.txStats));
+
+		// Pack all IDs into a flat vector for a single API call
+		const size_t totalStatsCount = (m_RxQueueCount + m_TxQueueCount) * m_StatsPerQueue;
+		std::vector<uint64_t> statIds(totalStatsCount);
+		std::vector<uint64_t> statValues(totalStatsCount);
```
Dimi1010, line 1601: nit, performance note: How often is collect to be called?

If it is called many times, we may want to profile and check if caching the scratch buffers to avoid the allocation cost on every collect gives noticeable improvement.

If it is only called 1~2 times per second tho, it probably isn't worth it tho.

Dimi1010 (approved):
Some minor notes. LGTM otherwise.

Pcap++/src/DpdkDevice.cpp
seladb: Maybe it's not super important, but calling:
```cpp
resolveStatId("rx_q" + std::to_string(qId) + "_packets", rxPacketIds[qId]);
```

Is faster than calling:
```cpp
rxPacketIds[qId] = resolveStatId("rx_q" + std::to_string(qId) + "_packets");
```

What do you think?

Pcap++/src/DpdkDevice.cpp
seladb: The reason they're copied is because in case stats ID fetching fail, we don't want to stay in an inconsistent state. I guess we can use the move assignment instead of copy, but it doesn't really matter because this only happens once in startup and not as part of the fast path

Pcap++/src/DpdkDevice.cpp
Dimi1010: Really? I would have thought they would be roughly the same.
I suppose it might be something with `rte_eth_xstats_get_id_by_name` writing directly to the array. 🤔 

Sure we can keep it as is then.

Pcap++/src/DpdkDevice.cpp
seladb: I think that stats shouldn't be collected every millisecond or so, they're probably collected 1-2 time per second or even less, so it probably doesn't matter

Pcap++/src/DpdkDevice.cpp
seladb: `rte_eth_xstats_get_id_by_name` will write directly to the array, vs using a string copy c'tor. Maybe it doesn't matter a lot, but still...

Pcap++/src/DpdkDevice.cpp
Dimi1010: Personally I would change them to moves then, but I guess it doesn't matter too much so its up to you.

seladb: Sure, updated in https://github.com/seladb/PcapPlusPlus/pull/2114/commits/9a2570f8976ce02542d03d141b3f7ff080943378

PR #2115, opened 2026-04-28
Improve range checks on RawPacket::insert and remove.

Dimi1010:
This PR adds guards for negative `atIndex` input in `RawPacket`.

Previously `atIndex` was only checked for upper bound, which allowed negative values.
That situation would have resulted dereferencing `basePtr /*m_Data */ + (-N) /* atIndex */` which is out of bounds.


seladb (approved):
LGTM

PR #2116, opened 2026-04-29
Add Ubuntu 26.04 and remove Ubuntu 20.04

seladb:
- Add Ubuntu 26.04 to CI
- Remove all images that are based on Ubuntu 20.04


Dimi1010 (approved):
LGTM. I assume the 2604 image has zst support by default and no need for separate image?

seladb:
> LGTM. I assume the 2604 image has zst support by default and no need for separate image?

I installed `libzstd-dev` as part of the image: https://github.com/seladb/PcapPlusPlus-DockerImages/blob/1db467887241eeed110511b1f1ab9d0f0c110633/Dockerfile-ubuntu2604#L29

PR #2117, opened 2026-04-30
Improve AF_XDP CI tests using veth interfaces

seladb:
AF_XDP tests are [very flaky](https://github.com/seladb/PcapPlusPlus/actions/runs/25093206821/job/73523840137) because the AF_XDP socket doesn't capture the packets replayed by tcpreplay to the same interface.

The solution is similar to what we do in [DPDK CI tests](https://github.com/seladb/PcapPlusPlus/pull/2106) which is create 2 veth interfaces, run tcpreplay on `veth1` and capture packets with AF_XDP on `veth0`.

### DIFF
--- a/Tests/Pcap++Test/main.cpp
+++ b/Tests/Pcap++Test/main.cpp
@@ -30,28 +30,29 @@ static struct option PcapTestOptions[] = {
 
 void printUsage()
 {
-	std::cout << "Usage: Pcap++Test -i ip_to_use | [-n] [-b] [-s] [-m] [-r remote_ip_addr] [-p remote_port] [-d "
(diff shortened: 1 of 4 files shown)


Tests/Pcap++Test/main.cpp
Dimi1010, line 1: 🤔 Wondering if it might be beneficial to split tests that require different frameworks into different executables? (e.g Core/Pcap, DPDK + KNI, XDP)

This argument list is getting somewhat long.

It also might simplify some potential use cases.
For example, running a test with DPDK enabled library, without actually running the dpdk test.

Dimi1010 (approved):
LGTM.

Tests/Pcap++Test/main.cpp
seladb: I'm not sure it's needed. It's nicer to have one test suite that has everything...

Issue #2101, opened 2026-04-13
Provide version checking functionality


KangLin:
@seladb 
The API was changed in #2056, but this project does not provide version checking functionality, which results in the application being unable to maintain compatibility(Supports both 25.05 and 24.09 etc)

Semantic Versioning: https://semver.org/

seladb:
@KangLin there is a version in this project, you can find all versions here: https://pcapplusplus.github.io/versions
We're using [CalVer](https://calver.org/) instead of SemVer.

Unfortunately we didn't make a new release for almost a year, and the latest stable version is v25.05, which was released in May 2025.

The PR you're referring to was merged in Jan 2026 and is part of the next release, which we hope to create soon. Our "dev" version is [`v25.05+`](https://github.com/seladb/PcapPlusPlus/blob/c3cc787829b794b84c6a63734d324c1c487cb696/Common%2B%2B/header/PcapPlusPlusVersion.h#L11) until we create the next release.

PcapPlusPlus does provide an API to get the version which you can find here: https://pcapplusplus.github.io/api-docs/next/_pcap_plus_plus_version_8h.html

KangLin:
@seladb Thank you for your reply.
The currently provided version information can be used for version checks during travel, but cannot be used for version checks during compilation. What I need is to perform version checks during compilation in order to achieve compatibility.
eg:
```
/*
   QT_VERSION is (major << 16) + (minor << 8) + patch.
*/
#define QT_VERSION      QT_VERSION_CHECK(QT_VERSION_MAJOR, QT_VERSION_MINOR, QT_VERSION_PATCH)
/*
   can be used like #if (QT_VERSION >= QT_VERSION_CHECK(4, 4, 0))
*/
#define QT_VERSION_CHECK(major, minor, patch) ((major<<16)|(minor<<8)|(patch))

#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
    #include <QScreen>
#endif
```

seladb:
[...]

PR #2118, opened 2026-04-30
Add compile-time version macros + update the version to `26.07-dev`

seladb:
Addresses #2101.

This PR adds version macros and introduces a patch version in case we need it:
```cpp
#define PCAPPLUSPLUS_VERSION_YEAR 26
#define PCAPPLUSPLUS_VERSION_MONTH 7
#define PCAPPLUSPLUS_VERSION_PATCH 0
#define PCAPPLUSPLUS_VERSION_DEV 1
```

It also includes macros to compare versions, for example:
```cpp
#if PCAPPLUSPLUS_VERSION_HIGHER_OR_EQUAL_THAN(26, 7, 0)
    /// Use new APIs introduces in v26.07
#endif
```


seladb:
@Dimi1010 this PR is still in draft, but I wanted to get your opinion on this approach

Common++/header/PcapPlusPlusVersion.h
```diff
@@ -8,9 +8,97 @@
...
+/// @brief The patch number component of the PcapPlusPlus version
+#define PCAPPLUSPLUS_VERSION_PATCH 0
+
+/// @def PCAPPLUSPLUS_VERSION_DEV
+/// @brief Development flag: non-zero for development/nightly builds, zero for official releases
+#define PCAPPLUSPLUS_VERSION_DEV 1
+
+/// @def PCAPPLUSPLUS_VERSION
+/// @brief Short version string (e.g., "25.05" for official release or "25.05+" for development)
 #define PCAPPLUSPLUS_VERSION "25.05+"
```
seladb: It's possible to create this string using macros, but it'll be very complex, so maybe it's easier to keep it as is...

Dimi1010: I suppose we can manually set it.

A potential alternative is to have the cmake build system generate the version file, like it does with the package config files. I think that is generally more flexible and we would only need to change the version in the CMakeLists.

---

Also, do you think we should include the patch number in all versions? (eg. 25.05.0), so people always expect it if they decide to do magic with the version string?

Dimi1010 (commented):
Looks ok. I am a bit iffy on `PCAPPLUSPLUS_VERSION_DEV`, but I don't have any better ideas for it, so... 🤷

Common++/header/PcapPlusPlusVersion.h
Dimi1010: > A potential alternative is to have the cmake build system generate the version file, like it does with the package config files. I think that is generally more flexible and we would only need to change the version in the CMakeLists.

@seladb Additional discussion may be needed separately, but I managed to get CMake to compile a version from the git tags. The versioning is based on the signature `major.minor.patch.tweak`.

It fetches the MAJOR, MINOR, PATCH from the most recent tag. (eg. last release)
It then calculates the Tweak version as the number of commits past the most recent tag the build is. 

If we are precisely on the release commit, the tweak version will be 0 and the full version `maj.min.patch.0`, if we are 1 commit past it it will be `maj.min.patch.1`. This can then be used to determine official releases (tweak 0) vs nightly / dev builds (tweak != 0).

The solution will allow us to have rolling versions unique for each development build commit, so we don't have just a nebulous "DEV" version flag. For example the current master would be `25.05.0.201`.

A extended version string can also be added with a commit hash, if we need to differentiate between commits on different branches, but I think most will be used from the `master` / `dev` line.

KangLin:
it is recommended to use Semantic format. the major for the API，the minor for year, the path for date.
eg:
v1.26.05-dev
1: major
26: minor, the year
05: patch, the date
dev: develop, it is can dev count. eg: 1, 2, 3 etc

seladb:
> it is recommended to use Semantic format. the major for the API，the minor for year, the path for date. eg: v1.26.05-dev 1: major 26: minor, the year 05: patch, the date dev: develop, it is can dev count. eg: 1, 2, 3 etc

@KangLin I don't think we want to switch from CalVer to SemVer, at least not now

Common++/header/PcapPlusPlusVersion.h
seladb: I'm not sure we need a build version because we never publish those versions. We don't even publish patch versions... I think `year.month.patch` is enough for the version.


> I managed to get CMake to compile a version from the git tags

I'm curious how does it work?
- When does it run and fetch the tag from git? In every commit? 
- Does it fetch the latest tag?

Dimi1010: > I think year.month.patch is enough for the version.

Sure. I guess that works.

> I'm curious how does it work?
>
> - When does it run and fetch the tag from git? In every commit?
> - Does it fetch the latest tag?

It is based on `git describe --tags --long --always`, which is called during the CMake configure step. The command results in a string like `v25.05-211-g1a2ad56b` which can then be reformatted. From what I have seen, it fetches the latest tag that the current commit is derived from. The command can be configured to select only annotated tags (default), or any tags (`--tags`).

Also, on second thoughts it does have some inconvenient drawbacks. Main one is that it establishes a dependency on the git repo being available at configure time to compute the correct version. This leads to some problems when remote compilation is used, like building on WSL since the default behaviour of `rsync` does not copy the `.git` folder since it is not needed. The requirement should only be relevant when the lib is build from source tho, since installs don't really run the configure the same way.

seladb: Yes I agree. Version should probably be static and not rely on CMake configuration or on git.

Should we stick with what we have in this PR?

Dimi1010: > Version should probably be static and not rely on CMake configuration or on git.

Hmm, it might be fine to rely on CMake configuration to generate the Version.h header as long as the version in the configuration is static? We could have the cmake project version to be the "source of truth" for version.
```cmake
project(MyLib VERSION 2.4.1)

configure_file(
    ${CMAKE_CURRENT_SOURCE_DIR}/version.h.in
    ${CMAKE_CURRENT_BINARY_DIR}/generated/version.h
    @ONLY)
```
As a sidenote, setting project version would also allow consuming packages to designate required versions for the package directly in the build system like this `find_package(PcapPlusPlus 25.04 ...)`.

Anyway, that can be handled in a separate PR, if we decide to do it as it does not strictly relate to the current PR.

Dimi1010 (approved):
LGTM. One comment on possible integration with cmake package versioning, but that can be addressed separately.

Common++/header/PcapPlusPlusVersion.h
seladb: I'm not sure I understand this comment. Do you suggest adding another file `version.h` in addition to `PcapPlusPlusVersion.h` that is based on CMake?

As you said, we can probably merge this PR as is and address this separately later.

seladb:
@Dimi1010 I updated this PR to reflect the new version and I'm also changing the non-release version string from `XX.YY+` to `XX.YY-dev`. Do you have any concerns or is it good to merge?

Common++/header/PcapPlusPlusVersion.h
```diff
@@ -8,9 +8,97 @@
...
+/// @def PCAPPLUSPLUS_VERSION_PATCH
+/// @brief The patch number component of the PcapPlusPlus version
+#define PCAPPLUSPLUS_VERSION_PATCH 0
+
+/// @def PCAPPLUSPLUS_VERSION_DEV
+/// @brief Development flag: non-zero for development/nightly builds, zero for official releases
+#define PCAPPLUSPLUS_VERSION_DEV 1
+
+/// @def PCAPPLUSPLUS_VERSION
+/// @brief Short version string (e.g., "25.05" for official release or "25.05+" for development)
```
Dimi1010: Should probably update docs too.
```suggestion
/// @brief Short version string (e.g., "25.05" for official release or "25.05-dev" for development)
```

Common++/header/PcapPlusPlusVersion.h
```diff
@@ -8,9 +8,97 @@
...
+/// @brief Check if the current version is higher than or equal to the specified version
+/// @param year The year component to compare
+/// @param month The month component to compare
+/// @param patch The patch number to compare
+#define PCAPPLUSPLUS_VERSION_HIGHER_OR_EQUAL_THAN(year, month, patch)                                                  \
+	PCAPPLUSPLUS_VERSION_NUM >= PCAPPLUSPLUS_MAKE_VERSION(year, month, patch)
+
+/// @def PCAPPLUSPLUS_VERSION_FULL
+/// @brief Full version string including version and release type (e.g., "v25.05+ (non-official release)")
 #define PCAPPLUSPLUS_VERSION_FULL "v" PCAPPLUSPLUS_VERSION " (" PCAPPLUSPLUS_VERSION_OFFICIAL ")"
```
Dimi1010, line 102: This would become `v26.07-dev (non-official release)`. Example still shows old pattern.

Dimi1010 (approved):
Some examples need updating in the doc comments to the new pattern.

I also wonder if "non-official release" wouldn't be better named as "development build", but it is not that important atm.

LGTM otherwise.

Common++/header/PcapPlusPlusVersion.h
seladb: Good catch! Updated in https://github.com/seladb/PcapPlusPlus/pull/2118/commits/ca7d0ab7ba454b519b06e87f171228683e64d06e

Common++/header/PcapPlusPlusVersion.h
seladb: Fixed in https://github.com/seladb/PcapPlusPlus/pull/2118/commits/ca7d0ab7ba454b519b06e87f171228683e64d06e

PR #2119, opened 2026-04-30
Add standard upgrade path for `enum` to `enum class` transition.

Dimi1010:
## Description

Add a standardized backwards compatible upgrade path from `enum` to `enum class` via macro.

Upgrades the following enums to `enum class` as proof of concept:
- `DpdkPMDType`
- `PcapLiveDevice::PcapDirection`

## Rationale

Generally `enum class` enumerator types are preferred in modern post C++11 code, due to better scoping mechanics and less pollution of the parent scope. However, the changes in the scoping mechanics introduce a backwards incompatible change when updating existing `enum` types to `enum class`.

## Proposal

A proposed solution to preserve backwards compatibility is to define `const auto Value = EnumT::Value;`. This effectively mimics the old behaviour of spilling the enum values. These constant values are additionally flagged with a `[[deprecated]]` attribute to encourage users to transition to the new `enum class` value definitions.

To reduce the required boilerplate the PR adds a new macro `PCPP_ENUM_CLASS_UPGRADE_COMPAT(cls, val)` that defines the a constant `val` to the value `cls::val`, along with an appropriate deprecation message.

The PR additionally updates the enums `DpdkPMDType` and `PcapDirection` as a proof of concept.

### DIFF
--- a/Common++/header/DeprecationUtils.h
+++ b/Common++/header/DeprecationUtils.h
@@ -9,7 +9,7 @@
 // Provides backwards compatibility for enum class types that have been upgraded.
 //
 // This macro should be once used in the header file of the upgraded enum class type with each enum value.
-// This will define the enum value as a const auto with the same name, mimicing the old enum value, but
+// This will define the enum value as a const auto with the same name, mimicking the old enum value, but
 // with a deprecation warning to encourage users to switch to the new enum class value.
 //
 #define PCPP_ENUM_CLASS_UPGRADE_COMPAT(cls, val)                                                                       \
--- a/Common++/header/DeprecationUtils.h
+++ b/Common++/header/DeprecationUtils.h
@@ -6,6 +6,16 @@
 #	define PCPP_DEPRECATED(msg) [[deprecated(msg)]]
 #endif
 
+// Provides backwards compatibility for enum class types that have been upgraded..
+//
+// This macro should be once used in the header file of the upgraded enum class type with each enum value.
+// This will define the enum value as a const auto with the same name, mimicing the old enum value, but
+// with a deprecation warning to encourage users to switch to the new enum class value.
+//
+#define PCPP_ENUM_CLASS_UPGRADE_COMPAT(cls, val)                                                                       \
+	PCPP_DEPRECATED("Enum class upgrade: Use " #cls "::" #val " instead.")                                             \
+	const auto val = cls::val
+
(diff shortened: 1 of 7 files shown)


seladb:
This solution is not completely backward compatible, for example:
1. Implicit integer conversion is gone, for example, this will not work:
   ```cpp
   int myInt = A;  // A is the enum value
   ```
2. Using an enum with a template won't work anymore, for example:
   ```cpp
   template <int N>
   void MyMethod() 
   {
      ...
   };

   MyMethod<A>();
   ```
3. Forward declarations of the old enum break

In addition, I'm pretty sure that `const static` will create variables for all of the enum values which is wasteful and might take more memory (but I'm not 100% sure about it).

Dimi1010:
@seladb Yeah, good points. I did not consider `int` conversion compatibility. I mostly focused on preserving the names compatibility since that is the largest surface of breaking changes that happen in such a change.

Although, it is my opinion that the `enum` types that are upgraded to `enum class` really shouldn't be relying on implicit `int` conversion in the first place?

For 1. and 2. I am not sure that is fixable... 🤔 I can try but wouldn't hold my breath. `enum class` was specifically made to prevent that.

For 3. I haven't used forward declared enums previously although when I tested it now, they seem to be pretty brittle in general since the forward declaration needs to behave as a complete type. Technically supporting that use case makes all changes to `enum (class)`'s underlying types breaking changes too, I suppose.

From this answer here https://stackoverflow.com/a/1280969 and some experiments, I found this:
  - They require the forward declaration to know the underlying type of the `enum` or `enum class` in question at the time of forward declaration. Which IMO is smelly and a big footgun, if it needs to be used for 3rd party dependency enums.
  - Yes, they do fail to compile if the forward declaration does not match with the definition.
  - Funnily enough, they DO compile if the compiler does NOT see the forward declaration in the same TU as the actual declaration. Provided they use the same underlying type, the linker apparently doesn't care? If they aren't, you shot yourself in the foot, because it DOES compile and the linker doesn't care.

So, I am not sure that can be fixed either, tbf. :/

> In addition, I'm pretty sure that const static will create variables for all of the enum values which is wasteful and might take more memory (but I'm not 100% sure about it).

Probably yes. It is possible to make them `constexpr` which should inline them in most locations like an `enum`.

seladb:
I think we should reconsider this approach. We try to avoid breaking changes as much as possible...

PR #2123, opened 2026-05-02
Replace Logger TLS std::string `m_LastError` to std::array.

Dimi1010:
## Bug

The replacement is due to an issue on MinGW with TLS variables that have non-trivial destructors. This causes crashes when threads that have initialized their thread local instance of the `m_LastError` variable attempt to finalize and exit.

The issue was discovered in the course of development of #2087.

## Proposed fix

The proposed fix replaces the last error std::string with a C-string array buffer. Due to std::array having trivial destructor semantics, the issues present with std::string do not materialize.

The length of the buffer of 255 was chosen from the previous reservation buffer to hopefully catch the majority of error messages without truncating the messages.

### DIFF
--- a/Common++/header/Logger.h
+++ b/Common++/header/Logger.h
@@ -311,7 +311,7 @@ namespace pcpp
 		/// @return Get the last error message
 		std::string getLastError() const
 		{
-			return m_LastError;
+			return { m_LastError.data() };
 		}
 
 		/// Suppress logs in all PcapPlusPlus modules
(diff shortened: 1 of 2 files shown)


Dimi1010:
Tbf, the issue is old and I think supposed to be fixed already? I am not sure why it is crashing for the current environment.

I did a small sample with just a TLS `std::string` and it runs fine, but the logger test driver crashes?

seladb (approved):

Issue #2129, opened 2026-05-05
Provide compile time feature availability querying capability.


Dimi1010:
## Description

Currently once the library is built (and installed), it does not provide an easy way for users to query which features are available at compile time. The only indication that the library has not been build without certain features in some cases is the `include` headers missing, which leads to compile time errors and issues if the consuming library must seamlessly support several optional feature sets.

The main features to be considered for this proposal are:
- zstd file support
- libpcap / winpcap / npcap
- dpdk + kni
- pfring
- windivert
- xdp

## Proposal

A possible solution leverage CMake's target based compile definitions machinery to provide `PUBLIC` defines in the form of `PCPP_HAS_XXX_SUPPORT=1`. The public definition will in turn propagate the define to targets that consume the library target allowing them to be used at compile time.

Example change to cmake configuration:
```cmake
add_library(Pcap++ ...)

if(PCAPPP_USE_DPDK)
  target_compile_definitions(Pcap++ PUBLIC PCPP_HAS_DPDK_SUPPORT=1)
endif()
```

Example usage:
```cpp
#ifdef PCPP_HAS_DPDK_SUPPORT
#include "DpdkDeviceList.h"
#endif
```

Dimi1010:
Added in #2136

PR #2130, opened 2026-05-05
Add runtime flag for zstd compression support.

### LINKED ISSUE #1973: Add indication if LightPcapNG backend is compiled with ZSTD compression support.
The backend for parsing / crafting PcapNG files can be compiled with integrated ZST compression support. Currently while we do support that, we have no way to indicate if the backend is compiled with that support enabled after build. 

As the backend is baked into the library, it would be nice to have a compile time constant that indicates if Zst support is available to enable early detection and error reporting if compression options are used without backend support.

Dimi1010:
Fixes #1973

Extracted functionality from #1962

Adds a static method that returns `true` or `false` depending on runtime support of zstd compression.

### DIFF
--- a/Pcap++/CMakeLists.txt
+++ b/Pcap++/CMakeLists.txt
@@ -109,6 +109,7 @@ target_link_libraries(
 )
 
 if(LIGHT_PCAPNG_ZSTD)
+  target_compile_definitions(Pcap++ PRIVATE PCPP_PCAPNG_ZSTD_SUPPORT)
   target_link_libraries(Pcap++ PRIVATE light_pcapng)
 endif()
 
--- a/Pcap++/header/PcapFileDevice.h
+++ b/Pcap++/header/PcapFileDevice.h
@@ -324,6 +324,10 @@ namespace pcpp
 		internal::LightPcapNgHandle* m_LightPcapNg;
 
 	public:
+		/// @brief A static method that checks if the device was built with zstd compression support
+		/// @return True if zstd compression is supported, false otherwise.
+		static bool isZstdSupported();
+
 		/// A constructor for this class that gets the pcap-ng full path file name to open. Notice that after calling
 		/// this constructor the file isn't opened yet, so reading packets will fail. For opening the file call open()
 		/// @param[in] fileName The full path of the file to read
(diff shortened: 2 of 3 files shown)


Pcap++/CMakeLists.txt
```diff
@@ -109,6 +109,7 @@ target_link_libraries(
 )
 
 if(LIGHT_PCAPNG_ZSTD)
+  target_compile_definitions(Pcap++ PRIVATE PCPP_PCAPNG_ZSTD_SUPPORT)
```
seladb, line 112: Maybe we should make it `PUBLIC` so users can do this:
```cpp
#ifdef PCPP_PCAPNG_ZSTD_SUPPORT
// do something only if Zstd is supported
#endif
```

seladb (approved):
See one comment, otherwise LGTM

Pcap++/CMakeLists.txt
Dimi1010: I thought about it. Decided to leave it private for now until we decide what the standard pattern for those macros should be in #2129.

seladb: Maybe we can use the convention we already have of using the `PCAPPP` prefix, like `PCAPPP_HAS_ZSTD_SUPPORT`?

We already have a flag of `-DLIGHT_PCAPNG_ZSTD` that we might want to change to `-DPCAPPP_USE_ZSTD` (not in this PR of course)

Dimi1010: We kinda have several exposed prefixes tbh.

- Cmake options use PCAPPP in cmake.
- The c++ macros use PCPP prefix
- The c++ version macros use PCAPPLUSPLUS.

PS: If the current implementation, perhaps we can merge it as is and modify it to public in a separate PR?

seladb: I don't see a reason to open a separate PR just to expose it to public. In CMake we usually use `PCAPPP` so I think that could be the best choice

Dimi1010: In cmake we use `pcappp`, but the define is a C++ preprocessor macro, where it would conflict with the other 2 cpp macro prefixes we already use, with `pcpp` being more common of the two in my experience, so maybe we could use that? IMO, we should try to avoid adding more prefixes to the preprocessor macros as it can get confusing.


And the reason I would prefer it to be changed to public in a separate PR is so we can do the entire compile time feature set options described in #2129 in a single PR instead of having it spread across multiple PRs piecewise, if possible.

seladb: Sure, it can be `PCPP`. If you insist making it public in a separate PR that's ok, but I don't see a reason why not making it public now... anyway the PR is approved so you can merge it

PR #2136, opened 2026-05-10
Add standardized feature set properties to Pcap++ target.

### LINKED ISSUE #2129: Provide compile time feature availability querying capability.

Dimi1010:
Implements #2129

The feature set options follow the pattern `PCPP_HAS_XXX_SUPPORT`, with XXX being the target feature set. Currently if a feature set is active the cmake Pcap++ target will have a property with the corresponding name set to a truthy value and a corresponding compile time definition will be added to the compilation. Otherwise the target will have a false value on the property and no compile definition will be added to the compilation.

This should allow consumers of the library to query the target at for features both at configure time and compile time.


Pcap++/CMakeLists.txt
```diff
@@ -109,10 +109,63 @@ target_link_libraries(
 )
 
 if(LIGHT_PCAPNG_ZSTD)
-  target_compile_definitions(Pcap++ PRIVATE PCPP_PCAPNG_ZSTD_SUPPORT)
   target_link_libraries(Pcap++ PRIVATE light_pcapng)
 endif()
 
+# Pcap++ Feature sets:
+# - PCPP_HAS_PCAPNG_ZSTD_SUPPORT - Whether Pcap++ was built with support for zstd-compressed pcapng files using the light_pcapng library
+# - PCPP_HAS_PCAP_FILE_SUPPORT - Whether Pcap++ was built with support for reading/writing pcap files (Pcap++ always has pcap file support)
```
Dimi1010: `PCPP_HAS_PCAP_FILE_SUPPORT` is for symmetry with `PCPP_HAS_PCAP_LIVE_SUPPORT` and `PCPP_HAS_PCAP_REMOTE_SUPPORT`.

seladb: I don't think this is necessary. We should only export features which can be included or excluded

seladb (approved):

PR #2141, opened 2026-05-16
Add sized overload for `Layer::copyData`

Dimi1010:
The PR adds a overload of `Layer::copyData` that offers improved bounding checks against buffer overruns. It also improves the documentation of the methods.


Packet++/src/Layer.cpp
```diff
@@ -53,6 +53,13 @@ namespace pcpp
 		memcpy(toArr, m_Data, m_DataLen);
 	}
 
+	size_t Layer::copyData(uint8_t* dest, size_t destSize) const
+	{
+		size_t bytesToCopy = (std::min)(destSize, m_DataLen);
```
Dimi1010, line 59: `(std::min)(destSize, m_DataLen)` this is written as such because `Windows.h` is fun and has `MIN(a, b)` macro, that normally conflicts with `std::min`. 🙂

Packet++/header/Layer.h
```diff
@@ -182,10 +182,29 @@ namespace pcpp
...
+		/// The method copies up to 'destSize' bytes of the layer's raw data into the provided destination array.
+		/// If the layer's data length is greater than 'destSize', only the first 'destSize' bytes will be copied.
+		///
+		/// To ensure sufficient space is available in the destination array, use getDataLen() to determine the actual
+		/// length of the layer's data before calling this method.
+		///
+		/// @param[out] dest The destination byte array
+		/// @param[in] destSize The maximum number of bytes to copy
+		/// @return The number of bytes copied to the destination array.
+		size_t copyData(uint8_t* dest, size_t destSize) const;
```
seladb, line 206: Can we add tests for this new method?

Dimi1010: Added ba39b93

seladb: Perhaps we should add it in `PacketTests.cpp`?
We already have `CopyLayerAndPacketTest` that kind of fits, or we can add another test in this file

Dimi1010: Moved the test to `PacketTests.cpp` 9f5e360f7a7383b416c465672b14c49720df0284
Kept the test separate because Ctrl+C, Ctrl+V was easier and `CopyLayerAndPacketTest` is very large as is.

Tests/Packet++Test/Tests/PacketTests.cpp
```diff
@@ -558,6 +558,55 @@ PTF_TEST_CASE(CopyLayerAndPacketTest)
...
 
+PTF_TEST_CASE(CopyLayerDataTest)
+{
+	// Tests Layer::copyData() method using IPv4 Layer as a test subject.
+	using namespace pcpp::literals;
+
+	pcpp::IPv4Layer ip4layer("192.168.0.1"_ipv4, "192.168.0.2"_ipv4);
+
+	auto dataLen = ip4layer.getDataLen();
+	PTF_ASSERT_EQUAL(dataLen, 20);  // IPv4 header without options is 20 bytes long
```
seladb, line 569: I'm not sure we need this assert. IPv4 layer length is always 20 bytes

seladb (approved):
Please see minor comment, otherwise LGTM

Tests/Packet++Test/Tests/PacketTests.cpp
Dimi1010: I added it just to be sure and so it is visible with regards to the rest of the test. It is more like a pre-condition than an actual test case. I think its fine to keep it for readability?

seladb: I guess you can keep it if you insist, but I don't think it's likely for the IPv4 layer data length to ever change...

PR #2024, opened 2025-11-17
Improve SIP packet detection using heuristic parsing

sorooshm78:
This PR improves SIP packet detection in PcapPlusPlus by introducing heuristic parsing based on Wireshark’s SIP dissector. SIP messages are now detected from the UDP payload itself, not only when using port 5060.

* Add `static bool dissectSipHeuristic(const uint8_t* data, size_t dataLen)` to detect SIP requests/responses from payload content (Wireshark-style logic)
* Use the new heuristic in `UdpLayer` so SIP packets on non-standard ports are correctly classified
* Preserve existing behavior for non-SIP payloads

Related issue: [https://github.com/seladb/PcapPlusPlus/issues/2022](https://github.com/seladb/PcapPlusPlus/issues/2022)

Note: I’m not yet fully familiar with PcapPlusPlus’ internal structure, so if there are better places, names, or patterns for this logic, I’m happy to adjust the PR based on your feedback

PR #2163, opened 2026-06-21
Use heuristic SIP parser by default.

Dimi1010:
The PR deduplicates the `parseSipLayer` logic by removing the overload that takes a port src / dst pair.

The existing two parser overloads perform effectively identical, with the exception that the port based overload also requires a specific port combination. Port numbers by themselves should generally only be considered hints for the next layer type and not a parsing error by themselves. As such, they shouldn't be part of downstream parsers that validate the data contents. Instead they should only be used during parser dispatch to select the most likely parser candidate.

### DIFF
--- a/Packet++/header/SipLayer.h
+++ b/Packet++/header/SipLayer.h
@@ -115,10 +115,10 @@ namespace pcpp
 		}
 
 		/// A static factory method that attempts to create a SIP layer from existing packet raw data
-		/// 
+		///
 		/// The method uses heuristics to determine whether the data contains a SIP request or response message and
 		/// creates the relevant layer type. If no message type can be identified, the method returns nullptr.
-		/// 
+		///
 		/// @param[in] data A pointer to the raw data
 		/// @param[in] dataLen Size of the data in bytes
 		/// @param[in] prevLayer A pointer to the previous layer
--- a/Packet++/header/SipLayer.h
+++ b/Packet++/header/SipLayer.h
@@ -115,18 +115,6 @@ namespace pcpp
 		}
 
 		/// A static factory method that attempts to create a SIP layer from existing packet raw data
-		/// The method first checks whether the source or destination port matches the SIP protocol.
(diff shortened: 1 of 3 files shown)


seladb:
@Dimi1010 the port-based parser does less work compared to the heuristic parser: it checks the port first - if it seems like a request it only tries to parse the method and then determines it's a SIP request. If it seems like a response it only tries to parse the version and then determines it's a SIP response.

The heuristic parser parses the first 3 chars, but then if it seems like a request or a response it parses the entire first line which is more than only checking the method / version.

We had a long discussion in https://github.com/seladb/PcapPlusPlus/pull/2024 and decided to build these 2 parsers because in most cases SIP is identified by its port and we'd like to parse it faster in this case.

Dimi1010:
@seladb I see. On closer look, I think version that parses the first line has a higher ceiling of optimization, tho.

- The port is being checked twice in the port version. Once in `UdpLayer` / `TcpLayer` and once in `parseSipLayer`.
- The first line is still parsed in its entirety even on the port based parser, it just happens in the SIP layer constructors.
Since the heuristic parser already parses the entire first line, there is the option to directly forward it to the actual layer, instead of requiring a reparse. This should be less work overall instead of having to do the parsing twice (even partially).

seladb:
@Dimi1010 you're right that the first line is parse in the SIP layer constructors, but now with the 2 parsers the first line is actually parsed twice (one in the parser and then again in the constructor). I think that a better optimization could be lazy parsing of the first line, but I don't think it's going to be easy

Dimi1010:
@seladb yeah, converted this to a draft for now because of that. I will try to implement the the reuse of the first line parse and then revisit. A lazy line parse could be better, but if the line is already needed to validate the parser, it might be of limited benefit.

seladb:
@Dimi1010 in the common case where the port is known, it might be better to lazy parse the first line. In the heuristic parser I agree it doesn't add anything because the first line parsing is needed to identify the layer

PR #2167, opened 2026-06-24
Simplify layer parsing loop in `Packet`

Dimi1010:
PR replaces a `while`-loop with a `for`-loop in `shortenLayer` to mirror similar loops in `extendLayer`.


seladb (approved):
@Dimi1010 I'd make the PR title a bit more informative, something like: "Simplify layer parsing loop in `Packet`"

PR #2178, opened 2026-07-01
Reuse a scratch buffer in light_write_packet to avoid per-packet allocation

MerinoSheep:
## Summary
- Reuse a scratch buffer kept on the `light_pcapng_t` handle when building enhanced
  packet blocks in `light_write_packet`, instead of `calloc`/`free`-ing one for every
  written packet. The buffer grows on demand and is released in `light_pcapng_close`.
- Only the `PADD32` padding tail is zeroed on reuse, since the fixed EPB header fields
  and the packet payload are fully overwritten on each write.
- Add `TestPcapNgFileWriteScratchReuse`, which writes packets of varying sizes (mixing
  buffer-grow and buffer-reuse transitions), then reads them back and verifies data and
  per-packet comments round-trip exactly.

## Performance
Measured writing 2,000,000 packets to a tmpfs file, 5 runs each, comparing this branch
against its base with only the patched file reverted:

| Packet size | Before (median) | After (median) | Δ |
|-------------|-----------------|-----------------|------|
| 64 B    | 4,584,602 pkts/s | 4,990,264 pkts/s | +8.8%  |
| 128 B   | 3,142,929 pkts/s | 3,449,104 pkts/s | +9.7%  |
| 256 B   | 2,236,526 pkts/s | 2,489,930 pkts/s | +11.3% |
| 512 B   | 1,489,529 pkts/s | 1,519,552 pkts/s | +2.0%  |
| 1500 B  |   439,037 pkts/s |   453,986 pkts/s | +3.4%  |
| 9000 B  |   128,028 pkts/s |   127,790 pkts/s | ~0%    |

Gains are largest for small-to-medium packets, where the per-packet allocation
overhead removed by this change is a larger fraction of total write cost.

## Test plan
- [x] `Bin/Pcap++Test -n` — all pcapng tests pass, including the new one
- [x] MemPlumber leak checks pass (run by default per test)
- [x] New test covers small/large buffer grow and reuse transitions, with and without
      per-packet comments
- [x] Full project (library, examples, tests) builds cleanly against latest `dev`
- [x] `clang-format` and `cppcheck` checked manually against the pre-commit hook
      invocations (no diffs / no warnings)

Made with [Cursor](https://cursor.com)


MerinoSheep:
hmmm, let me fix tests failures

Dimi1010 (approved):
LGTM.

There might be additional gains to be had by fine-tuning buffer growth strategy, but it is not critical to this PR. Might not even be worth it, because it would eventually hit the largest encountered packet and stay there.

Tests/Pcap++Test/Tests/FileTests.cpp
```diff
@@ -1598,6 +1601,89 @@ PTF_TEST_CASE(TestPcapNgFileReadWrite)
 
 }  // TestPcapNgFileReadWrite
 
+PTF_TEST_CASE(TestPcapNgFileWriteScratchReuse)
```
seladb: I'm not sure what this test adds on top of the existing tests to read and write pcapng files. Other tests also have variable size packets 🤔 

Is there anything specific in this test that is not covered by the other tests?

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng_ext.c
```diff
@@ -38,6 +38,13 @@ struct _light_pcapng_t
 	light_pcapng pcapng;
 	light_pcapng_file_info *file_info;
 	light_file file;
+	// PCPP patch
+	// A calloc/free pair per written packet showed up as measurable overhead on write-heavy
+	// workloads, so the buffer lives on the handle and is reused across light_write_packet
+	// calls instead.
+	uint8_t *write_block_scratch;
```
seladb, line 45: There is slight degradation in thread safety with this approach. Before `epb_memory` was allocated on the fly and could be thread safe, and now it's not. However I don't think it's a concern because writing packets with LightPcapNg isn't thread safe anyway so this change doesn't break it

Tests/Pcap++Test/Tests/FileTests.cpp
MerinoSheep: This should hopefully catch any issues with padding due to buffer reuse

seladb: @MerinoSheep can you explain how the test catches this issue? In other words - if we run the same test without the fix, will it fail?

MerinoSheep: No I believe it won't fail without the fix. I was just intending to prevent future work and more concretely validating this with any corruption or issues that would come up with changing to buffer approach. I can remove if you feel so.

seladb: We already have tests that write packets of various sizes to a pcapng file so I don't see a point in adding another one, I think we can remove this test

MerinoSheep: ok sounds good

seladb (approved):

seladb:
@MerinoSheep thank you for working on this fix 🙏

PR #2181, opened 2026-07-10
Fix out-of-bounds read in GtpV2Layer with truncated TEID-present header

### LINKED ISSUE #2171: PcapPlusPlus GTPv2 short TEID-present header sequence over-read
## Summary
PcapPlusPlus accepts a truncated GTPv2 packet whose TEID-present bit requires more fixed-header bytes than the supplied buffer contains. At commit `af83983b079d8225d5cf3d2b4ed3d3a40e194557`, `GtpV2Layer::isDataValid` accepts the 8-byte input, and later `GtpV2Layer::getSequenceNumber` reads four bytes past that attacker-controlled buffer. It is a valid OOB, but the security threat did not seem critical, which is why we are reporting via github issues following the instruction in Security.md.

## Affected
- Project: PcapPlusPlus
- Repo: https://github.com/seladb/PcapPlusPlus
- Pinned ref: af83983b079d8225d5cf3d2b4ed3d3a40e194557
- Severity: CVSS 3.1 5.3/10 — `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:L`

## Root cause
The raw-data constructor for `GtpV2Layer` stores the caller-supplied buffer and length through the base `Layer` constructor without expanding or revalidating the backing storage (`Packet++/header/GtpLayer.h:979`, `Packet++/header/Layer.h:228`). The GTPv2 validator only requires `sizeof(gtpv2_basic_header) + sizeof(uint32_t)` bytes and version 2 before returning true (`Packet++/src/GtpLayer.cpp:1020`, `Packet++/src/GtpLayer.cpp:1034`). That minimum is enough for the basic header plus a TEID, but not for the sequence/spare word that is also part of the TEID-present header form. Later, `getSequenceNumber` starts at `m_Data + sizeof(gtpv2_basic_header)`, advances another four bytes when `teidPresent` is set, and dereferences a 32-bit word at that position without checking `m_DataLen` (`Packet++/src/GtpLayer.cpp:1111`, `Packet++/src/GtpLayer.cpp:1117`). The result is the same partial-frame validation pattern as CVE-2025-7485: protocol bytes are accepted before the implementation proves the fixed fields needed by a later parser/accessor are present.

## Reproduction

<details>
<summary>PoC files</summary>

poc/inputs/gtpv2-short-teid.hex
```
48200008deadbeef
```

poc/gtpv2_probe.cpp
```cpp
#include <cctype>
#include <cstdint>
#include <cstdio>
#include <cstdlib>
#include <cstring>
#include <fstream>
#include <string>
#include <vector>

#include "GtpLayer.h"

static uint8_t hexByte(char hi, char lo) {
  auto nibble = [](char c) -> uint8_t {
    if (c >= '0' && c <= '9') return static_cast<uint8_t>(c - '0');
    if (c >= 'a' && c <= 'f') return static_cast<uint8_t>(c - 'a' + 10);
    if (c >= 'A' && c <= 'F') return static_cast<uint8_t>(c - 'A' + 10);
    std::fprintf(stderr, "invalid hex\n");
    std::exit(3);
  };
  return static_cast<uint8_t>((nibble(hi) << 4) | nibble(lo));
}

static std::vector<uint8_t> readHex(const char* path) {
  std::ifstream in(path);
  std::string text((std::istreambuf_iterator<char>(in)), std::istreambuf_iterator<char>());
  std::string hex;
  for (char c : text) {
    if (!std::isspace(static_cast<unsigned char>(c))) hex.push_back(c);
  }
  if (hex.size() % 2 != 0) {
    std::fprintf(stderr, "odd hex length\n");
    std::exit(4);
  }
  std::vector<uint8_t> bytes;
  for (size_t i = 0; i < hex.size(); i += 2) bytes.push_back(hexByte(hex[i], hex[i + 1]));
  return bytes;
}

int main(int argc, char** argv) {
  if (argc != 2) return 2;
  std::vector<uint8_t> fixture = readHex(argv[1]);
  uint8_t* data = new uint8_t[fixture.size()];
  std::memcpy(data, fixture.data(), fixture.size());

  if (!pcpp::GtpV2Layer::isDataValid(data, fixture.size())) {
    std::fprintf(stderr, "GTPv2 input rejected before accessor path\n");
    delete[] data;
    return 5;
  }

  pcpp::GtpV2Layer layer(data, fixture.size(), nullptr, nullptr);
  auto teid = layer.getTeid();
  std::printf("teid_present=%d teid=%u\n", teid.first ? 1 : 0, teid.second);
  std::printf("sequence=%u\n", layer.getSequenceNumber());
  return 0;
}

```
run.sh
```sh
set -euo pipefail

script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
source_dir="$script_dir/source"

UPSTREAM_REPO_URL="https://github.com/seladb/PcapPlusPlus"
UPSTREAM_PINNED_SHA="af83983b079d8225d5cf3d2b4ed3d3a40e194557"

# Run a function silently, dumping its log to stderr only on failure.
# Keeps the maintainer-facing terminal clean: success → no noise;
# failure → enough context to know what broke and where.
_silent() {
  local label="$1"; shift
  local log="$script_dir/.${label}.log"
  if ! "$@" > "$log" 2>&1; then
    echo "[run.sh] $label failed; log follows:" >&2
    cat "$log" >&2
    return 1
  fi
}

setup() {
  if [[ -d "$source_dir/.git" ]] \
     && [[ "$(git -C "$source_dir" rev-parse HEAD 2>/dev/null)" == "$UPSTREAM_PINNED_SHA" ]]; then
    echo "[run.sh] reusing existing $source_dir at $UPSTREAM_PINNED_SHA"
    return 0
  fi
  _setup_inner() {
    rm -rf "$source_dir"
    mkdir -p "$source_dir"
    cd "$source_dir"
    git init -q
    git remote add origin "$UPSTREAM_REPO_URL"
    if git fetch --depth 1 origin "$UPSTREAM_PINNED_SHA" -q 2>/dev/null; then
      git checkout -q FETCH_HEAD
    else
      cd "$script_dir"
      rm -rf "$source_dir"
      git clone -q "$UPSTREAM_REPO_URL" "$source_dir"
      git -C "$source_dir" checkout -q "$UPSTREAM_PINNED_SHA"
    fi
    local head
    head="$(git -C "$source_dir" rev-parse HEAD)"
    if [[ "$head" != "$UPSTREAM_PINNED_SHA" ]]; then
      echo "setup: HEAD=$head but expected $UPSTREAM_PINNED_SHA" >&2
      exit 1
    fi
  }
  _silent setup _setup_inner
  # One-line confirmation reaches the terminal — mirrors the
  # 'Pinned ref:' line in the report README so the maintainer can see
  # exactly which commit they're exercising before the proof emits.
  echo "[run.sh] reproducing against $UPSTREAM_REPO_URL @ $UPSTREAM_PINNED_SHA"
}

build() {
  _silent build env source_dir="$source_dir" script_dir="$script_dir" bash -c '
    set -euo pipefail
    build_dir="$script_dir/build/pcpp-build"
    bin="$script_dir/gtpv2_probe"
    mkdir -p "$script_dir/build/fakebin" "$build_dir"
    printf "#!/usr/bin/env sh\nexit 1\n" > "$script_dir/build/fakebin/git"
    chmod +x "$script_dir/build/fakebin/git"

    PATH="$script_dir/build/fakebin:$PATH" cmake -S "$source_dir" -B "$build_dir" \
      -DPCAPPP_BUILD_PCAPPP=OFF \
      -DPCAPPP_BUILD_TESTS=OFF \
      -DPCAPPP_BUILD_EXAMPLES=OFF \
      -DPCAPPP_USE_SANITIZER=AddressSanitizer \
      -DCMAKE_BUILD_TYPE=Debug
    cmake --build "$build_dir" --target Packet++ -j2
    c++ -std=c++17 -fsanitize=address -g -O1 \
      -I"$source_dir/Common++/header" \
      -I"$source_dir/Packet++/header" \
      -I"$source_dir/3rdParty/EndianPortable/include" \
      "$script_dir/gtpv2_probe.cpp" \
      "$build_dir/Packet++/libPacket++.a" \
      "$build_dir/Common++/libCommon++.a" \
      -o "$bin"
  '
}

trigger() {
  local log="$script_dir/.run.log"
  set +e
  "$script_dir/gtpv2_probe" "$script_dir/inputs/gtpv2-short-teid.hex" > "$log" 2>&1
  local status=$?
  set -e
  if [[ "$status" -eq 0 ]]; then
    echo "trigger: harness exited cleanly; expected AddressSanitizer abort" >&2
    cat "$log" >&2
    return 1
  fi
  grep -q "SUMMARY: AddressSanitizer: heap-buffer-overflow .*GtpLayer.cpp:1117 in pcpp::GtpV2Layer::getSequenceNumber" "$log" \
    || { echo "trigger: no matching AddressSanitizer proof in run output" >&2; cat "$log" >&2; return 1; }
  echo "AddressSanitizer: heap-buffer-overflow in GtpV2Layer::getSequenceNumber"
}

cmd="${1:-all}"
case "$cmd" in
  setup)         setup ;;
  build)         setup; build ;;
  trigger|all|"") setup; build; trigger ;;
  *)
    echo "usage: $0 {setup|build|trigger|all}" >&2
    exit 2
    ;;
esac

```

</details>

Run:
```bash
bash ./poc/run.sh
```
Expected:
```expected-output
AddressSanitizer: heap-buffer-overflow in GtpV2Layer::getSequenceNumber
```

This line is emitted only after the harness verifies that AddressSanitizer reported a heap-buffer-overflow at `Packet++/src/GtpLayer.cpp:1117` in `pcpp::GtpV2Layer::getSequenceNumber() const`. A non-zero exit without that verified ASan frame is treated by the script as setup or build failure, not as a successful reproduction.

## Impact
The attacker model is a remote unauthenticated GTPv2 peer that can send one crafted packet to software using PcapPlusPlus to decode untrusted GTPv2 traffic and then read the sequence number from accepted layers. The trigger satisfies the current non-null, 8-byte minimum and version-2 checks, sets the TEID-present bit, and follows the public raw-buffer constructor plus `getSequenceNumber` accessor path; no guard checks for the additional sequence/spare bytes required after the TEID, and the message length field is not used to prove the backing buffer is long enough. The demonstrated effect is a 4-byte out-of-bounds read immediately past an 8-byte heap allocation. 

## Suggested fix
```001-fix.diff
diff --git a/Packet++/src/GtpLayer.cpp b/Packet++/src/GtpLayer.cpp
--- a/Packet++/src/GtpLayer.cpp
+++ b/Packet++/src/GtpLayer.cpp
@@ -1020,16 +1020,21 @@ namespace pcpp
 	bool GtpV2Layer::isDataValid(const uint8_t* data, size_t dataSize)
 	{
 		if (!data || dataSize < sizeof(gtpv2_basic_header) + sizeof(uint32_t))
 		{
 			return false;
 		}
 
 		auto* header = reinterpret_cast<const gtpv2_basic_header*>(data);
 
 		if (header->version != 2)
 		{
 			return false;
 		}
 
+		if (header->teidPresent && dataSize < sizeof(gtpv2_basic_header) + 2 * sizeof(uint32_t))
+		{
+			return false;
+		}
+
 		return true;
 	}
```

*Reported by Team Atlanta.*

v25.05

_No response_

Linux

_No response_

### Compiler version

GCC 13.3.0 

_No response_

vsaraikin:
Fixes #2171.

`GtpV2Layer::isDataValid` only requires `sizeof(gtpv2_basic_header) + sizeof(uint32_t)` bytes (8 bytes). That is enough for the basic header plus the sequence-number word of a **TEID-absent** header, but when the TEID-present flag is set the fixed header additionally carries the 4-byte TEID *before* the sequence word:

- TEID absent: `[basic header (4)][seq word (4)]` → 8 bytes
- TEID present: `[basic header (4)][TEID (4)][seq word (4)]` → 12 bytes

So a TEID-present packet of only 8 bytes passes validation, and `getSequenceNumber()` (which advances an extra 4 bytes when `teidPresent` is set) then reads a 32-bit word past the buffer — the OOB read reported in the issue (PoC `48200008deadbeef`, version 2 + TEID-present, 8 bytes).

The fix makes `isDataValid` require the extra 4 bytes when the TEID-present flag is set, so a truncated TEID-present header is rejected before any accessor can read past the buffer. The flag lives in the basic header, which is already covered by the existing minimum-length check, so reading it here is safe.

## Testing

- Added a regression case to `GtpV2LayerParsingTest`: the 8-byte TEID-present PoC is now rejected by `isDataValid`, while the same header with the sequence word present (12 bytes) is accepted.
- Verified under AddressSanitizer: the new GtpV2 tests pass clean with the fix; without the fix `isDataValid` accepts the truncated packet (the assertion fails, and the underlying `getSequenceNumber` read is OOB).
- Full `Packet++Test` suite passes (258/258).


Dimi1010 (approved):
LGTM

Dimi1010:
Closing and reopen to run CI after base branch change.

seladb (approved):

PR #2182, opened 2026-07-10
Bound attacker-controlled block lengths in LightPcapNg (OOM/OOB in EPB parsing)

### LINKED ISSUE #2180: OOM in LightPcapNg EPB parsing due to unbounded `captured_packet_length`
## Summary

The vendored LightPcapNg pcapng parser can allocate an attacker-controlled amount of memory while parsing an Enhanced Packet Block (EPB).

In `parse_by_block_type()`, the `LIGHT_ENHANCED_PACKET_BLOCK` parser reads `captured_packet_length` from the block body, pads it with `PADD32()`, and uses the padded value directly in `calloc()`. This happens before the value is validated against `block_total_length` or the remaining bytes in the current block.

As a result, a malformed pcapng block can keep `block_total_length` small while setting `captured_packet_length` to a very large value, causing a multi-gigabyte allocation attempt.

The same length-trust pattern appears in `LIGHT_CUSTOM_DATA_BLOCK`, where the `len` field is also read from the file and used to size an allocation without first checking that the data fits inside the current block.

## Affected Code

```c
// 3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
case LIGHT_ENHANCED_PACKET_BLOCK:
{
    struct _light_enhanced_packet_block *epb = NULL;
    uint32_t interface_id = *local_data++;
    uint32_t timestamp_high = *local_data++;
    uint32_t timestamp_low = *local_data++;
    uint32_t captured_packet_length = *local_data++;
    uint32_t original_packet_length = *local_data++;
    uint32_t actual_len = 0;

    PADD32(captured_packet_length, &actual_len);

    epb = calloc(1, sizeof(struct _light_enhanced_packet_block) + actual_len);
    memcpy(epb->packet_data, local_data, captured_packet_length);
    local_data += actual_len / sizeof(uint32_t);
    // ...
}
```

Problems in this path:

- `captured_packet_length` is file-controlled input.
- `PADD32(captured_packet_length, &actual_len)` preserves large values as large padded allocation sizes.
- `calloc()` happens before checking whether `captured_packet_length` bytes actually exist in the EPB body.
- `memcpy()` then trusts the same unvalidated length.
- `local_data` is advanced by the padded length, which can move it past the current block.

## Root Cause

`block_total_length` describes the total size of a pcapng block:

```text
Block Type (4)
Block Total Length (4)
Block Body (variable)
Block Total Length (4)
```

For an EPB, the fixed body fields before packet data are:

```text
interface_id              4 bytes
timestamp_high            4 bytes
timestamp_low             4 bytes
captured_packet_length    4 bytes
original_packet_length    4 bytes
```

Therefore, EPB packet data must fit within:

```text
block_total_length - 12 - 20
```

where `12` is the pcapng block overhead (`block_type`, first `block_total_length`, trailing `block_total_length`) and `20` is the EPB fixed body size.

The parser currently dispatches to `parse_by_block_type()` after reading the block type and block length. Inside the EPB case, it trusts the inner `captured_packet_length` field independently of the enclosing `block_total_length`.

## Triggering Input Shape

A malformed pcapng file can trigger the issue with an EPB shaped like this:

```text
block_type              = 0x00000006  // LIGHT_ENHANCED_PACKET_BLOCK
block_total_length      = small valid block size, for example 92

interface_id            = any value
timestamp_high          = any value
timestamp_low           = any value
captured_packet_length  = very large value, for example 0xf3000000
original_packet_length  = any value

packet_data             = much shorter than captured_packet_length
options/trailing data   = enough bytes only to satisfy block_total_length
trailing block length   = same value as block_total_length
```

In this shape, `block_total_length` says the whole EPB is small, but `captured_packet_length` says the packet data is several gigabytes. The parser uses `captured_packet_length` for allocation before checking that the packet data fits inside the block.

This can lead to an allocation like:

```text
calloc(1, sizeof(struct _light_enhanced_packet_block) + 0xf3000000)
```

Under libFuzzer/ASan this is reported as an out-of-memory abort, for example:

```text
ERROR: libFuzzer: out-of-memory (malloc(4076863312))
```

## ASan / libFuzzer Report

```text
==7989== ERROR: libFuzzer: out-of-memory (malloc(4076863312))
    #0  __sanitizer_print_stack_trace
    #1  fuzzer::Fuzzer::HandleMalloc(unsigned long)
    #2  fuzzer::MallocHook(void const volatile*, unsigned long)
    ...
    #8  parse_by_block_type  light_pcapng.c:165
    #9  light_read_record    light_pcapng.c:377
    #10 light_get_next_packet light_pcapng_ext.c:384
    #11 PcapNgFileReaderDevice::getNextPacketInternal  PcapFileDevice.cpp:894
    #12 LLVMFuzzerTestOneInput  FuzzTarget.cpp:71
```

## Also Affected

`LIGHT_CUSTOM_DATA_BLOCK` has the same shape:

```c
case LIGHT_CUSTOM_DATA_BLOCK:
{
    uint32_t len = *local_data++;
    uint32_t actual_len = 0;

    PADD32(len, &actual_len);
    cnb = calloc(1, sizeof(struct _light_custom_nonstandard_block) + actual_len);
    memcpy(cnb->packet_data, local_data, len);
    // ...
}
```

Here, `len` is also a file-controlled value used for allocation, copy length, and pointer advancement before it is checked against the available bytes in the current block.

`LIGHT_SIMPLE_PACKET_BLOCK` is less directly affected by this specific issue because its allocation size is derived from `block_total_length`, not from an independent packet-data length field.

## Impact

Any PcapPlusPlus code path that reads pcapng files through `PcapNgFileReaderDevice` can reach this vendored LightPcapNg parser. A malformed pcapng input can therefore cause excessive memory allocation and process termination instead of being rejected as malformed.

Other (specify in the next question)

dev branch

Linux

_No response_

### Compiler version

clang version 22.1.8 (Fedora 22.1.8-1.fc44)

_No response_

vsaraikin:
Fixes #2180.

The vendored LightPcapNg parser trusts several length fields read straight from the input, which lets a malformed pcapng drive an attacker-controlled `calloc()`/`memcpy()`:

- In `parse_by_block_type()`, the `LIGHT_ENHANCED_PACKET_BLOCK` handler reads `captured_packet_length` from the block body, pads it, and uses it directly to size a `calloc()` and then `memcpy()` — before validating it against the block. A block can keep `block_total_length` small while setting `captured_packet_length` huge, causing a multi-gigabyte allocation and an out-of-bounds read past the input buffer. `LIGHT_CUSTOM_DATA_BLOCK` has the same pattern with its `len` field.
- The only length sanity checks in `__parse_mem_copy()` use the `DCHECK_*` macros, which are **no-ops in release builds** (`light_stop` expands to `(void)0`), so `block_total_length` itself is never bounded against the remaining buffer.

The fix adds real (release-active) bounds:

1. In `__parse_mem_copy()`, reject a block whose `block_total_length` is smaller than the mandatory framing or larger than the bytes left in the buffer, before dispatching to the per-block parser.
2. In the EPB and custom-block handlers, clamp `captured_packet_length` / `len` to the space the block can actually hold (mirroring how `LIGHT_SIMPLE_PACKET_BLOCK` already derives its length from `block_total_length`), so a bogus length can't over-size the `calloc()`/`memcpy()`.

## Testing (AddressSanitizer)

Verified against the parser (`light_read_from_memory`) directly, since the DCHECK-based checks are compiled out in release:

- **Malformed EPB** (`block_total_length = 32`, `captured_packet_length = 0x10000`): before the fix ASan reports `stack-buffer-overflow READ of size 65536` in `parse_by_block_type` (`light_pcapng.c:172`, the `memcpy`); after the fix it is bounded and clean.
- **Valid EPB** (`captured_packet_length = 8`, correctly framed): parses unchanged — `captured_packet_length` is preserved (not clamped) and the packet bytes are copied correctly, so well-formed captures are unaffected.

The bug originates from OSS-Fuzz (`FuzzTargetNg`); the project's fuzzers exercise this path going forward. Happy to also add a regression sample to `Tests/Fuzzers/RegressionTests` if you'd like one in the corpus.

## Note

`3rdParty/LightPcapNg` is vendored, so the change follows the surrounding vendored style and is marked with `// PCPP Patch (GH #2180)` comments consistent with the existing patches in this file.


seladb:
@vsaraikin this PR is marked as "draft". Are you still working on it?

vsaraikin:
> @vsaraikin this PR is marked as "draft". Are you still working on it?

Yes, it is done. Marking it ready for review now

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
```diff
@@ -276,6 +302,16 @@ static size_t __parse_mem_copy(struct _light_pcapng **iter, const uint32_t *memo
...
 
+      // PCPP Patch (GH #2180): block_total_length is read straight from the input and is
+      // attacker-controlled. The DCHECK_* macros above are no-ops in release builds, so
+      // reject here any block that is smaller than the mandatory framing or claims more
+      // bytes than are left in the buffer, otherwise the per-block parsers below would
+      // trust bogus lengths and over-allocate / read out of bounds.
+      if (current->block_total_length < 12 || current->block_total_length > remaining)
+      {
+         break;
+      }
```
seladb: This check is done after `current` was already `calloc`'ed and linked into the list. We should probably allocate and link it after the length check

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
seladb, line 1: > The bug originates from OSS-Fuzz (FuzzTargetNg); the project's fuzzers exercise this path going forward. Happy to also add a regression sample to Tests/Fuzzers/RegressionTests if you'd like one in the corpus.

I think we can add it to `Tests/Fuzzers/RegressionTests` so this fix is covered

3rdParty/LightPcapNg/LightPcapNg/src/light_pcapng.c
vsaraikin: Done. I read `block_type` and `block_total_length` into locals and check them
before allocating and linking, so a rejected block no longer leaves a
half-initialized entry in the list.

I left the allocation block itself as it was to keep the diff on the vendored
file small

seladb (approved):

seladb:
Thank you @vsaraikin for working on this fix 🙏

PR #2183, opened 2026-07-12
Fix packaging issues

seladb:
- Replace Ubuntu 20.04 with Ubuntu 26.04
- Change Visual Studio to `Visual Studio 18 2026` like we have in the build workflow
- Fix FreeBSD and macOS builds


tigercosmos (approved):
LGTM

PR #2184, opened 2026-07-12
Add macOS 26 to CI

seladb:
Remove macOS 14 and replace is with macOS 26

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -464,19 +464,14 @@ jobs:
     strategy:
       matrix:
         include:
-          - os-version: macos-14
-            host-arch: x86_64
-            arch: x86_64
-            config-zstd: OFF
-          # TODO: see https://github.com/seladb/PcapPlusPlus/issues/1989
-          # - os-version: macos-14
-          #   host-arch: x86_64
-          #   arch: x86_64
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -464,19 +464,14 @@ jobs:
...
-          # - os-version: macos-14
-          #   host-arch: x86_64
-          #   arch: x86_64
-          #   config-zstd: ON
-          - os-version: macos-14
+          - os-version: macos-26
             host-arch: arm64
             arch: arm64
             config-zstd: ON
+          - os-version: macos-26-intel
```
tigercosmos, line 471: Maybe we can drop Intel now.

seladb: I actually have an old Intel macOS laptop with macOS 26 🙂

tigercosmos: I see. I guess we can still support it for a while. However, Apple will not support Intel anyway from macOS 27.

tigercosmos (approved):

.github/workflows/build_and_test.yml
seladb: Yes, macOS is the last version supported on Intel devices 😕

PR #2187, opened 2026-07-14
Improve encapsulation of `TextBasedProtocol` fields.

Dimi1010:
The PR changes the visibility of the fields of `TextBasedProtocol` from `protected` to `private` and establishes methods to allow checked mutation. Notably all offsets are range validated prior to change.


Packet++/src/TextBasedProtocol.cpp
```diff
@@ -360,17 +360,38 @@ namespace pcpp
...
 		}
 	}
 
+	void TextBasedProtocolMessage::validateFieldsOffset(int offset) const
+	{
+		if (offset < 0 || offset > m_DataLen)
+			throw std::invalid_argument(
+			    "Field offset must be greater or equal to zero and less than or equal to the data length");
+	}
+
```
Dimi1010, line 394: Maybe this should be `std::out_of_range`?

seladb (approved):

Issue #1853, opened 2025-06-30
Port the code to C++14


tigercosmos:
Let's leverage C++14 features:

- Generic lambdas
- Lambda capture by move (`[=, x = std::move(y)]`)
- Return type deduction for functions
- Relaxed `constexpr` functions (allowing loops, conditionals, multiple statements)
- Variable templates
- Binary literals (`0b...`)
- Digit separators (e.g., `1'000'000`)
- `[[deprecated]]` attribute
- Extended `auto` in lambda parameters
- `std::make_unique`
- `std::integer_sequence` and `std::index_sequence`
- `std::exchange`
- `std::quoted` (quoted I/O manipulators)
- `std::shared_timed_mutex`
- `std::chrono::duration` user-defined literals (e.g., `5s`, `10ms`)
- Additional type traits:
  - `std::is_final`
  - `std::is_null_pointer`
  - And others

Old PR for C++11: https://github.com/seladb/PcapPlusPlus/issues/977

adityaxa:
Hey, I'd like to help with the C++14 modernization effort. 

Looking at the codebase, I found a few places where we could use std::make_unique. For example in  [Logger.cpp](https://github.com/seladb/PcapPlusPlus/blob/0588d88eca769a02e660161f8965ef6152ada90e/Common%2B%2B/src/Logger.cpp) line 70:
```cpp
return std::unique_ptr<internal::LogContext>(new internal::LogContext(level, source));
```
A couple questions before I dive in:

Should I stick to one module per PR? Was thinking Common++ would be a good start

I can start small with just the straightforward replacements and submit a PR for review.

Dimi1010:
@adityaxa I think we already have an open PR for a lot of the make_unique cases #1854. You are welcome to open a PR with one of the other items if you wish. :)

@tigercosmos Can we mark the [[deprecated]] attribute usage as completed? `PCPP_DEPRECATED` macro already maps to that?

tigercosmos:
@Dimi1010  true, let me remove it.

shikharish:
Hi @tigercosmos , whats the status of this issue? I would like to contribute.

seladb:
> Hi @tigercosmos , whats the status of this issue? I would like to contribute.

@shikharish this is an evergreen issue 🙂 feel free to contribute!

KenziVisor:
[...]

PR #2193, opened 2026-07-25
Replace safe Packet++ option macros with constexpr constants

KenziVisor:
## Summary

- Replace safe internal Packet++ option macros with typed `constexpr` constants.
- Preserve the original values and behavior.
- Keep the changes limited to constants that can be converted without affecting conditional compilation or macro expansion.

Refs #1853

## Rationale

These macros represent fixed numeric values used by Packet++ option-processing code.

Replacing them with typed `constexpr` constants:

- provides explicit type information;
- avoids preprocessor text substitution;
- limits the constants to normal C++ scoping rules;
- makes their compile-time nature clear;
- preserves the existing runtime behavior.

For one-byte option values, `uint8_t` is used because the corresponding protocol field is eight bits wide. Byte order is not relevant for a single-byte value.

## Validation

- Built the `Packet++Test` target successfully.
- Ran the Packet++ test suite on this branch.
- Ran the same test suite on `master`.
- Confirmed that the same pre-existing tests fail on both branches.
- No additional test failures were introduced by these changes.

## Commit

`Replace safe Packet++ option macros with constexpr constants`

### DIFF
--- a/Packet++/src/CiscoHdlcLayer.cpp
+++ b/Packet++/src/CiscoHdlcLayer.cpp
@@ -10,8 +10,8 @@ namespace pcpp
 {
 
 // Protocol types for Cisco HDLC
-#define CISCO_HDLC_TYPE_IP 0x0800
-#define CISCO_HDLC_TYPE_IPV6 0x86DD
+constexpr uint16_t CiscoHdlcTypeIp = 0x0800;
+constexpr uint16_t CiscoHdlcTypeIpv6 = 0x86DD;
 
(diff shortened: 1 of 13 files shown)


Dimi1010:
Hello @KenziVisor. Thanks for the PR. Could you change the PR base to target the `dev` branch instead of `master`?
We usually merge PRs to dev first, so CI is configured to run on PRs for it there. You may have to close and reopen the PR after the base change, so the CI system detects the change, tho.

KenziVisor:
Hi @Dimi1010 , I have opened a new pull request to dev branch. Thank you :)

Issue #1717, opened 2025-02-14
DPDK uses more than 32 cores


gyl30:
### Question

I noticed that both CoreMask and SystemCores are only 32, can we support more?

gyl30:
> How many cores do you have? I think maybe we can make it to 64?

My current project aims to use 32 cores. However, since the MasterCore occupies one core, the current implementation can only use 30 cores, as the number of usable cores needs to be an even number. More generally, when using multiple network cards, each card or queue may require a dedicated core. Therefore, I believe we should increase the number to 128 or 256 to accommodate a wider range of needs. Perhaps we can gather more opinions on this.

Dimi1010:
> 128 or 256

Having the CoreMask be >64 cores would require a change of the CoreMask from a `std::uint32_t` to a dedicated struct as the bitmask would be split up between 2 64-bit integers. 

Also I am not sure if DPDK can be initialized 256 cores. The highest I saw was initializing to core index 128.

gyl30:
> > 128 or 256
> 
> Having the `CoreMask` be >64 cores would require a change of the `CoreMask` from a `std::uint32_t` to a dedicated struct as the bitmask would be split up between 2 64-bit integers. Perhaps we can add a `CoreMaskLong` if we don't want to break code depending on `CoreMask` being integral type?
> 
> Also I am not sure if DPDK can be initialized 256 cores. The highest I saw in documentation was initializing to core index 128.

As you said, if we need to change the CoreMask limit, we may need to design the code properly

SesomB:
DPDK sets its core limit using the `RTE_MAX_LCORE` define in `rte_config.h`, but you can adjust it to support 512 or even 1024 cores by changing the configuration and recompiling.
I suggest updating PcapPlusPlus to support an 'unlimited' coremask for more flexibility.

I can take this on, but I won't be able to start until a couple of weeks from now.

tigercosmos:
@SesomB  Yes, it will be great. Thanks.

seladb:
@SesomB are you still planning to work on this task?

PR #2197, opened 2026-07-29
Add extended core mask for up to 256 cores.

Dimi1010:
## Summary
Adds infrastructure for extended core mask that allows addressing up to 256 machine cores. 
Part of #1717 

## Proposal

Current server grade processors frequently exceed 32 cores. The current core mask implementation cannot address these cores as it is limited to 32 bits. The PR proposes adding a new mask object `LongCoreMask` that is based on `std::bitmask` instead of `uint32_t` that can accommodate up to 256 machine cores. The maximum of 256 was chosen as it is the highest available count that would not require changing `SystemCore::CoreId` data type from `uint8_t` to `uint16_t`. Further expansion can happen in a future PR.

## Changes
- Added `LongCoreMask` to support larger machine core count.
- The field `SystemCore::Mask` has been deprecated to disambiguate between the short and long core masks. The field would also be unsafe for system cores with IDs over 32, as they can't be represented. It has been replaced by `SystemCore::getShortCoreMask()` which dynamically computes the mask from the current `SystemCore::CoreId`.
- Added an overload to `createCoreVectorFromCoreMask` which uses return value instead of out parameter.
- Added validation for functions `createCoreMaskFromCoreVector` and `createCoreMaskFromCoreIds`. They will now throw `std::out_of_range` exception if they encounter `CoreId >= 32`.
- Fixed documentation comments. Comments without `@brief` did not have their documentation added as short documentation.

## Not planned in this PR
- Add support for `LongCoreMask` in `PfRingDevice`.
- Add support for `LongCoreMask` in `DpdkDevice`.


.github/workflows/build_and_test.yml
Dimi1010, line 1: Bumped cppcheck version to 2.21.0 due to false positive during the pre-commit checks:
https://github.com/seladb/PcapPlusPlus/actions/runs/30527830962/job/90822810175#step:6:90

Also, updated the GH repository URL, since it has been moved.

Common++/header/SystemUtils.h
```diff
@@ -6,13 +6,15 @@
...
+#include <bitset>
+#include <stdexcept>
 
 /// @file
 
 // @todo Change to constexpr when C++17 is minimum supported version
 enum : uint8_t
 {
-	MAX_NUM_OF_CORES = 32
+	MAX_NUM_OF_CORES = 32,
```
seladb, line 17: nit: is this extra comma needed? 🤔

Common++/header/SystemUtils.h
```diff
@@ -23,32 +25,96 @@ int gettimeofday(struct timeval* tp, struct timezone* tzp);
...
+
+	DISABLE_WARNING_PUSH
+	DISABLE_WARNING_DEPRECATED
 
 	/// @struct SystemCore
-	/// Represents data of 1 CPU core. Current implementation supports up to 32 cores
+	/// @brief Represents data of 1 CPU core. Current implementation supports up to 256 cores.
+	///
+	/// Cores above 31 do not support the Mask attribute and require usage of LongCoreMask instead of CoreMask.
 	struct SystemCore
```
seladb: Why do we need to change `SystemCore` at all? Maybe we should just mark the entire struct as deprecated and ask users to use `LongCoreMask`

Common++/header/SystemUtils.h
```diff
@@ -119,33 +185,196 @@ namespace pcpp
...
 	};
 
-	using CoreMask = uint32_t;
+	/// @brief An extended core mask which can support up to 256 cores.
+	///
+	/// Intended to be used in cases where the system has more than 32 cores, as CoreMask is limited to 32 bits.
+	struct LongCoreMask
+	{
+		/// @brief The maximum number of cores supported by LongCoreMask.
+		static const size_t MaxCoreCount = 256;
```
seladb, line 194: Maybe it should be a `constexpr`?

Common++/header/SystemUtils.h
seladb, line 1: I'm not sure if we need all of these code changes. I'd make a simpler change of changing `SystemCore.Mask` to be a `std::bitset<MaxCoreCount>`. It's a breaking change but quite small because most users don't use it. We probably don't need `LongCoreMask` - it's a lot of code that just manages the bitset but it's not used much. We already have what we need in the different methods in `SystemUtils.`.

Instead I'd change the implementation of these methods to support more than 32 cores, then we can update `DpdkDevice` to support more than 32 cores

Dimi1010: My general idea is that `SystemCore.Mask` should be removed. It bloats the `SystemCore` class by 400% (5 bytes id + mask vs 1 byte id only), when it can be generated on the fly from the coreId. This problem gets worse as you expand the core mask, as every system core instance now needs to hold its ID + MaxCore bits of Mask. 256 cores bitmask would expand the class from 1 byte to 33 bytes, for every instance. The current implementation completely negates the point of keeping a sparse list of cores.

Additionally, the `SystemCores` implementation gets unwieldy as cores are scaled up. Do we plan to have 256 instances of CoreN? Personally, I would prefer not to... both because it would get tedious, and it would inflate the static memory footprint.

The `LongCoreMask` is to first separate the "core mask" concept from the "system core" identifier, and second to provide a typed object with a mask-like interface that can interact with itself and `SystemCore`. Yes, it currently uses a bitset internally, it might not use one in the future. Tbh, I was debating if the Mask field on the `LongCoreMask` should be public. The important thing is that the external API will (hopefully) keep.

The change would also allow the new core implementation to be an additive change for the device classes instead of a breaking change, which IMO would be a positive, even if its not used a lot.

In addition, I think the devices would need some refactoring for optimizations for larger core support. PfRing and Dpdk have just an `array<Config, MAX_NUM_CORES>` which I would like to optimize when I get the time.

PR #2170, opened 2026-06-25
OSS-Fuzz: Add new fuzzer targets X509 processing

arthurscchan:
This PR adds a new OSS-Fuzz fuzzer targeting X509 processing, together with the fixes of the CMakeList.txt and the OSS-Fuzz build script to accommodate the new fuzzer and the initial seed corpora preparation.

PR #2204, opened 2026-08-04
Fix size_t underflow when decoding malformed ASN.1 BitString records

arthurscchan:
In #2170, a new fuzzer is added and it discovers a new integer underflow problem. `Asn1BitStringRecord::decodeValue` computed `numBits` from the unused-bits octet without validating it, so a malformed BIT STRING underflowed `size_t` and made `BitSet` attempt a ~16 EiB allocation (reported as OOM by ASan and as an oversized allocation by MSan from the new fuzzer); a zero-length value also read the unused-bits octet one byte past the end of the buffer. This PR add validation for that the value must contain the initial octet, and the unused-bit count must be 0–7, std::invalid_argument will be thrown otherwise. The unit test are also fixed to include these cases.


seladb:
@arthurscchan the tests seem to fail, can you please check?

arthurscchan:
@seladb Sorry for that. It is the malformedData in the CryptoInvalidDataTest accidentally triggers the bug. It got undetected because originally the bug does not ends up in an assertion failure. Now I have fixed that and it should work. Please do restart the CI to check that and I am sorry for your time.

arthurscchan:
I am now checking on the failing of windivert build test.

Dimi1010:
> I am now checking on the failing of windivert build test.

@arthurscchan Windivert failure is likely due to environment. It fails CI sometimes. I started it again.
The changed code in this PR shouldn't be affecting windivert.

seladb (approved):

seladb:
Thank you @arthurscchan for the fix 🙏

PR #2207, opened 2026-08-05
Simplify RawSocketDevice `m_Socket` member variable and fix warnings.

Dimi1010:
The change removes the need for many of the casts inside `RawSocketDevice` by changing `m_Socket` to be property typed via opaque type. `SocketContainer` is moved under `RawSocketDevice` as a private struct, declared in the header and defined in the implementation file.

Additional notes:
- Added constructors to SocketContainer to fix uninitialized member variable warnings.

---
Cherrypicked from #2199

### DIFF
--- a/Pcap++/header/RawSocketDevice.h
+++ b/Pcap++/header/RawSocketDevice.h
@@ -2,6 +2,7 @@
 
 /// @file
 
+#include <memory>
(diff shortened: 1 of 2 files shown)


seladb (approved):
LGTM, thanks for the fix! 🙏

PR #2209, opened 2026-08-05
Add CI for FreeBSD 15.1 and drop 13.4

egecetin:
This PR drops FreeBSD 13.4 support and adds 15.1

(also PR should fix the issue in https://github.com/seladb/PcapPlusPlus/pull/2205)


seladb (approved):
LGTM

Issue #2206, opened 2026-08-05
HAS_PCAP_IMMEDIATE_MODE for TPACKET_V3


MerinoSheep:
### Question

I believe I discoverd a potential Linux performance regression linked to the code cleanup in #1772. Forcing `pcap_set_immediate_mode` causes libpcap to fall back from TPACKET_V3 to TPACKET_V2 in this part of the code
https://github.com/seladb/PcapPlusPlus/blob/80c73281201f6c7354e9605a35bb3831506f5151/Pcap%2B%2B/src/PcapLiveDevice.cpp#L679  
In my initial testing, disabling immediate mode significantly reduced CPU usage and increased throughput under load. I wanted to confirm what the intended behavior is currently. Thanks.

[Relevant libpcap code](https://github.com/the-tcpdump-group/libpcap/blob/f126e6a28b8741f871738dad24a76c9988c55dc0/pcap-linux.c#L2961)

Dimi1010:
I am unsure if we have a requirement that needs it forced on.
From what I can see in the libpcap source, it only uses tpacket_v2 if the switch is enabled.

I think the cleanest way to address this is to add a immediate mode switch added to `DeviceConfiguration`, so the user can choose.
I am unsure on what the default value to be. Possibly default it to `ON` to keep the current behavior with previous versions?

MerinoSheep:
Yeah that sounds good, I can make a PR for this then if needed.

Dimi1010:
Sure. Go for it.

PR #2210, opened 2026-08-05
Add ability to disable pcap immediate mode to allow TPACKET_V3 on Linux

MerinoSheep:
## Summary
- Addresses #2206.
- PcapPlusPlus always enables libpcap's immediate mode where supported, for minimal latency.
- On Linux this forces the less efficient TPACKET_V2 path instead of TPACKET_V3 (blocking, batched reads), hurting throughput/CPU at high packet rates. This became unconditional in #1772 (previously opt-in via `PCAPPP_ENABLE_PCAP_IMMEDIATE_MODE`, default `OFF`).
- Adds `DeviceConfiguration::disableImmediateMode` (default `false`, backward compatible) to opt out and let Linux use TPACKET_V3.

## Changes
- `PcapLiveDevice.h`: new `disableImmediateMode` field + constructor param (default `false`).
- `PcapLiveDevice.cpp`:
  - `doOpen()` skips `pcap_set_immediate_mode()` when set.
  - Falls back to a 100 ms timeout on Linux/Windows when disabled and no `packetBufferTimeoutMs` is given (their default `-1` is unpredictable without immediate mode). FreeBSD/macOS keep their existing `pcap_breakloop()` workaround timeouts untouched.
- `LiveDeviceTests.cpp`: `TestPcapLiveDeviceBlockingMode` now also runs against a `disableImmediateMode = true` config.


Pcap++/header/PcapLiveDevice.h
```diff
@@ -318,6 +318,15 @@ namespace pcpp
...
 			TimestampPrecision timestampPrecision;
 
+			/// Disable libpcap's "immediate mode", which PcapPlusPlus enables by default (where supported) for
+			/// minimal packet delivery latency. On Linux, immediate mode also forces a less efficient capture
+			/// path (TPACKET_V2 with non-blocking polling, instead of TPACKET_V3 with blocking, batched reads),
+			/// which can hurt throughput and CPU usage at high packet rates. Setting this to true trades some
+			/// added latency (up to packetBufferTimeoutMs, or 100 ms on Linux/Windows if not set) for those
+			/// throughput and CPU benefits.
+			/// Default value is false.
+			bool disableImmediateMode;
```
Dimi1010: IMO, positive flags are easier to reason about than negative flags since the reader does not have to negate the condition. `useImmediateMode` defaulted to `true` might be easier to reason about when looking through the docs.


Also, I am wondering, do you think it would be better to have the flag be an enum instead of a bool?

Something like:
```cpp
enum class BufferingMode
{
  Default, // let the device decide (current behavior)
  Immediate, // use immediate mode
  Buffered // use buffered mode
}
```

Default mode can also possibly have a graceful fallback to buffered, if immediate mode fails to enable for some reason.

MerinoSheep: yeah good idea, Ill change

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -677,10 +690,13 @@ namespace pcpp
...
-		if (ret != 0)
+		if (!useBufferedMode)
 		{
-			throw std::runtime_error("Cannot set immediate mode, error was: " + std::string(pcap.getLastError()));
+			ret = pcap_set_immediate_mode(pcap.get(), 1);
+			if (ret != 0)
+			{
+				throw std::runtime_error("Cannot set immediate mode, error was: " + std::string(pcap.getLastError()));
+			}
 		}
```
Dimi1010, line 700: If we are on "default" mode, we could possibly gracefully fallback to buffered mode if we fail to enable immediate, since the immediate mode was not explicitly requested by the user to be "on".

Dimi1010 (approved):
LGTM. Minor comment about possible graceful fallback to buffered mode on `BufferingMode::Default`, but it is good as it is currently too.

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -660,7 +664,16 @@ namespace pcpp
...
-		int timeout = (config.packetBufferTimeoutMs <= 0 ? LIBPCAP_OPEN_LIVE_TIMEOUT : config.packetBufferTimeoutMs);
+		bool useBufferedMode = config.bufferingMode == BufferingMode::Buffered;
+
+		int timeout = config.packetBufferTimeoutMs;
+		if (timeout <= 0)
+		{
+			// FreeBSD/macOS already have a positive LIBPCAP_OPEN_LIVE_TIMEOUT for the pcap_breakloop() workaround
+			bool useNonImmediateModeDefault = useBufferedMode && LIBPCAP_OPEN_LIVE_TIMEOUT <= 0;
+			// cppcheck-suppress knownConditionTrueFalse
+			timeout = useNonImmediateModeDefault ? NON_IMMEDIATE_MODE_DEFAULT_TIMEOUT : LIBPCAP_OPEN_LIVE_TIMEOUT;
```
seladb, line 675: I'm not sure I understand why the timeout we set depends on the buffering mode. Can we set a timeout when immediate mode is set? 🤔

PR #2217, opened 2026-08-09
Fix out-of-bounds read from a truncated IGMPv3 group record

vsaraikin:
A truncated IGMPv3 report reads off the end of the buffer. On `dev`, taking the multicast address of the record `getFirstGroupRecord()` hands back:

```
ERROR: AddressSanitizer: stack-buffer-overflow
READ of size 4
    #0 pcpp::igmpv3_group_record::getMulticastAddress() const  IgmpLayer.h:85
SUMMARY: stack-buffer-overflow IgmpLayer.h:85 in getMulticastAddress
```

Both accessors check the wrong thing. `getFirstGroupRecord()` only asks whether anything at all follows the report header:

```cpp
if (getHeaderLen() <= sizeof(igmpv3_report_header))
    return nullptr;
```

and `getNextGroupRecord()` only asks where the next record starts. One byte past the header is enough to get a pointer back. But `igmpv3_group_record` is 8 bytes before its source list, so `recordType`, `numOfSources` and `getMulticastAddress()` all read past the end. `isDataValid()` does not catch it either. It validates the 8 byte report header and nothing more.

Both now require the fixed part of the record to fit.

The reproducer is a 46 byte Ethernet + IPv4 packet whose report claims one group record and carries 4 bytes of it. With the fix `getFirstGroupRecord()` returns null.

Added to `Igmpv3ParsingTest`. Packet++Test is 259 passed, 0 failed. The igmp tests are clean under ASAN.

Same shape as #2203. The query side already does this: `IgmpV3QueryLayer::getSourceAddressAtIndex()` checks the offset against `getDataLen()` before reading

### DIFF
--- a/Packet++/src/IgmpLayer.cpp
+++ b/Packet++/src/IgmpLayer.cpp
@@ -333,8 +333,9 @@ namespace pcpp
 
 	igmpv3_group_record* IgmpV3ReportLayer::getFirstGroupRecord() const
 	{
-		// check if there are group records at all
-		if (getHeaderLen() <= sizeof(igmpv3_report_header))
+		// a group record is only usable if its fixed part fits inside the layer, otherwise reading
+		// the record type, the source count or the multicast address runs past the end of the data
+		if (getHeaderLen() < sizeof(igmpv3_report_header) + sizeof(igmpv3_group_record))
 			return nullptr;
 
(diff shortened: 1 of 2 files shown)


seladb (approved):

seladb:
Thank you @vsaraikin for working on this fix 🙏

PR #2219, opened 2026-08-15
Update C-style casts

Dimi1010:
The PR contains cherrypicked changes from #2199 which update C-style casts.
This is done to consolidate the changes into discrete PRs, as #2199 contains quite a lot of fixes.

The updates are achieved by the following changes:
- converting to a narrower Cpp style cast.
- removing the cast due to being unneeded (`DpdkTests.cpp`)

Some unrelated changes are also included as they were part of the cherripicked commits:
- `DeviceTeardown` has also been updated by removing the copy constructor and adding a move constructor.
- `prepareCoreConfiguration` vector parameters changed to `const&`
- range-for loop variable updated to `auto const&`
- SuppressLogs `getFilename` return value changed to `std::string const&`.

---
Cherrypicked from #2199

### DIFF
--- a/Common++/header/PointerVector.h
+++ b/Common++/header/PointerVector.h
@@ -177,10 +177,14 @@ namespace pcpp
 
 			// Release is called after the raw pointer is already inserted into the vector to prevent
 			// a memory leak if push_back throws.
-			// cppcheck-suppress danglingLifetime
+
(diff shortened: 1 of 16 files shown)


egecetin (approved):

seladb (approved):

tigercosmos (approved):
LGTM

PR #2220, opened 2026-08-15
Fix size_t underflow in the SSH handshake message length

vsaraikin:
`getSSHHandshakeMessageLength()` returns SIZE_MAX for a handshake message whose padding leaves no room for the message type.

```cpp
uint32_t length = be32toh(getMsgBaseHeader()->packetLength);
return static_cast<size_t>(length) - getMsgBaseHeader()->paddingLength - sizeof(uint8_t) * 2;
```

Both `packetLength` and `paddingLength` come from the packet. The parser checks that the padding is not larger than the message:

```cpp
if (msgBase->paddingLength > msgLength)
    return nullptr;
```

but it allows them to be equal, and the length subtracts two more bytes for the padding length and the message type. So `paddingLength` of `msgLength` or `msgLength - 1` underflows and the message is reported as 18446744073709551615 bytes long.

That matters because `getSSHHandshakeMessage()` and `getSSHHandshakeMessageLength()` are the documented pair for reading the message body, so a caller reads that many bytes from a 16 byte layer.

The length now returns 0 when the overhead does not fit, which is what the existing New Keys case already produces the honest way (packet length 12, padding 10).

Added to `SSHMalformedParsingTest`: a packet with packet length 12 and padding length 11. On `dev` the assert reads

```
Actual:   18446744073709551615
Expected: 0
```

Packet++Test is 259 passed, 0 failed


Tests/Packet++Test/Tests/SSHTests.cpp
```diff
@@ -126,6 +126,29 @@ PTF_TEST_CASE(SSHMalformedParsingTest)
 	PTF_ASSERT_NULL(sshMessageMalformedLenPacket.getLayerOfType<pcpp::SSHHandshakeMessage>());
 	PTF_ASSERT_NOT_NULL(sshMessageMalformedLenPacket.getLayerOfType<pcpp::SSHEncryptedMessage>());
 
+	// SSH message whose padding leaves no room for the message type, which used to underflow
+	// getSSHHandshakeMessageLength() and report a message of SIZE_MAX bytes
+	{
+		uint8_t underflow[] = { 0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x55, 0x44, 0x33, 0x22, 0x11, 0x08, 0x00,
```
egecetin: Most of the tests (maybe all of them) using `createPacketFromHexResource`. Instead of embedding array data of the packet using a dat file and reading it might be better for consistency.

vsaraikin:
Moved it to `PacketExamples/SSHMessage_MalformedPaddingUnderflow.dat` and it loads through `createPacketFromHexResource` like the rest, thanks.

Unrelated, but I noticed it while renumbering the locals in that test: the malformed type case builds its packet from the wrong raw packet.

```cpp
auto rawPacket4 = createPacketFromHexResource("PacketExamples/SSHMessage_MalformedType.dat");
pcpp::Packet sshMessageMalformedTypePacket(rawPacket3.get());
```

`rawPacket3` is the malformed padding length file, so `SSHMessage_MalformedType.dat` is never actually parsed. The asserts pass either way since both are expected to fall back to an encrypted message. Happy to fix it here or leave it for a separate PR

egecetin:
Nice catch! Separate PR is better for tracking history/changes for next release. Can you open a PR for it?

egecetin (approved):

Packet++/src/SSHLayer.cpp
```diff
@@ -109,8 +109,16 @@ namespace pcpp
 
 	size_t SSHHandshakeMessage::getSSHHandshakeMessageLength() const
 	{
-		uint32_t length = be32toh(getMsgBaseHeader()->packetLength);
-		return static_cast<size_t>(length) - getMsgBaseHeader()->paddingLength - sizeof(uint8_t) * 2;
+		size_t length = be32toh(getMsgBaseHeader()->packetLength);
```
seladb: `be32toh` should return `uint32_t`, I'd add `static_cast<size_t<` here

seladb (approved):
@vsaraikin please see one comment, otherwise LGTM

Packet++/src/SSHLayer.cpp
vsaraikin: Added, thanks

vsaraikin:
Opened it as #2232

Tests/Packet++Test/Tests/SSHTests.cpp
vsaraikin: Done, it loads from PacketExamples/SSHMessage_MalformedPaddingUnderflow.dat now

Issue #2221, opened 2026-08-16
BPF filters no longer work on LINKTYPE_RAW (101) pcap files (regression in v26.07)


ZByteCrafter:
### Description

Since PcapPlusPlus v26.07 (and current `master`), setting a BPF filter on a pcap/pcapng
file whose link-layer type is `LINKTYPE_RAW` (101) no longer matches any packets.
Before v26.07 this worked correctly.

### How to reproduce

The repository itself contains such a file: `Tests/Pcap++Test/PcapExamples/raw_ip.pcap`
(its file header stores linktype = 101, i.e. `LINKTYPE_RAW`).

```cpp
pcpp::PcapFileReaderDevice reader(R"(Tests/Pcap++Test/PcapExamples/raw_ip.pcap)");
reader.open();

// e.g. TcpReassembly / TLSFingerprinting examples behave exactly like this
if (!reader.setFilter("ip"))        // returns true (compiled against Ethernet by default)
    return;

pcpp::RawPacket packet;
int count = 0;
while (reader.getNextPacket(packet))  // always returns false -> 0 packets read
    count++;
```
Expected: all 50 IPv4 packets in `raw_ip.pcap` pass the `"ip"` filter (this is what v25.05 and earlier did). Actual: `setFilter` seems fine but `getNextPacket` returns 0 packets.

### Root cause

I believe this is a regression introduced by PR #2034 ("Read and write pcap files without libpcap", commit `44e90857`).

- The new reader parses the pcap header itself and keeps the __file LINKTYPE value__ verbatim: `PcapFileDevice.cpp`'s `toLinkLayerType()` returns `static_cast<LinkLayerType>(value)` for `LINKTYPE_RAW` (101).
- Previously the reader went through libpcap (`pcap_open_offline` + `pcap_datalink`), which converts the on-disk `LINKTYPE_RAW` (101) to the platform __DLT_RAW__ value (12 on Windows/Linux), i.e. `LINKTYPE_DLT_RAW1`.
- BPF compilation in `BpfFilterWrapper::compileFilter()` / `BpfFilterWrapper::matches()` calls `pcap_open_dead(linkType, ...)` + `pcap_compile(...)` with that stored value. `101` is not a valid DLT for libpcap, so compilation fails and every packet is dropped.
[...]

PR #2222, opened 2026-08-16
Fix packets with `Linktype_RAW` (101) not being recognized by the BPF filters correctly.

Dimi1010:
This fixes a regression during the transition to a non-libpcap parser described in #2221, where filters would stop working for packets of type RAW. The change also restores a test to keep track of the behaviour, with the change being that the packets are now reported as platform agnostic type (LINKTYPE_RAW) instead of the platform specific type (LINKTYPE_DLT_RAW), and only perform the conversion internally.


Pcap++/src/PcapFilter.cpp
```diff
@@ -131,13 +131,40 @@ namespace pcpp
...
+			switch (linktype)
+			{
+			// NOTE: Issue #2221
+			// Libpcap maps LINKTYPE_RAW to the old DLT_RAW. LINKTYPE_RAW is not that well supported for BPF filters.
+			// https://github.com/the-tcpdump-group/libpcap/blob/720fb235648dd06eaa5b0fc0e7b6bac84bf2bf28/pcap-common.c#L1559-L1560
+			case LinkLayerType::LINKTYPE_RAW:
+			{
+#ifdef __OpenBSD__
+				return LinkLayerType::LINKTYPE_DLT_RAW2;  // OpenBSD uses a different DLT_RAW value.
+#else
```
Dimi1010: Do we have an OpenBSD CI to test this?

egecetin: What about FreeBSD? If it has the same behavior. https://github.com/seladb/PcapPlusPlus/blob/25e1f9cb34230963a1a201c0eb1b0a81c8305c77/.github/workflows/build_and_test.yml#L831-L835

Dimi1010: Tbh, I am not sure if both have the same behavior. Not really familiar with them.
I took that define straight out of libpcap's code. 
[Define link](https://github.com/the-tcpdump-group/libpcap/blob/720fb235648dd06eaa5b0fc0e7b6bac84bf2bf28/pcap/dlt.h#L111-L115)
[Conversion location](https://github.com/the-tcpdump-group/libpcap/blob/720fb235648dd06eaa5b0fc0e7b6bac84bf2bf28/pcap-common.c#L1559-L1560)

ZByteCrafter:
@Dimi1010 Nice fix, thanks for the quick turnaround!

We verified the same approach on Windows/Npcap (v26.07):
without it the BPF filter matches 0/50 on raw_ip.pcap (test fails),
with it 50/50 pass and no regressions in the other filter tests.

Minor: files storing DLT_RAW2 (14) have the same issue on non-OpenBSD
(libpcap maps it to DLT_RAW=12 too). Our local patch covered the whole
raw family, FYI. Not blocking.

Thanks!

Dimi1010:
> files storing DLT_RAW2 (14) have the same issue on non-OpenBSD
(libpcap maps it to DLT_RAW=12 too). Our local patch covered the whole
raw family, FYI. Not blocking.

Sure, that can be patched too. Do you know if the same is true in reverse, eg 12 on OpenBSD?

Pcap++/src/PcapFilter.cpp
```diff
@@ -131,13 +131,40 @@ namespace pcpp
 #endif
 	}
 
+	namespace
+	{
+		/// @brief Applies linktype patches to ensure compatibility with BPF filters.
+		/// @param linktype The original link layer type.
+		/// @return The patched link layer type suitable for BPF filters.
+		PCPP_MAYBE_UNUSED LinkLayerType patchLinktype2dlt(LinkLayerType linktype)
```
seladb: Why do we need `PCPP_MAYBE_UNUSED`? These are just a few lines of code, we can move them to the `#ifdef USE_PCAP` section:
```cpp
#ifdef USE_PCAP

if (linkType == LinkLayerType::LINKTYPE_RAW)
{
#ifdef __OpenBSD__
	linkType = LinkLayerType::LINKTYPE_DLT_RAW2; 
#else
	linkType = LinkLayerType::LINKTYPE_DLT_RAW1;
#endif
}
auto pcap = std::unique_ptr<pcap_t, internal::PcapCloseDeleter>(
		    pcap_open_dead(patchLinktype2dlt(linkType), DEFAULT_SNAPLEN));

...
```

Pcap++/src/PcapFilter.cpp
seladb: According to Claude FreeBSD doesn't define `__OpenBSD__` so we can't really test it, but I guess it's fine to leave it untested because we don't officially support OpenBSD

Tests/Pcap++Test/Tests/FilterTests.cpp
```diff
@@ -836,25 +836,69 @@ PTF_TEST_CASE(TestPcapFilters_LinkLayer)
...
+						++validCounter;
+					}
+				}
+			}
 		}
+		PTF_ASSERT_EQUAL(validCounter, 50);
+		rawPacketVec.clear();
+	}
+
+	// pcpp::LINKTYPE_RAW layer with live capture
```
seladb: This is not live capture but rather setting a filter on a pcap file reader device

Pcap++/src/PcapFilter.cpp
Dimi1010: They aren't necessarily few, there may be more. According to this [comment](https://github.com/seladb/PcapPlusPlus/pull/2222#issuecomment-5307805706) DLT_RAW2 might also need to be patched to DLT_RAW1 on non-OpenBSD platforms (with potentially the reverse too).

I would prefer to keep patch code slightly separated from the base `compileFilter` code to be more distinguishable where the patch conversions happen instead of having to search them inside another function.

As for maybe unused. We don't necessarily need it. An alternative is to move the entire function under `ifdef USE_PCAP`. I chose the maybe unused due to the code not really having any specific functionality that fails to compile under non-pcap environment (other than it being unused).

seladb: I prefer to remove `PCPP_MAYBE_UNUSED`. Maybe moving `patchLinktype2dlt` to `#ifdef USE_PCAP` is a good idea

ZByteCrafter:
> > files storing DLT_RAW2 (14) have the same issue on non-OpenBSD
> > (libpcap maps it to DLT_RAW=12 too). Our local patch covered the whole
> > raw family, FYI. Not blocking.
> 
> Sure, that can be patched too. Do you know if the same is true in reverse, eg 12 on OpenBSD?

Good question, but I have to be honest - we only tested on Windows, no OpenBSD environment here, so I can't confirm it from experience.

Based on libpcap's linktype_to_dlt() mapping (12/14/101 all map to the platform's DLT_RAW), I'd expect 12 on OpenBSD (= DLT_RAW) to have the same issue. Mapping all three values to the platform DLT_RAW would cover it symmetrically. Maybe someone with OpenBSD can confirm.

Dimi1010:
> Based on libpcap's linktype_to_dlt() mapping (12/14/101 all map to the platform's DLT_RAW), I'd expect 12 on OpenBSD (= DLT_RAW) to have the same issue. Mapping all three values to the platform DLT_RAW would cover it symmetrically. Maybe someone with OpenBSD can confirm.

@ZByteCrafter Where did you see DLT mapping 14 to 12 in the libpcap code? Can you give a link, please? I can't find any mapping between the two. As far as I can see, libpcap DLT_RAW is either 12 or 14, but only 1 of the two "exists" in the code in a build.

Pcap++/src/PcapFilter.cpp
Dimi1010: Fixed 983bdab

Tests/Pcap++Test/Tests/FilterTests.cpp
Dimi1010: Fixed cd49de7

seladb (approved):

PR #2225, opened 2026-08-19
IGMPv3 tests: Move truncated example from inlined array to a hex resource file.

Dimi1010:
This moves the truncated array to a data file to conform with other tests and allow usage of the `pcpp_tests::createPacketFromHexResource` machinery.

Includes modifications to the `ResourceProvider` as it was used to generate the data file from the inlined array.
- Adds freeze and unfreeze actions to `ResourceProvider` to allow modifying the frozen setting.
- Changes `ResourceProvider::saveResource` to be non-const action.


Tests/PcppTestUtilities/Resources.h
```diff
@@ -50,8 +62,7 @@ namespace pcpp_tests
 			/// @param data Pointer to the data to be saved.
 			/// @param length The length of the data in bytes.
 			/// @throw std::runtime_error if the provider is frozen and does not allow saving.
-			void saveResource(ResourceType resourceType, const char* filename, const uint8_t* data,
-			                  size_t length) const;
+			void saveResource(ResourceType resourceType, const char* filename, const uint8_t* data, size_t length);
```
Dimi1010, line 65: Stripped the const since it conceptually mutates the resource store.

egecetin (approved):

seladb (approved):

PR #2228, opened 2026-08-20
fix bpf/xsk.h file not found

### LINKED ISSUE #2227: Cannot build with XDP support against libbpf > 1.0
### Bug description

**Describe the bug**
Building with XDP support enabled fails against libbpf > 1.0, since xsk.h was moved from libbpf to libxdp. Originally raised in #1635 

**Code example to reproduce**
```
$ cat /usr/include/bpf/libbpf_version.h | grep "LIBBPF_.*_VERSION"
#define LIBBPF_MAJOR_VERSION 1
#define LIBBPF_MINOR_VERSION 3

$ cat /etc/os-release | grep PRETTY_NAME
PRETTY_NAME="Ubuntu 24.04.4 LTS"

$ cmake .. -DPCAPPP_USE_XDP=ON
$ make
/home/user/code/PcapPlusPlus/Pcap++/src/XdpDevice.cpp:8:10: fatal error: bpf/xsk.h: No such file or directory
```

**Expected behavior**
Code should compile successfully, with XDP support

PcapPlusPlus master branch

_No response_

Linux

_No response_

### Compiler version

g++ 13.3.0

XDP

cconverse711:
xsk.h was moved from libbpf to libxdp at version 1.0

closes #2227


seladb:
@cconverse711 I'm sorry for the delayed review. The CI fails and according to Claude:

> Here's what's happening: `TestXdpDeviceReceivePackets` opens an AF_XDP socket on the interface/queue, runs its assertions, and tears the socket down at the end. But the kernel's unbind of the umem from the network device happens asynchronously after `xsk_socket__delete()` returns — the delete call returns before the kernel has actually finished releasing the queue. Then `TestXdpDeviceSendPackets` runs immediately after (same interface, same queue), calls `xsk_socket__create()`, and the kernel still considers the queue bound from the previous socket, so the bind fails with `-EBUSY` (-16).
> 
> This is a well-documented libbpf/libxdp behavior, not new to libxdp — it existed before the xsk.h move too, so libxdp being newly installed didn't cause it. The PR just exposed it because now the tests can actually run and reach this teardown/setup sequence (previously they were failing earlier at a build/include error).
> 
> **Fixes to look for or suggest in review:**
> 
> 1. **Add a delay between tests** — the pragmatic fix others hit this with: a small `sleep()` (even ~100ms–1s) or a retry loop with backoff between the teardown of one XDP socket test and the setup of the next, when they share the same interface/queue.
> 2. **Retry `xsk_socket__create()` on EBUSY** — wrap the create call in a small retry loop (a few attempts with short sleeps) specifically for `-EBUSY`, since this is a known transient condition rather than a real config error.
> 3. **Use separate queue IDs per test** — if the CI interface is multi-queue, having `ReceivePackets` and `SendPackets` bind to different queue IDs avoids the collision entirely.
> 4. **Check test ordering/isolation** — if these are separate test binaries or use a shared fixture, verify there's an explicit teardown synchronization point (not just relying on `xsk_socket__delete()` returning) before the next test's setup runs.
> 
> Given it's a fresh CI environment likely running a single-queue veth-style interface, option 1 or 2 (small retry/backoff around create) is probably the most robust fix without needing multiple queues.

If this is indeed the issue, I'd choose option 1 and add a loop of 100ms sleep for up to 2 seconds in each test if `device.open()` fails.
We can do it inside the tests rather than in the `XdpDevice` class itself because this is something that the user should handle and not the library.

.github/workflows/build_and_test.yml
```diff
@@ -959,15 +959,15 @@ jobs:
...
+    runs-on: ubuntu-latest
 
     steps:
       - name: Checkout code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Install dependencies
         run: |
-          sudo apt update && sudo apt -y install ccache libpcap-dev libbpf-dev tcpreplay
+          sudo apt update && sudo apt -y install ccache libpcap-dev libbpf-dev libxdp-dev tcpreplay
```
seladb: Is there a way to run both tests, one with an old libbpf version and another one with a newer version and libxdp?

cconverse711: Simplest approach would probably be to have two separate jobs; one on ubuntu-22.04 (like before) and one on ubuntu-latest. Alternatively we could install them from source like is done for the dpdk job(s)

seladb: Sure, I guess we can do that.

Please also notice CI is failing probably because of this: https://github.com/seladb/PcapPlusPlus/pull/2228#issuecomment-5407765487

seladb (approved):

seladb:
Thank you @cconverse711 for working on this 🙏

PR #2229, opened 2026-08-20
Add QUIC v1 layer

seladb:
Adds a `QuicLayer` for parsing QUIC v1 (RFC 9000) packets.

- New `QuicV1Layer` class hierarchy covering both packet forms:
  - **Long header**: Initial, 0-RTT, Handshake, Retry, and Version Negotiation
  - **Short header**: 1-RTT
- Parses fixed fields (version, header form, packet type) directly off the raw buffer, and exposes variable-length fields (Connection IDs, Token, Length, Retry Integrity Tag, supported versions) via pointer+length accessors rather than copying

Parsing only - no packet creation/editing support yet, and Initial-packet payloads (crypto frames, etc.) are not decrypted/parsed since that requires TLS key material not available from the packet alone.


Packet++/header/QuicLayer.h
```diff
@@ -0,0 +1,453 @@
...
+		/// @struct quic_long_header
+		/// The fixed-size prefix of every long-header QUIC v1 packet (Initial, 0-RTT, Handshake,
+		/// Retry and Version Negotiation): the common header byte, the 4-byte version, and the
+		/// Destination Connection ID length. The variable-length fields that follow (DCID, SCID,
+		/// Token, Length) are not part of this struct - they're read via pointer+length accessors
+		/// on QuicV1LongHeaderLayer and its subclasses instead
+		struct quic_long_header
+		{
+#if (BYTE_ORDER == LITTLE_ENDIAN)
+			uint8_t packetNumberLength : 2, reserved : 2, longPacketType : 2, fixedBit : 1, headerForm : 1;
```
Dimi1010, line 140: nit: Can this be spread across multiple lines? It can be a bit hard to read when all members are on a single line.

egecetin (approved):
I didn't check the protocol definition, but I took a look at the code.

Packet++/header/QuicLayer.h
```diff
@@ -0,0 +1,474 @@
...
+
+			/// Writes the ByteArray's hex-string representation (see @ref toString) to a stream
+			/// @param[in] os The output stream to write to
+			/// @param[in] byteArray The ByteArray to write
+			/// @return The same output stream, for chaining
+			friend std::ostream& operator<<(std::ostream& os, const ByteArray& byteArray)
+			{
+				return os << byteArray.toString();
+			}
+		};
```
Dimi1010: Do we need to inherit from vector here? I haven't had great experience with classes made from directly inheriting from vector (or other STL containers that aren't designed around it), so I am somewhat sceptical.

I don't think just the `toString` convinience overloads justify this, as they can be provided in a cleaner ways.

From what I can see, the entire `toString` is a wrapper over `byteArrayToHexString`. IMO, It would be better if we add a general purpose free function `std::string toHexString(std::vector<uint8_t> const& )` and/or possibly helper view objects (for `ostream` integration) that focus on transforming the buffer to a hex string at the point of need, instead of having a dedicated `ByteArray` container.

Dimi1010 (commented):
Overall, looks good. I left a comment about the `ByteArray`, since that can do with some refactoring, IMO.

Somewhat unsure about how QuicV2 will overlay with these classes, tho. So that might be a breaking change down the line. :/

Packet++/header/QuicLayer.h
seladb: > Do we need to inherit from vector here? I haven't had great experience with classes made from directly inheriting from vector (or other STL containers that aren't designed around it), so I am somewhat sceptical.

What bad experience did you have with it?


> I don't think just the `toString` convinience overloads justify this, as they can be provided in a cleaner ways.
> 
> From what I can see, the entire `toString` is a wrapper over `byteArrayToHexString`. IMO, It would be better if we add a general purpose free function `std::string toHexString(std::vector<uint8_t> const& )` and/or possibly helper view objects (for `ostream` integration) that focus on transforming the buffer to a hex string at the point of need, instead of having a dedicated `ByteArray` container.

I'm not entirely sure what you suggest. I'm looking for ways to provide both the hex string as well as the byte array which is needed in multiple places for QUIC, do you have a better suggestion of the API we should expose?

Dimi1010: > What bad experience did you have with it?

The issue I have is that those inherited containers tend to spread and create some friction across boundaries.
If you have an API that assigns to `ByteArray`, but you have `std::vector<uint8_t>` you either need a copy, or to implement a custom move constructor from `std::vector` to `ByteArray`, since the compiler doesn't do that for you. Same with a function that takes a `ByteArray` param (tho that isn't really relevant here).

They also require extra attention if the entity overrides any existing methods, since then you can have issues with the static dispatch.

Due to the above reasons, I prefer to avoid them if they aren't strictly needed, which in the current case I believe they aren't.

> I'm not entirely sure what you suggest. I'm looking for ways to provide both the hex string as well as the byte array which is needed in multiple places for QUIC, do you have a better suggestion of the API we should expose?

My idea is that instead of having a specific QUIC container type to provide byte array to hex string conversions for just this type, we should add better general purpose utilities for byte array to hex string. That way the utilities would be available for the entire library (and users) the next time it is needed instead of needing to be reimplemented.

The change would also allow the user to choose their preferred formatting option at the call site, instead of having the library force a certain format inside the `toString` implementation.

Here is some code for said proposal.
```cpp

// QuicLayer.hpp

class QuicLayer
{
  // Drop the special ByteArray container and return a generic vector<uint8_t>.
  std::vector<uint8_t> getConnectionId() const;
}

// GeneralUtility.hpp (or somewhere else)

// General purpose convenience function for converting a byte array vector to hex string.
std::string toHexString(std::vector<uint8_t> const& byteArr)
{
	return byteArrayToHexString(byteArr.data(), byteArr.size());
}

// General purpose view class to output a byte array like a hex to an ostream.
class ByteArrayHexView
{
public:
	ByteArrayHexView(uint8_t const* byteArr, size_t byteArrSize) : m_ByteArr(byteArr), m_Size(byteArrSize)
	{}

	std::string toString() const
	{
		return byteArrayToHexString(m_ByteArr, m_Size);
	}

	friend std::ostream& operator<<(std::ostream& os, ByteArrayHexView const& view)
	{
		// Potential optimization opportunity:
		// Skip creating the string and output to std::ostream directly.
		os << byteArrayToHexString(view.m_ByteArr, view.m_Size);
		return os;
	}

private:
	uint8_t const* m_ByteArr;
	size_t m_Size;
};

// General purpose utility to convert a byte array-like object to a hex view class.
ByteArrayHexView asHex(std::vector<uint8_t> const& bytes)
{
	return ByteArrayHexView(bytes.data(), bytes.size());
}

// Can be cleanly overloaded with different types if needed. Those do not need to be in the same header.
ByteArrayHexView asHex(pcpp::ScalarBuffer<uint8_t> const& bytes)
{
	// Somewhat off memory, but we do have a scalar buffer view struct.
	return ByteArrayHexView(bytes.data, bytes.len);
}

// Usage example.
void fooUsage()
{
	// Works for any byte array, instead of just for the specific type.
	std::vector<uint8_t> bytes = { 0xaa, 0x2b, 0x10 };
	// User choses the output format at the call site, instead of the PCPP choosing it for him inside the old
	// ByteArray.toString() method.
	std::cout << pcpp::asHex(bytes) << std::endl;
	
	pcpp::QuicLayer quic /* from somewhere */;
	std::cout << "Connection Id as hex:" << pcpp::asHex(quic.getConnectionId()) << std::endl;
	
	// Potential parallel implementation for formatting the array as raw binary (0/1) instead of hex.
	// std::cout << "Connection Id as bytes" << pcpp::asBinary(bytes) << std::endl;

}
```

seladb: Instead of adding these extra classes, won't it be simpler to add an overload to `byteArrayToHexString()`:

```cpp
std::string byteArrayToHexString(const uint8_t* byteArr, size_t byteArrSize, int stringSizeLimit = -1);
std::string byteArrayToHexString(const std::vector<uint8_t> byteArr);
// more overloads if we need?
```

However, ultimately the user still needs to call this method:
```cpp
auto connId = quicLayer->getConnectionId();
std::cout << "Connection ID: " << pcpp::byteArrayToHexString(connId);
```

IMO, the ideal interface could be:
```cpp
std::cout << "Connection ID: " << quicLayer->getConnectionId().toString();
```

Or even better:
```cpp
std::cout << "Connection ID: " << quicLayer->getConnectionId();
```

Maybe we can add move / copy constructors from `ByteArray` to `std::vector<uint8_t>` if you think it's better...

Dimi1010: > Instead of adding these extra classes, won't it be simpler to add an overload to `byteArrayToHexString()`:
> 
> ```c++
> std::string byteArrayToHexString(const uint8_t* byteArr, size_t byteArrSize, int stringSizeLimit = -1);
> std::string byteArrayToHexString(const std::vector<uint8_t> byteArr);
> // more overloads if we need?
> ```

That function overload only covers `toHexString`. It can work, but I think that the full name is a bit long and cumbersome to call inside a chain, as you have noted. That is why I suggested having a shorter named alternatives (or aliases) for ease of use.

Also, it does not allow the optimization opportunity when outputting to a stream that I mentioned in my example, as it would always require allocating and deallocating a string.

> However, ultimately the user still needs to call this method:
> 
> ```c++
> auto connId = quicLayer->getConnectionId();
> std::cout << "Connection ID: " << pcpp::byteArrayToHexString(connId);
> ```
> 
> IMO, the ideal interface could be:
> 
> ```c++
> std::cout << "Connection ID: " << quicLayer->getConnectionId().toString();
> ```
> 
> Or even better:
> 
> ```c++
> std::cout << "Connection ID: " << quicLayer->getConnectionId();
> ```

I agree that the interface would be nice. But I don't think its existence justifies an entire new container class that is very situational (only used inside `QuicLayer`) and only differs in formatting behaviour, when a similar thing can be done with a free conversion function.

Personally, I don't see much issue with the user needing to call a conversion function to get a certain formatting style provided the name is succinct. `std::cout << pcpp::asHex(quicLayer->getConnectionId())` is just as good as the `toString` variant IMO, and does not require the user working around additional custom types, which can become a bigger pain point. In how many places is output to a stream really needed, compared to using the a specific type? 
- In the view class option, the user would almost never have to store the view class somewhere. Or even know it exists as the primary interaction would be `pcpp::asHex(...)` inside the `std::cout <<` line.
- In the container class option, every time the user wants to store the connection id, it would either need to be stored as `pcpp::QuicLayer::ByteArray`, or converted to `std::vector<uint8_t>` where the user loses all benefits from the custom type if printing is needed down the line.

The existence of an extra array class also would require the users to look up how that object differs from the simple `vector<uint8_t>` buffers that we have used elsewhere in the library API, which it ultimately doesn't.

To clarify, my issue isn't really with the extra class for formatting API, it is that the class is a container class, which can persist for significantly more time (and be used in a lot more places) than a view class's common use and dispose pattern. As a result of that, they usually end up significantly harder to change if needed, and potential issues with them can be much more impactful, as they are more prevalent in user code.

> Maybe we can add move / copy constructors from `ByteArray` to `std::vector<uint8_t>` if you think it's better...

We can, and should, if you insist on having the custom byte array type for the formatting. But IMO, it is inconsistent with the existing library code, where a simple buffer is represented as `vector<uint8_t>` and the specific formatting is left at the user's discretion and/or free-function converters.

Dimi1010: Looking through the code, in similar situations, we have usually either supplied a custom object type specifying the type and holding the buffer as a data member and having the extra methods bound to the object type, or just returning the buffer as `vector`.

seladb: I guess a good middle ground would be to add `get...AsString()` everywhere and get rid of the `ByteArray`, for example: have `getDestinationConnectionIdAsString()` in addition to `getDestinationConnectionId()`.

I'll make this change

Packet++/header/QuicLayer.h
seladb: I tried changing it, but `clang-format` reverted it back to one line...

Dimi1010: Oh well... 😞

Packet++/header/QuicLayer.h
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2229/commits/d6345ae719a2f251ef5e935bc14e714b988f4c7c

Dimi1010 (approved):
LGTM

PR #2230, opened 2026-08-20
Fix raw data overflow in X509ExternsionDataDecoder

arthurscchan:
In #2170 a new fuzzer was added, and it discovered a new heap overflow problem. X509ExtensionDataDecoder::decodeAsn1Data decoded the extension value from a hex string into a buffer half that string's length, but passed rawData.size() as the buffer length to Asn1Record::decode, which is twice the real allocation. That defeated the decoder's own bounds check (m_TotalLength > dataLen), so an extension holding an ASN.1 record whose declared length exceeded the extension value would read past the end of the buffer, causing a heap overflow given a sufficiently long hex string. This PR fixes the length that is passed in, and adds a new test case for this bug.


seladb (approved):

seladb:
Thank you @arthurscchan for this fix 🙏

PR #2231, opened 2026-08-21
Fix typos

cconverse711:


seladb (dismissed):

cconverse711:
Huh, not sure why cppcheck would start complaining about those things now. Looks like adding the `--include` option caused cppcheck to start analyzing the header files, which it seems like it should have been doing all along. I also tried using `--library` instead with a custom config that teaches cppcheck about the PCPP_DEPRECATED macro, and it had the same effect as `--include`.

seladb:
> Huh, not sure why cppcheck would start complaining about those things now. Looks like adding the `--include` option caused cppcheck to start analyzing the header files, which it seems like it should have been doing all along. I also tried using `--library` instead with a custom config that teaches cppcheck about the PCPP_DEPRECATED macro, and it had the same effect as `--include`.

Yes, I think adding `--include` to cppcheck is causing it, I'd remove it...

cconverse711:
`--include` was added because otherwise cppcheck reports a syntax error caused by the deprecation macro. Using the equivalent  deprecation attribute directly seems to be ok though. 

I suspect that the additional cppcheck argument just revealed that certain headers were not being analyzed as expected. I'll try to investigate further.

seladb:
> `--include` was added because otherwise cppcheck reports a syntax error caused by the deprecation macro. Using the equivalent deprecation attribute directly seems to be ok though.
> 
> I suspect that the additional cppcheck argument just revealed that certain headers were not being analyzed as expected. I'll try to investigate further.

Maybe we could suppress these lines to cppcheck ignore them? 🤔

cconverse711:
You mean, to suppress the `syntaxError` lines? Perhaps, though my understanding is that when a `syntaxError` occurs (even if it's suppressed), cppcheck will abort processing that file, so that seems undesirable. 

As for the newly reported cppcheck defects, they seem valid, right? (as in, cppcheck should've detected them before)

```
Packet++/header/Asn1Codec.h:462:5: performance: Variable 'm_Value' is assigned in constructor body. Consider performing initialization in initialization list. [useInitializationList]
    m_Value = initFromInt(value);
    ^
Packet++/header/X509ExtensionDataDecoder.h:48:13: style: Class 'X509ExtendedKeyUsagePurpose' has a constructor with 1 argument that is not explicit. [noExplicitConstructor]
  constexpr X509ExtendedKeyUsagePurpose(Value value) : m_Value(value)
            ^
Packet++/header/X509ExtensionDataDecoder.h:138:4: style: Class 'X509SubjectKeyIdentifierDataDecoder' has a constructor with 1 argument that is not explicit. [noExplicitConstructor]
   X509SubjectKeyIdentifierDataDecoder(const std::string& keyIdentifier) : m_KeyIdentifier(keyIdentifier)
   ^
Packet++/header/X509ExtensionDataDecoder.h:160:4: style: Class 'X509KeyUsageDataDecoder' has a constructor with 1 argument that is not explicit. [noExplicitConstructor]
   X509KeyUsageDataDecoder(const std::string& keyUsage) : m_KeyUsage(keyUsage)
   ^
Packet++/header/Asn1Codec.h:462:5: performance: Variable 'm_Value' is assigned in constructor body. Consider performing initialization in initialization list. [useInitializationList]
    m_Value = initFromInt(value);
    ^
Pcap++/header/Device.h:56:33: style: Parameter 'filter' can be declared as reference to const [constParameter]
  bool setFilter(GeneralFilter& filter)
                                ^
Pcap++/header/Device.h:56:33: style: Parameter 'filter' can be declared as reference to const [constParameter]
  bool setFilter(GeneralFilter& filter)
                                ^
```

seladb:
> As for the newly reported cppcheck defects, they seem valid, right? (as in, cppcheck should've detected them before)

Maybe they're valid, but I wonder if we want to fix them in this PR... 🤔

cconverse711:
Probably makes sense to fix them in a separate PR. I'm still confused as to how the new flag(s) cause those new defects to appear; might be some bug in cppcheck that's resolved in a newer version.

seladb (approved):

PR #2236, opened 2026-08-24
Add Windows 11 ARM64 support

egecetin:


.github/workflows/build_and_test.yml
```diff
@@ -785,6 +785,64 @@ jobs:
...
+  # WinPcap has no arm64 support, so this job only runs when NPcap credentials are available. Additionally OpenCppCoverage
+  # not works with arm64 currently so coverage calculation is disabled for Windows ARM builds.
+  visual-studio-arm64:
+    needs: define-windows-pcap-backend-matrix
+    if: ${{ needs.define-windows-pcap-backend-matrix.outputs.npcap-or-default == 'npcap' }}
+    strategy:
+      matrix:
+        include:
+          - os: windows-11-arm
+            platform: "Visual Studio 17 2022"
```
seladb, line 797: Do we want to add `Visual Studio 18 2026`?

.github/workflows/build_and_test.yml
```diff
@@ -785,6 +785,64 @@ jobs:
...
+            arch: "arm64"
```
seladb: Do we really need `arch` as a parameter? This job is only meant for arm64...

cmake/modules/FindPacket.cmake
```diff
@@ -40,21 +40,44 @@
...
-  # For the WinPcap and Npcap SDKs, the Lib subdirectory of the top-level directory contains 32-bit libraries; the
-  # 64-bit libraries are in the Lib/x64 directory.
-  #
-  # The only way to *FORCE* CMake to look in the Lib/x64 directory without searching in the Lib directory first appears
-  # to be to set CMAKE_LIBRARY_ARCHITECTURE to "x64".
-  #
-  set(CMAKE_LIBRARY_ARCHITECTURE "x64")
+# This logic only applies to the WinPcap/Npcap SDK's Lib directory layout.
+if(WIN32)
+  if(CMAKE_SIZEOF_VOID_P EQUAL 4)
```
seladb: Why not keeping the original condition `if(CMAKE_SIZEOF_VOID_P EQUAL 8)`?

An alternative would be:
```
if(NOT CMAKE_SIZEOF_VOID_P EQUAL 4)
  ...
endif()
```

cmake/modules/FindPCAP.cmake
```diff
@@ -33,16 +33,37 @@
...
-  # For the WinPcap and Npcap SDKs, the Lib subdirectory of the top-level directory contains 32-bit libraries. The
-  # 64-bit libraries are in the Lib/x64 directory.
-  #
-  # The only way to *FORCE* CMake to look in the Lib/x64 directory without searching in the Lib directory first appears
-  # to be to set CMAKE_LIBRARY_ARCHITECTURE to "x64".
-  #
-  set(CMAKE_LIBRARY_ARCHITECTURE "x64")
+# This logic only applies to the WinPcap/Npcap SDK's Lib directory layout.
+if(WIN32)
+  if(CMAKE_SIZEOF_VOID_P EQUAL 4)
```
seladb: ditto

cmake/modules/FindPacket.cmake
egecetin: It is possible but still need outer if for win32 otherwise mingw or linux (i cant remember which one) fails to compile.

seladb: Sure, we can leave the outer `if` statement

cmake/modules/FindPCAP.cmake
```diff
@@ -33,16 +33,30 @@
...
-  # For the WinPcap and Npcap SDKs, the Lib subdirectory of the top-level directory contains 32-bit libraries. The
-  # 64-bit libraries are in the Lib/x64 directory.
-  #
-  # The only way to *FORCE* CMake to look in the Lib/x64 directory without searching in the Lib directory first appears
-  # to be to set CMAKE_LIBRARY_ARCHITECTURE to "x64".
-  #
-  set(CMAKE_LIBRARY_ARCHITECTURE "x64")
+# This logic only applies to the WinPcap/Npcap SDK's Lib directory layout.
+if(WIN32)
+  if(NOT CMAKE_SIZEOF_VOID_P EQUAL 4)
```
seladb: Are we sure we need the nested `if`, can't we do:
```cmake
if(WIN32 AND NOT CMAKE_SIZEOF_VOID_P EQUAL 4)
  ...
endif()
```

Which is very similar to what we had before this change 🤔

cmake/modules/FindPacket.cmake
```diff
@@ -40,21 +40,38 @@
...
-  # 64-bit libraries are in the Lib/x64 directory.
-  #
-  # The only way to *FORCE* CMake to look in the Lib/x64 directory without searching in the Lib directory first appears
-  # to be to set CMAKE_LIBRARY_ARCHITECTURE to "x64".
-  #
-  set(CMAKE_LIBRARY_ARCHITECTURE "x64")
+# This logic only applies to the WinPcap/Npcap SDK's Lib directory layout.
+if(WIN32)
+  if(NOT CMAKE_SIZEOF_VOID_P EQUAL 4)
+    #
```
seladb: ditto

cmake/modules/FindPacket.cmake
```diff
@@ -1,4 +1,4 @@
-# ~~~
+#
```
seladb: nit: why did you remove the `~~~`?

egecetin: I get the latest version from wireshark repo so it is because of that I will revert some changes to minimize

seladb (approved):

PR #2238, opened 2026-08-27
Fix TLVRecord type safety in getValueAs() and setValue() (#2237)

### LINKED ISSUE #2237: Restrict TLVRecord getValueAs/setValue to trivially copyable types
## Describe the bug

`TLVRecord::getValueAs<T>()` and `TLVRecord::setValue<T>()` currently accept arbitrary types, although both functions use `memcpy` to copy raw bytes to and from objects of type `T`.

This implementation is only safe for types whose object representation can be copied byte-for-byte, such as trivially copyable types. However, the current API does not enforce this requirement at compile time.

For example, the following code compiles successfully:

```cpp
auto value = dhcpLayer
                ->getOptionData(pcpp::DHCPOPT_DHCP_LEASE_TIME)
                .getValueAs<std::string>();
```

`std::string` is not a trivially copyable type, so using it with `getValueAs<T>()` results in invalid behavior.

### 1. Insufficient-data path

When the record data is smaller than `sizeof(T) + offset`, `getValueAs<T>()` currently returns:

```cpp
return 0;
```

For `T = std::string`, the integer literal `0` can be converted to a null pointer and then used to initialize a `std::string` through its `const char*` constructor.

For example:

```cpp
std::string str(nullptr);
```

With GCC's `libstdc++`, this results in a runtime exception:

```text
basic_string: construction from null is not valid
```

Changing the return expression to:

```cpp
return static_cast<T>(0);
```

does not solve the problem, because `static_cast<std::string>(0)` is still accepted.

The insufficient-data path should instead use value-initialization:

```cpp
return T{};
```

### 2. Sufficient-data path

When the record contains enough data, `getValueAs<T>()` performs:

```cpp
memcpy(&result, ..., sizeof(T));
```

For non-trivially copyable types, this overwrites the internal object representation directly, bypassing constructors and object invariants. For types such as `std::string`, this can corrupt internal state and results in undefined behavior.

`setValue<T>()` has the same underlying issue because it also performs a raw `memcpy` without constraining `T`.

### C++ version note

This invalid usage compiles without a compile-time error in C++11, C++14, C++17, and C++20.

The problem is therefore not specific to a particular C++ standard mode: the API currently allows types that are incompatible with its `memcpy`-based implementation.

## Code example to reproduce

```cpp
// getValueAs<T>() currently accepts a non-trivially copyable type.
// This compiles, but the implementation is not valid for std::string.
auto value = dhcpLayer
                ->getOptionData(pcpp::DHCPOPT_DHCP_LEASE_TIME)
                .getValueAs<std::string>();
```

## Minimal reproduction of the runtime failure

The following reproduces the behavior of the insufficient-data path for `T = std::string`:

```cpp
#include <iostream>
#include <string>

int main()
{
    std::cout << "sizeof(std::string) = " << sizeof(std::string) << "\n";
    std::string str(0);
    return 0;
}
```

Example compilation and output:

```shell
❯ clang++ main.cc -std=c++14 -Wall -Wextra -O0
main.cc:4:14: warning: unused parameter 'argc' [-Wunused-parameter]
    4 | int main(int argc, char* argv[]) {
      |              ^
main.cc:4:26: warning: unused parameter 'argv' [-Wunused-parameter]
    4 | int main(int argc, char* argv[]) {
      |                          ^
2 warnings generated.
❯ ./a.out
sizeof(std::string) = 32
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string: construction from null is not valid
[1]    87732 IOT instruction (core dumped)  ./a.out
```

## Expected behavior

`getValueAs<T>()` and `setValue<T>()` should reject unsupported types at compile time because their implementations rely on raw `memcpy` of object representations.

In particular:

* Non-trivially copyable types such as `std::string` should fail at compile time with a clear diagnostic near the call site.
* `getValueAs<T>()` should use value-initialization for the insufficient-data path:

```cpp
return T{};
```

PcapPlusPlus master branch

_No response_

Linux

_No response_

### Compiler version

clang version 22.1.8 (Fedora 22.1.8-4.fc44)

_No response_

alacrity-aya:
## Summary

Fixes #2237.

`TLVRecord::getValueAs<T>()` and `setValue<T>()` use `memcpy` to copy raw bytes to and from objects, but previously accepted arbitrary types.

This could lead to undefined behavior when non-trivially copyable types such as `std::string` were used.

### Changes

- Add a `static_assert` to `getValueAs<T>()` and `setValue<T>()` to require `T` to be trivially copyable and non-pointer.
- Change the insufficient-data return path in `getValueAs<T>()` from `return 0;` to `return T{};`.

### Rationale

Since these APIs rely on raw `memcpy` of the object representation, accepting non-trivially copyable types is unsafe.

Using `static_assert` makes this requirement explicit and provides a direct compile-time diagnostic at the call site instead of allowing invalid types to compile and fail at runtime.

For valid trivially copyable types, `return T{};` also provides the intended value-initialized result when the record does not contain enough data.

### DIFF
--- a/Packet++/header/TLVData.h
+++ b/Packet++/header/TLVData.h
@@ -155,20 +155,24 @@ namespace pcpp
 		/// A templated method to retrieve the record data as a certain type T. For example, if record data is 4B long
 		/// (integer) then this method should be used as getValueAs<int>() and it will return the record data as an
 		/// integer.<BR> Notice this return value is a copy of the data, not a pointer to the actual data
-		/// @tparam T A non-pointer, trivially copyable type
+		/// @tparam T A non-pointer, default-constructible, trivially copyable type
 		/// @param[in] offset The offset in the record data to start reading the value from. Useful for cases when you
 		/// want to read some of the data that doesn't start at offset 0. This is an optional parameter and the default
 		/// value is 0, meaning start reading the value at the beginning of the record data
 		/// @return The record data as type T
 		template <typename T> T getValueAs(size_t offset = 0) const
 		{
-			static_assert(std::is_trivially_copyable<T>::value && !std::is_pointer<T>::value,
-			              "TLVRecord::getValueAs<T>() requires T to be a non-pointer, trivially copyable type");
+			static_assert(std::is_trivially_copyable<T>::value && std::is_default_constructible<T>::value &&
+			                  !std::is_pointer<T>::value,
+			              "TLVRecord::getValueAs<T>() requires T to be a non-pointer, "
+			              "default-constructible, trivially copyable type");
 
 			if (getDataSize() < sizeof(T) + offset)
+			{
 				return T{};
+			}
 
-			T result{};
+			T result;
 			memcpy(&result, m_Data->recordValue + getValueOffset() + offset, sizeof(T));
 			return result;
 		}
@@ -187,7 +191,9 @@ namespace pcpp
 			              "TLVRecord::setValue<T>() requires T to be a non-pointer, trivially copyable type");
 
 			if (getDataSize() < sizeof(T))
+			{
 				return false;
+			}
 
 			memcpy(m_Data->recordValue + getValueOffset() + valueOffset, &newValue, sizeof(T));
 			return true;
--- a/Packet++/header/TLVData.h
+++ b/Packet++/header/TLVData.h
@@ -3,6 +3,7 @@
 #include "Layer.h"
 #include "IpAddress.h"
 #include <string.h>
+#include <type_traits>
 
 /// @file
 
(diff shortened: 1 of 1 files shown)


Packet++/header/TLVData.h
```diff
@@ -154,29 +155,37 @@ namespace pcpp
...
 		{
+			static_assert(std::is_trivially_copyable<T>::value && !std::is_pointer<T>::value,
+			              "TLVRecord::getValueAs<T>() requires T to be a non-pointer, trivially copyable type");
+
 			if (getDataSize() < sizeof(T) + offset)
-				return 0;
+				return T{};
 
-			T result;
+			T result{};
```
seladb: I think the initialization here is redundant because the next line does `memcpy`

Packet++/header/TLVData.h
```diff
@@ -154,29 +155,37 @@ namespace pcpp
...
 		/// integer.<BR> Notice this return value is a copy of the data, not a pointer to the actual data
+		/// @tparam T A non-pointer, trivially copyable type
 		/// @param[in] offset The offset in the record data to start reading the value from. Useful for cases when you
 		/// want to read some of the data that doesn't start at offset 0. This is an optional parameter and the default
 		/// value is 0, meaning start reading the value at the beginning of the record data
 		/// @return The record data as type T
 		template <typename T> T getValueAs(size_t offset = 0) const
 		{
+			static_assert(std::is_trivially_copyable<T>::value && !std::is_pointer<T>::value,
+			              "TLVRecord::getValueAs<T>() requires T to be a non-pointer, trivially copyable type");
```
seladb: Shouldn't we add `std::is_default_constructible<T>::value` because we run the default constructor?

seladb (approved):

seladb:
@alacrity-aya thank you for working on this, much appreciated 🙏

Issue #1712, opened 2025-02-11
PSP/GENEVE Layers


RShvadron:
### Question

Hi,
I'm currently working on an application that receive a pcap file that might contain packets with PSP/GENEVE protocols.
Ate you guys planning to add pcpp::Layer support for these protocols ?
If not, would you consider to insert a pull request ?

Thank you

tigercosmos:
@RShvadron We don't have a plan on our side, but it's welcome to submit any PRs.

PR #2239, opened 2026-08-27
Add GENEVE protocol layer

alacrity-aya:
## Summary

  Add parsing, validation, crafting, and editing support for GENEVE packets.

  ## Features

  - Add `GeneveLayer` parsing and packet creation support.
  - Add accessors for VNI and encapsulated protocol EtherType.
  - Add endian-safe accessors for GENEVE option classes and fields.
  - Add a non-owning range and iterator interface for GENEVE options.
  - Support adding, finding, iterating, and removing GENEVE options.
  - Detect GENEVE traffic on UDP destination port 6081.
  - Parse encapsulated Ethernet II, IEEE 802.3, IPv4, IPv6, ARP, VLAN, and MPLS payloads.
  - Validate GENEVE version, EtherType protocol values, option lengths, critical flags, and truncated packets.
  - Add tests for parsing, packet creation, option editing, ARP payloads, and malformed packets.

  ## API Example

  ```cpp
  pcpp::GeneveLayer geneveLayer(
      0x123456,                 // 24-bit VNI
      PCPP_ETHERTYPE_IPV4,     // Encapsulated protocol
      true                      // OAM flag
  );

  const uint8_t optionData[] = {1, 2, 3, 4};

  geneveLayer.addOption(
      pcpp::GeneveOptionBuilder(
          0x0102,               // Option class
          0x03,                 // Option type
          optionData,
          sizeof(optionData),
          true                  // Critical option
      )
  );

  geneveLayer.setVNI(0xabcdef);

  for (const auto& option : geneveLayer.getOptions())
  {
      std::cout << "Class: " << option.getOptionClass()
                << ", Type: " << static_cast<int>(option.getType())
                << ", Critical: " << option.isCritical()
                << ", Data size: " << option.getDataSize()
                << std::endl;
  }

  auto option = geneveLayer.getOptions().find(0x0102, 0x03);
  if (option != geneveLayer.getOptions().end())
      geneveLayer.removeOption(0x0102, 0x03);
  ```
  ## Validation

  Malformed packets are rejected when they contain:

  - Unsupported GENEVE versions.
  - Protocol values outside the EtherType range.
  - Options exceeding the declared options area.
  - Invalid option lengths.
  - Critical options without the base GENEVE critical flag.

  ## Notes

  Partially addresses #1712.

   ## TODO

  missing translation

### DIFF
--- a/Packet++/header/GeneveLayer.h
+++ b/Packet++/header/GeneveLayer.h
@@ -1,0 +1,346 @@
+#pragma once
+
+#include "Layer.h"
+#include "TLVData.h"
+
+#include <iterator>
+#include <vector>
+
+/// @file
+
+/// @namespace pcpp
+/// @brief The main namespace for the PcapPlusPlus lib
+namespace pcpp
+{
+	class GeneveLayer;
+	class GeneveOptionIterator;
+
+	/// @struct geneve_header
+	/// Represents the fixed part of a GENEVE protocol header
+#pragma pack(push, 1)
+	struct geneve_header
+	{
+#if (BYTE_ORDER == LITTLE_ENDIAN)
+		/// Options length in 4-byte units
+		uint8_t optionsLength : 6;
+		/// Protocol version
+		uint8_t version : 2;
+		/// Reserved bits
+		uint8_t reserved1 : 6;
+		/// Critical options present flag
+		uint8_t criticalFlag : 1;
+		/// Operations, administration, and maintenance packet flag
+		uint8_t oamFlag : 1;
+#else
+		/// Protocol version
+		uint8_t version : 2;
+		/// Options length in 4-byte units
+		uint8_t optionsLength : 6;
+		/// Operations, administration, and maintenance packet flag
+		uint8_t oamFlag : 1;
+		/// Critical options present flag
+		uint8_t criticalFlag : 1;
+		/// Reserved bits
+		uint8_t reserved1 : 6;
+#endif
+		/// EtherType of the encapsulated protocol
+		uint16_t protocolType;
+		/// Virtual network identifier
+		uint8_t vni[3];
+		/// Reserved byte
+		uint8_t reserved2;
+	};
+#pragma pack(pop)
+	static_assert(sizeof(geneve_header) == 8, "geneve_header size is not 8 bytes");
+
+	/// @struct geneve_option_header
+	/// Represents a GENEVE option header
+#pragma pack(push, 1)
+	struct geneve_option_header
+	{
+		/// Option namespace assigned by IANA
+		uint16_t optionClass;
+		/// Option type. The most significant bit is the critical bit
+		uint8_t type;
+#if (BYTE_ORDER == LITTLE_ENDIAN)
+		/// Option data length in 4-byte units
+		uint8_t length : 5;
+		/// Reserved bits
+		uint8_t reserved : 3;
+#else
+		/// Reserved bits
+		uint8_t reserved : 3;
+		/// Option data length in 4-byte units
+		uint8_t length : 5;
+#endif
+	};
+#pragma pack(pop)
+	static_assert(sizeof(geneve_option_header) == 4, "geneve_option_header size is not 4 bytes");
+
+	/// @class GeneveOption
+	/// A non-owning view of a GENEVE option. The view is invalidated when its underlying data is destroyed or moved,
+	/// including when options are added to or removed from the containing GeneveLayer
+	class GeneveOption
+	{
+		friend class GeneveLayer;
+		friend class GeneveOptionIterator;
+		friend class GeneveOptionRange;
+
+	private:
+		geneve_option_header* m_Data;
+
+		explicit GeneveOption(geneve_option_header& optionData) : m_Data(&optionData)
+		{}
+
+		static bool canAssign(const uint8_t* optionRawData, size_t optionDataLen);
+
+	public:
+		/// @return Option class in host byte order
+		uint16_t getOptionClass() const;
+
+		/// @return The 7-bit option type without the critical bit
+		uint8_t getType() const;
+
+		/// @return True if the option is critical
+		bool isCritical() const;
+
+		/// @return Option data length in bytes
+		size_t getDataSize() const;
+
+		/// @return Total option size including its 4-byte header
+		size_t getTotalSize() const;
+
+		/// @return A pointer to the option data
+		uint8_t* getData() const;
+
+		/// @return A pointer to the option header
+		uint8_t* getRecordBasePtr() const
+		{
+			return reinterpret_cast<uint8_t*>(m_Data);
+		}
+	};
+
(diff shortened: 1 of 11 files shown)


Packet++/header/ProtocolType.h
```diff
@@ -259,6 +259,9 @@ namespace pcpp
 	/// MySQL protocol
 	const ProtocolType MySQL = 63;
 
+	/// GENEVE protocol
+	const ProtocolType Geneve = 65;
```
seladb, line 263: Can you move it below `QUICv1` to keep the numerical order?

Packet++/header/GeneveLayer.h
seladb, line 1: Can you please update the `README.md` file and all of its translations?

Packet++/header/GeneveLayer.h
```diff
@@ -0,0 +1,451 @@
...
+			return (static_cast<uint32_t>(vni[0]) << 16) | (static_cast<uint32_t>(vni[1]) << 8) | vni[2];
+		}
+
+		/// @param[in] value The 24-bit virtual network identifier
+		void setVNI(uint32_t value)
+		{
+			vni[0] = static_cast<uint8_t>((value >> 16) & 0xff);
+			vni[1] = static_cast<uint8_t>((value >> 8) & 0xff);
+			vni[2] = static_cast<uint8_t>(value & 0xff);
+		}
```
seladb, line 81: We usually keep the struct simple and keep these methods only on the layer class which is the main interface for the protocol. Moreover, we can keep the struct private inside `GeneveLayer`.

Same goes for `geneve_option_header` that we can keep as private inside `GeneveOption` and move the struct's methods to the class.

Packet++/src/GeneveLayer.cpp
```diff
@@ -0,0 +1,402 @@
...
+	}
+
+	void GeneveLayer::setVNI(uint32_t vni)
+	{
+		getGeneveHeader()->setVNI(vni);
+	}
+
+	uint16_t GeneveLayer::getProtocolType() const
+	{
+		return be16toh(getGeneveHeader()->protocolType);
```
seladb, line 209: We have wrapper methods in `SystemUtils.h` so we don't need to include `"EndianPortable.h"` directly

Packet++/src/GeneveLayer.cpp
```diff
@@ -0,0 +1,402 @@
...
+
+	void GeneveLayer::setProtocolType(uint16_t protocolType)
+	{
+		getGeneveHeader()->protocolType = htobe16(protocolType);
+	}
+
+	size_t GeneveLayer::getOptionsLength() const
+	{
+		if (m_Data == nullptr || m_DataLen < sizeof(geneve_header))
+			return 0;
```
seladb, line 220: This check is probably not needed because PcapPlusPlus or users shouldn't be able to generate a malformed layer.

Ditto in `getHeaderLen()` and in `getOptions()`

Packet++/header/GeneveLayer.h
```diff
@@ -0,0 +1,451 @@
...
+		bool operator!=(const GeneveOptionIterator& other) const
+		{
+			return !operator==(other);
+		}
+	};
+
+	/// @class GeneveOptionRange
+	/// A non-owning range of GENEVE options. The range and all iterators obtained from it are invalidated when the
+	/// containing GeneveLayer is modified or destroyed
+	class GeneveOptionRange
```
seladb, line 271: The pattern we usually use in PcapPlusPlus is `getFirstOption()` and `getNextOption(option)` directly on the layer. We usually also expose `getOption(type)` on the layer. This approach requires much less code (which makes it simpler) and achieves the same goals

Packet++/header/GeneveLayer.h
```diff
@@ -0,0 +1,451 @@
...
+	public:
+		/// Construct a GENEVE option builder
+		/// @param[in] optionClass Option namespace assigned by IANA
+		/// @param[in] optionType The 7-bit option type
+		/// @param[in] optionData A read-only buffer containing option data
+		/// @param[in] optionDataLen Option data length in bytes. The maximum supported length is 124 bytes
+		/// @param[in] critical Set the option critical bit
+		GeneveOptionBuilder(uint16_t optionClass, uint8_t optionType, const uint8_t* optionData, uint8_t optionDataLen,
+		                    bool critical = false)
+		    : TLVRecordBuilder(optionType, optionData, optionDataLen), m_OptionClass(optionClass), m_Critical(critical)
```
seladb, line 332: I'd vote for a "builder" class only if we have multiple c'tors with different types of data. Currently there's only one c'tor that accepts a byte array. If we plan to keep it like this we can probably move these parameters to `addOption()` and build it there directly

Packet++/src/GeneveLayer.cpp
```diff
@@ -0,0 +1,402 @@
...
+		if (be16toh(header->protocolType) < 0x0600)
+			return false;
+
+		auto optionsLength = header->getOptionsLength();
+		if (optionsLength > dataLen - sizeof(geneve_header))
+			return false;
+
+		const uint8_t* option = data + sizeof(geneve_header);
+		size_t remaining = optionsLength;
+		while (remaining > 0)
```
seladb, line 182: I don't think we need to parse all the options inside of `isDataValid()` which is part of the fast path. Instead, the methods to parse and return the options should take into account that the data might be malformed

Packet++/header/GeneveLayer.h
```diff
@@ -0,0 +1,451 @@
...
+		/// @param[in] optionClass Option class in host byte order
+		/// @param[in] optionType The 7-bit option type
+		/// @return True if an option was found and removed
+		bool removeOption(uint16_t optionClass, uint8_t optionType);
+
+		/// Remove all options
+		/// @return True if all options were removed
+		bool removeAllOptions();
+
+		/// Parse the encapsulated protocol according to the Protocol Type field
```
seladb, line 431: I'd mention which layers we support parsing as the next layer

Tests/Packet++Test/Tests/GeneveTests.cpp
seladb, line 1: This tests is written in an older style that we don't use anymore. I'd propose a few changes:
1. Separate the parsing, creation and update to different tests
2. For the parsing test, get a sample pcap with real packet(s), create `.dat` files for each packet using Wireshark's "Copy as a hex stream" (see the screenshot), read them inside the test and make sure the packet is parsed correctly. You can see other tests as references, try to look at newer tests that were written more recently
3. We use a "sub-test" convention to separate different things in the same test. You can look at `QuicTests.cpp` as an example
4. Try to make sure you cover as much as the code as possible. You can look at the Codecov report to see which lines and use-cases are covered

Image

alacrity-aya:
@seladb Thanks for the review! I'm a little tied up at the moment, but I'll update the PR as soon as I have a free slot over the next few days.

PR #2243, opened 2026-08-29
Fix Heap-buffer-overflow READ 4 · pcpp::GREv1Layer::getAcknowledgmentNum

egecetin:
Fixes ossfuzz issue https://oss-fuzz.com/testcase-detail/6042318394949632


seladb:
@egecetin CI fails on many platforms...

seladb (approved):

PR #2244, opened 2026-08-30
inform cppcheck about pcpp macros, fix newly reported cppcheck defects

cconverse711:
This PR is spun off of #2231

### DIFF
--- a/cppcheck-pcpp.cfg
+++ b/cppcheck-pcpp.cfg
@@ -1,4 +1,0 @@
-<?xml version="1.0"?>
-<def format="2">
-  <define name="PCPP_DEPRECATED(msg)" value=""/>
-</def>
--- a/cppcheck-pcpp.cfg
+++ b/cppcheck-pcpp.cfg
@@ -1,0 +1,4 @@
+<?xml version="1.0"?>
+<def format="2">
+  <define name="PCPP_DEPRECATED(msg)" value=""/>
+</def>
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,7 +36,7 @@ repos:
         args: ["--style=file", "-i"]  # Use the .clang-format file for configuration and apply all fixes
         files: ^(Common\+\+|Packet\+\+|Pcap\+\+|Tests|Examples)/.*\.(cpp|h)$
       - id: cppcheck
-        args: ["--std=c++14", "--language=c++", "--suppressions-list=cppcheckSuppressions.txt", "--inline-suppr", "--force"]
+        args: ["--std=c++14", "--language=c++", "--suppressions-list=cppcheckSuppressions.txt", "--inline-suppr", "--force", "--library=cppcheck-pcpp.cfg"]
         exclude: ^3rdParty/json
   - repo: https://github.com/BlankSpruce/gersemi
     rev: 0.26.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,7 +36,8 @@ repos:
         args: ["--style=file", "-i"]  # Use the .clang-format file for configuration and apply all fixes
         files: ^(Common\+\+|Packet\+\+|Pcap\+\+|Tests|Examples)/.*\.(cpp|h)$
       - id: cppcheck
-        args: ["--std=c++14", "--language=c++", "--suppressions-list=cppcheckSuppressions.txt", "--inline-suppr", "--force", "--library=cppcheck-pcpp.cfg"]
+        # TODO: Add -I for each directory that contains header files that are included by the source files, for more thorough checking.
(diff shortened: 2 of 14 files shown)


cppcheck-pcpp.cfg
seladb, line 1: Why is this needed? 🤔 
Can't `cppcheck` see the definition of the macro?

cconverse711: Ok so I realized that the `--include` approach that I tried in #2231 was not quite right; that's intended to mirror gcc's `--include`, which is for including special files first, such as PCH's. That's why I went with the "library" approach here. But I think you're right, cppcheck _should_ be able to see the macro definitions, but it's not because its preprocessor is not locating the DeprecationUtils.h file as it should, because cppcheck needs to be told about the include paths with `-I`. In the end I think that's what was causing certain cppcheck defects to not be reported: cppcheck was choking on the macros and giving up silently because of `unknownMacro:*` in cppcheckSuppressions.txt

cconverse711:
Rather than using `--library` with a custom cfg file, I've gone ahead with the `-I` approach to ensure cppcheck knows about the macros in `DeprecationUtils.h`. In addition to the single `-I` added in this PR, also I tried adding `-I` for each of the include paths that _should_ be used for thorough analysis, but tons more defects were reported, so those should be taken care of incrementally in separate PRs. 

Additionally, I should note that when I tried adding`-ITests/PcppTestFramework` so that cppcheck would know about the test framework macros, cppcheck got hung. I suspect some bug in cppcheck; might be resolved in a newer version of cppcheck.

seladb:
> Rather than using `--library` with a custom cfg file, I've gone ahead with the `-I` approach to ensure cppcheck knows about the macros in `DeprecationUtils.h`. In addition to the single `-I` added in this PR, also I tried adding `-I` for each of the include paths that _should_ be used for thorough analysis, but tons more defects were reported, so those should be taken care of incrementally in separate PRs.
> 
> Additionally, I should note that when I tried adding`-ITests/PcppTestFramework` so that cppcheck would know about the test framework macros, cppcheck got hung. I suspect some bug in cppcheck; might be resolved in a newer version of cppcheck.

Something still doesn't make sense to me 🤔 Why do we need to explicitly specify the include path of `Common/header`? There are other includes inside of `Pcap++` and `Packet++` to headers in `Common++` which are not `DeprecationUtils.h` and cppcheck didn't complain. Why does it complain about the macros in `DeprecationUtils.h`? Maybe because they're macros? 🤔

cconverse711:
Cppcheck's preprocessor is unable to find the header files unless given the `-I` flag. I believe cppcheck can find headers that are in the same directory as a given cpp file that includes them, however due to this repo being laid out with a separate directory per library and separate header/src subdirectories, it cannot. Cppcheck is not complaining about this fact due to the `missingInclude:*` line in `cppcheckSuppressions.txt`

As for the macros in `DeprecationUtils.h`, the `unknownMacros:*` suppression seems to have allowed cppcheck to stay happy until I added those new deprecations in #2231. I suspect that the `unknownMacros` suppression is enough to allow cppheck to ignore other uses of the `PCPP_DEPRECATED` macro because they were applied in fairly typical contexts (i.e., as function attributes). Whereas for the #2231, the problematic application of the macro in `enum OsiModelLayer` caused a `syntaxError` (i.e. not just `unknownMacro` warning) because the `PCPP_DEPRECATED` macro was applied to an enumerator within the enum, which I'd consider somewhat more exotic (and thus perhaps not properly handled/fully supported by cppcheck even when ignoring unknown macros).

seladb (approved):

seladb:
> Cppcheck's preprocessor is unable to find the header files unless given the `-I` flag. I believe cppcheck can find headers that are in the same directory as a given cpp file that includes them, however due to this repo being laid out with a separate directory per library and separate header/src subdirectories, it cannot. Cppcheck is not complaining about this fact due to the `missingInclude:*` line in `cppcheckSuppressions.txt`
> 
> As for the macros in `DeprecationUtils.h`, the `unknownMacros:*` suppression seems to have allowed cppcheck to stay happy until I added those new deprecations in #2231. I suspect that the `unknownMacros` suppression is enough to allow cppheck to ignore other uses of the `PCPP_DEPRECATED` macro because they were applied in fairly typical contexts (i.e., as function attributes). Whereas for the #2231, the problematic application of the macro in `enum OsiModelLayer` caused a `syntaxError` (i.e. not just `unknownMacro` warning) because the `PCPP_DEPRECATED` macro was applied to an enumerator within the enum, which I'd consider somewhat more exotic (and thus perhaps not properly handled/fully supported by cppcheck even when ignoring unknown macros).

Thank you @cconverse711 for the explanation. To be honest, I'm not super familiar of cppcheck details, but what you say make sense 👍

seladb:
Thank you @cconverse711 for working on this, much appreciated! 🙏

PR #2249, opened 2026-09-02
Fix FiveTupleSplitter file collision on TCP session reuse

### LINKED ISSUE #2248: Bug: FiveTupleSplitter causes file truncation/corruption on TCP session reuse (PcapSplitter)
### Bug description
A critical file truncation and data corruption bug exists in `FiveTupleSplitter` when a TCP connection is reused (i.e., a new `SYN` packet arrives for an already tracked 5-tuple hash). 

While `FiveTupleSplitter` correctly attempts to isolate the new logical session by allocating a brand-new file number (`fileNum`), `FiveTupleSplitter::getFileName` completely ignores this `fileNumber` parameter. As a result, the new session generates the exact same output path string as the previous session. Because `main.cpp` caches and tracks writers by their integer `fileNum` rather than their file path, it treats this as a brand-new file stream and opens it without the append flag, truncating the existing file or causing an uncoordinated write race condition between two active writer handles.

#### 1. File Number Allocation on Session Reuse (`ConnectionSplitters.cpp`)
When a fresh `SYN` arrives for a 5-tuple hash that has been seen before, `FiveTupleSplitter::getFileNumber` deliberately assigns a **new file number** to isolate the logical session:
```cpp
if (isSyn && m_TcpFlowTable.find(hash) != m_TcpFlowTable.end() && m_TcpFlowTable[hash] == false)
{
    m_FlowTable[hash] = getNextFileNumber(filesToClose);   // <-- Allocates a NEW file number
}
```

#### 2. The Filename Disconnect (`ConnectionSplitters.cpp`)
Right below this logic, `FiveTupleSplitter::getFileName` overrides the base class implementation but completely ignores the `fileNumber` parameter passed into it:
```cpp
std::string getFileName(pcpp::Packet& packet, const std::string& outputPcapBasePath, int fileNumber)
{
    // ...
    sstream << "connection-";
    if (packet.isPacketOfType(pcpp::TCP)) {
        // ...
        // Filename is derived strictly from the packet's IP/Port values:
        updateStringStream(sstream, getSrcIPString(packet), srcPort, getDstIPString(packet), dstPort);
        return outputPcapBasePath + sstream.str();   // <-- fileNumber parameter is UNUSED
    }
    // ...
}
```
Because `fileNumber` is dropped, the new session produces the **identical filename string** as the older session.

#### 3. The Truncation / Race Condition (`main.cpp`)
In `main.cpp`, the active writer cache map (`outputFiles`) is keyed by the integer `fileNum`, not the filepath string. When the newly allocated `fileNum` is evaluated, it goes down the fresh initialization branch:
```cpp
// Since fileNum is a newly generated integer, it won't be found in the map
if (outputFiles.find(fileNum) == outputFiles.end()) 
{
    std::string fileName = splitter->getFileName(parsedPacket, outputPcapFileName, fileNum) + outputFileExtension;

    if (isReaderPcapng) {
        outputFiles[fileNum].reset(new pcpp::PcapNgFileWriterDevice(fileName));
    } else {
        outputFiles[fileNum].reset(new pcpp::PcapFileWriterDevice(fileName, rawPacket.getLinkLayerType()));
    }

    // CRITICAL: Plain open — no append flag — on a path that already holds valid session data!
    if (!outputFiles[fileNum]->open()) 
        break;
}
```

### Impact & Consequences
Depending on the state of the LRU cache mechanism, this results in two fatal outcomes:
1. **File Truncation:** If the old `fileNum` was already evicted/closed by the LRU mechanism, the new writer blindly triggers a standard file open on the existing path, **completely wiping out (truncating to 0 bytes)** the pcap data captured from the previous session.
2. **Write Race / Corruption:** If the old `fileNum` is still open and active in the `outputFiles` map, two distinct `IFileWriterDevice` objects will simultaneously hold active, uncoordinated write handles to the exact same file on disk. This breaks the sequential layout of the PCAP/PCAPNG format and leaves the file corrupted and unreadable.

### Suggested Fix
Modify `FiveTupleSplitter::getFileName` to include the unique `fileNumber` in the file string to guarantee uniqueness across connection reuses (similar to how other splitters behave), or append a counter/index when the file number changes for the same 5-tuple connection state.

PcapPlusPlus master branch

_No response_

Linux (Alpine 3.22)

_No response_

### Compiler version

g++ (Alpine 14.2.0) 14.2.0

_No response_

rshayer95:
## Summary

Fixes #2248.

`FiveTupleSplitter::getFileNumber()` deliberately allocates a new `fileNumber` when a fresh SYN arrives on a 5-tuple hash that's already been seen, to isolate the new logical session. But `FiveTupleSplitter::getFileName()` ignored the `fileNumber` parameter entirely, deriving the output path purely from the packet's own IP/port values — so the new session produced the exact same path string as the old one.

`main.cpp`'s writer cache (`outputFiles`) is keyed by `fileNumber`, not by path. On collision, it treats the new `fileNumber` as never-seen and opens its writer fresh (no append flag) against a path an earlier, still-tracked writer object may still be using — two independent file handles on one inode, with whichever one flushes last silently discarding the other's data.

## Fix

Fold `fileNumber` into the generated path for both the TCP and UDP branches, matching how every other splitter in this file avoids the problem by keeping `fileNumber` and filename 1:1. UDP's `getFileNumber()` never reallocates on hash reuse today, so the collision isn't reachable there, but `fileNumber` is included anyway for naming consistency and to stay correct if that ever changes.

## Verification

Built both an unpatched and patched copy (Alpine 3.22, g++ 14.2.0) and ran both against:

1. **A synthetic capture with a deliberately reused 5-tuple** (same 5-tuple, two SYNs 10 unrelated connections apart):
   - Unpatched: reports "written ... to 12 files", only 11 exist on disk. The file for the reused 5-tuple contains only session 1's 4 packets — session 2's data was silently discarded.
   - Patched: 12 files claimed, 12 exist. The reused 5-tuple now produces two separate files (`...-0001.pcap`, `...-0012.pcap`), each complete and valid (48/48 packets accounted for across all 12 files, 0 read errors).

2. **A real 126,502-packet / 95-connection capture with no reused 5-tuples** (regression check): 95/95 files clean, all packets accounted for, identical before and after the patch.

## Notes

This is one of two independent resource-ceiling issues found while investigating the corruption reports in #2248 (the other being fd-exhaustion in the underlying pcapng writer on very high connection counts, unrelated to this code path). This PR addresses the file-collision mechanism specifically, which is the one with a deterministic, reproducible root cause and fix.

### DIFF
--- a/Examples/PcapSplitter/ConnectionSplitters.h
+++ b/Examples/PcapSplitter/ConnectionSplitters.h
@@ -189,7 +189,20 @@ public:
 				{
 					updateStringStream(sstream, getSrcIPString(packet), srcPort, getDstIPString(packet), dstPort);
 				}
-				return outputPcapBasePath + sstream.str();
+
+				// Append the file number so a reused 5-tuple (a new SYN on the same
+				// ports as an earlier, already-tracked session) can't collide with the
+				// previous session's output file. getFileNumber() above deliberately
+				// allocates a new fileNumber to isolate the new session, but this
+				// method previously ignored that parameter entirely, so both sessions
+				// produced the exact same path string. main.cpp's writer cache is
+				// keyed by fileNumber, not by path, so on collision it would open the
+				// second session's file fresh (no append flag) against a path the
+				// first session's writer might still be using -- truncating or
+				// corrupting it. See seladb/PcapPlusPlus#2248.
+				std::ostringstream numStream;
+				numStream << "-" << std::setw(4) << std::setfill('0') << fileNumber;
+				return outputPcapBasePath + sstream.str() + numStream.str();
 			}
 		}
 		else if (packet.isPacketOfType(pcpp::UDP))
@@ -210,7 +223,15 @@ public:
 				std::string secondIP = (srcPort < dstPort) ? getSrcIPString(packet) : getDstIPString(packet);
 
 				updateStringStream(sstream, firstIP, firstPort, secondIP, secondPort);
-				return outputPcapBasePath + sstream.str();
+
+				// getFileNumber() never reallocates a new fileNumber for a UDP flow
+				// that reuses an existing hash (only the TCP/SYN path above does), so
+				// this isn't reachable as a collision today -- fileNumber is folded in
+				// here anyway to keep naming consistent with the TCP branch and to
+				// stay correct if that ever changes.
+				std::ostringstream numStream;
+				numStream << "-" << std::setw(4) << std::setfill('0') << fileNumber;
+				return outputPcapBasePath + sstream.str() + numStream.str();
 			}
 		}
 


seladb:
@rshayer95 I found the commit when I made this change: https://github.com/seladb/PcapPlusPlus/commit/63791006db94bc07aeb88f55a649d80dc121cbc2

To be honest, I don't remember why I made this change - in a 5-tuple splitter, it makes sense to include all packets of the same 5-tuple in one file...

In any case, I'm not sure your fix is the right one because it add a number to each file generated from now on. Maybe we should just revert my PR and let all packets to be written to the same file?

rshayer95:
> @rshayer95 I found the commit when I made this change: [6379100](https://github.com/seladb/PcapPlusPlus/commit/63791006db94bc07aeb88f55a649d80dc121cbc2)
> 
> To be honest, I don't remember why I made this change - in a 5-tuple splitter, it makes sense to include all packets of the same 5-tuple in one file...
> 
> In any case, I'm not sure your fix is the right one because it add a number to each file generated from now on. Maybe we should just revert my PR and let all packets to be written to the same file?

@seladb Thanks for finding the original commit.

Looking at it, the intent was clearly separate files per session — the comment says as much. The filename side just never got updated to match, which is where the collision comes from.

Fair point on adding the number to every file — that changes output for everyone, including captures with no reuse.

On reverting: it fixes the collision but merges two distinct sessions into one file with no indication they're separate. Silent correctness issue rather than a visible one.

We can have a middle ground by implementing suffixes only when an actual collision occurs. The first session can keep its current filename while subsequent ones are assigned numbers. This way captures without reuse remain unchanged and important sessions stay separated where necessary.

Happy to rework it that way, or revert if you had prefer simpler.

seladb:
> We can have a middle ground by implementing suffixes only when an actual collision occurs. The first session can keep its current filename while subsequent ones are assigned numbers. This way captures without reuse remain unchanged and important sessions stay separated where necessary.

Sure, I think that could work 👍 
When making this change, you can add tests for it. We have tests written in Python for the examples: https://github.com/seladb/PcapPlusPlus/blob/master/Tests/ExamplesTest/tests/test_pcapsplitter.py

rshayer95:
@seladb Updated to suffix only on an actual collision (tracking which fileNumber first claims a given base filename, appending the fileNumber only when a different one collides with it). Captures with no reused 5-tuples keep unchanged filenames.

Also added a test (`test_split_by_connection_with_reused_five_tuple`) with a small fixture capture containing a deliberately reused 5-tuple — verifies no file count collapse, no packet loss, the first session keeps its unsuffixed name, and only the colliding session gets suffixed. Full `test_pcapsplitter.py` suite (17 tests) passes.

Tests/ExamplesTest/tests/test_pcapsplitter.py
```diff
@@ -268,6 +268,48 @@ def test_split_by_connection(self, tmpdir):
...
+
+        # exactly one base name repeats: the first session keeps it unsuffixed,
+        # the second (colliding) session gets a "-<fileNumber>" suffix
+        reused_base = [n for n in base_names if "40001" in n]
+        assert len(reused_base) == 2
+        unsuffixed = [n for n in reused_base if not n.split("-")[-1].isdigit()]
+        suffixed = [n for n in reused_base if n.split("-")[-1].isdigit()]
+        assert len(unsuffixed) == 1
+        assert len(suffixed) == 1
+        assert suffixed[0] == unsuffixed[0] + "-" + suffixed[0].split("-")[-1]
```
seladb, line 307: Won't it be easier to assert the file names? These file names shouldn't change between runs and it'd make the test shorter and easier to read

rshayer95:
@seladb Done — asserts the exact filename set now instead of the suffix heuristic. Shorter and matches what's actually deterministic about the output. All 17 tests in test_pcapsplitter.py still pass.

seladb (approved):

seladb:
Thank you @rshayer95 for working on this fix, much appreciated 🙏

PR #2255, opened 2026-09-11
cppcheck fixes for constParameter

cconverse711:
another set of cppcheck fixes as a follow-on to https://github.com/seladb/PcapPlusPlus/pull/2244

### DIFF
--- a/Pcap++/header/PcapLiveDevice.h
+++ b/Pcap++/header/PcapLiveDevice.h
@@ -717,7 +717,7 @@ namespace pcpp
 		/// @return True if raw packet matches the filter or false otherwise
 		/// @deprecated This method is deprecated, use GeneralFilter::matches(...) method directly.
 		PCPP_DEPRECATED("Prefer GeneralFilter::matches(...) method directly.")
-		static bool matchPacketWithFilter(GeneralFilter& filter, RawPacket* rawPacket);
+		static bool matchPacketWithFilter(const GeneralFilter& filter, RawPacket* rawPacket);
 
 	protected:
 		/// @brief Called before starting a capture to prepare the device for capturing packets.
--- a/Pcap++/src/DpdkDevice.cpp
+++ b/Pcap++/src/DpdkDevice.cpp
@@ -1281,7 +1281,7 @@ namespace pcpp
 		return packetSent;
 	}
 
-	bool DpdkDevice::sendPacket(Packet& packet, uint16_t txQueueId, bool useTxBuffer)
+	bool DpdkDevice::sendPacket(const Packet& packet, uint16_t txQueueId, bool useTxBuffer)
 	{
 		return sendPacket(*(packet.getRawPacket()), txQueueId);
 	}
--- a/Pcap++/src/KniDevice.cpp
+++ b/Pcap++/src/KniDevice.cpp
@@ -773,7 +773,7 @@ namespace pcpp
 		return sent;
 	}
 
-	bool KniDevice::sendPacket(Packet& packet)
+	bool KniDevice::sendPacket(const Packet& packet)
 	{
 		return sendPacket(*packet.getRawPacket());
 	}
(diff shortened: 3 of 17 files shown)


seladb (commented):
Adding `const` to methods that are exposed in the public API is a potential breaking change (although it probably won't break anything for most users). Maybe we can avoid it in the public API and keep it in anything that is not public and in the examples/tutorials?

cconverse711:
Hmm yes I suppose you're right that it would break ABI.  Ultimately these changes would help with const correctness so I think it would be beneficial overall. What if we overloaded the public methods with const-parameter versions, and deprecated the non-const-parameter versions?

seladb:
@cconverse711 I guess we can, but maybe it'd be easier to just avoid it for public methods? I'm not sure it's worth the effort

Pcap++/src/PcapLiveDevice.cpp
```diff
@@ -283,7 +283,7 @@ namespace pcpp
 #endif
 	}
 
-	static void setTimestampProvider(internal::PcapHandle& pcap,
+	static void setTimestampProvider(const internal::PcapHandle& pcap,
 	                                 const PcapLiveDevice::TimestampProvider timestampProvider)
```
Dimi1010, line 287: IIRC, `PcapHandle` was deliberately kept non-const due to modifying the underlying handle state.

Pcap++/header/PcapLiveDevice.h
```diff
@@ -717,7 +717,7 @@ namespace pcpp
 		/// @return True if raw packet matches the filter or false otherwise
 		/// @deprecated This method is deprecated, use GeneralFilter::matches(...) method directly.
 		PCPP_DEPRECATED("Prefer GeneralFilter::matches(...) method directly.")
-		static bool matchPacketWithFilter(GeneralFilter& filter, RawPacket* rawPacket);
+		static bool matchPacketWithFilter(const GeneralFilter& filter, RawPacket* rawPacket);
```
Dimi1010, line 720: I don't think it worth to change the signature of a deprecated method?

Pcap++/src/KniDevice.cpp
```diff
@@ -773,7 +773,7 @@ namespace pcpp
 		return sent;
 	}
 
-	bool KniDevice::sendPacket(Packet& packet)
+	bool KniDevice::sendPacket(const Packet& packet)
 	{
 		return sendPacket(*packet.getRawPacket());
 	}
```
Dimi1010, line 779: IIRC, this is deliberately kept non-const, due to the KniDevice packet implementation.

It is `getRawPacket` that has the defect of returning non-const reference for the const Packet::getRawPacket. And changing that call breaks a lot of API.

Pcap++/src/DpdkDevice.cpp
```diff
@@ -1281,7 +1281,7 @@ namespace pcpp
 		return packetSent;
 	}
 
-	bool DpdkDevice::sendPacket(Packet& packet, uint16_t txQueueId, bool useTxBuffer)
+	bool DpdkDevice::sendPacket(const Packet& packet, uint16_t txQueueId, bool useTxBuffer)
```
Dimi1010, line 1284: ditto: Deliberately kept non-const due to Dpdk implementation (same as Kni).

Dimi1010 (commented):
Can we split this PR a bit, as it changes const-ness of the API on several different levels?

I would categorize the changes into the following:
- Examples fixes - It has the lowest severity, as we don't keep API compatibility guarantees there.
- Packet++ layers - General considerations about API compatibility.
- PcapFilter - General considerations about API compatibility.
- Pcap/Dpdk/Kni Device - Those have quirks as I have noted in the comments, and need special attention with regards to implementation review to change the const-ness on the API, and may require more changes.

In general I don't recommend changing the const-ness of the Device classes in a general PR and have more specialized PRs for the devices if needed. I think most of the changes in those classes were kept deliberately non-const tho.

PR #2256, opened 2026-09-11
clean up pre-commit dependency setup

cconverse711:
This PR optimizes the pre-commit step:
- Builds cppcheck with proper optimizations to improve runtime
- Caches cppcheck artifacts to avoid unnecessary rebuilds

And performs some cleanup and minor tweaks:
- Separates linting into new workflow
- Updates cppcheck links
- Adds clang-format as pre-commit hook additional dependency to be auto-installed
- Moves clang-tidy steps into separate linting job


.pre-commit-config.yaml
```diff
@@ -35,10 +29,12 @@ repos:
       - id: clang-format
         args: ["--style=file", "-i"]  # Use the .clang-format file for configuration and apply all fixes
         files: ^(Common\+\+|Packet\+\+|Pcap\+\+|Tests|Examples)/.*\.(cpp|h)$
+        additional_dependencies: [clang-format==19.1.6]
       - id: cppcheck
         # TODO: Add -I for each directory that contains header files that are included by the source files, for more thorough checking.
         args: ["--std=c++14", "--language=c++", "--suppressions-list=cppcheckSuppressions.txt", "--inline-suppr", "--force", "-ICommon++/header"]
         exclude: ^3rdParty/json
+        additional_dependencies: [cppcheck-py==2.9]
```
seladb: Is this the project you're maintaining? https://github.com/cconverse711/cppcheck-wheel

I'd rather not pull in `cppcheck-py` here since it's a new package you maintain yourself, and it'd end up executing on every contributor's machine and CI job. For a library that gets embedded downstream, I'd like to keep cppcheck's provenance tied to the upstream project. Could we revert to installing it the previous way?

cconverse711: Yep that's mine, the goal is to make it easy to install different to avoid conflicts from having different cppcheck versions needed across different projects. But yeah your concern is totally understandable, and using the python package as a listed additional dep is just one way to achieve that. 

Another side effect/secondary goal of using the python package was to reduce the pre-commit job's runtime by avoiding the need to build cppcheck from source each time, and by ensuring that cppcheck is built with proper optimizations. I'll tweak the build args to accomplish the latter and set up caching to accomplish the former.

cconverse711:
You'll see I broke out the pre-commit job into a separate linting workflow to allow for caching the cppcheck artifacts. I left the jobs using the seladb/alpine317 container, though I'm not really sure why/if that's actually necessary.

For further speedup, additional caching could be applied to the pip cache and the pre-commit cache as well, but caching the pre-commit cache would require resolving this warning: `[WARNING] The 'rev' field of repo 'https://github.com/crate-ci/typos' appears to be a mutable reference (moving tag / branch).  Mutable references are never updated after first install and are not supported.  See https://pre-commit.com/#using-the-latest-version-for-a-repository for more details.  Hint:`pre-commit autoupdate` often fixes this.`  Alternatively/additionally, [prek](https://github.com/j178/prek) could be used as a drop-in replacement for pre-commit.

seladb:
> You'll see I broke out the pre-commit job into a separate linting workflow to allow for caching the cppcheck artifacts. I left the jobs using the seladb/alpine317 container, though I'm not really sure why/if that's actually necessary.
> 
> For further speedup, additional caching could be applied to the pip cache and the pre-commit cache as well, but caching the pre-commit cache would require resolving this warning: `[WARNING] The 'rev' field of repo 'https://github.com/crate-ci/typos' appears to be a mutable reference (moving tag / branch). Mutable references are never updated after first install and are not supported. See https://pre-commit.com/#using-the-latest-version-for-a-repository for more details. Hint:`pre-commit autoupdate` often fixes this.` Alternatively/additionally, [prek](https://github.com/j178/prek) could be used as a drop-in replacement for pre-commit.

Why do we need to split it into a separate workflow rather than add the cppcheck caching inside the existing job? 🤔 
All jobs are running in parallel anyway, it's like this job is the bottleneck of our current CI pipeline

cconverse711:
I split it into a separate workflow so that the workflow file itself could be hashed and act as (part of) the cache key, to be safe and avoid invalid cache hits, such as in the case where the container was changed, or cmake args were modified.  If we didn't want to hash on the workflow file itself, we'd probably end up with a cache key like `cppcheck-${{ runner.os }}-${{ env.CPPCHECK_VERSION }}` , which doesn't protect us from those invalid cache hit scenarios that I just mentioned.

We could certainly take the same approach in the `build_and_test.yml` workflow file by hashing that file to act as the cache key, but that would result in cache misses any time the build_and_test.yml file is modified, which is probably much more frequently than just the lint-only workflow file would be modified.  It may also be personal preference, but I think it makes more sense to separate linting out from build/test anyway. It's your call, of course. If you prefer to keep it all in build_and_test.yml and just hash on that yml (or something else you have in mind) I can adjust it.

cconverse711:
Also I understand that this update would be in conflict with #2208. I generally think caching cppcheck is preferable to baking it into the image because it makes upgrading cppcheck as simple as modifying the CPPCHECK_VERSION environment variable.

Dimi1010:
I think its fine to have the linting separated into its own workflow.

As for the conflict with #2208. I agree that the current caching solution is preferable as it would result less maintenance needed in the long run, so no issues there IMO.

seladb:
@cconverse711 I think we can have the container image hash as part of the cache key so if the container changes the cache will be invalidated

cconverse711:
You mean instead of hashing the workflow file, or in addition to it? Since the workflow file is being hashed, the container hash is implicitly part of the hash key now.

seladb:
What I meant was to use the image hash instead of just `seladb/aplint317`, like `seladb/alpine317@sha256:abc123...` and have this as part of the key - so whenever the image hash changes it will invalidate the cache. To be honest, we rarely change the image so it shouldn't happen much

Dimi1010:
> What I meant was to use the image hash instead of just `seladb/aplint317`, like `seladb/alpine317@sha256:abc123...` and have this as part of the key - so whenever the image hash changes it will invalidate the cache. To be honest, we rarely change the image so it shouldn't happen much

This seems more inconvinient than having an extra workflow tbh, even if we do change the containers rarely.

seladb:
@Dimi1010 why is it inconvenient? 🤔 
The jobs run in parallel anyway so it's easier to have as many of them in the same file

Dimi1010:
> @Dimi1010 why is it inconvenient? 🤔 
> The jobs run in parallel anyway so it's easier to have as many of them in the same file

@seladb Having to deal with the sha hashes would be inconvinient. I suppose tags would work too, but isn't the point that we use the latest image, so we don't have to worry about that?

Yes, the jobs run in parallel, so it is same either way from that perspective. But I am of the opinion that having the workflows separated by purpose would be better than adding extra complexity while trying to keep them all in the same file.

The build/test workflow config is long enough as it is, so I don't think splitting the linting part into a separate file would hurt. (And arguably keeps the workflows more focused)

Overall, I am fine with either, but I would prefer not having to deal with the image hashes, if the alternative is just having a separate file.

seladb:
Actually when thinking about it more, maybe the approach in #2208 is better so we don't need to install cppcheck at all in these workflows and have it baked into the image already. We rarely update cppcheck, and when we do we can introduce a new image / a new tag of the image and test it there. It's also easier to read and understand than GitHub Actions caching

PR #2259, opened 2026-09-16
Add packet serialization API

seladb:
Add a generic serialization framework to Packet++ and support serializing packets and layers into structured formats.

What's included:
- Add a generic `ISerializer` interface and serialization scopes.
- Add support for serializing collections of fields and nested objects/arrays.
- Add a JSON serializer implementation.
- Add serialization support to Packet and Layer.

This is the first PR, more will follow soon


Common++/header/Serializers.h
```diff
@@ -0,0 +1,280 @@
...
+	struct FieldDescriptor
+	{
+		/// Stable numeric identifier of the field.
+		const uint16_t id;
+
+		/// Name of the field.
+		const std::string name;
+
+		/// Function returning the descriptors of the field's children, or nullptr if the field has no children.
+		std::vector<FieldDescriptor> (*const children)();
```
Dimi1010, line 26: Curious why function pointer instead of std::functon?

Also, maybe we can name it `getChildren` eince it is a method?

Common++/header/Serializers.h
```diff
@@ -0,0 +1,280 @@
...
+	class ObjectScope;
+
+	namespace internal
+	{
+		template <typename T>
+		using EnableIfSignedIntegral = typename std::enable_if<
+		    std::is_integral<T>::value && std::is_signed<T>::value && !std::is_same<T, bool>::value, int>::type;
+		template <typename T>
+		using EnableIfUnsignedIntegral = typename std::enable_if<
+		    std::is_integral<T>::value && std::is_unsigned<T>::value && !std::is_same<T, bool>::value, int>::type;
```
Dimi1010, line 69: nit: `std::enable_if<...>::type` -> `std::enable_if_t<...>`

Common++/header/Serializers.h
```diff
@@ -0,0 +1,280 @@
...
+		}
+
+		virtual void writeHexField(const FieldDescriptor& field, uint64_t value) = 0;
+
+		template <typename T,
+		          typename std::enable_if<std::is_integral<T>::value && std::is_unsigned<T>::value, int>::type = 0>
+		void writeHexField(const FieldDescriptor& field, T value)
+		{
+			writeHexField(field, static_cast<uint64_t>(value));
+		}
```
Dimi1010, line 127: Fun fact: Technically, according to [RFC 8259 Section 2](https://www.rfc-editor.org/info/rfc8259/#section-2), the `json` syntax allows for single field values at the root node.

Not sure if it will be needed in the library, and if it will translate well for other structured formats tho.

Common++/header/Serializers.h
```diff
@@ -0,0 +1,280 @@
...
+		/// @param[in] field The descriptor of the object field
+		ObjectScope(ISerializer* serializer, const FieldDescriptor& field);
+
+		/// Close the object.
+		~ObjectScope() override;
+	};
+
+	/// @class JsonSerializer
+	/// Serializer that writes JSON.
+	class JsonSerializer : public ISerializer
```
Dimi1010, line 243: Do we want to mark this as `final`? I don't think this will require further derived classes?

Common++/header/Serializers.h
```diff
@@ -0,0 +1,280 @@
...
+/// @namespace pcpp
+/// @brief The main namespace for the PcapPlusPlus lib
+namespace pcpp
+{
+	/// @struct FieldDescriptor
+	/// Describes a serializable field.
+	struct FieldDescriptor
+	{
+		/// Stable numeric identifier of the field.
+		const uint16_t id;
```
Dimi1010, line 20: Could you explain why the field ids are needed and how they work? I don't think I see them used in the PR.

Packet++/header/ProtocolType.h
```diff
@@ -267,6 +268,148 @@ namespace pcpp
 
 	/// @}
 
+	/// Convert a protocol type to its string representation.
+	/// @param[in] protocolType The protocol type to convert.
+	/// @return The string representation of the protocol type.
+	inline const char* protocolTypeToString(ProtocolType protocolType)
```
Dimi1010, line 274: I think this should be able to be marked `constexpr`. It should allow direct replacement to the string literal during optimization.

Dimi1010, line 274: @seladb @tigercosmos @clementperon @egecetin 
What are your opinions on inline implementation files ( .ipp / .tpp)? I get the feeling functions such as this (`protocolTypeToString`), unnecessarily bloat the public header, while at the same time still benefit from inlining in the header. Which can be a benefit of the inline implementation files, due to keeping the definition and declaration separate.

Packet++/header/RawPacket.h
```diff
@@ -250,6 +251,237 @@ namespace pcpp
 		LINKTYPE_INVALID = 0xFFFF
 	};
 
+	/// Convert a link layer type to its string representation.
+	/// @param[in] linkLayer The link layer type to convert.
+	/// @return The string representation of the link layer type.
+	inline const char* linkLayerToString(LinkLayerType linkLayer)
```
Dimi1010, line 257: ditto: constexpr
ditto: ipp files

Common++/header/Serializers.h
seladb: Field descriptors are static, per-type metadata, so `getChildren` never needs captured state. A plain function pointer keeps `FieldDescriptor` trivially copyable, pointer-sized for this member, and allocation-free. However, if you insist I can switch to `std::function`

Common++/header/Serializers.h
seladb: Yes, that's why `JsonSerializer` moves these methods to be `protected` rather than `public`. Other serializers may or may not have this restriction...

Common++/header/Serializers.h
seladb: I'm not sure we should do that... if someone doesn't like this specific implementation and wants to tweak it a bit, maybe they can inherit it and change what they don't like. Of course we can make (non breaking) changes in the class itself, but why not give the users this flexibility?

Common++/header/Serializers.h
seladb: The interface should be as generic as possible and allow users to implement their own serializers. I may add XML and YAML in later PRs, but I assume users will have their own serializers later on.

Serializers that implement CBOR, MessagePack, Protocol Buffers, etc might prefer the field names to be IDs rather than strings

Common++/header/Serializers.h
Dimi1010: Ok, we can keep it as function pointer. For the record, Field descriptor shouldn't be trivially copiable because it contains `std::string`.

Packet++/header/ProtocolType.h
seladb: I don't have a strong opinion either way, the code is the same either way, just in different files...

Common++/header/Serializers.h
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2259/commits/bbd6b157aafc078e825b94d9713d18f4f55116e5

Packet++/header/ProtocolType.h
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2259/commits/bbd6b157aafc078e825b94d9713d18f4f55116e5

Packet++/header/RawPacket.h
seladb: Updated in https://github.com/seladb/PcapPlusPlus/pull/2259/commits/bbd6b157aafc078e825b94d9713d18f4f55116e5

Packet++/header/Layer.h
```diff
@@ -559,6 +587,9 @@ namespace pcpp
 		{
 			return data != nullptr && dataLen >= sizeof(T);
 		}
+
+		virtual void serializeLayer(ObjectScope& serializer) const
+		{}
```
Dimi1010, line 592: Should this be documented since its an extension point?

Dimi1010 (approved):
Overall LGTM.

Somewhat concerned about the serializers file tho. If we plan on having many serializers, do we want to have the concrete serializers (json / xml / etc) in separate headers? That way, they can be selected by the users on demand, and aren't included everywhere.

Could also make a subfolder for the whole serialization subsystem too, to organize the headers a bit better.

seladb:
> Somewhat concerned about the serializers file tho. If we plan on having many serializers, do we want to have the concrete serializers (json / xml / etc) in separate headers? That way, they can be selected by the users on demand, and aren't included everywhere.
> 
> Could also make a subfolder for the whole serialization subsystem too, to organize the headers a bit better.

I'll add a few serializers, probably XML and YAML, but I don't think I'll add more. Users can/should create the own serializers. If this feature gets popular enough that people add more and more serializers we can split the header files later

Packet++/header/Layer.h
seladb: Good point, added in https://github.com/seladb/PcapPlusPlus/pull/2259/commits/9737c872eb7b1135c7da73949330b42b645f6859

PR #2261, opened 2026-09-18
Add `serializeLayer()` for Eth, IPv4, IPv6, TCP and UDP layers

seladb:
Adds concrete `serializeLayer()` implementations for the core protocol layers, building on the serialization framework from #2259.

What's included:
- `EthLayer`: source/destination MAC address, EtherType
- `IPv4Layer`: source/destination IP, ID, protocol, total length, fragmentation, options
- `IPv6Layer`: source/destination IP, payload length, fragmentation, next header, extension headers
- `TcpLayer`: ports, sequence number, flags, window size, checksum, options
- `UdpLayer`: ports, checksum

Base branch is `packet-serialization-foundation` (#2259)


.github/workflows/build_and_test.yml
```diff
@@ -3,7 +3,7 @@ on:
   push:
     branches: ["master", "dev"]
   pull_request:
-    branches: ["dev"]
+    branches: ["dev", "packet-serialization-foundation"]
```
seladb, line 6: This is temporary for run CI on stacked branches

PR #2266, opened 2026-09-20
Add `BluetoothHciEventLayer` for parsing Bluetooth HCI *Event* packets

### LINKED ISSUE #1179: Bluetooth HCI packet trace support

smhussa22:
Related Issues: 
Refs #1179 
Refs #1754 

Overall Summary:
Added initial Bluetooth HCI support with a new `BluetoothHciEventLayer` that parses HCI Event packets. 
HCI is a suite of 5 packet types (Command, ACL Data, SCO Data, Event, ISO Data). This PR covers Event packets only, as the other packets are definitely a larger endeavour.
Claimed `ProtocolType` numbers for the remaining packet types, but they remain unimplemented.
    
Test Data:
Test packets are sourced from Wireshark SampleCaptures. The original file was a `.cap`, which was converted to a `.pcap` using `tshark`. 
Two `.dat` files were created; one with and one without a pseudo header describing the packet direction after inspecting the packet in Wireshark:
image
A third `.dat` file (Command Status) covers an event type that is identified but not decoded yet.

Not in this PR:
Decoding of further event types (Command Status, Connect Complete, ...)
Command, ACL, SCO and ISO data packet layers

All Packet++Test tests pass locally, behaving correctly for valid and invalid inputs. Methods such as the string representation are visually correct as well:

image


smhussa22:
Just checked over the code coverage; will adjust the tests to cover the missing branches when I get the chance.

Dimi1010:
> Just checked over the code coverage; will adjust the tests to cover the missing branches when I get the chance.

Code cov looks good enough tbh. Most misses are in the test file, and a no-op functions?

smhussa22:
> > Just checked over the code coverage; will adjust the tests to cover the missing branches when I get the chance.
> 
> Code cov looks good enough tbh. Most misses are in the test file, and a no-op functions?

Yeah, it's missing coverage for a no-op, a trivial string return function etc.
Adding the coverage would be easy if needed but should be good regardless.

Packet++/header/ProtocolType.h
```diff
@@ -265,6 +265,21 @@ namespace pcpp
...
+	const ProtocolType BluetoothHciAclData = 66;
+
+	/// Bluetooth HCI Protocol SCO Data Packet
+	const ProtocolType BluetoothHciScoData = 67;
+
+	/// Bluetooth HCI Protocol Event Packet
+	const ProtocolType BluetoothHciEvent = 68;
+
+	/// Bluetooth HCI Protocol ISO Data Packet
+	const ProtocolType BluetoothHciIsoData = 69;
```
Dimi1010: Is it necessary to have all the Bluetooth HCI types to be at the top level? The way pcpp is currently written, they only use one byte for identifying protocols, so they are somewhat limited at the moment.

From what I can see in the spec, all HCI packet types are preceded by a identifier octet (0x01 to 0x04 + 0x09)[[Link]](https://software-dl.ti.com/simplelink/esd/simplelink_cc13x2_sdk/1.60.00.29_new/exports/docs/ble5stack/vendor_specific_guide/BLE_Vendor_Specific_HCI_Guide/hci_interface.html). I think it should be possible to have a hierarchical structure:
- BluetoothHciLayer (possibly abstract)
  - BluetoothHciEventLayer
  - BluetoothHci***Layer
  - ...

That way the only one top layer protocol is used `BluetoothHci`. The base Bluetooth HCI layer can expose a method to identify the specific types, as well as a static factory method `BluetoothHicLayer* parseLayer(uint8_t*, size_t, Packet*, /* extra args */)` that performs secondary decoding to the correct HCI sub-type after the primary identification is performed.

We have some examples of this in [`TcpLayer.cpp:454-478`](https://github.com/seladb/PcapPlusPlus/blob/b5bda25b48d53e65bdce1b0427e324bbb156c5ef/Packet%2B%2B/src/TcpLayer.cpp#L454-L478) and all layers that are being constructed via `Layer::tryConstructNextLayerFromFactory` and `Layer::tryConstructNextLayerFromFactoryWithFallback`.

Downside of this is that the user has a 2 step downcast, instead of 1 step tho. :/ Although it might be possible to provide a utility function inside the base class, that bypasses a second dynamic cast, if internal guarantees are kept.

smhussa22: As for the `ProtocolType`, completely slipped my mind that it would be an issue, thanks. It's completely possible to just claim one type and use byte peeking to determine the correct type of packet.

As for the hierarchy, that makes sense and I'm happy to restructure it that way. The plan would be:

1) A single `BluetoothHci` protocol type
2) A `BluetoothHciLayer` base class that owns the parts shared by every HCI packet type: the optional 4-byte direction pseudo-header, direction accessor, and a method to identify the packet type from the H4 indicator octet.
3) A static `BluetoothHciLayer::parseLayer()` factory that peeks the indicator and constructs the matching subclass, e.g. 0x04 maps to BluetoothHciEventLayer (right now for everything else it would map to `nullptr` until those subtypes are implemented)
4) `Packet.cpp` would call the factory for both link types instead of constructing `BluetoothHciEventLayer` directly
5) `BluetoothHciEventLayer` becomes a subclass keeping only the event-specific parsing

Regarding the two-step downcast: since `parseLayer()` is the only thing that ever constructs these objects, the indicator octet is a reliable guarantee of the concrete type (e.g. 0x04 = `BluetoothHciEventLayer`), so the base class can expose typed accessors that check the octet and `static_cast`, avoiding a second `dynamic_cast`, something like this:

```cpp 
// in BluetoothHciLayer
BluetoothHciEventLayer* asEventLayer()
{
  return getPacketType() == BluetoothHciPacketType::Event ? 
    static_cast<BluetoothHciEventLayer*>(this) : nullptr;
}
```
And this way user code can stay at one cast:

```cpp
auto* hci = packet.getLayerOfType<pcpp::BluetoothHciLayer>();
if (auto* event = hci->asEventLayer())
{
    // stuff to do with event packet
}
```

Let me know if this sounds good to you or if you have any suggestions, and I can work on it when I get the chance. Thanks so much for the insight btw.

Dimi1010: Yeah, that sounds good to me. 👍 

@seladb, what do you think?

seladb: I can see it was already changed to one protocol type (65)

Packet++/CMakeLists.txt
seladb: nit: these files are order alphabetically so `BluetoothHciLayer` should come before `BgpLayer`

Tests/Packet++Test/PacketExamples/bluetoothHciCommandStatusWithPhdr.dat
seladb: Can you add a pcap file with these packets so it's easier to view them in Wireshark?

Tests/Packet++Test/CMakeLists.txt
seladb: nit: ditto: `BluetoothHciTests` should come before `BgpTests`

Packet++/src/BluetoothHciLayer.cpp
```diff
@@ -0,0 +1,120 @@
...
+	BluetoothHciLayer* BluetoothHciLayer::parseLayer(uint8_t* data, size_t dataLen, Packet* packet,
+	                                                 bool hasDirectionHeader)
+	{
+		if (!isDataValid(data, dataLen, hasDirectionHeader))
+		{
+			return nullptr;
+		}
+
+		size_t directionHeaderLen = hasDirectionHeader ? sizeof(bluetooth_hci_direction_header) : 0;
+		switch (static_cast<BluetoothHciPacketType>(data[directionHeaderLen]))
```
seladb: `static_cast` may be undefined if `data[directionHeaderLen]` is not one of the enum values. We should check that the value is in the enum range

Packet++/src/BluetoothHciLayer.cpp
```diff
@@ -0,0 +1,120 @@
...
+			return false;
+		}
+
+		size_t directionHeaderLen = hasDirectionHeader ? sizeof(bluetooth_hci_direction_header) : 0;
+		if (dataLen < directionHeaderLen + sizeof(uint8_t) + sizeof(bluetooth_hci_event_header))
+		{
+			return false;
+		}
+
+		return static_cast<BluetoothHciPacketType>(data[directionHeaderLen]) == BluetoothHciPacketType::Event;
```
seladb: ditto: I'd cast `BluetoothHciPacketType::Event` to `uint8_t` and compare the value.

Ditto in other places we do `static_cast`

Packet++/src/BluetoothHciLayer.cpp
Dimi1010: There is some nuance to this. According to the standard, an integral type remains unchanged if the enum is of a fixed underlying type of the same type, since it follows the integral conversion rules to the underlying type prior to being casted to the enum.

https://eel.is/c++draft/expr.static.cast#8

So it would convert with defined behaviour and fall down the default branch.

seladb: Ah ok, I didn't know that. Still, I think it'd be better to check the values first

PR #2267, opened 2026-09-21
Add multi-packet serialization (`PacketSerializer`, `serializePackets`)

seladb:
Adds serialization support for collections of packets, building on the packet/layer serialization framework from #2259.

What's included:

- `PacketSerializer`: serializes a collection of packets into one JSON array, via `addPacket()`/`addPackets()`
- `serializePackets()`: reads every packet from an `IFileReaderDevice` and serializes them into one array, returning the packet count
- Tests for both, in `PacketUtilsTests.cpp` and `FileTests.cpp`

Depends on #2261.

Part of the packet serialization series (see #2259).


Packet++/header/PacketUtils.h
```diff
@@ -62,4 +63,45 @@ namespace pcpp
...
+	class PacketSerializer
+	{
+	public:
+		/// Opens the "packets" array field on the given serializer.
+		/// @param[in] serializer The serializer to write the array into
+		explicit PacketSerializer(ISerializer& serializer) : m_Packets(serializer.writeArray(Packets))
+		{}
+
+		PacketSerializer(const PacketSerializer&) = delete;
+		PacketSerializer& operator=(const PacketSerializer&) = delete;
```
Dimi1010, line 82: Can it be moved?

Issue #2241, opened 2026-08-28
clang-23 with -Werror fails compilation


anders-wind:
### Bug description

**Describe the bug**
Compilation errors:
- `[-Wimplicit-exception-spec-mismatch]` in `3rdParty/MemPlumber/MemPlumber/memplumber.cpp`
- `[-Werror,-Wunused-template]` mostly in `Common++/header/GeneralUtils.h`

Partial build output:
```
...
[80/113] /usr/bin/ccache /usr/bin/clang++-23 -DGIT_BRANCH=\"\" -DGIT_COMMIT=\"\" -DHAS_PCAP_IMMEDIATE_MODE -DHAS_TIMESTAMP_PRECISION_ENABLED -DHAS_TIMESTAMP_TYPES_ENABLED -DUSE_PCAP -Ivcpkg/buildtrees/pcapplusplus/src/v26.07-3364840bc7.clean/3rdParty/MemPlumber/MemPlumber -fPIC -g -std=gnu++14 -fPIC -Wall -MD -MT 3rdParty/MemPlumber/MemPlumber/CMakeFiles/memplumber.dir/memplumber.cpp.o -MF 3rdParty/MemPlumber/MemPlumber/CMakeFiles/memplumber.dir/memplumber.cpp.o.d -o 3rdParty/MemPlumber/MemPlumber/CMakeFiles/memplumber.dir/memplumber.cpp.o -c vcpkg/buildtrees/pcapplusplus/src/v26.07-3364840bc7.clean/3rdParty/MemPlumber/MemPlumber/memplumber.cpp
vcpkg/buildtrees/pcapplusplus/src/v26.07-3364840bc7.clean/3rdParty/MemPlumber/MemPlumber/memplumber.cpp:413:6: warning: function previously declared with an explicit exception specification redeclared with an implicit exception specification [-Wimplicit-exception-spec-mismatch]
  413 | void operator delete(void* pointer, std::size_t size) {
      |      ^
/usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/new:140:6: note: previous declaration is here
  140 | void operator delete(void*, std::size_t) _GLIBCXX_USE_NOEXCEPT
      |      ^
[...]

PR #2268, opened 2026-09-21
Remove static from templated align function

anders-wind:
In #2241 I describe that pcapplusplus is having issues with compilation on the new clang-23. This was due to among other reasons a `-Wunused-template`:
```
In file included from vcpkg/buildtrees/pcapplusplus/src/v26.07-3364840bc7.clean/Packet++/src/PacketTrailerLayer.cpp:2:
vcpkg/buildtrees/pcapplusplus/src/v26.07-3364840bc7.clean/Common++/header/GeneralUtils.h:50:38: error: unused function template 'align' [-Werror,-Wunused-template]
   50 |         template <int alignment> static int align(int number)
```

I think this is due to the unnecessary static in the template definition of align.

PR #2269, opened 2026-09-21
Dont add memplumber to builds without test or example

anders-wind:
In #2241 I describe how the new clang-23 cannot compile pcapplusplus without warnings. One of these warnings were `-Wimplicit-exception-spec-mismatch`: 
```
[80/113] /usr/bin/ccache /usr/bin/clang++-23 -DGIT_BRANCH=\"\" -DGIT_COMMIT=\"\" -DHAS_PCAP_IMMEDIATE_MODE -DHAS_TIMESTAMP_PRECISION_ENABLED -DHAS_TIMESTAMP_TYPES_ENABLED -DUSE_PCAP -Ivcpkg/buildtrees/pcapplusplus/src/v26.07-3364840bc7.clean/3rdParty/MemPlumber/MemPlumber -fPIC -g -std=gnu++14 -fPIC -Wall -MD -MT 3rdParty/MemPlumber/MemPlumber/CMakeFiles/memplumber.dir/memplumber.cpp.o -MF 3rdParty/MemPlumber/MemPlumber/CMakeFiles/memplumber.dir/memplumber.cpp.o.d -o 3rdParty/MemPlumber/MemPlumber/CMakeFiles/memplumber.dir/memplumber.cpp.o -c vcpkg/buildtrees/pcapplusplus/src/v26.07-3364840bc7.clean/3rdParty/MemPlumber/MemPlumber/memplumber.cpp
vcpkg/buildtrees/pcapplusplus/src/v26.07-3364840bc7.clean/3rdParty/MemPlumber/MemPlumber/memplumber.cpp:413:6: warning: function previously declared with an explicit exception specification redeclared with an implicit exception specification [-Wimplicit-exception-spec-mismatch]
  413 | void operator delete(void* pointer, std::size_t size) {
      |      ^
/usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/new:140:6: note: previous declaration is here
  140 | void operator delete(void*, std::size_t) _GLIBCXX_USE_NOEXCEPT
      |      ^
```

There are two solutions to this problem and this PR contains both of them.
- memplumber.h and memplumber.cpp is only ever used in examples and tests. A production build should not include it, therefore one solution is to not add the code unless tests or examples are required.
- The real cause of the problem is that the definitions of operator new and delete in memplumber is not in line with the stdlib: https://en.cppreference.com/cpp/memory/new/operator_delete. `delete` operators should be `noexcept`since c++11 and new operators should be `noexcept(false)`. I have updated the functions. This however might not be wanted since it is modifying 3rd party code.


Dimi1010:
IMO, both fixes are good. 

MemPlumber is technically first party, since @seladb maintains it too. Although do mark the changes from the MemPlumber code with `// PCPP Patch` as it makes it easier to see what is changed from the base, when its ported back. You can see examples in LightPcapNg.

---
@seladb Tangential, but while we are on the subject, the `new` / `delete` implementations of memplumber need a refactor as they aren't thread-safe which is a requirement since C++11.

Dimi1010:
Also, @anders-wind  please change base branch to `dev`, which is where we merge PRs. It will also run the CI automatically.
The PR may need a close and reopen after the base branch change similar to what I have done with #2268.

3rdParty/MemPlumber/MemPlumber/memplumber.cpp
```diff
@@ -378,60 +378,60 @@ const char* getCaller() {
...
     return MemPlumberInternal::getInstance().allocateMemory(size, file, line);
 }
 
-void* operator new[](std::size_t size, const char* file, int line) {
+void* operator new[](std::size_t size, const char* file, int line) noexcept(false) { // PCPP Patch
     return operator new(size, file, line);
 }
 
-void* operator new[](size_t size) _THROW_BAD_ALLOC {
+void* operator new[](size_t size) noexept(false) { // PCPP Patch
```
Dimi1010:
```suggestion
void* operator new[](size_t size) noexcept(false) { // PCPP Patch
```

3rdParty/MemPlumber/MemPlumber/memplumber.cpp
```diff
@@ -378,60 +378,60 @@ const char* getCaller() {
...
 	return operator new(size, getCaller(), 0);
 }
 
-void* operator new(size_t size) _THROW_BAD_ALLOC {
+void* operator new(size_t size) noexept(false) { // PCPP Patch
```
Dimi1010:
```suggestion
void* operator new(size_t size) noexcept(false) { // PCPP Patch
```

3rdParty/MemPlumber/MemPlumber/memplumber.cpp
anders-wind: hehe sorry silly mistake: I was too quick about it.

Dimi1010 (approved):
LGTM

seladb (approved):

PR #2271, opened 2026-09-21
Make MemPlumber allocations thread-safe.

Dimi1010:
Adds a spin lock based on atomic flag to MemPlumber's implementation to satisfy C++11's requirements for new/delete operators to be thread safe. Previously, while `std::malloc` / `std::free` were thread safe, access to the hash table wasn't.

A spin lock is chosen because memory allocation contention would be rare and short lived.


3rdParty/MemPlumber/MemPlumber/memplumber.cpp
```diff
@@ -105,8 +111,28 @@ class MemPlumberInternal {
...
+    // PCPP Patch: C++11 thread safety
+    struct SpinLockGuard
+	{
+		std::atomic_flag& lock;
+		SpinLockGuard(std::atomic_flag& lock) noexcept : lock(lock)
+		{
+			while (lock.test_and_set(std::memory_order_acquire))
+			{
+				// spin until the lock is released
+			}
```
Dimi1010, line 123: Design: Possibly yield the timeslice?

seladb:
Since MemPlumber is basically a debug tool, is there a reason to implement a spin-lock rather than use a simple `std::lock_guard<std::mutex>`? 🤔

Dimi1010:
> Since MemPlumber is basically a debug tool, is there a reason to implement a spin-lock rather than use a simple `std::lock_guard<std::mutex>`? 🤔 

Truthfully, I wasn't sure if the mutex implementations use new / delete anywhere (i think they don't but wasn't sure), so I went with the simpler option from that point of view, since atomic flag is basically just a variable.

seladb:
Sure, I think we can go with you implementation 👍

seladb (approved):

